PR #78, opened 2018-07-10
server: basic testcases

ecthiender:


shahidhk:
@ecthiender add your coverage reports to `/build/_server_output/` if you want them to be accessible post build.

PR #84, opened 2018-07-10
Adds license information in README

dsandip:
Adds a tiny note about licenses


shahidhk:
@dsandip Why can't we add the contents of the gist in the readme itself?

Issue #97, opened 2018-07-11
urls are wrong when access key is set and console redirects


shahidhk:
console redirects to `/login` instead of `/console/login` when being served on server

PR #107, opened 2018-07-11
fix schema dropdown ui, redirect with urlPrefix

praveenweb:
Also fixes the graphiql rendering where on header change, requests were being made.


praveenweb:
Fixes #97

PR #145, opened 2018-07-17
add option to serve console assets from a local directory

shahidhk:


shahidhk:
Need to a redirect from `/` to `/console` on cli served console

PR #160, opened 2018-07-18
supporting uuid in console

### LINKED ISSUE #142: support uuid type in console
While `UUID` type is natively supported in Postgres, `pgcrypto` extension has to be enabled to add a default value.

Without default value: 
```plpgsql
CREATE TABLE article ( 
  id UUID PRIMARY KEY,
  title TEXT NOT NULL
)
```

With default value:
```plpgsql
CREATE EXTENSION IF NOT EXISTS pgcrypto;
CREATE TABLE article (
  id UUID DEFAULT gen_random_uuid() PRIMARY KEY,
  title TEXT NOT NULL
);
```

wawhal:
Fix #142

### DIFF
--- a/console/cypress/helpers/dataHelpers.js
+++ b/console/cypress/helpers/dataHelpers.js
@@ -4,6 +4,7 @@ export const dataTypes = [
   'bigserial',
   'integer',
   'bigint',
+  'uuid',
   'text',
   'numeric',
   'date',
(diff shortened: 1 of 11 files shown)


shahidhk:
@praveenweb please review

coco98:
@praveenweb can we add docs on this please. In the console I cant see the default UUID generation function if the text box is too small.

PR #167, opened 2018-07-18
replaced official graphiql with a fork

wawhal:
[View fork](https://github.com/hasura/graphiql).


shahidhk:
GraphiQL is not accepting any PRs? Until they do, we can use the fork.

PR #169, opened 2018-07-19
add `_inc` and jsonb operators to update_mutation (close #159)

### LINKED ISSUE #159: support various operators in update mutations
For example: 

```graphql
mutation update {
  update_tags ( 
    _concat: { tags: ["a", "b"] }, 
    where { id: { _eq: some_id } }
  ) {
    affected_rows
  }
}
```
The operators that I can think of which will be useful during updates:

### `_inc`
increments the column by a value

```graphql
_inc: { clicks: 1 }
```

## jsonb columns:
https://www.postgresql.org/docs/10/static/functions-json.html
### `_concat`
Equivalent to `||` postgres operator

### `_delete_key`
Equivalent to `-` postgres operator

### `_delete_elem` 
Equivalent to `-` postgres operator

### `_delete_at_path`
Equivalent to `#-` postgres operator

rakeshkky:


server/src-lib/Hasura/RQL/Types/SchemaCache.hs
```diff
@@ -183,6 +185,16 @@ data PGColInfo
 
 $(deriveToJSON (aesonDrop 3 snakeCase) ''PGColInfo)
 
+isIntegerType :: PGColInfo -> Bool
```
0x777: Define this in SQL/Types.hs?

0x777:
@rakeshkky Let's add `_append` and `_prepend` instead of `_concat`. 

Also, let's use the description from Postgres docs for the operators. 
https://www.postgresql.org/docs/10/static/functions-json.html

PR #181, opened 2018-07-20
Convert function expression to an arrow function

cos:
Just a few cases of converting a traditional function to ES6 arrow functions. Pure refactoring.

### DIFF
--- a/console/webpack/webpack-isomorphic-tools.js
+++ b/console/webpack/webpack-isomorphic-tools.js
@@ -53,7 +53,7 @@ module.exports = {
 
     style_modules: {
       extensions: ['scss'],
-      filter: function(module, regex, options, log) {
+      filter: (module, regex, options, log) => {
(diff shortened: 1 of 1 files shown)


cos:
hm, I'm a bit skeptical of signing a CLA enforced in India -- while enjoyable, that would be quite a bit of travel 😄

shahidhk:
This is an automated tool scanning repos and making PRs with certain linting suggestions. @praveenweb if we want to promote the style of using arrow functions instead on traditional function definition, we should setup linting for the same with existing tools that we use, or setup new tools to do the same.

cos:
@shahidhk it's actually a bit less automated, just because I dislike spam as much as the next person. We are selecting repos by hand, running the tool. and reviewing all fixes before making a PR.

shahidhk:
Thanks @cos.

PR #188, opened 2018-07-23
Dev rename table col

zaxboy24:


0x777:
We'll merge this feature when it is ready

PR #201, opened 2018-07-24
add limit to select permissions

praveenweb:
Adds limit input/checkbox under select permissions.


shahidhk:
Needs server side changes.

praveenweb:
PR merged in #237

PR #209, opened 2018-07-25
check for updates every 24 hrs in background (fix #204)

### LINKED ISSUE #204: check for updates everyday in a background thread
https://releases.hasura.io/graphql-engine?agent=server&version=v1.0.0-alpha10

ecthiender:


server/src-lib/Hasura/Server/CheckUpdates.hs
```diff
@@ -0,0 +1,57 @@
...
+
+
+newtype UpdateInfo
+  = UpdateInfo
+  { uiLatest :: T.Text
+  } deriving (Show, Eq)
+
+$(A.deriveJSON (A.aesonDrop 2 A.snakeCase) ''UpdateInfo)
+
+checkForUpdates :: IO ()
```
0x777: Do not create another loggerset. We can use the one created in main

ecthiender:
I'm exporting the `LoggerCtx` data constructor, to retrieve the logger set. Is that fine? 

Ideally, should we have a different instance for logs which are not JSON? `UserMessageLog` or something, and `mkLogger` is parameterized (should work on `EngineLog` and `UserMessageLog`) ?

0x777:
LGTM

PR #211, opened 2018-07-25
filter schema identifiers to conform to GraphQL naming scheme (close #134)

### LINKED ISSUE #134: filter schema identifiers to conform to graphql naming scheme
Filter out tables, columns, relationships etc which does not conform to graphql naming scheme.

This ensures GraphiQL initialisation works properly for existing databases.

rakeshkky:
Filter out tables, columns, relationships etc which does not conform to
GraphQL naming scheme. This ensures GraphiQL initialisation works properly for existing
databases.

Reference: http://facebook.github.io/graphql/June2018/#sec-Names

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Utils.hs
+++ b/server/src-lib/Hasura/GraphQL/Utils.hs
@@ -15,6 +15,7 @@ module Hasura.GraphQL.Utils
   , mkMapWith
   , onLeft
   , showNames
+  , isGraphQLConform
   ) where
 
 import           Hasura.RQL.Types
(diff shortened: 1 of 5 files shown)


server/src-lib/Hasura/GraphQL/Utils.hs
```diff
@@ -87,3 +89,10 @@ onLeft e f = either f return e
 showNames :: (Foldable t) => t G.Name -> Text
 showNames names =
   T.intercalate ", " $ map G.unName $ toList names
+
+-- Ref: http://facebook.github.io/graphql/June2018/#sec-Names
+isGraphQLConform :: G.Name -> Bool
```
0x777: Change it to `isValidName`

PR #215, opened 2018-07-26
add version update checker

praveenweb:
Adds update checker for a new version of server.

Things to test out for
- [ ] If releases api is down, console should ignore
- [x] Once update banner is closed, the state is in localstorage. But if a newer version is available, it should prompt notification again.
- [x] Changelog should link to github's releases tag of that version.

### DIFF
--- a/console/src/components/Main/Actions.js
+++ b/console/src/components/Main/Actions.js
@@ -3,6 +3,7 @@ import globals from 'Globals';
 import defaultState from './State';
 import Endpoints from '../../Endpoints';
 import requestAction from '../../utils/requestAction';
+import requestActionPlain from '../../utils/requestActionPlain';
 import { globalCookiePolicy } from '../../Endpoints';
 import { saveAccessKeyState } from '../AppState';
 import {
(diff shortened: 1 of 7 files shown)


console/src/components/Main/Actions.js
```diff
@@ -41,6 +47,63 @@ const loadMigrationStatus = () => dispatch => {
...
+      } catch (e) {
+        console.error(e);
+      }
+      dispatch({
+        type: SET_SERVER_VERSION_SUCCESS,
+        data: parsedVersion.version,
+      });
+    },
+    error => {
+      console.error(error);
```
karthikvt26: Do we want to show to the user that there is `something wrong` with fetching the current API version

`please get back to us if this keeps happening. This can happen if the graphql engine api is down or is returning a 400.` might help

console/src/components/Main/Actions.js
```diff
@@ -41,6 +47,63 @@ const loadMigrationStatus = () => dispatch => {
...
+        console.error(e);
+      }
+      dispatch({
+        type: SET_LATEST_SERVER_VERSION_SUCCESS,
+        data: parsedVersion.latest,
+      });
+    },
+    error => {
+      console.error(error);
+      dispatch({ type: SET_LATEST_SERVER_VERSION_ERROR, data: null });
```
karthikvt26: Do we want to show to the user that there is `something wrong` with fetching the latest API version

`please get back to us if this keeps happening. This can happen if the releases.hasura.io api is down or is returning a 400.` might help

console/src/routes.js
```diff
@@ -48,7 +73,11 @@ const routes = store => {
       <Route
         path=""
         component={Main}
-        onEnter={composeOnEnterHooks([requireSchema, requireMigrationStatus])}
+        onEnter={composeOnEnterHooks([
+          requireSchema,
+          requireMigrationStatus,
+          requireServerVersion,
```
karthikvt26: We need to move it inside the layout instead of onEnter because if if that hook doesn't succeed, then console will be down. We can leave it if this is the intended behaviour.

console/src/components/Main/Main.js
```diff
@@ -3,119 +3,203 @@ import { connect } from 'react-redux';
...
+      isUpdateAvailable = semver.gt(
+        this.props.latestServerVersion,
+        this.props.serverVersion
+      );
+      const isClosedBefore = window.localStorage.getItem(
+        this.props.latestServerVersion + '_BANNER_NOTIFICATION_CLOSED'
+      );
+      if (isClosedBefore === 'true') {
+        isUpdateAvailable = false;
+        this.state.showBannerNotification = false;
```
karthikvt26: Change it to `this.setState` instead

PR #229, opened 2018-07-30
don't allow creating foreign-key based relationships from/to a table that isn't tracked (fix #185)

### LINKED ISSUE #185: Add a foreign-key based relationship to/fro a table that isn't tracked
Create profile table (via sql, don't track)
```sql
CREATE TABLE profile (id serial primary key, name text);
```

Create address table via console:
`address (id serial, user_id integer, address text)`

Create a foreign key via sql
```sql
alter table profile
add constraint profile_user_fkey foreign key (user_id) references user (id);
```

Go to profile table and add suggested object relationship. Tail hasura logs and try to use graphiql.

rakeshkky:


0x777:
@rakeshkky can we add test cases for this?

0x777:
@praveenweb does the console handle this too? Otherwise an error is seen.

PR #237, opened 2018-07-31
add limit in select permission and override query limit (close #178)

### LINKED ISSUE #178: provide an option to override limit in select permissions
Let's say a limit of `100` is enforced for `anonymous` role on the `article` table. 

When a query `q` on `article` table is made with role `anonymous`, the behavior will be as follows:
1. When `q.limit <= limit`: query goes through.
2. When `q.limit is null OR q.limit > limit`: `q.limit = limit`.

From [delynn](https://app.intercom.io/a/apps/rucirpb3/inbox/inbox/all/conversations/17572719226) on intercom

rakeshkky:


server/src-lib/Hasura/RQL/DML/Internal.hs
```diff
@@ -281,3 +281,10 @@ simplifyError txErr = do
       -- invalid parameter value
       ("22023", msg) -> return msg
       _              -> Nothing
+
+
+-- validate limit and offset int values
+onlyPositiveInt :: MonadError QErr m => Maybe Int -> m ()
+onlyPositiveInt Nothing = return ()
+onlyPositiveInt (Just i) = when (i < 0) $ throw400 NotSupported
+  "negative values are not supported"
```
0x777: `"unexpected negative value"`

This function should take `Int` not `Maybe Int`. Use `mapM/forM` to apply it on a container of type `t a`.

0x777:
Also we seem to be losing the path information. Can you check where and why?

PR #255, opened 2018-08-06
add support for jwt authorization (close #186)

### LINKED ISSUE #186: Enable JWT support for auth

ecthiender:
The API:
1. HGE has `--jwt-secret` flag or `HASURA_GRAPHQL_JWT_SECRET` env var. The value of which is a JSON.

2. The structure of this JSON is: `{"type": "<standard-JWT-algorithms>", "key": "<the-key>"}`
`type` : Standard JWT algos : `HS256`, `RS256`, `RS512` etc. (see jwt.io).
`key`:
  i. Incase of symmetric key, the key as it is.
  ii. Incase of asymmetric keys, only the public key, in a PEM encoded string or as a X509 certificate.

3. The claims in the JWT token must contain the following:
  i. `x-hasura-default-role` field: default role of that user
  ii. `x-hasura-allowed-roles` : A list of allowed roles for the user. The default role is overriden by `x-hasura-role` header.

4. The claims in the JWT token,  can have other `x-hasura-*` fields where their values can only be strings.

5. The JWT tokens are sent as `Authorization: Bearer <token>` headers.

---
To test:
1. Generate a shared secret (for HMAC-SHA256) or RSA key pair.
2. Goto https://jwt.io/ , add the keys
3. Edit the claims to have `x-hasura-role` (mandatory) and other `x-hasura-*` fields. Add permissions related to the claims to test permissions.
4. Start HGE with `--jwt-secret` flag or `HASURA_GRAPHQL_JWT_SECRET` env var, which takes a JSON string: `{"type": "HS256", "key": "mylongsharedsecret"}` or `{"type":"RS256", "key": "<PEM-encoded-public-key>"}`
5. Copy the JWT token from jwt.io and use it in the `Authorization: Bearer <token>` header.

---
TODO: Support EC public keys. It is blocked on frasertweedale/hs-jose#61


server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -0,0 +1,225 @@
...
+     , MonadError QErr m)
+  => JWK
+  -> HTTP.RequestHeaders
+  -> m UserInfo
+processJwt key headers = do
+
+  -- try to parse JWT token from Authorization header
+  let mAuthzHeader = find (\h -> fst h == CI.mk "Authorization") headers
+  (_, authzHeader) <- maybe missingAuthzHeader return mAuthzHeader
+  let tokenParts = BLC.words $ BL.fromStrict authzHeader
```
0x777: pattern match to throw errors instead of checking for length and accessing an element with an index

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -0,0 +1,225 @@
...
+        claims ^. unregisteredClaims
+
+  -- transform the map of text:aeson-value -> text:text
+  metadataWithRole <- decodeJSON $ A.Object claimsMap
+
+  -- throw error if role is not in claims
+  let mRole = Map.lookup userRoleHeader metadataWithRole
+  role <- maybe missingRoleClaim return mRole
+
+  -- delete the x-hasura-role key from this map
```
0x777: why? Let `x-hasura-role` be also passed as params?

ecthiender: The `UserInfo` type already has role info. So I though we should remove it ?

server/src-lib/Hasura/Server/Auth.hs
```diff
@@ -9,42 +9,46 @@
...
 import           Hasura.Server.Logging
+import           Hasura.Server.Utils
+
+import qualified Hasura.Logging         as L
 
-import qualified Hasura.Logging           as L
 
-bsToTxt :: B.ByteString -> T.Text
-bsToTxt = TE.decodeUtf8With TE.lenientDecode
+type AccessKey = T.Text
```
0x777: Let's wrap them in newtypes. I've noticed couple of functions which seem to take `.. Maybe Text -> Maybe Text ..`

PR #258, opened 2018-08-06
remove manifests directory from init (close #254)

### LINKED ISSUE #254: remove install scripts from init directory
The directory create by CLI on doing `hasura init` contains `install_scripts` directory. This is confusing to users as they intend to do `init` on server where they would like to deploy the engine. `init` is supposed to be done on a client, connecting to an engine instance, just to enable migrations through CLI.

shahidhk:


shahidhk:
Needs changes on the deploy script.

PR #263, opened 2018-08-06
add legacy boolean operators to permissions builder

rikinsk:
handles legacy bool operators in permissions builder


0x777:
@rikinsk might as well add support for all the legacy operators?

PR #267, opened 2018-08-07
fix error code for when insertion check constraint fails (fix #257)

### LINKED ISSUE #257: incorrect error code when insertion check constraint fails
```json
{
  "errors": [
    {
      "path": "$",
      "error": "Check constraint violation. insert check constraint failed",
      "code": "postgres-error"
    }
  ]
}
```

It shouldn't be `postgres-error`, should either be `permission-error` or `access-denied`.

rakeshkky:


server/src-lib/Hasura/RQL/DML/Internal.hs
```diff
@@ -247,11 +247,13 @@ validateHeaders depHeaders = do
     unless (hdr `elem` map T.toLower headers) $
     throw400 NotFound $ hdr <<> " header is expected but not found"
 
-simplifyError :: Q.PGTxErr -> Maybe T.Text
+simplifyError :: Q.PGTxErr -> Maybe (Code, T.Text)
 simplifyError txErr = do
   stmtErr <- Q.getPGStmtErr txErr
   codeMsg <- getPGCodeMsg stmtErr
-  extractError codeMsg
+  errTxt <- extractError codeMsg
```
0x777: Let's change `extractError` itself to return both `code` and `message`. We'll probably need to use the correct codes for the rest of the Postgres errors too

PR #291, opened 2018-08-17
fix upsert queries to work on non admin roles (fix #239)

### LINKED ISSUE #239: upsert queries do not work for non admin roles
1. Create a table `user` table with types `id, name` and primary key: `id`, 
2. Create a permission for the user allowing any insert and allowing upserts too

This query is valid per GraphiQL:
```
X-Hasura-Role: user
X-Hasura-Role: 1

mutation {
  insert_user(
    objects:[{name:"test"}],
    on_conflict: {
      constraint: user_pkey
      action: update
    }
  ) {
    affected_rows
  }
}
```

But returns this error from postgres:

```
{
  "errors": [
    {
      "path": "$",
      "error": "constraint \"user_pkey\" for table \"user__insert__public__user\" does not exist",
      "code": "postgres-error"
    }
  ]
}
```

rakeshkky:


server/src-lib/Hasura/GraphQL/Resolve/Mutation.hs
```diff
@@ -107,35 +112,60 @@ parseConstraint obj = do
...
+    setConflictCtxTx conflictCtxM = do
+      let t = maybe "{}" conflictCtxToJSON conflictCtxM
+          setVal = toSQL $ S.SELit t
+          setVar = BB.string7 "SET LOCAL hasura.conflict_clause = "
+          q = Q.fromBuilder $ setVar <> setVal
+      Q.unitQE defaultTxErrorHandler q () False
+
+    conflictCtxToJSON (act, constrM) =
+      let ctx = InsertTxConflictCtx act constrM
+      in CS.cs $ A.encode $ A.object ["conflict_clause" A..= ctx]
```
0x777: https://hackage.haskell.org/package/aeson-1.0.2.1/docs/Data-Aeson-Text.html#v:encodeToLazyText

and convert it to strict text instead of using `cs`.

PR #296, opened 2018-08-21
update tests to use access key

### LINKED ISSUE #113: update tests to use access key
cli tests should have cases where access key is present

arvi3411301:
Closes #113


shahidhk:
need to update circleci tests with access key and without access key

PR #306, opened 2018-08-24
select fields with primary key columns as input arguments (fix #304)

### LINKED ISSUE #304: Add support for arguments based directly on field names (simplified where clause)
Basically,

```
query {
  profile (id: 1) {
    id
    name
  }
}
```

as syntactic sugar for:

```
query {
  profile (where: {id: {_eq: 1}}) {
    id
    name
  }
}
```

This would make it much easier for new folks coming straight from the GraphQL docs who expect to see a similar syntax.

(via David Ventimiglia on intercom/email)

rakeshkky:


server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -275,6 +283,27 @@ mkSelFld tn =
...
+): table!
+-}
+mkSelFldPKey
+  :: QualifiedTable -> [PGColInfo]
+  -> ObjFldInfo
+mkSelFldPKey tn cols =
+  ObjFldInfo (Just desc) fldName args ty
+  where
+    desc = G.Description $ "fetch data from the table: " <> tn
+           <<> " with primary key columns as input fields"
```
0x777: change it to `"using primary key columns"`

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -275,6 +283,27 @@ mkSelFld tn =
...
+    fldName = qualTableToName tn <> "_by_pkey"
```
0x777: Let's change `"_by_pkey"` to `"_by_pk"` everywhere.

PR #311, opened 2018-08-25
add export metadata button in the console (fix #293)

### LINKED ISSUE #293: Feature request: UI Buttons for export and import permission rules
Hasura currently allows us to export metadata with this [api](https://docs.hasura.io/1.0/graphql/manual/hasura-cli/hasura_metadata_export.html), for example
```bash
curl -XPOST -H "x-hasura-access-key: mysecretkey" -d '{"type":"export_metadata", "args": null }' https://hge.herokuapp.com/v1/query
```
The downside is that: If we create our tables in pure sql scripts (not via UI), then when we apply metadata, information is lost.

So this feature will allow us : 

- Separate tables creation from permission rules metadata exporting/importing
- Convenient to export permission rules right from UI (not via cli)

aswinmprabhu:


coco98:
@aswinmprabhu @praveenweb @karthikvt26 @surendran82 Please remember the golden rule. 
There is only one yellow button on every page. We do this so that users gradually build a sense that the yellow button is probably the most important button to look for on any page. Now we have 2 yellow buttons on the schema page.

PR #320, opened 2018-08-27
add req_user_id as alias to x-hasura-user-id (fix #317)

### LINKED ISSUE #317: add req_user_id as alias to x-hasura-user-id
This is to support migration from Hasura platform < 0.15

rakeshkky:


server/src-lib/Hasura/RQL/DDL/Permission/Internal.hs
```diff
@@ -203,12 +209,16 @@ getDependentHeaders boolExp = case boolExp of
...
-  (String t, ty) ->
-    if isXHasuraTxt t
-    then return $ S.SEUnsafe $
+  (String t, ty)
+    | isXHasuraTxt t -> return $ S.SEUnsafe $
       "current_setting('hasura." <> dropAndSnakeCase t
        <> "')::" <> T.pack (show ty)
-    else txtRHSBuilder ty val
+
+    | isReqUserId t -> return $ S.SEUnsafe $
```
0x777: why is this needed here? Why wouldn't dropAndSnakeCase do the same thing?

PR #323, opened 2018-08-28
Issue 293

karthikvt26:
Adds metadata export/import via console.


coco98:
@karthikvt26 Why was my comment on the previous PR ignored?

coco98:
@aswinmprabhu @praveenweb @karthikvt26 @surendran82 Please remember the golden rule.
There is only one yellow button on every page. We do this so that users gradually build a sense that the yellow button is probably the most important button to look for on any page. Now we have 2 yellow buttons on the schema page.

The UI will become a mess if we arbitrarily add things.

@rikinsk @dsandip we quickly need to come up with a UI for where this button has to go. Coming up soon is a schema reload button also that the console needs to have.

console/src/components/Services/Data/Metadata/Metadata.js
```diff
@@ -18,23 +18,21 @@ class Metadata extends Component {
...
-        <hr />
         <div className={metaDataStyles.intro_note}>
-          <h4 className={metaDataStyles.intro_note_heading}>
-            What is GraphQL engine metadata?
-          </h4>
           <div className={metaDataStyles.content_width}>
-            Information about your tables, relationships, and permissions that
-            is used to power the GraphQL API.
+            Hasura metadata stores information about your tables, relationships,
+            and permissions that are used to generate the GraphQL schema and
```
dsandip: ..that **is** used to generate..?

dsandip:
@praveenweb another quick review, please?

0x777:
How does this make sense when we have `events` tab? Shouldn't this go under a separate `Export/Import` tab? @coco98 @karthikvt26 @praveenweb

shahidhk:
@praveenweb @0x777 @coco98 Since we are adding `Events` tab, which is also part of `Metadata`, I suggest we add a new tab called `Schema/Metadata` where we can put the import/export metadata stuff, along with migration mode toggle (for cli).

praveenweb:
@karthikvt26 - static assets (images) are referenced from rstatic/dist which cannot be served by graphql-engine server or cli. These images need to come from CDN link (google storage).

coco98:
@karthikvt26 Many small issues:
1. Link only works if I click on text and not on the entire tab
2. Metadata page has a weird left/right padding
3. Metadata icon is not actually a dropdown, but it's a page. That means it should have the yellow highlighted page styling
4. Twitter message should be
```
Just deployed a GraphQL backend with @HasuraHQ! ❤️ 🚀
https://github.com/hasura/graphql-engine

#graphql #postgres
```
5. I'm not sure about having different colours:
screen shot 2018-09-05 at 12 59 23 pm

karthikvt26:
@praveenweb I have added the version check. Please review it once and provide feedback.

PR #339, opened 2018-08-31
allow mutations on views only if they are allowed by postgres (fix #232)

### LINKED ISSUE #232: enable insert/update/delete on views
for simple views, postgres supports inserts/updates etc. server/console needs to enable this. `pg_catalog` has the required info

rakeshkky:
Similar behaviour observed while creating permissions on views.


0x777:
upserts should not be allowed on views.

rakeshkky:
@shahidhk  This PR requires few console changes. Console has to allow creating `insert/update/delete` permissions based on data from `information_schema.views` table. 
@praveenweb ^

shahidhk:
@rakeshkky Please resolve server conflicts here too.

karthikvt26:
Test steps:

1. Test the editable view flow

- Create a table author with `id` and `name`
- Create the following view to test `editable view`:
```
CREATE OR REPLACE VIEW sample."author_article" AS 
 SELECT a.name,
    a.id
   FROM author a
  ORDER BY a.id;
```
- Track the view
- Navigate to the view > permission and you should see insert/update/delete columns activated.
- Add permission and test in GraphiQL

2. Test the non-editable view flow

- Create article table with `id`, `author_id` and `description`
- Create the following view to test `non-editable view`:
```
CREATE OR REPLACE VIEW public."author_article" AS 
 SELECT a.name,
    a.id, ar.description
   FROM author a
   JOIN article ar
   ON a.id = ar.author_id
  ORDER BY a.id;
```
- Track the view
- Navigate to the view > permission and you shouldn't see insert/update/delete columns.

Test can be performed across the schemas. 

@rakeshkky Please update in case I have missed any use case.

dsandip:
@karthikvt26 Let's add a note in the permissions UI to indicate why configuration for insert/update/delete is not available as shown below. Also, if it's not too much of a change, show the columns for insert/update/delete like you would for a regular table but disable the options in them.

permissions for view

praveenweb:
@karthikvt26 - can we enable Insert tab (next to Browse rows) for views that allows mutations?

PR #341, opened 2018-08-31
generate mutation returning types only if permissions are defined (fix #340)

### LINKED ISSUE #340: do not generate returning type in mutation response if select permission is not defined
For the non-admin role, introspection fails if `insert/delete/update` permission is defined and `select` permission is not defined with the following error.

```json
{
  "errors": [
    {
      "path": "$",
      "error": "type info not found for: 'author'",
      "code": "unexpected"
    }
  ]
}
```

rakeshkky:


server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -319,10 +319,13 @@ type table_mutation_response {
 -}
 mkMutRespObj
   :: QualifiedTable
+  -> Maybe () -- sel perm
```
0x777: Use a Bool

PR #344, opened 2018-09-01
do not update all rows of a table on upsert when action is update (fix #342)

### LINKED ISSUE #342: Primary key changing on Upsert
Hi. I've discussed this in discord with `tanmaig` aka @coco98.

I have the following table structure.
2018-08-31_14h03_41

A tweet can contain many symbols and a symbol can be in many tweets. Classic many to many relationship.

The problem. If I try to `upsert` anything to an existing row in the `symbol` table graphql wants to change the primary key which will break the relationship to tweets.

```javascript
  mutation upsert_symbol {
    insert_symbol(
      objects: [{ symbol: "INTC", is_stock: true, is_crypto: false }]
      on_conflict: { constraint: symbol_symbol_key, action: update }
    ) {
    returning {
      id
      symbol
    }
      affected_rows
    }
  }
```

In this case, if `INTC` already exists and `is_crypto` is being changed then the row is deleted and recreated with a new primary key. 😳

The same thing happens **even if `[{ symbol: "INTC", is_stock: true, is_crypto: false }]` already have the same values**.

@coco98 and @0x777 have discussed and plan on releasing a fix shortly.  
2018-08-31_14h14_10
2018-08-31_14h17_44

They asked me to make this issue so they have something to attach to the PR.

Thanks!
Jeremy

rakeshkky:


coco98:
@0x777 In this case if I provide multiple objects with a different schema then there's a query error per Rakesh. Is that ok for now?

server/src-lib/Hasura/GraphQL/Resolve/Mutation.hs
```diff
@@ -68,49 +69,65 @@ convertRowObj val =
...
+    actionM <- parseAction obj
+    constraint <- parseConstraint obj
+    updColsM <- parseUpdCols obj
+    -- consider "action" if "update_columns" is not mentioned
+    return $ case (updColsM, actionM) of
+      (Just [], _)             -> RI.CCDoNothing $ Just constraint
+      (Just cols, _)           -> RI.CCUpdate constraint cols
+      (Nothing, Just CAIgnore) -> RI.CCDoNothing $ Just constraint
+      (Nothing, Just CAUpdate) -> RI.CCUpdate constraint inpCols
+      (Nothing, Nothing)       -> RI.CCDoNothing $ Just constraint
```
0x777: this should also be same as the previous case, when `update_columns` is missing, we take it that they want to update the specified columns

PR #365, opened 2018-09-06
update logs, add tooltips, prettify settings payload for event triggers

praveenweb:


shahidhk:
@praveenweb Is there an associated issue for this PR?

shahidhk:
included in #373

PR #368, opened 2018-09-06
make sure only events for existing triggers are fetched

tirumaraiselvan:


shahidhk:
@tirumaraiselvan Is there an associated issue for this PR?

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -255,7 +255,11 @@ fetchEvents =
   map uncurryEvent <$> Q.listQE defaultTxErrorHandler [Q.sql|
       UPDATE hdb_catalog.event_log
       SET locked = 't'
-      WHERE id IN ( select id from hdb_catalog.event_log where delivered ='f' and error = 'f' and locked = 'f' LIMIT 100 )
+      WHERE id IN ( SELECT l.id
+                    FROM hdb_catalog.event_log l
+                    JOIN hdb_catalog.event_triggers e
+                    ON (l.trigger_id = e.id) and l.delivered ='f' and l.error = 'f' and l.locked = 'f'
```
0x777: change it to
```
ON (l.trigger_id = e.id) 
WHERE l.delivered ='f' and l.error = 'f' and l.locked = 'f'
```

PR #374, opened 2018-09-07
add parallel tests and update dependencies

arvi3411301:


praveenweb:
Lets remove test_complete.js. And cache cypress module.

PR #383, opened 2018-09-09
update checks on ci systems (close #319)

### LINKED ISSUE #319: update checks on ci systems
tests need to be run with these checks disabled, more preferably, disable update checks when `CI=true` env var is set. 

@0x777 thoughts?

shahidhk:


shahidhk:
Need some more update

PR #386, opened 2018-09-09
export event triggers component

praveenweb:


shahidhk:
@wawhal @karthikvt26 Needs review

PR #387, opened 2018-09-10
python based tests

0x777:
currently tests hge's event framework


0x777:
@shahidhk 
```bash
cd server
stack build --fast --coverage

pytest -vv --hge-url="http://127.0.0.1:8080" --pg-url="postgresql://admin@127.0.0.1:5432/hge_tests"

# stop the server and then
stack hpc report graphql-engine.tix
```

0x777:
exit codes of pytest
https://docs.pytest.org/en/latest/usage.html#possible-exit-codes

shahidhk:
@nizar-m You need to sign the CLA or add your email address to github account.

shahidhk:
@ecthiender Looks like stack is downloading again, can you take a look?

shahidhk:
@ecthiender Can you also check why the HPC report is empty?

0x777:
@arvi3411301 Can you disable the `start_duplicate` test? Let's get some clarity on the spec before disabling the check which causes the test to fail.

PR #419, opened 2018-09-11
add custom headers for webhooks, refactor retry logic

tirumaraiselvan:


server/src-lib/Hasura/RQL/Types/Subscribe.hs
```diff
@@ -60,6 +61,18 @@ data RetryConf
 
 $(deriveJSON (aesonDrop 2 snakeCase){omitNothingFields=True} ''RetryConf)
 
+data HeaderType = FromEnv | FromValue deriving (Show, Eq, Lift)
+$(deriveJSON defaultOptions ''HeaderType)
+
+data HeaderConf
```
0x777:
```haskell
data HeaderConf = HeaderConf HeaderName HeaderValue
data HeaderValue 
  = HVValue Text
  | HVEnv Text
```

The json structure can be as follows:
```json
{
  "name" : "Authorization",
  "valueFromEnv": "ET_AUTH_KEY"
}
```
or 
```json
{
  "name" : "X-Source",
  "value": "hasura-et"
}
```

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -214,18 +219,20 @@ tryWebhook pool e _ = do
     Nothing -> return $ Left $ HOther "table or event-trigger not found"
     Just et -> do
       let webhook = etiWebhook et
+          headerConfs = fromMaybe [] $ etiHeaders et
           createdAt = eCreatedAt e
           eventId =  eId e
       eeCtx <- asks getter
-
+      mheaders <- liftIO $ mapM getHeader headerConfs
```
0x777: Don't read these in every webhook call? We don't expect them to change once the server starts. So you might as well validate them and store the resolved header in the cache.

shahidhk:
Needs docs

tirumaraiselvan:
Added `headers` key to the existing API definition. There are two types of headers `value` and `value_from_env`

```
{
"type": "create_event_trigger", 
"args":
    {
    "name": "testtriggerheaders", 
    "table": "users", 
    "update": { "columns" : ["username", "real_name"]},"insert": { "columns" : ["username"]}, 
    "webhook": "https://httpbin.org/post" , 
    "headers":[{
        "name": "X-Hasura-From-Val", 
        "value": "myvalueNEW"
        }, 
        {
        "name": "X-Hasura-From-Env", 
        "value_from_env": "EVENT_WEBHOOK_HEADER"
        }
        ]} 
}
```

PR #423, opened 2018-09-11
better handling for pgcrypto extension (fix #338)

### LINKED ISSUE #338: create pgcrypto extension by default for uuid support
Right now, when a UUID column is created from console, console executes `CREATE EXTENSION IF NOT EXISTS pgcrypto WITH SCHEMA public;`

But if the user try to create a table with UUID column from run SQL or psql, as shown in [this blog](https://blog.hasura.io/graphql-and-uuid-type-on-postgres-767f016479e9) it will fail.

We should create the pgcrypto extension on graphql engine init step.

ecthiender:


server/src-exec/Ops.hs
```diff
@@ -71,10 +70,18 @@ initCatalogStrict createSchema initTime =  do
...
+    flExtExists <- isExtAvailable "first_last_agg"
+    if flExtExists
+      then Q.unitQ "CREATE EXTENSION first_last_agg SCHEMA hdb_catalog" () False
+      else Q.multiQ $(Q.sqlFromFile "src-rsr/first_last.sql") >>= \(Q.Discard _) -> return ()
+
+  pgcryptoExtExists <- Q.catchE defaultTxErrorHandler $ isExtAvailable "pgcrypto"
+  if pgcryptoExtExists
+    -- only if we created the schema, create the extension
+    then when createSchema $ Q.unitQE needsPgCryptoExt "CREATE EXTENSION IF NOT EXISTS pgcrypto" () False
+    else liftIO $ putStrLn "FATAL: Could not find extension pgcrytpo. This extension is required."
```
0x777: shouldn't we be throwing an error here and exit?

PR #440, opened 2018-09-13
update custom resolvers docs

rikinsk:


shahidhk:
@rikinsk Is this closing any issue?

PR #447, opened 2018-09-14
indicate access key is set in the console context (close #426)

### LINKED ISSUE #426: indicate access key is set in the console context
server:

add a variable `isAccessKeySet: true` to the context if access key is set so that console can prompt the login page.

console:

show login page when the variable is set to true

rakeshkky:
Requires console changes. 
^@praveenweb


shahidhk:
@praveenweb please open another PR for console

PR #452, opened 2018-09-14
use mdillon/postgis to provide postgis support

patcito:
Works as advertised here.


shahidhk:
@patcito Thanks for the PR. Can you move this to a new directory named `docker-compose-postgis`? Let's keep `docker-compose` as it is with official postgres. :smiley:

PR #481, opened 2018-09-19
show access key prompt when server indicates it's set (close #426)

### LINKED ISSUE #426: indicate access key is set in the console context

karthikvt26:


karthikvt26:
@praveenweb Please test out all the flows once.

console/src/components/AppState.js
```diff
@@ -8,11 +8,10 @@ const saveAppState = state => {
   window.localStorage.setItem(stateKey, JSON.stringify(state));
 };
 
-const loadAccessKeyState = () =>
-  window.localStorage.getItem('CONSOLE_ACCESS_KEY');
+const loadAccessKeyState = () => window.localStorage.getItem(stateKey);
```
karthikvt26: Changed it to `stateKey` it was `CONSOLE_ACCESS_KEY` any reasons? same for saveAccessKeyState also.

console/src/components/validateLogin.js
```diff
@@ -0,0 +1,86 @@
+import { loadAccessKeyState, clearState } from './AppState';
```
praveenweb: Shouldn't this file be placed under a folder? Like src/components/Common/validateLogin.js ?

console/src/components/AppState.js
praveenweb: Revert back to 'CONSOLE_ACCESS_KEY' because stateKey variable is used for other localstorage info like history.

praveenweb:
isAccessKeySet: ‘true’ was the expected state. Basically env variables are supposed to be strings and not booleans. Should i change this?

shahidhk:
It'll be easier to make the change on console rather than making a new server release. We'll setup a convention for the future.

shahidhk:
@karthikvt26 I was a data page -> `console/data/schema/public` something like that. I refresh after setting access key, and it takes me to `/console/console/login` and the screen stays at Loading...

shahidhk:
Where is the access key stored? Should there be a logout equivalent?

PR #493, opened 2018-09-19
fix typos in metadata command

Maxpain177:


shahidhk:
Thanks for the PR @Maxpain177

Issue #465, opened 2018-09-17
JWKs should be picked up from well-known URLs of well known providers


ecthiender:
E.g -
1. Auth0 publishes JWK under: https://<your-auth-domain>.auth0.com/.well-known/jwks.json
2. Google publishes it under: https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com

PR #527, opened 2018-09-25
jwt config now takes a jwk url

ecthiender:
### Description
JWT config now takes an optional `jwk_url` parameter (which points to published JWK Set). This is useful for providers who rotate their JWK Set.

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content

### Related Issue

#465 

### Solution and Design

Optional `jwk_url` parameter is taken. The published JWK set under that URL should be in standard JWK format (https://tools.ietf.org/html/rfc7517#section-4.8).

If the response contains an `Expires` header, the JWK set is automatically refreshed.

### Tests

Tests for this would need the current test setup to change (right now tests assume graphql-engine is running in dev mode with full admin access). Tests would be submitted as a separate PR.

### Type

### Checklist:


server/src-exec/Main.hs
```diff
@@ -101,48 +102,88 @@ printJSON = BLC.putStrLn . A.encode
...
       "Fatal Error: Both webhook and JWT mode cannot be enabled at the same time"
     (Just _, Just _, Just _)     -> throwError
       "Fatal Error: Both webhook and JWT mode cannot be enabled at the same time"
 
+-- is this correct ?
+mkJwtMode
+  :: ( MonadIO m
+     , MonadError T.Text m
+     )
+  => AccessKey
```
0x777: Remove the access key argument to this function. It is only used to construct AuthMode. This function can instead return JWTCtx without taking in accesskey.

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -20,29 +27,61 @@ import           Data.ASN1.Encoding         (decodeASN1')
...
+  , jcKey      :: !(Maybe JWK)
+  , jcJwkUrl   :: !(Maybe N.URI)
+  , jcClaimNs  :: !(Maybe T.Text)
+  , jcAudience :: !(Maybe T.Text)
+  -- , jcIssuer   :: !(Maybe T.Text)
+  } deriving (Show, Eq)
+
+data JWTCtx
+  = JWTCtx
+  { jcxKey      :: IORef JWKSet -- should it have strictness anno?
```
0x777: It shouldn't matter here. You can add it for consistency

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -211,30 +308,45 @@ instance A.FromJSON JWTConfig where
...
-      "RS384" -> parseRsaKey rawKey keyType claimNs
-      "RS512" -> parseRsaKey rawKey keyType claimNs
-      -- TODO: support ES256, ES384, ES512, PS256, PS384
-      _       -> invalidJwk ("Key type: " <> T.unpack keyType <> " is not supported")
+    aud     <- o A..:? "audience"
+    jwkUrl  <- o A..:? "jwk_url"
+
+    case mRawKey of
+      Nothing -> return $ JWTConfig keyType Nothing jwkUrl claimNs aud
+      Just rawKey -> do
```
0x777: wrap this in something like `mkJWKFromTxt`. You can then use, 
```haskell
jwkM <- mapM mkJWKFromTxt mRawKey
```

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -20,29 +27,61 @@ import           Data.ASN1.Encoding         (decodeASN1')
...
+import qualified Network.Wreq               as Wreq
 
 
-type RawJWT = BL.ByteString
+newtype RawJWT = RawJWT BL.ByteString
+
+data JWTConfig
+  = JWTConfig
+  { jcType     :: !T.Text
+  , jcKey      :: !(Maybe JWK)
```
0x777: This type will allow both `jcKey` and `jcJwkUrl` to be Nothing which is not a valid `JWTConfig`. It should instead be `jcKey :: Either JWK N.URI`

0x777: The json parsing should account for this.

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -39,6 +42,40 @@ import           Hasura.Prelude
 import           Hasura.RQL.Types.Error
 import           Hasura.Server.Utils
 
+data JwkRefreshLog
```
0x777: Ideally specific logging logic should move to their corresponding modules.

shahidhk:
@0x777 Needs review.

ecthiender:
@shahidhk I need to push some fixes

PR #589, opened 2018-09-30
Made it so that nullable radio defaults correctly (fix #545)

### LINKED ISSUE #545: tabbing through input values should work in insert row page
Let's say I have name (text), description (nullable, text)

I insert something in text, and then I tab and insert text in description. The radio button doesn't get selected on the tab. That means that the actual value that gets inserted is NULL for description and not the text I tried to type.

zacbraddy:
### Description
Made it so that when you focus the text input for a text field that is nullable then if the value in the text field is `undefined`, `null`, or an empty string then the null radio box for that field with be checked.

Also whilst typing in the text input of a nullable text field if you add characters, thus making the value not an empty string, then the radio to the left of the field is checked and the null radio is unchecked. If you then remove all characters leaving an empty string the radio to the left is unchecked and the null radio is checked.

The radio buttons themselves can be changed after having entered or removed a value in the text input but the behaviour above allows for the UI to choose sensible defaults for the user.

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


karthikvt26:
LGTM. @praveenweb can you please take a look?

praveenweb:
@zacbraddy - Hey, just verified your implementation in the above sample heroku app. But doesn't seem like it has been fixed. When i try to type in a text field which is nullable, the left radio has to be selected (typical onChange handler). But it doesn't get selected as expected. Can you recheck?

praveenweb:
@zacbraddy - Just to clarify, your implementation works only for text fields. The expectation is that it should work for all types of fields, like integer, timestamp and boolean fields etc. Let me know if you want to proceed working for that implementation :)

shahidhk:
@zacbraddy Thanks for your contribution :sparkles:. Works as expected for fields marked `Null`.

Do you think it will be easy to add the same behaviour to `Default` also? i.e. If we start typing in a field which has a default value marked, the input box radio button should be selected.

@praveenweb What do you think? I know it's out of scope for the current issue. But would be nice to have.

praveenweb:
@zacbraddy - Thanks. The code refactor looks nice. The giant if-else being rewritten to switch case was neat. The unused code you mentioned was used before when there was a Rich Text Editor functionality for text fields. We had commented out that now, except for this function.

As @shahidhk mentioned in the above comment, can you also add the same behaviour if the field is `Default` also? I think you are conditionally allowing it if its nullable and not auto-increment, so you can also add if isDefault or something? Could be a simple change, i assume.

zacbraddy:
@praveenweb @shahidhk Yeah guys I can absolutely add something like that. The change isn't massive, however, it's more the "How things should work?" that makes doing it complex. Am I correct in saying that a field can be nullable **as well as** having a default. If this is the case which should take precedence when the text changes. The questions that need answering are below. I've guessed at the ones I could I've left the ones I'm unsure of for you to answer.

**A field is not nullable or defaulted**
No functionality applied, the radio boxes stay the same regardless of changes.

**A field is nullable, but not defaulted or auto incremented**
onFocus: 
- If there is a value in the field then put the radio to the left on and turn off the nullable radio
onChange: 
- If length greater than 0 then the left radio is turned on and the nullable radio is turned off
- If length 0 then the left radio is turned off and the right radio is turned on

**A field is nullable, and auto incremented (implicit default implied here)**
- No functionality applied, the radio boxes stay the same regardless of changes.

**A field is nullable, but not auto incremented and defaulted**
- ???
- Profit.

praveenweb:
@zacbraddy 
> Am I correct in saying that a field can be nullable as well as having a default

Yes this scenario seems to be possible when you want a field to assume default value and accept `null` values too when required. 

1. IMO, it makes sense to prioritise `default` over `null` if both were enabled for the field while it is created. Here is the screenshot which depicts the issue. I think this occurs because `null` condition is valid. 

screenshot 2018-10-03 at 4 19 23 pm


Now if at all user wants to have control on the radio button selection he can use arrow keys to navigate between radio buttons.

 2. Now when you have a field with only default option selected. User can tab through the UI and start typing on the field. Now when this happens `default` radio button needs to be unselected and the input box radio button needs to be selected accordingly.

zacbraddy:
There you go, as requested 😃

shahidhk:
Awesome @zacbraddy :sunglasses: 

Looks like some tests are failing for console. @praveenweb @karthikvt26 Can you help out @zacbraddy to fix the tests?

zacbraddy:
Oh bugger, it's cool I'll look into the failing tests and see if I can fix them. I'm going on holiday for a long weekend this weekend so I won't be able to look at them until next week. But if there's no urgency on this then I'm happy to take a look into fixing these when I get back.

shahidhk:
@wawhal Can you quickly check why the tests are failing here and comment with a possible fix?

shahidhk:
@zacbraddy Good news, the tests have been fixed. Finally we can merge this :tada: :clap: 

Thanks for the contribution :pray:. You'll receive a comment from our bot to fill up your address etc. so that we can send some goodies. Don't forget to fill that when you're back from your vacation. :smile:

zacbraddy:
Ah just got back from holiday, sorry to leave you with the tests but thanks for doing that for me @shahidhk was fun working with you guys, you might see me again. Cheers!

shahidhk:
You're always welcom!

PR #620, opened 2018-10-03
update versions in requirements.txt

### LINKED ISSUE #619: Update requirements.txt
Currently requirements.txt has a few python libraries that are not up to date. We have more recent and stable versions for the same. I think it's better to update them.

@shahidhk I'd like to work on this and will be sending a PR for the same in a few minutes.

rahulkumaran:
fixes #619 
### Description

As mentioned in #619, most libraries were not up to date with the latest stable version. I've updated all the libraries in requirements.txt in the docs folder with their latest versions.

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

### Solution and Design

I just checked the latest versions of each library and updated the same in requirements.txt

### Type

- [x] Docs update
- [ ] Community content

### Checklist:


shahidhk:
@rahulkumaran Thanks for your contribution :sparkles: .

How have you tested this change to make sure docs are working as expected?

rahulkumaran:
So I ran all the python files in `docs/_ext` directory and the conf.py file in the `docs` directory with the upgraded versions that I've mentioned in requirements.txt and each of those files ran without raising any errors.

shahidhk:
@rahulkumaran Awesome :trophy: 

@rikinsk Please review the PR.

rikinsk:
@rahulkumaran The Sphinx version cannot be updated to 1.8 as it has breaking changes from 1.6 which we use. 
Also another error that is thrown after the update is: "sphinx-autobuild 0.7.1 has requirement port-for==0.3.1, but you'll have port-for 0.4 which is incompatible"

You should take a look at virtualenv to set up an independent python environment and try setting up docs in that virtual environment. It will not work.

In general, it's not a good idea to randomly update versions of packages a project depends on unless there is real need for it (like a bug in the old version or the need of a new functionality). This ensures the stability of the project and proofing it from breaking changes in newer versions of dependencies.

The other version updates don't seem to cause any issues at first glance but this seems like a high risk PR to merge as a lot more testing needs to be done before we can be sure that nothing is broken.

shahidhk:
Thanks for your efforts @rahulkumaran :pray: :sparkles: . Sadly, as @rikinsk mentioned, we cannot merge this PR.

Feel free to pick up any issue from https://github.com/hasura/graphql-engine/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Ahacktoberfest and give it a go.

PR #670, opened 2018-10-08
add borders for brand assets

edawine:
### Description

Add borders for brand assets in README.md and assets/brand/README.md
for better visibility whether the background color is changeable or not.

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Solution and Design

Add borders to the brand assets by using HTML kbd tags.

### Type

- [x] Docs update
- [ ] Community content

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,8 +164,12 @@ application/website etc. We'd be thrilled if you add the "Powered by Hasura"
 badge to your applications built using Hasura. ❤️
 
 <div style="display: flex;">
-  <img src="assets/brand/powered_by_hasura_black.svg" width="150px"/>
-  <img src="assets/brand/powered_by_hasura_white.svg" width="150px"/>
+  <kbd>
+    <img src="assets/brand/powered_by_hasura_black.svg" width="150px" style="background-color: white;"/>
+  </kbd>
+  <kbd>
+    <img src="assets/brand/powered_by_hasura_white.svg" width="150px" style="background-color: black;"/>
+  </kbd>
 </div>
 
 ```html
--- a/assets/brand/README.md
+++ b/assets/brand/README.md
@@ -2,7 +2,7 @@
 
 All assets are available on a Google Cloud Storage Bucket CDN at `https://graphql-engine-cdn.hasura.io/img/<name>`.
 
-*Note: some logos maybe invisible depending on the page background color.*
+*Note: some logos may be invisible depending on the page background color.*
 
 | name | asset |
 | ---- | ----- |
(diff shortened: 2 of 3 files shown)


shahidhk:
Thanks your contribution @edawine.

I feel like the images look better the way they are right now.
I appreciate the motivation, but all assets come with a transparent background.

Also, the tag you have used is `kbd` which is for indicating a keyboard stroke, like <kdb>Ctrl</kbd>. For adding borders, `border` CSS element is recommended.

ref: https://html.com/tags/kbd/

edawine:
@shahidhk FYI, I did this because the powered_by_hasura_white.svg is invisible by default GitHub user (see below). By giving a border to the picture (yes, it is a hack-ish border (ab)using kbd tag), at least the readers are aware on the image's presence. The same image is not visible in assets/brand/README.md.

hasura_readme md

edawine:
@shahidhk Yeah the code sample indicates that there is a white version of the logo, so I think that's fine.

I've seen your attempt to color the images with style tag, but unfortunately GitHub markdown doesn't [parse style tag](https://github.github.com/gfm/#disallowed-raw-html-extension-). The other solution that I've thought of is to create a colored background images, just for demonstration purposes.

What do you think @shahidhk ? Do you have other ideas?

shahidhk:
@edawine Yes, I had tried that and given up. :rofl: 

I don't think adding images just for demo purposes is a good idea. It will unnecessarily increase the repo size.

Another idea I can think of is to order them in such a manner that the striped table takes care of the background. If you have noticed, only the last logo is hidden by white bg since the table is striped. Maybe by renaming the mascot logos appropriately or by modifying the `for` loop in the script we can get an output where all white logos appear on the grey row.

How about that?

edawine:
@shahidhk I think modifying the `for` loop is a better option compared to renaming the files.

Renaming the files will be harder if more brand assets are added in the future.

My idea of modifying the `for` loop is have two `for` loops. One iterates the logos that has `black` or `white` in their file names, to ensure the logos will have their matching backgrounds. The other loop runs for the rest of the logo (maybe hard coding the file names?). The problem I see with my idea is that the maintainers will have to remember to update the list in the script.

shahidhk:
Adding the file names to the script defeats the purpose of having the scrips. I think as a stop-gap measure, we can just manually edit the md file and switch the position of last two files. What say?

shahidhk:
Sorry about the delay @edawine. This looks good. Go ahead and submit a PR.

edawine:
Will do, thank you!

Issue #536, opened 2018-09-26
alert thrown when access key is set on server, but cli doesn't know


shahidhk:
PFA the screenshot. Access key is set on server. But, cli is not aware. Can we not show the alert and prompt login?

![screenshot from 2018-09-26 14-16-16](https://user-images.githubusercontent.com/4124733/46068736-7e6b8c00-c197-11e8-9368-598623b2e1c9.png)

n1arash:
i'm working on this issue.

PR #679, opened 2018-10-08
Removed alert box (solve #536)

vipulgupta2048:
- The alert now comes up only when there is access key involved. I think that would get the job done. 

:hatching_chick:


vipulgupta2048:
Sorry for the noise, I should have read the issue completely.

PR #802, opened 2018-10-17
Update README.md

tinafanson:
updating the language support for C#.


shahidhk:
@tinafanson The readme is supposed to show what languages have boilerplate already added to this repo. Since we do not have a C# boilerplate, we cannot check that column. :smiley:

PR #813, opened 2018-10-19
add digital ocean one-click docs

shahidhk:
docs for DO one click


shahidhk:
@rikinsk The url cannot be changes since it's already added to the snapshot image.

PR #1003, opened 2018-11-09
add graphql server boilerplates

shahidhk:


shahidhk:
@wawhal Are you adding a JS boilerplate?

Issue #997, opened 2018-11-08
Create trigger failed


nathanstitt:
I receive an error whenever I attempt to add a trigger.  The message indicates I'm using invalid characters but that doesn't seem to be the case.  In the example I'm just using "test".

Based on the error I think the error originates from https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Hasura/RQL/Types/Subscribe.hs#L129

When I look at the request from the console it's sending:

```json
{"name":"create_trigger_test","up":[{"type":"create_event_trigger","args":{"name":"test","table":{"name":"submissions","schema":"public"},"webhook":"https://localhost:9292/hook","insert":{"columns":["id","embed_id","created_at","created_by_id"]},"headers":[]}}],"down":[{"type":"delete_event_trigger","args":{"name":"test"}}]}
```

I'm using `v1.0.0-alpha28`

<img width="888" alt="screen shot 2018-11-07 at 9 41 36 pm" src="https://user-images.githubusercontent.com/79566/48176520-f4562d80-e2d5-11e8-8ded-549fd3e3d5f5.png">

0x777:
@nathanstitt We couldn't reproduce this on our end. Can you post the server log for this request?

nathanstitt:
Sure, happy to do so.

I should also add that I'm running the server directly, i.e. without docker.  I haven't had any other issues with that though.  

I also slightly altered the line to print the name so I could be sure it's actually seeing the right value and it appears to be correct
```
      Nothing -> fail (printf "only alphanumeric and underscore allowed for name: '%s'" (T.unpack name))
```
I don't see why the regex wouldn't match.  My Haskel is really shaky but am happy to try anything else you suggest.

I see the logged after clicking "create trigger", note that the last line does show the name is "test"
```
{"timestamp":"2018-11-08T09:47:05.539-0600","level":"info","type":"http-log","detail":{"status":200,"query_hash":"f57ddec8d7517d9634f77a23dc15a0839b6b6b5a","http_version":"HTTP/1.1","query_execution_time":0.537491,"request_id":null,"url":"/v1/query","ip":"127.0.0.1","response_size":53,"user":{"x-hasura-role":"admin","x-hasura-access-key":"mysecretkey"},"method":"POST","detail":null}}
[...]

PR #1007, opened 2018-11-11
Fix creating triggers on OSX

nathanstitt:
The other version was failing on OSX, most likely because posix regex
is poorly supported there and the line-start/end characters where
not matching the strings?

### Description

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

https://github.com/hasura/graphql-engine/issues/997

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:

- [ ] I have read the **[contributing guide](https://github.com/hasura/graphql-engine/blob/master/CONTRIBUTING.md)** and my code conforms to the guidelines.

### DIFF
--- a/server/src-lib/Hasura/RQL/Types/Subscribe.hs
+++ b/server/src-lib/Hasura/RQL/Types/Subscribe.hs
@@ -29,7 +29,7 @@ import           Hasura.Prelude
 import           Hasura.SQL.Types
 import           Language.Haskell.TH.Syntax (Lift)
 import           Text.Regex                 (matchRegex, mkRegex)
-
+import           Text.Printf
 import qualified Data.Text                  as T
 
 type TriggerName = T.Text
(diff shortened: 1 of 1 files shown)


server/src-lib/Hasura/RQL/Types/Subscribe.hs
```diff
@@ -122,11 +122,11 @@ instance FromJSON CreateEventTriggerQuery where
     webhook   <- o .: "webhook"
     headers   <- o .:? "headers"
     replace   <- o .:? "replace" .!= False
-    let regex = mkRegex "^\\w+$"
+    let regex = mkRegex "\\W"
```
tirumaraiselvan: @nathanstitt Can you please change this to the following condition:
`^[A-Za-z]+[A-Za-z0-9_\\-]*$`

Not only is this more safe (vanilla regex) but it also allows for `-` (hyphen). Must be merged after https://github.com/hasura/graphql-engine/pull/1012/ .

tirumaraiselvan: Seems like we will need to change the whole regex package. Turns out we need a native Haskell package like regex-tdfa : https://wiki.haskell.org/Regex_Posix

```
If you want a bug-free and/or portable POSIX extended regular expression library to use from Haskell, then regex-posix will not help you. You should use the regex-tdfa package instead.
```

@nathanstitt If you are up for it, you may try changing the package from `regex-compat` to `regex-tdfa` . Otherwise I'll queue this up in my list.

Issue #657, opened 2018-10-05
Descending order does not utilize index


jasonmorganson:
Descending queries use `NULLS LAST` which does not allow them to utilize the same index as ascending order queries.

## Query
Given this GraphQL query:
```graphql
{
  article (limit: 10, order_by: date_desc) {
    link
  }
}
```

## Index
```sql
create index on article (date);
```

## Using NULLS LAST
It produces this SQL:
```sql
EXPLAIN ANALYSE SELECT coalesce(json_agg((SELECT "e" FROM (SELECT "r"."link" AS "link") AS "e")), '[]')
FROM (SELECT "link" AS "link"
      FROM "public"."article"
      WHERE (('true') AND (('true') AND ('true')))
      ORDER BY "date" DESC NULLS LAST
      LIMIT ('10') :: int) AS "r"
```
Which results in this explain:
```
Aggregate  (cost=2402457.72..2402457.73 rows=1 width=32) (actual time=41347.092..41347.092 rows=1 loops=1)
  ->  Limit  (cost=2402457.47..2402457.50 rows=10 width=92) (actual time=41346.861..41346.875 rows=10 loops=1)
        ->  Sort  (cost=2402457.47..2458049.25 rows=22236712 width=92) (actual time=41346.861..41346.868 rows=10 loops=1)
              Sort Key: article.date DESC NULLS LAST
              Sort Method: top-N heapsort  Memory: 26kB
              ->  Seq Scan on article  (cost=0.00..1921930.12 rows=22236712 width=92) (actual time=0.202..37852.110 rows=22236711 loops=1)
  SubPlan 1
    ->  Result  (cost=0.00..0.01 rows=1 width=32) (actual time=0.005..0.005 rows=1 loops=10)
Planning time: 1.761 ms
Execution time: 41347.308 ms
```

## Using NULLS FIRST
If you change the query to order by `NULLS FIRST` it does a backward scan using the same index.
```sql
EXPLAIN ANALYSE SELECT coalesce(json_agg((SELECT "e" FROM (SELECT "r"."link_web" AS "link_web") AS "e")), '[]')
FROM (SELECT "link_web" AS "link_web"
      FROM "public"."acris_simple"
      WHERE (('true') AND (('true') AND ('true')))
      ORDER BY "recorded_date" DESC NULLS FIRST
      LIMIT ('10') :: int) AS "r"
```
```
[...]

PR #1009, opened 2018-11-12
change descending ordering to nulls first (fix #1008)

### LINKED ISSUE #1008: Very slow select queries (Ignoring Indexing?)
I have a table with 1M rows that has the following index:

```
psql -U user -d db -c “SELECT * FROM pg_indexes WHERE tablename = ‘messages’;”
schemaname |  tablename  |         indexname         | tablespace |                                          indexdef
------------+-------------+---------------------------+------------+---------------------------------------------------------------------------------------------
public     | messages| updates_timestamp_idx |            | CREATE INDEX updates_timestamp_idx ON public.updates USING btree (“timestamp” DESC)
```

When I do a select query for the first 10 entries in desc order on the timestamp key on PSQL it runs in 0.039s while hasura takes 4.735s. 

image

This difference is orders of magnitude bigger. Any idea what's the reason? 

P.S. It returns a different result as well. Maybe it doesn't use the indexes at all? 

Also, example times from a query on a non indexed column from 1. hasura, 2. psql, 3. postgrest

image

slowr:
### Description

~~Auto-generated queries when there is an order function they include `NULLS FIRST` or `NULLS LAST`.~~
Descending ordering auto-generated queries include `NULLS LAST` instead of postgres default; `NULLS FIRST`
This tends to mess-up indexing tables as mentioned in #1008 and #657

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

#1008
#657

### Solution and Design

~~A new enum is added to avoid NULLS ordering (FIRST, LAST) in the auto-generated queries. Decided to introduce new enum and not change the default implementation to keep backwards compatibility.~~
Change the default NULLs ordering for desc to be `NULLS FIRST` so it follows postgres specification.

### Type

- [x] Docs update
- [ ] Community content

### Checklist:


0x777:
@slowr We decided to change the default behavior of `desc` to use `null first` instead of the current `nulls last`. The enums would therefore be:

```graphql
asc
asc_nulls_last # same as asc
asc_nulls_first
desc
desc_nulls_first # same as desc
desc_nulls_last
```

Will you be willing to make these changes? If so, you don't need to add another value called `NNull` in the SQL AST. You can just map `desc` to `NFirst` and `asc` to `NLast`.

slowr:
@0x777 Should I make the change to the file on this PR and change descriptions etc or just a commit will be sufficient?

PR #1072, opened 2018-11-20
do not add extra s after table name when generating array relationship name

elitan:
…p name

### Description
do not add extra s after table name when generating array relationship name

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

None

### Solution and Design

Some (most) use plural when naming their tables. This make the extra added, hard coded, `s` after a table name confusing. Even if the tables are named in singular, just adding a `s` in the end is not the correct plural naming. Like `company` which in plural is `companies`. Just adding a `s` in the end would result in an incorrect naming: `companys`. 

### Type
- [ ] Docs update
- [ ] Community content

### Checklist:

### DIFF
--- a/community/tools/json2graphql/src/import/relationships.js
+++ b/community/tools/json2graphql/src/import/relationships.js
@@ -7,7 +7,7 @@ const getObjRelationshipName = dep => {
 };
 
 const getArrayRelationshipName = (table, parent) => {
-  const relName = `${table}sBy${parent[0].toUpperCase()}`;
+  const relName = `${table}By${parent[0].toUpperCase()}`;
   return parent.length === 1 ? `${relName}Id` : `${relName}${parent.substring(1, parent.length)}Id`;
 };
 


elitan:
hmm, do I have to take any further action on this?

shahidhk:
Hi @elitan, thanks for the contribution :pray: 

The change you have added is for `community/tools/json2graphql`, which is different from the console.

@praveenweb @karthikvt26, can you point @elitan to the right place where he needs to make this change?

My suggestion is that we add a `s` for array realationships if the table name doesn't end with a `s` already (e.g. `order` -> `ordersByUserId`). If the table name already contains `s`, we don't add another `s` (e.g. `orders` -> `ordersByUserId`).

elitan:
Ok, yes please nock me in the right direction. :)

But first maybe we should discuss if we should or shouldn't add the extra `s`. As I described above, English is more complex than just add a `s` in the end to transform any world to its plural form.

With your idea with adding a `s` if the word does not end with `s` does not work if you would name your tables in singular form and use, let's say, `company`.

This would result in `companysBy...` Not very pretty nor correct.

shahidhk:
Makes sense @elitan. What about using something like [`pluralize`](https://github.com/blakeembrey/pluralize)? It should cover most of the use cases, right?

For array relationships, we take the remote table, pluralize it, and then append the `ByColumnName`. :thinking:

shahidhk:
@elitan Sorry, lost track of this issue. @praveenweb What do you think?

praveenweb:
@shahidhk @elitan - Related issue #1444 . 

The idea is to generate singular names for object relationships and plural names for array relationships using the inflection module.

We can close this PR.

shahidhk:
Thanks for the PR @elitan, as @praveenweb mentioned, we'll use the node module inflection to do this. Feel free to take up #1444 if you are interested :smile:

PR #1073, opened 2018-11-20
postgres functions as custom queries (close #333)

### LINKED ISSUE #333: expose postgres functions as queries and mutations
This is kind of an escape hatch for some of the features that we do not support out of the box.

rakeshkky:
### Description

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [x] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

close #333 

### Solution and Design

For table `article`.

<p>

```sql
create table article (id serial primary key, title text not null, content text not null, author_id integer not null)
```

</p>

Define a function `search_articles` as following:-

```sql
      create function search_articles(search text)
      returns setof article as $$
          select *
          from article
          where
          title ilike ('%' || search || '%') or
          content ilike ('%' || search || '%')
      $$ language sql stable;
```
Then GraphQL Engine generates custom query field on `search_articles` :- 

```graphql
search_articles(
args: search_articles_args! 
where: artcle_bool_exp
order_by: [article_order_by!] 
limit: Int
offset: Int
): [article!]! 

input search_article_args {
search: String!
}
```
Sample GraphQL query:-

```graphql
query search_articles {
  search_articles(
    args: {search: "Hasura"}
  ){
    id
    title
    content
    author_id
  }
}
```
Metadata API:-

Adding a function

```http
   POST /v1/query HTTP/1.1
   X-Hasura-Role: admin

   {
       "type": "track_function",
       "args": {
           "schema": "public",
           "name": "search_articles"
       }
   }

```
Removing a function

```http
   POST /v1/query HTTP/1.1
   X-Hasura-Role: admin

   {
       "type": "untrack_function",
       "args": {
           "schema": "public",
           "name": "search_articles"
       }
   }

```

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


revskill10:
How about sql function as `mutation`  ?

shahidhk:
@revskill10 We'll add them inccrementally.

revskill10:
So as i see, there won't be a subscription for the same query with custom sql function ?
e.g 
```
subscription search_articles {
  search_articles(
    args: {search: "Hasura"}
  ){
    id
    title
    content
    author_id
  }
}
```

rakeshkky:
@revskill10 Subscription allows custom queries as well.

Sach97:
> @revskill10 Subscription allows custom queries as well.

Does this mean we will be able to subscribe to a function that return a trigger like in this [article](https://tapoueh.org/blog/2018/07/postgresql-listen/notify/) ?

jasonmorganson:
@rakeshkky this would be a really useful feature. Any idea on when this will land?

shahidhk:
@jasonmorganson This should take another week or two.

karthikvt26:
@praveenweb Please review schema sections and version check
@arvi3411301 Please review run sql regular expression

rakeshkky:
@Sach97 

Instead of creating a composite type for `post_type`, create a table 
```sql
CREATE TABLE post_type (pid INT, p_title TEXT)
```
Add this table using the console. 

Now, In your function definition change `RETURNS setof post` to `RETURNS setof post_type`

server/src-exec/Ops.hs
```diff
@@ -241,6 +241,20 @@ from6To7 = liftTx $ do
...
+from7To8
+  :: (MonadTx m, HasHttpManager m, CacheRWM m, UserInfoM m, MonadIO m)
+  => m ()
+from7To8 = do
+  -- migrate database
+  Q.Discard () <- liftTx $ Q.multiQE defaultTxErrorHandler
+    $(Q.sqlFromFile "src-rsr/migrate_from_7_to_8.sql")
+  -- migrate metadata
+  void $ runQueryM migrateMetadataFrom7
+  setAsSystemDefined
```
0x777: `setAsSystemDefined` is too broad. Use a statement which will only make the modifications for the added metadata.

server/src-lib/Hasura/RQL/DDL/Schema/Function.hs
```diff
@@ -0,0 +1,196 @@
...
+  adminOnly
+  rawSchemaCache <- askSchemaCache
+  when (M.member qf $ scFunctions rawSchemaCache) $
+    throw400 AlreadyTracked $ "function already tracked : " <>> qf
+
+trackFunctionP2Setup :: (QErrM m, CacheRWM m, MonadTx m)
+                     => QualifiedFunction -> m ()
+trackFunctionP2Setup qf = do
+  fi <- withPathK "name" $ liftTx $ getFunctionInfo qf
+  void $ askTabInfo $ fiReturnType fi
```
0x777: Is this to check if the return table is tracked? If so, it should catch the error and return a more useful error message along these lines: "the return table: tname has to be tracked before the function can be tracked"

console/src/components/Services/Data/Function/Modify/ModifyCustomFunction.js
```diff
@@ -0,0 +1,217 @@
...
+} from '../customFunctionReducer';
+
+import { SET_SQL } from '../../RawSQL/Actions';
+
+class ModifyCustomFunction extends React.Component {
+  constructor() {
+    super();
+    this.state = {};
+    this.state.deleteConfirmationError = null;
+    this.state.sampleSQL = `create function search_posts(search text)
```
praveenweb: Is this required?

console/src/components/Services/Data/Function/customFunctionReducer.js
```diff
@@ -0,0 +1,365 @@
...
+      credentials: globalCookiePolicy,
+      headers: dataHeaders(getState),
+      body: JSON.stringify(finalReqBody),
+    };
+
+    const onSuccess = data => {
+      if (globals.consoleMode === 'cli') {
+        dispatch(loadMigrationStatus()); // don't call for server mode
+      }
+      // dispatch(loadTriggers());
```
praveenweb: remove this?

console/src/components/Services/Data/Function/customFunctionReducer.js
```diff
@@ -0,0 +1,365 @@
...
+    const options = {
+      credentials: globalCookiePolicy,
+      method: 'POST',
+      headers: dataHeaders(getState),
+      body: JSON.stringify(bulkQuery),
+    };
+    dispatch({ type: FETCHING_INDIV_CUSTOM_FUNCTION });
+    return dispatch(requestAction(url, options)).then(
+      data => {
+        if (data[0].length > 0 && data[1].length > 0) {
```
praveenweb: What if data is an empty array? It will throw an exception.

karthikvt26: Its a bulk query. I am guessing the base value will be `[[],[]]`.

console/src/components/Services/Data/Function/customFunctionReducer.js
```diff
@@ -0,0 +1,365 @@
...
+      }
+    );
+  };
+};
+
+const deleteFunctionSql = () => {
+  return (dispatch, getState) => {
+    const currentSchema = getState().tables.currentSchema;
+    const functionName = getState().functions.functionName;
+    const sqlDropView =
```
praveenweb: change to sqlDropFunction

console/src/components/Services/Data/Function/customFunctionReducer.js
```diff
@@ -0,0 +1,365 @@
...
+      '"' +
+      functionName +
+      '"';
+    const sqlUpQueries = [
+      {
+        type: 'run_sql',
+        args: { sql: sqlDropView },
+      },
+    ];
+    // const sqlCreateView = ''; //pending
```
praveenweb: why is the down migration not generated?

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -39,8 +40,8 @@ const migrationNameTip = (
 );
 const trackTableTip = (
   <Tooltip id="tooltip-tracktable">
-    If you are creating a table/view, you can track them to query them with
-    GraphQL
+    If you are creating a table/view/function, you can track them to query them
```
praveenweb: Semver check for this?

console/src/helpers/semver.js
```diff
@@ -13,6 +13,7 @@ const componentsSemver = {
   webhookEnvSupport: '1.0.0-alpha29',
   insertPermRestrictColumns: '1.0.0-alpha28',
   permHideUpsertSection: '1.0.0-alpha32',
+  customFunctionSection: '1.0.0-alpha35',
```
praveenweb: Shouldn't this be alpha36?

PR #1153, opened 2018-12-01
fix preloader refresh on events page (close #1000)

### LINKED ISSUE #1000: Annoying UI loader on events
Hello,

When watching events is on, the events table flickers every few seconds because of an extra loader, like this:

image

image

This becomes quite tiring for the eyes after a while, so I am posting this for your consideration to remove the loader from that position.

WickedBrat:
Fixes #1000 
This PR fixes the annoying flicker while watching events.


praveenweb:
@WickedBrat - Can you also remove the loader from pending events page? Also remove any code written for the loading spinner on those 2 pages.

WickedBrat:
@praveenweb I've updated. PTAL

PR #1230, opened 2018-12-18
added query params for preloading queries (fix #1179)

### LINKED ISSUE #1179: accept query params for preloading queries/mutations in console graphiql
It would be nice to have the console accept query parameters to preload queries/mutations.

Right now `query_file` pointing to a github link works. But lets also accept `query` parameter directly with support for variables too.

So it would support the following:

1) `query` param for actual query
2) `headers` param for a list of headers to be templated

CodHeK:
Added query params for preloading queries/mutations/subscriptions, variables and headers. 
Fixes: #1179 

### Description

- [x] Console

- [x] Console

### Solution and Design

Test link : [demo url](http://localhost:3000/?query_string=query+nurseQuery%28%24room_no%3A+String%21%29+%7B%0D%0A++++nurse+%28%0D%0A++++++where%3A+%7B+room_no%3A+%7B_eq%3A+%24room_no+%7D%7D%0D%0A++++%29+%7B%0D%0A++++++id%0D%0A++++++name%0D%0A++++++room_no%0D%0A++++%7D%0D%0A++%7D&graphiql_headers=x-hasura-role%3Auser%2C+x-hasura-user-id%3A1&query_variables=%7B+%0D%0A++++%22room_no%22+%3A+10%2C%0D%0A%7D)
Result: 
screenshot 2018-12-18 at 3 28 57 pm

query_string : 
```
query nurseQuery($room_no: String!) {
    nurse (
      where: { room_no: {_eq: $room_no }}
    ) {
      id
      name
      room_no
    }
  }
```

query_variables : 
```
{ 
    "room_no" : 10,
}
```

graphiql_headers : 
`x-hasura-role:user` `x-hasura-user-id:1`

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


praveenweb:
@CodHeK - This is working except for the case where there are no tables. In that case, only headers are templated and query isn't.

Also need a way to generate this URL with query params so that people can share.

CodHeK:
@praveenweb  in case of no tables, it should not right? 

I'll add a share button too then ?

wawhal:
@CodHeK templating should be done irrespective of whether the tables are present or not.

CodHeK:
Demo of the the feature added : 

ezgif com-video-to-gif

CodHeK:
That was by mistake, I installed a package in the root by mistake, I’ll just remove and push again 


On Thursday, January 10, 2019, 12:32 PM, Praveen Durairaj <notifications@github.com> wrote:


@praveenweb requested changes on this pull request.

Why is package-lock.json file created in the root of the repo?

Stephve:
How can I write codes?

wawhal:
@CodHeK there are two issues:

1. When there are no tables, the editor does not work. It goes into some buggy state.
2. GraohiQL is making an introspection request for every character typed in the editor. I think this is because you are re-rendering graphiql everytime something is typed. You will have to figure out a way to not let this happen.

Sorry for the delayed review.

CodHeK:
@wawhal I've corrected the behaviour of the editor when there are no tables, for the second point, how would you suggest I should proceed, because for displaying the text typed on the editor, I need to update the state and re-render.

wawhal:
@CodHeK you need not store the variables and queries in state. You can directly operate over the variables you get in the prop functions.

wawhal:
@CodHeK you get the absolute values in that prop function. That should be enough to update the URL. You don't need to store them separately because you aren't doing anything else with them right?

CodHeK:
@wawhal anyway, if I get rid of the states, GraphiQL will still have to make those api calls to get the latest value in the editor right ? And the state I just maintained it for it to look organised and could be accessed anywhere, if needed.

Stefmore02:
Thank you for this PR and we're sorry it took us so long to review. Unfortunately, we'll close it since a lot has changed and it's no longer relevant.

PR #1280, opened 2019-01-01
allow specifying network interface with "server-host" option

nathanstitt:
### Description

This adds an option to allow specifying the network address to bind to.  Previously graphql-engine would bind to the wildcard interface, while I'd prefer it only listen to `localhost`

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Solution and Design
I'm very much a Haskel neophyte and pretty much just copied how `server-port` worked.  I'm happy to rework if this could be accomplished in a better manner.

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:

   * **note** I did not document this option since `server-port` is not documented.  If you'd like I can add a mention for both of these to the README

### DIFF
--- a/server/src-exec/Main.hs
+++ b/server/src-exec/Main.hs
@@ -18,7 +18,6 @@ import qualified Data.Yaml                  as Y
 import qualified Network.HTTP.Client        as HTTP
 import qualified Network.HTTP.Client.TLS    as HTTP
 import qualified Network.Wai.Handler.Warp   as Warp
-import qualified Data.String                as DataString
 
 import           Hasura.Events.Lib
 import           Hasura.Logging             (Logger (..), defaultLoggerSettings,
@@ -126,8 +125,7 @@ main =  do
       (app, cacheRef) <- mkWaiApp isoL loggerCtx pool httpManager
                          am corsCfg enableConsole
 
-      let hostS = DataString.fromString(host)
-      let warpSettings = Warp.setPort port $ Warp.setHost hostS Warp.defaultSettings
+      let warpSettings = Warp.setPort port $ Warp.setHost host Warp.defaultSettings
 
       -- start a background thread to check for updates
       void $ C.forkIO $ checkForUpdates loggerCtx httpManager
--- a/server/src-exec/Main.hs
+++ b/server/src-exec/Main.hs
@@ -18,6 +18,7 @@ import qualified Data.Yaml                  as Y
 import qualified Network.HTTP.Client        as HTTP
 import qualified Network.HTTP.Client.TLS    as HTTP
 import qualified Network.Wai.Handler.Warp   as Warp
+import qualified Data.String                as DataString
 
 import           Hasura.Events.Lib
 import           Hasura.Logging             (Logger (..), defaultLoggerSettings,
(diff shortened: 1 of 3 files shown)


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -33,6 +33,7 @@ type RawAuthHook = AuthHookG (Maybe T.Text) (Maybe AuthHookType)
 data RawServeOptions
   = RawServeOptions
   { rsoPort          :: !(Maybe Int)
+  , rsoHost          :: !(Maybe String)
```
0x777: Instead of `String` we can directly use `HostPreference` (http://hackage.haskell.org/package/warp-3.2.25/docs/Network-Wai-Handler-Warp.html#t:HostPreference)

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -55,6 +56,7 @@ type CorsConfig = CorsConfigG T.Text
 data ServeOptions
   = ServeOptions
   { soPort          :: !Int
+  , soHost          :: !String
```
0x777: Same as above, use `HostPreference`.

server/src-exec/Main.hs
```diff
@@ -129,8 +131,9 @@ main =  do
       putStrLn $ "server: running on port " ++ show port
       (app, cacheRef) <- mkWaiApp isoL loggerCtx pool httpManager
                          am corsCfg enableConsole
-      let warpSettings = Warp.setPort port Warp.defaultSettings
-                         -- Warp.setHost "*" Warp.defaultSettings
+
+      let hostS = DataString.fromString(host)
+      let warpSettings = Warp.setPort port $ Warp.setHost hostS Warp.defaultSettings
```
0x777: You won't need to do the conversion as `ServerOptions` will have the value of correct type. You can get rid of `Data.String` import.

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -492,6 +502,12 @@ parseServerPort = optional $
          help (snd servePortEnv)
        )
 
+parseServerHost :: Parser (Maybe String)
```
0x777: Change the type to `Parser (Maybe HostPreference)`. `HostPreference` is an instance of `IsString` type class so this should work without any other change.

nathanstitt:
Thanks for the review @0x777 - I've made the changes to store value as HostPreference as soon as it's read.  Note that I also had to extend the `FromEnv` class, which is hopefully correct.

0x777:
This looks good! Can you document the cli flag and the environment variable here: https://github.com/hasura/graphql-engine/blob/master/docs/graphql/manual/deployment/graphql-engine-flags/reference.rst

nathanstitt:
Oh neat! I had no idea those were documented.  I only found the port option by grepping the code 😁 

just pushed a update to docs, happy to reword if you'd like

Pomazan-Bogdan:
Very very good.
Just figured out that the lack of this option does not allow to run several dockers from hasura, because there is a conflict over the IP address due to the default 0.0.0.0

PR #1384, opened 2019-01-15
add recommendation note for URL usage ( fix #970 )

### LINKED ISSUE #970: console warnings for header values in migrations
Needed for event triggers and schema stitching:
The console should warn if you are using values directly instead of environment variables for webhooks and remote schema URLs. Because if you have multiple environments, the webhook URLs could be different.
This can just be a warning note between the title and the dropdown like this:

console warning

The text for note: 
a) event triggers:`Specifying the webhook URL via an environmental variable is recommended if you have different URLs for multiple environments.`
b) remote schemas: `Specifying the server URL via an environmental variable is recommended if you have different URLs for multiple environments.`

@shahidhk Will this work?

gopal-virtual:
### Description

- Add recommendation note to use URLs from environment variable for webhooks and remote schema creation
- Change the initial state of reducer to use environment variable as URL type.
- Respective files are prettified in the process

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

https://github.com/hasura/graphql-engine/issues/970

### Solution and Design

- Add a warning for the user to understand, that in the case of multiple environments it's safer to use URL from env files
- Due to our recommendation, the initial state for reducer also changes

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


coco98:
@arvi3411301 @karthikvt26 @praveenweb Any idea why the tests are failing? Seems like a text change PR that shouldn't be affecting anything.

console/src/components/Services/CustomResolver/state.js
```diff
@@ -14,8 +14,8 @@ const listState = {
 };
 
 const addState = {
-  manualUrl: '',
-  envName: null,
+  manualUrl: null,
```
karthikvt26: Could you please explain this change? and the motivation behind it?

console/src/components/Services/EventTrigger/Add/AddState.js
```diff
@@ -6,7 +6,7 @@ const defaultState = {
   operations: { insert: [], update: [], delete: [] },
   selectedOperations: { insert: false, update: false, delete: false },
   webhookURL: '',
-  webhookUrlType: 'url',
+  webhookUrlType: 'env',
```
karthikvt26: The default option is intentionally made to be `url` as it will serve most of the use cases unless user would like to enter something `secret`. Could you also explain what is the motivation behind this?

karthikvt26:
@coco98 this changes bunch of default states which are assumed to be there for the tests to be successful.

coco98:
I see. And changing the default states is not recommended because we want
the default to be url right? Not env var.

Correct?

karthikvt26:
Yes.

gopal-virtual:
@karthikvt26 @nathanstitt Got it. I changed it to use env var, because the bug related to this change recommends the user to use env var in case of multiple environments, so having the recommendation and default option set to env, made sense to me.

karthikvt26:
@gopal-virtual The issue points out a scenario to let user know of a **recommended** way for a very specific scenario (when user has multiple envs: dev/stg/prod. Configration would be easy in that scenario). Currently this may not be the recommended way to go for all the scenarios.

@dsandip your thoughts?

dsandip:
@karthikvt26 @gopal-virtual  Agree with Karthik - let's leave the default at URL instead of env var. URL is the easiest way to get started.

dsandip:
@karthikvt26 please review and close/give feedback.

karthikvt26:
@dsandip The text is of the lighter shade. Is it fine? 
2) Should we add the text in the modify section of the event triggers?

screenshot 660

dsandip:
@karthikvt26 
1) The lighter shade is fine. 
2) Yeah, we can add it the modify/edit sections too. But let's add it separately and merge these set of changes?

PR #1401, opened 2019-01-17
add anonymous telemetry

shahidhk:
The Hasura GraphQL Engine collects anonymous telemetry data that helps the
Hasura team in understanding how the product is being used and in deciding
what to focus on next.

The data collected is minimal and, since there is no *sign-in* associated with
the GraphQL Engine, it **cannot be used to uniquely identify any user**.
Furthermore, data collected is strictly statistical in nature and
**no proprietary information is collected** (*please see the next section*).

As a growing community, we greatly appreciate the telemetry data users
send to us, as it is very valuable in making GraphQL Engine a better product
for everyone. If you are worried about privacy, you can choose to disable
sending telemetry as described below.

   Access to collected data is strictly limited to the Hasura team and not shared with 3rd parties.

### Server

The server periodically sends the number of tables, views, relationships,
permission rules, event triggers and remote schemas tracked by GraphQL Engine,
along with randomly generated UUID per database and per instance. The
server version is also sent.

Here is a sample row from the telemetry database:

```json
   {
      "id": 12,
      "timestamp": "2019-01-21T19:43:33.63838+00:00",
      "db_uid": "dddff371-dab2-450f-9969-235bca66dab1",
      "instance_uid": "6799360d-a431-40c5-9f68-24592a9f07df",
      "version": "v1.0.0-alpha35",
      "metrics": {
        "views": 1,
        "tables": 2,
        "permissions": {
          "roles": 1,
          "delete": 2,
          "insert": 1,
          "select": 2,
          "update": 2
        },
        "relationships": {
          "auto": 2,
          "manual": 0
        },
        "event_triggers": 0,
        "remote_schemas": 1
      }
    }
```

### Console

The console is a React-Redux UI. Redux action names along with anonymized
route names are sent without any identifiable information or payload. Console
also records the UUID of the server/CLI that it is connected to.

Here is a sample:

```json
   {
     "id": 902,
     "timestamp": "2019-01-21T10:00:23.849202+00:00",
     "url": "/data/schema/SCHEMA_NAME/tables/TABLE_NAME/modify",
     "event_type": "ModifyTable/RESET",
     "console_mode": "server",
     "server_uuid": "79485a57-fca5-40f3-a31b-78c0d211314b",
     "server_version": "v1.0.0-alpha35",
     "cli_uuid": null
   }
```

### CLI

The CLI collects each execution event, along with a randomly generated UUID.
The execution event contains the command name, timestamp and whether the
execution resulted in an error or not. **Error messages, arguments and flags
are not recorded**. CLI also collects the server version and UUID that it
is talking to. The operating system platform and architecture is also
noted along with the CLI version.

Sample data:

```json
   {
     "id": 115,
     "timestamp": "2019-01-21T11:36:07.86783+00:00",
     "uuid": "e462ce20-42dd-40fd-9549-edfb92f80455",
     "execution_id": "ddfa9c33-0693-457d-9026-c7f456c43322",
     "version": "v0.4.27",
     "command": "hasura version",
     "is_error": false,
     "os_platform": "linux",
     "os_arch": "amd64",
     "server_uuid": "a4d66fb2-f88d-457b-8db1-ea7a0b57921d",
     "server_version": "v1.0.0-alpha35",
     "payload": null
   }
```

## Where is the data sent?

The data is sent to Hasura's servers addressed by `telemetry.hasura.io`.

## How do I turn off telemetry (opt-out)?

You can turn off telemetry on the server and on the console hosted by server
by setting the following environment variable on the server or by using
the flag `--enable-telemetry=false`:

```bash
   HASURA_GRAPHQL_ENABLE_TELEMETRY=false
```

In order to turn off telemetry on CLI and on the console served by CLI,
you can set the same environment varibale on the machine running CLI.
You can also set `"enable_telemetry": false` in the JSON file created
by the CLI at `~/.hasura/.config.json` to perisist the setting.


server/src-exec/Main.hs
```diff
@@ -141,6 +140,16 @@ main =  do
         mkGenericStrLog "event_triggers" "starting workers"
       void $ C.forkIO $ processEventQueue hloggerCtx logEnvHeaders httpSession pool cacheRef eventEngineCtx
 
+      -- start a background thread to check for updates
+      void $ C.forkIO $ checkForUpdates loggerCtx httpManager
+
+      -- start a background thread for telemetry
+      when enableTelemetry $ do
+        unLogger logger $ mkGenericStrLog "telemetry" telemetryNotice
+        res <- getUniqIds ci
```
0x777: We can merge `initialise` and `migrate` functions into a single function, say `initialise`. This should return the unique ids (database, instance). These unique ids then can be directly used here.

server/src-lib/Hasura/Server/Telemetry.hs
```diff
@@ -0,0 +1,226 @@
...
+
+    relsOfTbl :: TableInfo -> [RelInfo]
+    relsOfTbl = rights . Map.elems . Map.map fieldInfoToEither . tiFieldInfoMap
+
+    permsOfTbl :: TableInfo -> [(RoleName, RolePermInfo)]
+    permsOfTbl = Map.toList . tiRolePermInfoMap
+
+
+-- TODO: use a 3rd party library?
+generateFingerprint :: Q.TxE QErr Text
```
0x777: We already depend on uuid library. We need not hit the database for this

server/src-exec/Main.hs
```diff
@@ -184,19 +192,30 @@ main =  do
...
     prepareEvents (Logger logger) ci = do
       logger $ mkGenericStrLog "event_triggers" "preparing data"
       res <- runTx ci unlockAllEvents
       either printErrJExit return res
 
+    getUniqIds ci =
+      runTx ci $ do
+        dbId <- getDbId
+        fp   <- liftIO generateFingerprint
+        return (dbId, fp)
```
0x777:
```haskell
      dbid <- runTx ci getDbId
      fp   <- liftIO generateFingerprint
      return (dbId, fp)
```

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -293,7 +298,7 @@ serveCmdFooter =
     examplesDoc = mkExamplesDoc examples
     examples =
       [
-        [ "# Start GraphQL Engine on default port (8080) with console enabled"
+        [ "# Start GraphQL Engine on default port (8080) with console cnabled"
```
0x777: How do these changes happen? @ecthiender

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -316,13 +321,17 @@ serveCmdFooter =
         , "graphql-engine --database-url <database-url> serve --access-key <secretaccesskey>"
           <> " --auth-hook https://mywebhook.com/post --auth-hook-mode POST"
         ]
+      , [ "# Start GraphQL Engine with telemetry disabled"
+        , "graphql-engine --database-url <database-url> serve --disable-telemetry"
```
0x777: Help text still refers to `--disable-telemetry`

server/src-lib/Hasura/Server/Init.hs
ecthiender: @shahidhk

shahidhk:
@praveenweb @wawhal Can you check the console? Throws the error:

image

hrj:
:-1:  Not a fan of opt-out telemetry. 

Somewhere, some dev, who hasn't read all the pages of the documentation, is inadvertently going to fail to opt-out.

Thankfully, I see that there is `notice` text included in this PR. Can you please tell where is this notice visible? Is it visible in the command line / console whenever the telemetry setting is on?

coco98:
@hrj Yep. There are notices on the server logs, on the console and on the CLI. 
Disabling telemetry is setting up an env var on the server and a config var for the console. 
https://docs.hasura.io/1.0/graphql/manual/guides/telemetry.html#how-do-i-turn-off-telemetry-opt-out

PR #1406, opened 2019-01-18
optimise 'run_sql' query

### LINKED ISSUE #1362: optimise Run SQL query
Confine metadata check to queries of kind `DROP` and `ALTER`.
Add `enforce_state_check` optional field (default `False`) to `args` of `run_sql` query in `/v1/query` API to ignore SQL query kind and perform metadata check.

rakeshkky:
### Description

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

close #1362 

### Solution and Design

For `"type": "run_sql"` in `/v1/query`, check whether input `SQL` contains any `DDL` statement of type `DROP`, `ALTER` and `REPLACE`. If present, perform `schema diff` and apply those changes in HGE `metadata`. If not present, just execute and return response of `SQL` statement. 

Optional argument to enforce `schema diff` check:-
Introduce new optional field `check_metadata_consistency` (takes bool value, default `false`) in `args` section of payload. This enforces performing `schema diff` irrespective of `SQL` statement.

Ex:-

```http
POST /v1/query
{
       "type": "run_sql",
	"args": {
		"sql": "select id from geom_table",
		"check_metadata_consistency": true
	}
}
```

Regex used to match against `SQL` string:- 
`alter table|drop table|alter view|drop view|replace view`

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -440,19 +450,25 @@ runSqlP2 (RunSQL t cascade) = do
...
 
+isAltrOrDrop :: QErrM m => T.Text -> m Bool
+isAltrOrDrop = either throwErr return . matchRegex regex False
   where
-    rawSqlErrHandler :: Q.PGTxErr -> QErr
-    rawSqlErrHandler txe =
-      let e = err400 PostgresError "query execution failed"
-      in e {qeInternal = Just $ toJSON txe}
+    throwErr s = throw500 $ "compiling regex failed: " <> T.pack s
+    regex = "alter table|drop table|alter view|drop view|replace view"
```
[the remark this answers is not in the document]
rakeshkky: This regex will catch all SQL having `alter`, `drop` and `replace` words. We explicitly need `alter table`, `replace view` etc.

server/src-lib/Hasura/Server/Utils.hs
```diff
@@ -111,3 +113,16 @@ _2 (_, y, _) = y
 
 _3 :: (a, b, c) -> c
 _3 (_, _, z) = z
+
+-- regex related
+matchRegex :: String -> Bool -> T.Text -> Either String Bool
```
0x777: Text -> Bool -> Text -> Either String Bool

Use the Bytestring module of TDFA instead of String internally.

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -375,8 +376,9 @@ buildSchemaCache = do
 
 data RunSQL
   = RunSQL
-  { rSql     :: T.Text
-  , rCascade :: !(Maybe Bool)
+  { rSql               :: T.Text
+  , rCascade           :: !(Maybe Bool)
+  , rEnforceStateCheck :: !(Maybe Bool)
```
0x777: Let's call this `check_metadata_consistency`.

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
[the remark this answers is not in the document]
0x777: Our metadata can get inconsistent by dropping schemas/constraints too. Instead of explicitly listing all the objects it will be simpler to just use `alter|drop|replace`.

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -389,10 +391,18 @@ data RunSQLRes
...
 
+execRawSQL :: (MonadTx m) => T.Text -> m RunSQLRes
+execRawSQL =
+  liftTx . Q.multiQE rawSqlErrHandler . Q.fromText
+  where
+    rawSqlErrHandler txe =
+      let e = err400 PostgresError "query execution failed"
+      in e {qeInternal = Just $ toJSON txe}
+
 runSqlP2
```
0x777: Let's rename this to `execWithMetadataCheck`

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -440,19 +450,27 @@ runSqlP2 (RunSQL t cascade) = do
...
+    regex = "alter|drop|replace"
 
 runRunSQL
   :: (QErrM m, UserInfoM m, CacheRWM m, MonadTx m, MonadIO m, HasHttpManager m)
   => RunSQL -> m RespBody
-runRunSQL q =
-  adminOnly >> runSqlP2 q
+runRunSQL q@(RunSQL t _ mChkMDCnstcy) = do
+  adminOnly
+  encode <$> case mChkMDCnstcy of
```
0x777: can be simplified further:
```
  chkMdCnstncy <- case mChkMdCnstncy of
    Nothing -> isAltrDropReplace t
    Just b  -> return b
  encode <$> bool (execRawSQL t) (execWithCheck q) chkMdCnstncy
```

shahidhk:
Just checking in to make sure we're doing case insensitive match.

PR #1473, opened 2019-01-25
support column presets in update mutation and restrict altering type of preset columns (close #1449, fix #1464)

### LINKED ISSUE #1449: Support for column presets in UPDATES
From discussion on Discord:
If you want to automatically set a column to a session variable (such as the current X-Hasura-User-Id), you can do this on **INSERT** with the column presets functionality below
https://docs.hasura.io/1.0/graphql/manual/schema/default-values/column-presets.html

Is there a specific reason, why the same functionality is not supported on **UPDATE**? If not, could this feature be added?
My use case: User A creates record 1. User B later updates record 1. Ideally, there is a column who tracks who created and who last edited the record.

### LINKED ISSUE #1464: Meta-data in column presets not updating after modification of a column type
Minor bug, but thought I would share it nevertheless:

When you modify the column type in a table (via the console, but I assume the same occurs with raw SQL...) and you have already existing column presets, it seems that the meta-data of the presets is not updated to the new type (which then results in an error when a mutation is submitted) nor it stops you when you modify the column type and warns you that there are dependencies on this column type (as it would do in case you have permission checks dependent on that column).

rakeshkky:
### Description

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [x] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

close #1449 
fix #1464

### Solution and Design

1. Allow specifying predefined values for columns in update permissions.
2. Restrict altering type of preset column through `run_sql` (fixes issue #1464)
3. `args` field of `create_update_permission` query type has optional `set` field.
It takes `json` object and defaults to `empty json object ({})` 

Metadata API:-

```http
POST /v1/query
{
	"type": "create_update_permission",
	"args": {
		"table": "author",
		"role": "user-state",
		"permission": {
			"filter": {},
			"columns": "*",
			
			"set": {
				"state": "karnataka"
			}
		}
	}
}
```

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


console/src/components/Services/Data/TablePermissions/Permissions.js
```diff
@@ -54,20 +54,27 @@ class Permissions extends Component {
...
+    this.state.showUpdatePresets = false;
   }
   componentDidMount() {
     if (this.props.serverVersion) {
-      this.checkSemVer(this.props.serverVersion).then(() =>
-        this.checkPrefixVer(this.props.serverVersion)
-      );
+      this.checkSemVer(this.props.serverVersion).then(() => {
+        this.checkPrefixVer(this.props.serverVersion);
+        this.checkUpdatePresetVer(this.props.serverVersion);
```
karthikvt26: There could be a possibility where states gets overwritten. Both the functions are updating state. It would be nice if states are updated one after the other deterministically

console/src/components/Services/Data/TablePermissions/Permissions.js
```diff
@@ -96,13 +103,14 @@ class Permissions extends Component {
 
   componentWillReceiveProps(nextProps) {
     if (nextProps.serverVersion !== this.props.serverVersion) {
-      this.checkSemVer(nextProps.serverVersion).then(() =>
-        this.checkPrefixVer(nextProps.serverVersion)
-      );
+      this.checkSemVer(nextProps.serverVersion).then(() => {
+        this.checkPrefixVer(nextProps.serverVersion);
+        this.checkUpdatePresetVer(nextProps.serverVersion);
```
karthikvt26: Here too

console/src/components/Services/Data/TablePermissions/Permissions.js
```diff
@@ -236,39 +245,51 @@ class Permissions extends Component {
     }
     return Promise.resolve();
   }
-  deleteSetKeyVal(e) {
+  checkUpdatePresetVer(version) {
+    let showUpdatePresets = false;
+    try {
+      showUpdatePresets = semverCheck('permUpdatePresets', version);
```
karthikvt26: semverCheck returns either true or false. I think in the previous version of semverCheck which was throwing an exception when the version is not in correct format.

Is it possible to change this ?

server/src-lib/Hasura/RQL/DDL/Permission.hs
```diff
@@ -258,6 +265,7 @@ instance IsPerm SelPerm where
 data UpdPerm
   = UpdPerm
   { ucColumns :: !PermColSpec -- Allowed columns
+  , ucSet     :: !(Maybe Object) -- Preset columns
```
0x777: Why not `(Maybe (M.HashMap PGCol Value)`? You then can just use `forM` in `procSetObj`

PR #1536, opened 2019-01-30
add support for multiple domains in cors config (fix #1436)

### LINKED ISSUE #1436: Request multiple domains support for --cors-domain
Currently `--cors-domain` appears to support only a single domain.

Our application is served from various domains depending on the context, some of which are wildcard domains.

I would like to be able to provide a list of domains, potentially like this

`foo.example.com,bar.example.com,*.baz.example.com`

I wonder if this would be easiest to support as a regex?

For now I've just removed the cors config to allow all domains, since it's ignored for websockets anyway I'm not sure what benefit it really brings.

ecthiender:
### Description

Support for multiple domains (as CSV) in the `--cors-domain` flag and `HASURA_GRAPHQL_CORS_DOMAIN` env var.

Following are all valid configurations (must include scheme and optional port):
```shell
HASURA_GRAPHQL_CORS_DOMAIN="https://*.foo.bar.com:8080"
HASURA_GRAPHQL_CORS_DOMAIN="https://*.foo.bar.com, http://*.localhost, https://example.com"
HASURA_GRAPHQL_CORS_DOMAIN="*"
HASURA_GRAPHQL_CORS_DOMAIN="http://example.com, http://*.localhost, http://localhost:3000, https://*.foo.bar.com, https://foo.bar.com"
```

**Note**: top-level domains are not considered as part of wildcard domains. You have to add them separately. E.g - `https://*.foo.com` doesn't include `https://foo.com`.

The default (if the flag or env var is not specified) is `*`. Which means CORS headers are sent for all domains.

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

#1436 
### Type

- [ ] Docs update
- [ ] Community content

### Checklist:

### DIFF
--- a/server/src-lib/Hasura/Server/Cors.hs
+++ b/server/src-lib/Hasura/Server/Cors.hs
@@ -1,26 +1,34 @@
+{-# LANGUAGE DeriveAnyClass #-}
 -- | CORS (Cross Origin Resource Sharing) related configuration
 
 module Hasura.Server.Cors where
+
+import           Hasura.Prelude
+import           Hasura.Server.Utils  (fmapL)
+
+import           Control.Applicative  (optional)
 
 import qualified Data.Aeson           as J
 import qualified Data.Aeson.Casing    as J
 import qualified Data.Aeson.TH        as J
 import qualified Data.Attoparsec.Text as AT
-import qualified Data.ByteString      as B
 import qualified Data.HashSet         as Set
 import qualified Data.Text            as T
 
-import           Control.Applicative  (optional)
-import           Data.List            (partition)
 
-import           Hasura.Prelude
-import           Hasura.Server.Utils
+data DomainParts =
+  DomainParts
+  { wdScheme :: !Text
+  , wdHost   :: !Text -- the hostname part (without the *.)
+  , wdPort   :: !(Maybe Int)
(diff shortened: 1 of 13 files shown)


server/src-lib/Hasura/Server/Cors.hs
```diff
@@ -0,0 +1,154 @@
...
+
+isCorsDisabled :: CorsConfig -> Bool
+isCorsDisabled = \case
+  CCDisabled -> True
+  _          -> False
+
+readCorsDomains :: String -> Either String CorsConfig
+readCorsDomains str
+  | str == "*"               = pure CCAllowAll
+  | str == "" || str == ","  = Left "invalid domain"
```
0x777: This check is not exhaustive (what if the string is " " or ",,"). Why not let the domain parsing logic handle empty strings post the split on `,`?

ecthiender: Yes, this check is not necessary. It was remnant of older regex-based parsing logic. Removed it.

Issue #1456, opened 2019-01-24
behavior of explicit null vs not providing a value for a variable in update mutations


0x777:
We'll need to check this behavior. If this is the mutation:
```graphql
mutation (
  $objectId: uuid!
  $objectName: String!
  $objectNo: String
) {
  update_objects (
    where: {
      object_id: {
        _eq: $objectId
      }
    }
    _set: {
      name: $objectName
      object_no: $objectNo
    }
  ) {
        affected_rows
  }
}
```
The behavior should be different when `variableValues` are as follows:
```json
{
  "objectId": "7f916c99-9f06-46b5-9533-c5a9df862048",
  "objectName": "Example object",
  "objectNo": null
}
```
In this case, the `objectNo` field should be set to `NULL`.

```json
{
  "objectId": "7f916c99-9f06-46b5-9533-c5a9df862048",
  "objectName": "Example object"
}
```
In this case, `objectNo` shouldn't be updated.

Currently, the behavior is same in both the cases. From grgurev on discord.

marioishikawa:
What is the expected behavior in this case when variable objectId is null? I'm not sure but I think I've seen cases in which if the variable inside where clause is null, Hasura does an update without where, changing the entire set. Does it really happen?

larsblumberg:
@0x777 Thanks for creating issue, I'm confronting the exact same situation. Due to then nature of Hasura's current behavior, we need to write quiet some boilerplate code. Either having two mutations, one for each scenario, potentially having even more mutations if there are multiple variations of which variables are passed and which not.

I find the current behavior quiet unexpected from a SQL perspective where you can update just a subset of all columns. Sure that's also possible with Hasura today but requires defining multiple (Apollo) mutations on the client side.

Are you engineers at Hasura planning to address this issue?

hixus:
I also stumbled upon similar issue. Had created_at as optional field in mutation and originally though that just giving null would be interpreted as `use fields default value now()`. I think my original assumption was just wrong but related to explicit nulls there is also open discussion in [https://github.com/graphql/graphql-spec/issues/542]() and [https://github.com/graphql/graphql-spec/issues/476]()

lughino:
Is there an update regarding this issue?
At the moment, I haven't found any workaround other than changing the mutation on the fly in the client.

PR #1661, opened 2019-02-25
add a json path argument to query values inside json columns (close #1598)

### LINKED ISSUE #1598: GraphQL Query JSON?
It'd be really cool if we could declare a field as JSON/B and query it per usual. My reasons for thinking this is that sifting through 50 columns in a table for misc stuff—or having 10 tables all revolving around the same concept—seems messy and hard to navigate.

In contrast, dumping things that are related to a concept, but not big enough to warrant their own table, might clean a db up. For instance, a _User_ might have a _profile_ with some addresses, names, and such in it... A _User_ might also have a bunch of stuff related to _auth_ (e.g. tokens, providers, roles, etc.). If you break those out into their own columns on the _User_ table, you end up with 50 columns. If you break them out to their own tables, you end up stitching together all those tables with relationships.

Open to discussion/suggestions/critique, but the generic query syntax might look like this:

(https://www.db-fiddle.com/f/UsVwfCin7LFLZbAuW691G/2)

```gql
{
  user {
    profile {
      name
    }
  }
}
```

And `where` syntax might look like this:

```gql
{
  user {
    profile(where: {
      addresses: {
        _contains: {
          type: "billing"
        }
      }
    }) {
      name
    }
  }
}
```

So I suppose arrays could do `_contains` where the param is part of an object?

Feel free to poke holes in this. No rush. I'm slightly concerned what performance would be, and if perf is fine, then nailing the syntax might be tricky if there isn't some GraphQL spec for it already.

---

**Resources:**

- https://blog.hasura.io/postgres-json-and-jsonb-type-support-on-graphql-41f586e47536/ (can `where`; can't generic query)
- https://github.com/taion/graphql-type-json (not sure what this is exactly, but it has 200k+ weekly downloads so might be relevant 😅)

hgiasac:
### Description

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

#1598 
### Solution and Design

- Define GraphQL Argument `path` for `json`/`jsonb` type
- Support jsonPath parsing (WIP)
- Convert arguments to DML Select
- If there is the argument, convert to `SEOpApp "->"` operator

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:


server/src-lib/Hasura/SQL/DML.hs
```diff
@@ -209,6 +209,9 @@ incOp = SQLOp "+"
 mulOp :: SQLOp
 mulOp = SQLOp "*"
 
+jsonbGetOp :: SQLOp
+jsonbGetOp = SQLOp "->"
```
rakeshkky: Use jsonb operator `#>` to extract at specified path. You can take help of `SEArray` constructor of `SQLExp` data type in `Hasura/SQL/DML.hs` module to generate array expression on RHS.

server/src-lib/Hasura/RQL/DML/Select/Internal.hs
```diff
@@ -153,6 +153,25 @@ buildJsonObject pfx parAls arrRelCtx flds =
                      ANIField (fldAls, arrSel)
         in S.mkQIdenExp arrPfx fldAls
 
+    toSQLCol :: PGColInfo -> S.SQLExp
+    toSQLCol col = toJSONableExp (pgiType col) $
+                    S.mkQIdenExp (mkBaseTableAls pfx) $ pgiName col
+
+    toSQLColArgs :: PGColInfo -> AnnArgs -> S.SQLExp
+    toSQLColArgs col args
+      | isJSONType colTy || isJSONBType colTy =
```
rakeshkky: You don't need to check `isJSONType` or `isJSONBType` since it is guaranteed that`PGColArg` is applied to only `jsonb` and `json` column types from `GraphQL` schema generation.

hgiasac: I got it. However, would we leave that condition for future support for custom argument types? IMO

server/src-lib/Hasura/RQL/DML/Select/Types.hs
```diff
@@ -76,9 +76,11 @@ data ArrSel
   deriving (Show, Eq)
 
 type ArrSelFlds = Fields ArrSel
+type AnnArgs = Fields PGColValue
 
 data AnnFld
   = FCol !PGColInfo
+  | FColArg !PGColInfo !AnnArgs
```
rakeshkky: Let's merge `FCol` and `FColArg` and make it more generic. Something like `FCol !(Maybe ColOp)`. This conveys, we're selecting a `column` that `may` required an `operation` to be applied. 
```haskell
data ColOp
  = ColOp
  { _coOp :: !S.SQLOp -- operation to be applied
  , _coExp :: !S.SQLExp -- RHS expression
  }
```

hgiasac: Ok. I create new constructor because I am afraid of breaking changes with old code base.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
```diff
@@ -44,6 +44,15 @@ withSelSet selSet f =
     res <- f fld
     return (G.unName $ G.unAlias $ _fAlias fld, res)
 
+paramsToAnnArgs :: ArgsMap -> RS.AnnArgs
```
rakeshkky: We've to resolve `ArgsMap` to `ColOp` here. 

```haskell
argsToColOp 
  :: ArgsMap -> Maybe RS.ColOp
```
`lookup` for `path` field in `ArgsMap` if present then resolve `AnnGValue` to an array `SQLExp` else `Nothing`. I would suggest to split path value `(path: "<path-value>")` on `.` (`T.splitOn "."`).  Convert `Text` array to `SQLExp` array expression using `SEArray`.

server/src-lib/Hasura/RQL/DML/Select/Internal.hs
```diff
@@ -153,6 +153,25 @@ buildJsonObject pfx parAls arrRelCtx flds =
...
+          case (getJSONArgumentPath args) of
+            Nothing -> toSQLCol col
+            Just pArg -> toJSONableExp colTy $
+                          S.SEOpApp S.jsonbGetOp
+                            [ S.mkQIdenExp (mkBaseTableAls pfx) $ pgiName col
```
rakeshkky: The logic here would be pretty straight forward if we use `ColOp` kind of data type.

server/src-lib/Hasura/RQL/DML/Select/Types.hs
```diff
@@ -225,3 +227,16 @@ mergeArrNodes lNode rNode =
   where
     ArrNode lExtrs colMapping lBN = lNode
     ArrNode rExtrs _          rBN = rNode
+
+getJSONArgumentPath :: AnnArgs -> Maybe T.Text
```
rakeshkky: We may not require this function if we use `ColOp` kind of data type.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
```diff
@@ -56,7 +65,10 @@ fromSelSet f fldTy flds =
       _ -> do
         fldInfo <- getFldInfo fldTy fldName
         case fldInfo of
-          Left colInfo -> return $ RS.FCol colInfo
+          Left colInfo -> if (null $ _fArguments fld)
```
rakeshkky: We may not require `(null $ _fArguments fld)` check if we use `argsToColOp` kind of function mentioned in previous comment.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
hgiasac: So we only support simple JSON Path such as ```user.address.street```. I intend to support another JSON query spec such as array ```users[0].address```. Does it make sense? Or I just skip it.

server/src-lib/Hasura/RQL/DML/Select/Internal.hs
rakeshkky: We're generating `GraphQL` schema based on column types. However, if we use `ColOp` kind of data type I suggested, then we don't require any column type check. And, `toSQLCol` and `toSQLColArgs` functions will get merged. In this case, the `operation` type and the RHS `SQLExp` are decided at the time of resolving the query in `Resolve/Select.hs` module. Currently we're generating only one operator ie. `path` for `json` and `jsonb` types.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
rakeshkky: Nice idea! You should somehow resolve `Text` value to appropriate SQL array using `SEArray`.

rakeshkky: For example, `users[0].address` to `['users', 0, 'address']`

hgiasac:
@rakeshkky I update the code with JSONPath parser and test cases. Can you review it? Thanks

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
```diff
@@ -44,6 +45,27 @@ withSelSet selSet f =
...
+  Right jPaths -> return $ S.SEArray $ map elToColExp jPaths
+  where
+    elToColExp (Key k)   = S.SELit k
+    elToColExp (Index i) = S.SELit $ T.pack (show i)
+
+
+argsToColOp :: (MonadError QErr m) => ArgsMap -> m (Maybe RS.ColOp)
+argsToColOp args = case (Map.lookup "path" args) of
+  Nothing  -> return Nothing
+  Just val -> toOp val
```
rakeshkky: Use `mapM toOp $ Map.lookup "path" args` for precise code.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
```diff
@@ -44,6 +45,27 @@ withSelSet selSet f =
...
+  where
+    toJsonPathExp t = jsonPathToColExp t
+      >>= (return . Just . (RS.ColOp S.jsonbPathOp))
+    toOp = \case
+      AGScalar _ (Just (PGValText t)) -> toJsonPathExp t
+      AGScalar _ (Just (PGValVarchar t)) -> toJsonPathExp t
+      _ -> return Nothing
```
rakeshkky: We're leaving values other than `PGValText` and `PGValVarchar`. So, in case of values other than `String` kinds we have to `500 Internal error`. Use `throw500`. However, GraphQL schema validation makes sure it will be `String`.

I suggest you write a function in `GraphQL/Resolve/InputValue.hs` module
```haskell
onlyText
  :: (MonadError QErr m)
  => PGColValue -> m Text
onlyText = \case
  PGValText t -> return t
  _           -> throw500 "expecting text for asPGColText"
asPGColTextM
  :: (MonadError QErr m)
  => AnnGValue -> m (Maybe Text)
asPGColTextM val = do
  pgColValM <- fmap snd <$> asPGColValM val
  mapM onlyText pgColValM
```
use `onlyText` function in `asPGColText` to replace pattern matching code.

server/src-lib/Hasura/RQL/DML/Select/Internal.hs
```diff
@@ -153,6 +152,20 @@ buildJsonObject pfx parAls arrRelCtx flds =
                      ANIField (fldAls, arrSel)
         in S.mkQIdenExp arrPfx fldAls
 
+    toSQLCol :: PGColInfo -> S.SQLExp
+    toSQLCol col = toJSONableExp (pgiType col) $
+                    S.mkQIdenExp (mkBaseTableAls pfx) $ pgiName col
+
```
rakeshkky: Merge `toSQLCol` and `toSQLColArgs`.

```haskell
toSQLCol col colOpM =
  toJSONableExp (pgiType col) $ case colOpM of
      Nothing -> colNameExp
      Just (ColOp op cExp) -> S.mkSQLOpExp op colNameExp cExp
  where
     colNameExp = S.mkQIdenExp (mkBaseTableAls pfx) $ pgiName col
```
This code looks precise. `toJSONableExp` logic is not redudant.

Note:-
  Wherever you feel you're repeating a logic (piece of code), try to unify them.

server/src-lib/Hasura/RQL/DML/Select/Types.hs
rakeshkky: Is this function relevant?

rakeshkky:
@hgiasac please apply `stylish haskell` on files you updated.

server/src-lib/Data/Parser/JSONPath.hs
```diff
@@ -0,0 +1,51 @@
...
+  , JSONPathElement(..)
+  ) where
+
+import           Control.Applicative  ((<|>))
+import           Data.Attoparsec.Text
+import           Prelude
+
+import qualified Data.Text            as T
+
+data JSONPathElement
```
rakeshkky: Why new `JSONPathElement` type since it is already defined in `aeson` library?
I suggest you define a parser for the type present in `aeson`.

server/src-lib/Data/Parser/JSONPath.hs
```diff
@@ -0,0 +1,51 @@
...
+
+childPathElement :: Parser JSONPathElement
+childPathElement = skip (== '.') *> pathElement
+
+indexPathElement :: Parser JSONPathElement
+indexPathElement = do
+  skip (== '[')
+  content <- many1' digit <?> "Invalid array JSON path. Expected digits, got empty or valid characters"
+  skip (== ']')
+  return $ Index (read content)
```
rakeshkky: `read` is error prone. Use `readMaybe` instead.

server/src-lib/Data/Parser/JSONPath.hs
```diff
@@ -0,0 +1,51 @@
...
+  remain <- many' ( letter
+                <|> digit
+                <|> satisfy (`elem` ("-_$" :: String))
+                )
+  return $ Key $ T.pack (firstChar:remain)
+
+childPathElement :: Parser JSONPathElement
+childPathElement = skip (== '.') *> pathElement
+
+indexPathElement :: Parser JSONPathElement
```
rakeshkky: `indexPathElement` should always parse an `Index` (`Int`). 
So I suggest,
```haskell
indexParser :: Parser Int
```
We should construct `JSONPathElement` in the function where this is being called. 

Similarly for `pathElement` function. 
```haskell
keyParser :: Parser T.Text
```
Note:-
A function signature should convey its behaviour. Both `indexPathElement` and `pathElement` gives `Parser JSONPathElement` which is not explicit.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
```diff
@@ -44,6 +45,27 @@ withSelSet selSet f =
     res <- f fld
     return (G.unName $ G.unAlias $ _fAlias fld, res)
 
+jsonPathToColExp :: (MonadError QErr m) => T.Text -> m S.SQLExp
+jsonPathToColExp t = case parseJSONPath t of
+  Left s       -> throwVE $ T.pack s
```
rakeshkky: `throwVE` gives `validation-failed` exception. It is thrown only in case of GraphQL schema validation errors. Here it is parsing error. 

Use `throw400 ParseFailed $ "json path parse error: " <> T.pack s`

You're free to use any convenient error message other than above.

server/src-lib/Data/Parser/JSONPath.hs
hgiasac: Because the result can be multiple digit, the index will be larger than 10. `many1' digit ~ Parser String`, not Char

server/src-lib/Data/Parser/JSONPath.hs
hgiasac: My idea is avoiding dependency. I will change it

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
hgiasac: When I saw in `GraphQL/Resolve/InputValue.hs ` module, `asPGColText` did exist already. I think it can be reused.

https://github.com/hasura/graphql-engine/blob/f794653b69279f7b123353997e0428d00f9c7814/server/src-lib/Hasura/GraphQL/Resolve/InputValue.hs#L124

server/src-lib/Data/Parser/JSONPath.hs
rakeshkky: Oh! I see. But I'm assuming we're missing few corner cases here. If content is of type `[Char]` , what happens if it is empty array? This occurs when we trying to parse `name.addresses[]` (any `[]`).

hgiasac: `many1` will check empty for it, from the documentation. However, the error is not friendly, so I need to improve a bit

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
rakeshkky: `asPGColText` is used in case of `!String` graphQL type. Here, we're parsing `path: String`. Currently we're treating absence of a key and explicit `null` value to a key as two different things( refer https://github.com/hasura/graphql-engine/issues/1456 ). So, you'll need to use `asPGColTextM` in to handle `{path: null}`.

server/src-lib/Data/Parser/JSONPath.hs
rakeshkky: We don't encourage using pure code functions prone to errors. `read` functions throws `error` exception, though `digit` parser guarantees a `number`. I suggest you use `readMaybe` and throw parse error when it is `Nothing`

server/src-lib/Hasura/RQL/DML/Select/Types.hs
hgiasac: Sorry, I forgot

hgiasac:
@rakeshkky I update new code. Please check if you have time, thanks.

Attoparsec prefers performance, so some error messages are skipped. I work around with some manual code. Please leave some advice If you get any better improvement. Thanks

Nice weekend!

server/src-lib/Hasura/GraphQL/Resolve/InputValue.hs
```diff
@@ -121,11 +122,27 @@ parseMany fn v = case v of
...
+  PGValText t -> return t
+  PGValVarchar t -> return t
+  _           -> throw500 "expecting text for onlyText"
+
 asPGColText
   :: (MonadError QErr m)
   => AnnGValue -> m Text
 asPGColText val = do
   (_, pgColVal) <- asPGColVal val
   case pgColVal of
```
rakeshkky: You can do
```haskell
onlyText pgColVal
```
Redundant logic found in `asPGColVal` and `onlyText`. Use `onlyText` in `asPGColVal` to replace pattern match.

server/src-lib/Hasura/RQL/DML/Select/Internal.hs
```diff
@@ -153,6 +152,13 @@ buildJsonObject pfx parAls arrRelCtx strfyNum flds =
                      ANIField (fldAls, arrSel)
         in S.mkQIdenExp arrPfx fldAls
 
+    toSQLCol :: PGColInfo -> Maybe ColOp -> S.SQLExp
+    toSQLCol col colOpM = toJSONableExp strfyNum (pgiType col) $ case colOpM of
```
rakeshkky: Move code from `toJSONable...` to new line. In that way code looks pretty :smile:

server/src-lib/Hasura/RQL/DML/Select/Types.hs
```diff
@@ -76,9 +76,16 @@ data ArrSel
   deriving (Show, Eq)
 
 type ArrSelFlds = Fields ArrSel
+type AnnArgs = Fields PGColValue
```
rakeshkky: We can remove this.

server/src-lib/Hasura/SQL/Types.hs
```diff
@@ -384,6 +384,10 @@ isNumType PGDouble  = True
 isNumType PGNumeric = True
 isNumType ty        = isIntegerType ty
 
+isJSONType :: PGColType -> Bool
```
rakeshkky: We can remove this.

hgiasac:
@rakeshkky I removed unused code. If there aren't much things to change, could I move it to `ready for review` from `draft`? 
Another thing to do is updating documentation. Which section would I update? Currently I see JSON type is mentioned in API reference https://docs.hasura.io/1.0/graphql/manual/api-reference/postgresql-types.html#json

rakeshkky:
@hgiasac 
Regarding Docs:- 
- [Here](https://docs.hasura.io/1.0/graphql/manual/api-reference/query.html#simpleobject), you can update `Simple Object` to have `json-field` with `path` argument. Also update the example.
- [Here](https://docs.hasura.io/1.0/graphql/manual/queries/simple-object-queries.html), add one more example with json fields.

rakeshkky:
@hgiasac 
These are my observations with `parseJSONPath` function in `GHCI`
```shell
1. test> parseJSONPath "random"
Right [Key "random"]
2. test> parseJSONPath ".random"
Left "the first character of property name must be a letter, an underscore (_) or a dollar sign ($)."                                                
3. test> parseJSONPath "$random"
Right [Key "$random"]
4. test> parseJSONPath "random[0]"
Right [Key "random",Index 0]
5. test> parseJSONPath "$.random[0]"
Right [Key "$",Key "random",Index 0]
```
The 2nd, 3rd and 5th are giving unexpected results. 

Let's keep path syntax very simple.
- path can start with dot (`.`) or `char` or `[` . 
- thereafter each `key` should start with `.` and index should enclosed in `[]`

Examples:-
1. `[12]` -> `[Index 12]`
2. `.name` -> `[Key "name"]`
3. `name` -> `[Key "name"]`
4. `name.projects[3]` -> `[Key "name", Key "projects", Index 3]`
5. `[0].name` -> `[Index 0, Key "name"]`

Code must be precise and readable. 

Also document `jsonpath` spec with examples [here](https://docs.hasura.io/1.0/graphql/manual/api-reference/query.html#object) in a `Note`

rakeshkky:
I would suggest you to define 3 core functions.
1. parsing first element `firstElement :: Parser JSONPathElement`.
2. parsing key `parseKey :: Parser Text` (should always start with `.`)
3. parsing index `parseIndex :: Parser Int` (should always enclosed in `[]`)

Use type alias `JSONPath` for `[JSONPathElement]`.

Test the function in `stack repl` with all inputs.

hgiasac:
@rakeshkky Thanks for your advice. So we won't parse `$` character, just dash `-` and underscore `_`, right?

server/tests-py/queries/graphql_query/basic/select_query_test_types.yaml
```diff
@@ -93,7 +104,11 @@ query:
         c30_inet
         c31_macaddr
         c32_json
+        c32_json_child_prop: c32_json(path: "obj.c1")
+        c32_json_array_item: c32_json(path: "arr[0]")
         c33_jsonb
+        c33_jsonb_child_prop: c33_jsonb(path: "obj.e1")
```
rakeshkky: use `.obj.e1` to cover a case when path starting with `.`

rakeshkky:
> @rakeshkky Thanks for your advice. So we won't parse `$` character, just dash `-` and underscore `_`, right?

I think we shouldn't take special care of other characters. while parsing `firstElement` parse until a `.` or a `[`. Only dot (`.`) and `[]` are important to us. 

Also a small improvisation. Parse `.[<index>]` to `Index <index>`. 

Examples:-
1. `.[2]` -> `[Index 2]`
2. `.key.[2]` -> `[Key "key", Index 2]`

Hope you got this.

coco98:
@shahidhk @hgiasac @rakeshkky I would like to see the docs review app so that I can try this out with heroku PR app.

shahidhk:
@coco98 This PR was opened before Netlify hooks were setup. So, it won't generate a preview deploy.

@hgiasac There are some console changes in this PR. Can you merge latest master and squash all the commits?

coco98:
Can't we force it to happen? Maybe when it moves out of draft PR considering the heroku app is not here? If that's the case case then we should close this PR and then open a new one?

@shahidhk Please advise on what we should do.

hgiasac:
I am merging and rebuild to check if there is any error. I will push new commit about 10 minutes later

shahidhk:
@coco98 looks like preview app is now deployed. https://deploy-preview-1661--hasura-docs.netlify.com/graphql/manual/queries/simple-object-queries.html#fetch-json-fields

shahidhk:
@arvi3411301 Can you check why Heroku preview app was never created for this PR?

arvi3411301:
@shahidhk Review apps will be deployed only when the circleci tests were run under hasura org.

shahidhk:
@arvi3411301, what does that mean? This PR's tests are running on a different org?

shahidhk:
@hgiasac Can you update master and fix the conflicts? Thanks. This is ready to be merged. @dsandip has made some docs updates.

arvi3411301:
@shahidhk yes. https://circleci.com/gh/hgiasac/graphql-engine/219?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

shahidhk:
@hgiasac Can you disable CircleCI on your repo?

hgiasac:
Done. Sorry about that

shahidhk:
Thanks! :+1:

PR #1664, opened 2019-02-25
forward response headers from remote servers (fix #1654)

### LINKED ISSUE #1654: Hasura doesn't forward Set-Cookie headers set by remote schemas
Hasura should forward `Set-Cookie` headers back to the client that are set in Remote Schemas, so that the consumer of the GraphQL API receives them.

* In order to use cookie-based authentication with Hasura, we need to make sure `Cookie:` headers are successfully passed to the GraphQL endpoint, so our auth webhook can extract useful information and pass in `X-Hasura-*` headers back to Hasura.
* One of the primary reasons to use cookies is the `HttpOnly` flag, which prevents against vendor javascript running on the same page from stealing our sessions.
* If `Set-Cookie:` headers don't come from the same origin as Hasura, something like `credentials: 'include'` is necessary for cookies to actually be included in GraphQL requests, at least over HTTP (Queries and Mutations).
* Unfortunately, WebSockets don't seem to have any ability to `credentials: 'include'` and simply will not pass cookies from different origins no matter what, which means that there is *currently no way to use http-only cookie-based authentication and subscriptions at the same time with Hasura*.
* Manually adding the `Cookie:` header to the WebSocket client could be an option, but we lose the ability to use `HttpOnly` cookies as client-side javascript would be required to parse them to build this header.
* This also allows a cleaner front-end coding experience; all calls can be done via GraphQL.

ecthiender:
### Description

Forward response headers received from remote schema servers, to client.

**Note**: This won't send back any remote response headers when the websocket transport is used.

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

#1654
### Solution and Design

Introduced new data type `HResponse` to wrap hasura specific response. Changed `mkSpockAction` to take a handler of signature `HResponse`. Changed corresponding handlers. 

It filters the following headers from the remote: 
```
[ "Server"
, "Transfer-Encoding"
, "Cache-Control"
, "Access-Control-Allow-Credentials"
, "Access-Control-Allow-Methods"
, "Access-Control-Allow-Origin"
, "Content-Type"
]
```

### Type

- [ ] Docs update
- [ ] Community content

### Checklist:

### DIFF
--- a/server/src-lib/Hasura/Server/Utils.hs
+++ b/server/src-lib/Hasura/Server/Utils.hs
@@ -163,4 +163,4 @@ filterResponseHeaders = filterHeaders respHeaders
                   ]
 
 filterHeaders :: Set.HashSet HTTP.HeaderName -> [HTTP.Header] -> [HTTP.Header]
-filterHeaders list = filter (\(n, _) -> n `Set.member` list)
+filterHeaders list = filter (\(n, _) -> not $ n `Set.member` list)
--- a/server/src-lib/Hasura/Server/Utils.hs
+++ b/server/src-lib/Hasura/Server/Utils.hs
@@ -14,6 +14,7 @@ import qualified Data.Text.Encoding           as TE
 import qualified Data.Text.Encoding.Error     as TE
 import qualified Data.Text.IO                 as TI
 import qualified Language.Haskell.TH.Syntax   as TH
+import qualified Network.HTTP.Types           as HTTP
 import qualified Text.Ginger                  as TG
 import qualified Text.Regex.TDFA              as TDFA
 import qualified Text.Regex.TDFA.ByteString   as TDFA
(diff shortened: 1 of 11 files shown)


server/src-lib/Hasura/Server/Utils.hs
```diff
@@ -134,3 +135,29 @@ matchRegex regex caseSensitive src =
...
+  where
+    respHeaders = [ "Server", "Transfer-Encoding", "Cache-Control"
+                  , "Access-Control-Allow-Credentials"
+                  , "Access-Control-Allow-Methods"
+                  , "Access-Control-Allow-Origin"
+                  , "Content-Type"
+                  ]
+
+filterHeaders :: [HTTP.HeaderName] -> [HTTP.Header] -> [HTTP.Header]
+filterHeaders list = filter (\(n, _) -> n `notElem` list)
```
0x777: Use a set instead of looking up in the list

PR #1690, opened 2019-03-04
remove localStorage usage with auth0

praveenweb:
### Description
Auth0 best practices recommend to not store tokens in localStorage. [Reference](https://auth0.com/docs/security/store-tokens#don-t-store-tokens-in-local-storage)

Updated code to make use of [checkSession](https://auth0.com/docs/libraries/auth0js/v9#using-checksession-to-acquire-new-tokens) which makes use of cookies.

### Affected components 

- Community Content

### DIFF
--- a/community/sample-apps/react-apollo-todo/src/apollo.js
+++ b/community/sample-apps/react-apollo-todo/src/apollo.js
@@ -5,28 +5,20 @@ import { WebSocketLink } from "apollo-link-ws";
 import { split } from "apollo-link";
 import { getMainDefinition } from "apollo-utilities";
 import { SubscriptionClient } from "subscriptions-transport-ws";
-import { setContext } from "apollo-link-context";
 
 import { GRAPHQL_URL, REALTIME_GRAPHQL_URL } from "./utils/constants";
 import auth from "./components/Auth/Auth";
 
 const getHeaders = () => {
-  const headers = {
-    authorization: `Bearer ${auth.getIdToken()}`
-  };
+  const headers = {};
+  const token = auth.getIdToken();
+  if (token) {
+    headers.authorization = `Bearer ${token}`;
+  }
   return headers;
 };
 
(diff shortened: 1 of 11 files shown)


coco98:
@praveenweb I will be reviewing this.

community/sample-apps/react-apollo-todo/src/apollo.js
```diff
@@ -8,41 +8,44 @@ import { SubscriptionClient } from "subscriptions-transport-ws";
 import { setContext } from "apollo-link-context";
 
 import { GRAPHQL_URL, REALTIME_GRAPHQL_URL } from "./utils/constants";
+import auth from "./components/Auth/Auth";
 
 const getHeaders = () => {
-  const token = localStorage.getItem("auth0:id_token");
   const headers = {
-    authorization: token ? `Bearer ${token}` : ""
+    authorization: `Bearer ${auth.getIdToken()}`
```
coco98: You changed the ternary check to not return an empty string. Will this be ok if there's no auth header?
Or shouldn't this be:

```javascript
const headers = {};
const token = auth.getIdToken();
if (token) {
headers.authorization = `Bearer ${token}`;
}
return token;
```

community/sample-apps/react-apollo-todo/src/apollo.js
```diff
@@ -8,41 +8,44 @@ import { SubscriptionClient } from "subscriptions-transport-ws";
...
   const authLink = setContext((_, { headers }) => {
-    const token = localStorage.getItem("auth0:id_token");
     return {
       headers: {
         ...headers,
-        authorization: token ? `Bearer ${token}` : ""
+        authorization: `Bearer ${auth.getIdToken()}`
       }
     };
   });
```
coco98: My feeling is that `makeApolloClient` should probably be removed entirely?

PR #1716, opened 2019-03-07
Improve error log when insert check constraint fails (close #1649)

### LINKED ISSUE #1649: provide more information in the logs when insert check constraint fails
Currently we throw this error message: `Check constraint violation. insert check constraint failed` which only hints at the problem. The logs should output the values of the row for which the constraint failed and the insert constraint as defined. This information along with the session variables should help debug the problem.

nizar-m:
### Affected components 

- Server

### Related Issues

#1649 

### Solution and Design

- On check constraint violation, the trigger on the insert view returns a json as message;
- This contains the inserted row and check_constraint, which is parsed and shown as an internal error message
- When EnvVar `HASURA_GRAPHQL_VERBOSE_API_RESPONSE=true` is set (or with flag `--verbose-api-response`) this extra information is send as part of the API response

### Steps to test and verify

- Run graphql-engine with flag `--verbose-api-response`
- Create table foo and insert permissions for the user
```
echo '
type: bulk
args:
- type: run_sql
  args:
    sql: |
      create table foo(id integer primary key);
- type: track_table
  args:
    schema: public
    name: foo
- type: create_insert_permission
  args:
    table: foo
    role: user
    permission:
      check:
        id: X-Hasura-User-Id
' | yaml2json - | curl -d @- http://localhost:8080/v1/query
```
- Try to insert with id != X-Hasura-User-Id
```
echo '
query: |
  mutation foo {
    insert_foo (
      objects: [ { id: 3 } ],
    ) {
      affected_rows
    }
  }
' | yaml2json - | curl -d @- http://localhost:8080/v1alpha1/graphql \
-H 'X-Hasura-Role: user' -H 'X-Hasura-User-Id: 5'
```
- The error message would now show the input row and check constraint
```
{
  "errors": [
    {
      "extensions": {
        "internal": {
          "input_row": {
            "id": 3
          },
          "check_constraint": "((((NEW.\"id\") = (((current_setting('hasura.user')::json->>'x-hasura-user-id'))::integer)) OR (((NEW.\"id\") IS NULL) AND ((((current_setting('hasura.user')::json->>'x-hasura-user-id'))::integer) IS NULL))) AND ('true')) AND ('true')"
        },
        "path": "$.selectionSet.insert_foo.args.objects",
        "code": "permission-error"
      },
      "message": "Insert permission check failed"
    }
  ]
}
```


server/src-lib/Hasura/RQL/DDL/Permission/Triggers.hs
```diff
@@ -44,4 +44,5 @@ buildInsTrigFn fn tn be = do
     tmpltVals = J.object [ "function_name" J..= toSQLTxt fn
                          , "table_name" J..= toSQLTxt tn
                          , "check_expression" J..= toSQLTxt be
+                         , "check_expression_literal" J..= pgFmtLit (toSQLTxt be)
```
rakeshkky: Do not use `pgFmtLit` or `pgFmtIden`, except in modules where it is defined. 
Here you can make use of `SELit` constructor.

nizar-m: Done

shahidhk:
@nizar-m Please add ok-to-merge label or otherwise.

shahidhk:
@nizar-m Are we planning to get this merged as another PR?

Wasted a lot of time recently on this.

PR #1744, opened 2019-03-12
add a json path argument to query values inside json columns - hgiasac (close #1598)

### LINKED ISSUE #1598: GraphQL Query JSON?

dsandip:
### Affected components 

- Server
- Console
- CLI
- Docs
- Community Content
- Build System
- Tests
- Other (list it)


coco98:
@dsandip I'm not sure if the docs are clear. It's also within the first docs section of simple queries, which makes it more confusing. As a new user looking for how to query stuff, this would be very surprising, it seems like this ought to be more "advanced postgres user" documentation?

Perhaps we should have a separate page for this, or maybe we should include it in the reference documentation? Maybe we should have a page for the different GraphQL types we generate.

Let's also update the JSON blogpost? https://blog.hasura.io/postgres-json-and-jsonb-type-support-on-graphql-41f586e47536/

shahidhk:
@dsandip The commits in this PR are out of sync with master - i.e. some console changes are included. Need to figure out a way to fix that.

shahidhk:
Also, why is this a new PR? What happened to the old one?

shahidhk:
closing in favour of https://github.com/hasura/graphql-engine/pull/1661

PR #1758, opened 2019-03-13
invoke event triggers manually on rows aka Manual Triggers (close #1687)

### LINKED ISSUE #1687: Console Triggers
Hasura already has a feature to browse your data, as well as a way to Trigger webhooks when CRUD operations are fired. The idea is to combine the two features to support building CMS-y like actions (flag, ban, activate, view, etc) that a user can run from the console, for a specific table/row.

V1 could be as simple as marking an event trigger's operation as "console".
--- 
image

... and then adding a drop-down that fires the event, on the 'browse rows' screen.
---

image

There's a lot of bells and whistles that could be added, but the above is probably a good MVP.

tirumaraiselvan:
### Affected components 

- Server
- Console
- Docs

### Related Issues

#1687 

### Solution and Design

- Create a new v1 query type: invoke_event_trigger
- Create a function which takes row data and related metadata and inserts into event_log.
- Use above function to take params from invoke_event_trigger

### DIFF
--- a/server/src-lib/Hasura/RQL/DDL/Subscribe.hs
+++ b/server/src-lib/Hasura/RQL/DDL/Subscribe.hs
@@ -83,11 +83,13 @@ getTriggerSql op trid trn qt allCols strfyNum spec =
         INSERT -> S.SEUnsafe "NULL"
         UPDATE -> getRowExpression OLD scs
         DELETE -> getRowExpression OLD scs
+        MANUAL -> S.SEUnsafe "NULL"
     renderNewDataExp op2 scs =
       case op2 of
         INSERT -> getRowExpression NEW scs
         UPDATE -> getRowExpression NEW scs
         DELETE -> S.SEUnsafe "NULL"
+        MANUAL -> S.SEUnsafe "NULL"
     getRowExpression opVar scs =
       case scs of
         SubCStar -> applyRowToJson $ S.SEUnsafe $ opToTxt opVar
--- a/server/src-lib/Hasura/RQL/DDL/Subscribe.hs
+++ b/server/src-lib/Hasura/RQL/DDL/Subscribe.hs
@@ -3,8 +3,9 @@ module Hasura.RQL.DDL.Subscribe
   , runCreateEventTriggerQuery
   , DeleteEventTriggerQuery
   , runDeleteEventTriggerQuery
-  , DeliverEventQuery
-  , runDeliverEvent
+  , RedeliverEventQuery
+  , runRedeliverEvent
+  , runInvokeEventTrigger
 
   -- TODO: review
   , delEventTriggerFromCatalog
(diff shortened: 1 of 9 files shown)


server/src-rsr/migrate_from_11_to_12.sql
```diff
@@ -0,0 +1,29 @@
+CREATE OR REPLACE FUNCTION hdb_catalog.insert_event_log(schema_name text, table_name text, trigger_name text, trigger_id text, op text, row_data json) RETURNS void AS $$
```
tirumaraiselvan: should we return affected rows?

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -208,7 +209,8 @@ queryNeedsReload qi = case qi of
 
   RQCreateEventTrigger _       -> True
   RQDeleteEventTrigger _       -> True
-  RQDeliverEvent _             -> False
+  RQRedeliverEvent _           -> False
```
tirumaraiselvan: do we need to support older api also?

server/src-rsr/migrate_from_11_to_12.sql
tirumaraiselvan: we should return event id actually so that the frontend can poll the response immediately.

server/src-lib/Hasura/RQL/Types/Subscribe.hs
```diff
@@ -126,6 +128,7 @@ instance FromJSON CreateEventTriggerQuery where
     insert         <- o .:? "insert"
     update         <- o .:? "update"
     delete         <- o .:? "delete"
+    manual         <- o .:? "manual" .!= True
```
tirumaraiselvan: manual triggers are enabled by default

server/src-lib/Hasura/RQL/DDL/Subscribe.hs
```diff
@@ -213,7 +216,7 @@ markForDelivery eid =
           |] (Identity eid) True
 
 subTableP1 :: (UserInfoM m, QErrM m, CacheRM m) => CreateEventTriggerQuery -> m (QualifiedTable, Bool, EventTriggerConf)
-subTableP1 (CreateEventTriggerQuery name qt insert update delete retryConf webhook webhookFromEnv mheaders replace) = do
+subTableP1 (CreateEventTriggerQuery name qt insert update delete _manual retryConf webhook webhookFromEnv mheaders replace) = do
```
tirumaraiselvan: when only manual triggers are present (no insert, update, delete) then `{"path":"$.args","error":"no trigger sql generated","code":"unexpected"}`

shahidhk:
@tirumaraiselvan Can you close this PR and open a Draft PR?

tirumaraiselvan:
Opened draft PR here: https://github.com/hasura/graphql-engine/pull/1872/files

PR #1770, opened 2019-03-14
add jwt analyzer on console (close #1369)

### LINKED ISSUE #1369: Add jwt inspector to hasura console
It would be super nice if we had a little JWT inspector within the console and also a test endpoint to test if the JWT was being verified and accepted by Hasura (to verify hasura configuration). 
Maybe a little inspect icon next to the JWT authorization header if the console detects it?

@rikinsk @karthikvt26 @praveenweb @wawhal thoughts?

karthikvt26:
Fix #1369 

### Description

JWT header as below can be analyzed
screenshot 775

### Affected components 

- Console

### Related Issues

close #1369 

### Steps to test and verify

Create a jwt token and add a header `Authorization: Bearer <token` in the explorer and click on the `+` icon to analyze.


console/src/components/ApiExplorer/Actions.js
```diff
@@ -186,15 +187,39 @@ const analyzeFetcher = (url, headers, analyzeApiChange) => {
...
-const changeRequestHeader = (index, key, newValue, isDisabled) => ({
-  type: REQUEST_HEADER_CHANGED,
-  data: {
-    index: index,
-    keyName: key,
-    newValue: newValue,
-    isDisabled: isDisabled,
-  },
-});
+const setInitialHeaderState = headerObj => dispatch => {
```
wawhal:
```js
const setInitialHeaderState = headerObj => ({
  type: SET_INITIAL_HEADER_DATA,
  data: headerObj
})
```

console/src/components/ApiExplorer/ApiRequest.js
```diff
@@ -1,6 +1,14 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 
+import jwt from 'jsonwebtoken';
+
+import Modal from 'react-bootstrap/lib/Modal';
```
wawhal: Let us implement a common modal at the root at the app that subscribes to title, content and `isOpen` in redux. Then we can change these properties from anywhere and the modal becomes reusable.

karthikvt26: Hey I am not able to wrap my head around this pattern where `CommonModal` can be used as 

`this.props.dispatch({ type: COMMON_MODAL_OPEN, title: <react-element>, content: <react-element>})`

instead of 
```
<CommonModal title={ title }>
    <Body>
</CommonModal>
```

The body can be as complicated as it can. @wawhal Let me know what you think.

coco98:
I think it would be important to highlight a few things in the JWT analyser to make this useful:

1. Show me if Hasura has been setup to verify this signature. If hasura has not been setup then prompt me to setup Hasura and link to the docs so that I can.
2. Highlight the Hasura claims (once hasura jwt config is setup you'll know which claims are the right claims to look at). Further tell me if Hasura specific claims are missing. `allowed-roles` and `default-role` are mandatory.

Without these 2 things this is not useful. Also, the `+` icon should be an `i`(information) or `?`(help) icon right?
@rikinsk

coco98:
Also, set tabwidth to 2 spaces? It looks like it's 8 right now.
Screenshot 2019-03-18 at 11 28 37 PM

karthikvt26:
It is possible for me to achieve both. But not in a straight forward way. I have mentioned them below.

> Show me if Hasura has been setup to verify this signature. If hasura has not been setup then prompt me to setup Hasura and link to the docs so that I can.

I will make `introspection` query without the `Authorization` header and parse the response to come to the conclusion.

> Highlight the Hasura claims (once hasura jwt config is setup you'll know which claims are the right claims to look at). Further tell me if Hasura specific claims are missing. allowed-roles and default-role are mandatory.

I  can parse the payload to find out `x-hasura-*` inside the object to get the relevant header. I can highlight and show it to the user.

The above issues will become straight forward if for 1, the server hydrates a flag `isJWTSet` it will be useful. For the second scenario, if it can provide the `claims_namespace` then it will become straight forward to implement it. 

@coco98 @rikinsk 
Let me know what you think?

coco98:
Why can’t the server template the entire jwt config for the console?

That way you have everything.

karthikvt26:
I think the problem is with the authentication. Server will not know
whether the user is authenticated or not on page load before the user is
asked to login.

The jwt stuff should be there in env only for authenticated requests.

@wawhal thoughts?

ecthiender:
The server can template `is_jwt_set` and `claims_namespace`. Console does
not need the exact key anyway.

coco98:
But having key is good so that console can tell you if the jwt is valid per
Hasura. Otherwise capturing feedback about invalid jwt key only happens via
logs :(

ecthiender:
> But having key is good so that console can tell you if the jwt is valid per Hasura. Otherwise capturing feedback about invalid jwt key only happens via logs :(

For invalid jwt, the response from server has `invalid-jwt` as a code with a reason message.

But in any case, the key can't be templated on the console html, because then anyone can grab the jwt secret. Server can provide a separate admin only endpoint to dump the config. Does that make sense @coco98 ?

ecthiender:
According to the latest discussion with @dsandip and @rikinsk we are doing the following:
cc @karthikvt26 

#### Server

1. Server will not dump the entire jwt config to the console. It will have an endpoint which will return `is_jwt_set`, `claims_namespace` and `claims_format`. (should this endpoint be available only on dev-mode?)

2. Server will need to improve current error messages when JWT is invalid. And also return more details in the reponse, when the verbose-api response flag is enabled.

#### Console
1. Console need not replicate the entire logic for verifying the JWT (because the jwt config can also contain jwk url not just the key, the verification becomes more complex). It can hit the server  to verify the JWT and display the error messages returned by the server.

2. Console needs to use the `claims_namespace` and `claims_format` to highlight the hasura claims correctly.

Please add if I missed anything.

shahidhk:
closed in favour of https://github.com/hasura/graphql-engine/pull/1925

PR #1776, opened 2019-03-14
document unique/primary key constraint requirement for mutation returning and insert input object

rakeshkky:
### Description

Update docs with following info:-
- Querying relationships in `returning` field of mutation response and inserting relational objects are allowed only if atleast a primary key or unique constraint defined on the table. 
- Otherwise, user can query only `columns` in `returning` field and can insert only `columns`.

This is follow up to https://github.com/hasura/graphql-engine/pull/1703

### Affected components 

- Docs


rikinsk:
I don't agree with the changes made to the auto generated schema sections. For a user who has simply come to learn how to insert / delete, etc. its too much unnecessary and complicated information. Just an addition to the notes section under ir should suffice.

dsandip:
@rikinsk made the changes you requested. please to review again.

PR #1791, opened 2019-03-15
query validation now throws error over websocket  (fix #1790)

### LINKED ISSUE #1790: connection error is thrown over websocket instead of query error
during query validation, in some cases `connection_error` is thrown instead of only `error`.

ecthiender:
### Description

Over websocket transport, in some cases of query validation errors, `connection_error` was thrown (which is for protocol errors) instead of `error`.

### Affected components 

- Server

### Related Issues

#1790 
### Solution and Design

Simple fix to use the correct function to send `GQL_ERROR` and not `GQL_CONNECTION_ERROR`.


ecthiender:
CircleCI not building

Issue #1452, opened 2019-01-24
Allow arrays as data for access control


danielcompton:
In the application I'm building, each database row has a field which indicates which group can view it. Each user can be a member of one or more groups simultaneously.

It would be very handy to be able to provide a JSON array of groups that a user is part of in an `X-Hasura-Groups` header/JWT claim. I could then add a custom permissions check (perhaps with `_in`?) to be able to say that the group for a row has to be one of the groups in the user's authentication headers.

It is possible to also express this by setting up relationships, but this feels like it could get quite repetitive over time.

revskill10:
Hi @danielcompton , flexible access control is hard to be done right. I have two related issues for this problem as well.
https://github.com/hasura/graphql-engine/issues/1342 --> This one allows for flexible returning columns .
https://github.com/hasura/graphql-engine/issues/877 --> This one is the same with your issue as well to allow array of roles to be accepted.

coco98:
Closing this in favour of #1333.

Issue #1333, opened 2019-01-07
support user info variables as lists to be used in permissions


ecthiender:
Right now the user info variables (as headers, from JWT claims or from webhook response) are converted into map of strings to strings. Support should be added for lists as well. Then in the permission system `_in` can be used with session variables.

See #1314

lanex-victor:
Definitely looking forward for this to be added. Hope sooner..

belgiancoder:
For string columns you can send `"X-Hasura-Variable"`:`(${IdArray.join("|")})`
Then restrict
```
{
    id: _similar: "X-Hasura-Variable"
}
```
It is a REGEX-like. `id1` will match `(id1|id2|id3)` but not `id4`
Only works for strings tho.

0x777:
added in https://github.com/hasura/graphql-engine/commit/f1cf6d0b172e99e439c0a4d5157293422f4a4e59. See #2475

PR #1799, opened 2019-03-18
Support JSON arrays in session variables

nizar-m:
### Description

Setting the permission
```
echo '
type: create_insert_permission
args:
  role: user
  permission:
    set: {}
    check:
      id:
        _in: X-Hasura-Allowed-Ids
    columns:
    - id
  table:
    schema: public
    name: foo
' | yaml2json - | curl  -H 'content-type: application/json' -d @- 'http://localhost:8080/v1/query'
```
The following query will fail (because id: `28` is NOT in  `X-Hasura-Allowed-Ids: [21,23,25,27]`)
```
echo '
query: |
  mutation f {
    insert_food(objects: [{id: 28}]) {
      affected_rows
    }
  }' | yaml2json - | curl -H 'x-hasura-allowed-ids: [21,23,25,27]' -H 'x-hasura-role: user'   -H 'content-type: application/json' -d @- 'http://localhost:8080/v1alpha1/graphql'  
``` 
with the following response
```
{
  "errors": [
    {
      "extensions": {
        "path": "$.selectionSet.insert_food.args.objects",
        "code": "permission-error"
      },
      "message": "Check constraint violation. insert check constraint failed"
    }
  ]
}
```
while the following query gets through (as id: `23` is in  `X-Hasura-Allowed-Ids: [21,23,25,27]`)
```
echo '
query: |
  mutation f {
    insert_food(objects: [{id: 23}]) {
      affected_rows
    }
  }' | yaml2json - | curl -H 'x-hasura-allowed-ids: [21,23,25,27]' -H 'x-hasura-role: user'   -H 'content-type: application/json' -d @- 'http://localhost:8080/v1alpha1/graphql'  
```
### Affected components 

- Server
- Console
- Docs
- Tests

### Related Issues

#1452 
#1333 

### Solution and Design

- [x] @nizar-m: Make server handle JSON arrays in session variables
    - For array session variables as input for IN operator, use the following as the boolean expression
  `column_value IN (SELECT json_array_elements_text(sessionVar))`
- [x] @rikinsk: Make console accept session variable name (string) also as input for `_in` and `_nin` boolean expressions 
- [ ] @dsandip: Docs for JSON arrays in session variables


coco98:
@nizar-m Can you share the permission rule you used to set this up as well?

coco98:
@dsandip This should potentially also have an associated section in auth docs that talk about how and when to use this feature. Especially important in enterprise type scenarios when ownership information is _outside_ the database.

Like: 
- role: manager
- permission: select from this table if account_id in x-hasura-allowed-accounts

And the list of allowed accounts for the manager comes from an external database.

coco98:
@nizar-m @dsandip Has this been reviewed by @rakeshkky or @ecthiender ?

dsandip:
@coco98 No, not yet

rikinsk:
Added console code to handle session variable value for array operators. 

The UX seems slightly odd. Couldn't come up with something more intuitive. Please share feedback / suggestions if any.

im-semir:
This is really nice! When will it be ready?

valstu:
Great, this will solve many current workarounds 👍

0x777:
added in #2475

gugaevkirill:
Sorry, is somebody working on this?

PR #1800, opened 2019-03-18
use bytestring builder to represent encoded json

0x777:
So far we've been using a 'bytestring' to represent encoded json as we mostly fetch encoded json from Postgres. This PR introduces a new type called `EncJSON` and a lot of utilities to manipulate and construct encoded json. 

### Affected components 

- Server

### DIFF
--- a/server/src-lib/Hasura/EncJSON.hs
+++ b/server/src-lib/Hasura/EncJSON.hs
@@ -1,0 +1,79 @@
+-- A module for representing encoded json
+-- and efficient operations to construct them
+
+module Hasura.EncJSON
+  ( EncJSON
+  , encJToLBS
+  , encJFromBuilder
+  , encJFromJValue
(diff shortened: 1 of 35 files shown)


server/src-lib/Hasura/EncJSON.hs
```diff
@@ -0,0 +1,76 @@
...
+import qualified Data.ByteString.Lazy    as BL
+import qualified Data.Text.Encoding      as TE
+
+-- encoded json
+newtype EncJSON
+  = EncJSON { unEncJSON :: BB.Builder }
+  deriving (Semigroup, Monoid, IsString)
+
+encJToLBS :: EncJSON -> BL.ByteString
+encJToLBS = BB.toLazyByteString . unEncJSON
```
ecthiender: is it not required to inline this?

PR #1801, opened 2019-03-18
use inflection to suggest relationship names (close #1665)

### LINKED ISSUE #1665: Improve auto/suggested relationship names generated via console
1. For any FK in an object relationship, generate a relationship name that is the singular form of the table name (remove `s` if that is the last letter).
2. For any FK in an array relationship, generate a relationship name that is the singular form of the table name (add `s` if that is _not_ the last letter).
3. If there is already a relationship to the same table via another FK use the current naming scheme (XByYId)

This should be used in "track all relationships" and in the relationships pane as well.

karthikvt26:
### Description

**CHANGES PREVIOUS BEHAVIOUR**
Auto-generated relationship names are changed
**Will have to call this out in the release notes if the tracking bit is being used *wrongly* for CI/CD, especially auto-track**
cc: @shahidhk @rikinsk 

Fixes issue #1665 

### Affected components 

- Console

### Related Issues

#1665 
### Solution and Design

Uses inflection API's
### Steps to test and verify

- Create two tables author and article
- Add foreign key from article to author
- Navigate to article and verify if the suggested relationship is `author` for `author_id -> id` FK
- Navigate to the schema page and click on the button `Track all relationtions` and `individual track buttons` and validate the relationship names


0x777:
@karthikvt26 @rikinsk Can we use the column names to generate the names of object relationships? In the above example, I should see `parent` and `sibling` rather than `user` and `user1`?

rikinsk:
@0x777 I had suggested that in the issue. Problem is we dont know the format of the foreign key column. We dont have a logic to convert parent_id -> parent or parentId -> parent.

The only proper solution I can think of this is to let the user decide the pattern. Google analytics lets you do this to customize page names that appear on your dashboard.

Screenshot from 2019-03-19 11-23-34

Its a little complicated but we can try to implement something similar using table_name and foreign_key column names to derive the relationship name

coco98:
I feel like this is a little too complicated. The cases when there are multiple foreign keys to the same table from a particular table are rare and having them be user and user1 is not a big deal because the user can rename them pretty easily.

@rikinsk @0x777 Am I understanding this right?

@karthikvt26 The singularisation and pluralisation seems to be off.
I created a table called authors and articles. The relationship one way should be author and the other way should be articles?
Screenshot 2019-03-18 at 11 05 04 PM
Refer: https://hge-ci-pull-1801.herokuapp.com/console/data/schema/public

karthikvt26:
@coco98 The ones shown are not the relationship names I think they are just the indications of the potential relationships?

`authors - articles :: author_id -> id ` ~ from authors table, there is a potential array relationship to articles table using `author_id -> id` FK

karthikvt26:
@rikinsk @dsandip @coco98 I will take in the existing relationships and see if there is any conflict, if yes I will fallback to how we were generating before (`authorByAuthorId`) and if there is no conflict the inflection API will work as expected.

coco98:
@karthikvt26 Hit track all on the console. The article to author relationship that is generated is an empty string.

karthikvt26:
@coco98 Yeah the return statement got missed somehow while refactoring. Should be there in the next preview app.

shahidhk:
@arvi3411301 We need to check this logic for the CLI-based name generation.

rikinsk:
@karthikvt26 For what case was the latest commit added? Share an example please.

karthikvt26:
@rikinsk Lets say the default suggested `relName` and fallback `relName` exists for some reason. In that scenario added this change: https://github.com/hasura/graphql-engine/pull/1801/files#diff-3a82f6f6996d129c98b55c7e38b6d08fR398

To suffix with numbers to fix the collision.

PR #1826, opened 2019-03-20
add rest-wrapper remote schema boilerplate

tirumaraiselvan:
### Description

A boilerplate to show how to turn a REST API like the following:

```
  GET /users
  GET /users/:userId
  POST /users
```
into a GraphQL API like the following:

```
  type User {
    id:       String!
    name:     String!
    balance:  Int!
  }

  type Query {
    getUser(id: String!): User
    users(name: String): [User]
  }

  type Mutation {
    addUser(name: String!, balance: Int!): User
  }
```

### Affected components 

- Community Content


community/boilerplates/remote-schemas/rest-wrapper/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+## Deployment (Using Heroku)
+
+You need a Heroku account and heroku-cli installed.
+
+1. Log into Heroku
+
+```bash
+$ heroku login
```
shahidhk: Can we remove all `$` from commands? Easier to copy paste.

PR #1832, opened 2019-03-21
fix bash completion example docs on cli

simoneb:
### Description

Fix completion directory name (missing leading dot) for bash on Windows

### Affected components 

- Docs


shahidhk:
Thanks for the PR @simoneb. The documentation is auto-generated from the CLI help. Can you also make the change here: https://github.com/hasura/graphql-engine/blob/master/cli/commands/completion.go#L29 ?

simoneb:
@shahidhk done!

shahidhk:
Thanks a ton @simoneb :+1:

PR #1841, opened 2019-03-22
[minor breaking] drop id from event_triggers table (fix #1840)

### LINKED ISSUE #1840: Allow metadata apply to retain web hooks and other stateful config if there is no change
Currently `hasura metadata apply` will remove and replace the metadata inside graphql-engine with the metadata in the `metadata.yaml` file. For most features this is fine as there is no state to keep between resets. However the webhook execution history and retry state is useful information to be able to see over time.

As previously discussed, Hasura doesn't actually delete the old webhook history or retry state when apply. Instead it creates new webhooks with new ids and doesn't show the old webhooks in the UI.

**It would be great if there was a way for webhook configuration to be persisted across apply's if nothing has changed in the webhook configuration.** I can see a few ways this could be done:

* Setting the webhook ID in the metadata.yaml. The next time the metadata was applied, the webhook would be updated to the same configuration and keep the same ID. This would let you see previous webhook executions.
* Checking if all of the webhook configuration values are the same as the ones that exist. If so, leave the webhook configuration untouched, instead of deleting and recreating it.

This would only be possible if Hasura was able to booted up enough to read the existing metadata. Relates to #231 and #1839.

tirumaraiselvan:
### Description

An auto-generated trigger_id was being generated for each event_trigger. This trigger_id was causing problems as described in #1840 . We solve this by removing the trigger_id and using trigger_name as the identifier for the event_trigger everywhere.

### Affected components 

- Server
- Console
- Docs

### Related Issues

#1840 

This change is breaking in the following way:

- There will no longer be an `id` key in the `trigger` section of the event payload i.e.

```
  "trigger": {
      "name": "<name-of-trigger>",
      "id": "<uuid>"
  }
```

changes to: 

```
  "trigger": {
      "name": "<name-of-trigger>"
  }
```


tirumaraiselvan:
@wawhal Should also check down_migration behaviour.

console/src/components/Services/EventTrigger/Modify/ActionButtons.js
```diff
@@ -1,25 +1,120 @@
...
+        <div className={`${styles.display_flex}`}>
+          <div>
+            <input
+              type="checkbox"
+              checked={shouldCascade}
+              onChange={toggle}
+              className={styles.add_mar_right_small}
+            />
+          </div>
+          <div>Drop pending events</div>
```
tirumaraiselvan: change this to `Delete existing events (including pending events)`

server/src-lib/Hasura/RQL/DDL/EventTrigger.hs
```diff
@@ -140,45 +137,52 @@ addEventTriggerToCatalog
...
-delEventTriggerFromCatalog trn = do
+delEventTriggerFromCatalog :: TriggerName -> Bool -> Q.TxE QErr ()
+delEventTriggerFromCatalog trn cascade = do
   Q.unitQE defaultTxErrorHandler [Q.sql|
            DELETE FROM
                   hdb_catalog.event_triggers
            WHERE name = $1
                 |] (Identity trn) True
   delTriggerQ trn
+  bool (return ()) (deleteAllEventsForTrigger trn) cascade
```
tirumaraiselvan: is there a constant function for `returnNil` or something?

server/src-exec/Migrate.hs
```diff
@@ -258,6 +258,25 @@ from10To11 = liftTx $ do
...
+     , UserInfoM m
+     , MonadIO m
+     )
+  => m ()
+from11To12 = do
+  -- Migrate database
+  Q.Discard () <- liftTx $ Q.multiQE defaultTxErrorHandler
+    $(Q.sqlFromFile "src-rsr/migrate_from_11_to_12.sql")
+  -- Migrate metadata
+  migrateMetadata True migrateMetadataFrom11
```
tirumaraiselvan: Should this be True or False?

danielcompton:
I attempted to run this with the server running `hasura/graphql-engine:pull1841-1a3f01e` and the CLI built from the same branch/commit. After running `hasura metadata apply`, I can see the triggers, but whenever I try and look at their details I get a 404 error and this in the logs:

```json
{
  "timestamp": "2019-03-25T02:10:25.195+0000",
  "level": "info",
  "type": "http-log",
  "detail": {
    "status": 400,
    "query_hash": "af926a1a566bf2b28ce50037d865b4777da0542e",
    "http_version": "HTTP/1.0",
    "query_execution_time": 0.000301981,
    "request_id": null,
    "url": "/v1/query",
    "ip": "172.19.0.9",
    "response_size": 110,
    "user": {
      "x-hasura-role": "admin"
    },
    "method": "POST",
    "detail": {
      "error": {
        "path": "$.args[0].args.columns[1].name",
        "error": "relationship \"events\" does not exist",
        "code": "not-exists"
      },
      "request": "{\"type\":\"bulk\",\"args\":[{\"type\":\"select\",\"args\":{\"columns\":[\"*\",{\"name\":\"events\",\"columns\":[\"*\",{\"name\":\"logs\",\"columns\":[\"*\"],\"order_by\":[\"-created_at\"]}],\"where\":{\"$or\":[{\"delivered\":{\"$eq\":true}},{\"error\":{\"$eq\":true}}]},\"order_by\":[\"-created_at\"],\"limit\":10}],\"where\":{\"name\":\"trig_1\"},\"table\":{\"name\":\"event_triggers\",\"schema\":\"hdb_catalog\"}}},{\"type\":\"count\",\"args\":{\"columns\":[\"name\"],\"where\":{\"$and\":[{\"trigger_name\":\"trig_1\"},{\"$or\":[{\"delivered\":{\"$eq\":true}},{\"error\":{\"$eq\":true}}]}]},\"table\":{\"name\":\"event_log\",\"schema\":\"hdb_catalog\"}}}]}"
    }
  }
}
```
 
Just some early feedback, I'm aware it might not be ready to merge yet.

tirumaraiselvan:
@danielcompton Thanks for the early feedback :) We are still pending few tests especially on the `metadata apply` path on existing instances. Will check this out.

server/src-lib/Hasura/RQL/DDL/EventTrigger.hs
rakeshkky: Use `when` function.
```haskell
when cascade $ deleteAllEventsForTrigger trn
```

rikinsk:
@wawhal @tirumaraiselvan There is a bug in console. Pending event throws 404.

rikinsk:
Add tooltip explaining what data wont be deleted means.

tirumaraiselvan:
Closing this PR and submitting a new one which breaks this PR into 2:

1. Bug fix
2. Delete Event Trigger flags.

PR #1855, opened 2019-03-25
refactor server code

0x777:
Reduces duplication and cleaner module exports. This does not change any functionality.


server/src-lib/Hasura/GraphQL/Explain.hs
```diff
@@ -128,30 +126,23 @@ explainGQLQuery
...
+    E.GExPHasura gCtx rootSelSet ->
+      return (gCtx, rootSelSet)
+    E.GExPRemote _ _  ->
+      throw400 InvalidParams "only hasura queries can be explained"
+  case rootSelSet of
+    GV.RQuery selSet -> do
+      let tx = mapM (explainField userInfo gCtx sqlGenCtx) (toList selSet)
+      plans <- liftIO (runExceptT $ runTx tx) >>= liftEither
+      return $ encJFromJValue plans
+    GV.RMutation _ ->
```
ecthiender: we can use wildcard pattern match `_` if its not `RQuery`?

0x777: It helps to list all the available constructors sometimes. For example, if we happen to introduce a new constructor later, with `_` we get no help from the compiler.

PR #1857, opened 2019-03-25
breaking: drop id from event_triggers table (fix #1840)

### LINKED ISSUE #1840: Allow metadata apply to retain web hooks and other stateful config if there is no change

tirumaraiselvan:
### Description

An auto-generated trigger_id was being generated for each event_trigger. This trigger_id was causing problems as described in #1840 . We solve this by removing the trigger_id and using trigger_name as the identifier for the event_trigger everywhere.

### Affected components 

- Server
- Console
- Docs

### Related Issues

#1840 

This change is breaking in the following way:

- There will no longer be an `id` key in the `trigger` section of the event payload i.e.

```
  "trigger": {
      "name": "<name-of-trigger>",
      "id": "<uuid>"
  }
```

changes to: 

```
  "trigger": {
      "name": "<name-of-trigger>"
  }
```

### DIFF
--- a/console/src/components/Services/EventTrigger/EventActions.js
+++ b/console/src/components/Services/EventTrigger/EventActions.js
@@ -242,7 +242,7 @@ const loadEventLogs = triggerName => (dispatch, getState) => {
                     columns: ['*'],
                   },
                 ],
-                where: { event: { trigger_id: triggerData[0].id } },
+                where: { event: { trigger_name: triggerData[0].name } },
                 order_by: ['-created_at'],
                 limit: 10,
               },
--- a/console/src/components/Services/EventTrigger/Modify/Info.js
+++ b/console/src/components/Services/EventTrigger/Modify/Info.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import Tooltip from './Tooltip';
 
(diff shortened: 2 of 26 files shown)


tirumaraiselvan:
Check invocation logs page as well.

danielcompton:
Thanks, I've verified this patch fixes our issues.

PR #1872, opened 2019-03-26
Manual event triggers: invoke event triggers manually on rows (close #1687)

### LINKED ISSUE #1687: Console Triggers

tirumaraiselvan:
### Description

The new API adds an optional `manual` BOOLEAN key to the spec:

```
{
    "type" : "create_event_trigger",
    "args" : {
        "name": "testtrigger",
        "table": "users",
        "webhook": "https://httpbin.org/post",
        "insert": {
            "columns": "*"
        },
        "manual": true
    }
}
```

### Affected components 

- Server
- Console
- Docs

### Related Issues

#1687 

### Solution and Design

- Create a new v1 query type: invoke_event_trigger
- Create a function which takes row data and related metadata and inserts into event_log.
- Use above function to take params from invoke_event_trigger

### Steps to test and verify

1. Create a table:

```
curl 'http://localhost:8080/v1/query'  -H 'Content-Type: application/json'  -d '{"type":"bulk","args":[{"type":"run_sql","args":{"sql":"CREATE TABLE public.\"users\"(\"username\" text NOT NULL, \"real_name\" text NOT NULL, PRIMARY KEY (\"username\") );"}},{"type":"add_existing_table_or_view","args":{"name":"users","schema":"public"}}]}' 
```

2. Create an event trigger with manual spec:

```
curl -H 'Content-Type: application/json' -X POST -d '{"type": "create_event_trigger", "args": {"name": "testtrigger", "table": "users", "update": { "columns" : ["username", "real_name"]},"insert": { "columns" : ["username"] }, "manual": true, "webhook": "https://httpbin.org/post" } }' localhost:8080/v1/query
```

3. Invoke an event manually:

```
curl -d '{"type": "invoke_event_trigger", "args": {"name": "testtrigger", "new": { "username": "manualusername", "real_name": "manualrealname"}, "old": null}}' localhost:8080/v1/query
```


hmexx:
Tested the feature on the review app with glitch. Seems to work correctly!

tirumaraiselvan:
Closing this draft PR for actual PR

PR #1897, opened 2019-03-28
read version from env var during server build time (close #1398)

### LINKED ISSUE #1398: read version from env var during build time
Right now, server build process takes the version by executing `scripts/get-version.sh`. Instead, server should take env var and the build system should set this accordingly.

shahidhk:
### Description
This PR modifies the server code to read the version from an env var `VERSION` during build time. Earlier it was done by execution `scripts/get-version.sh`. Also adds `build` and `exec` rules to the makefile.

If the env var is not set, it defaults to the output of `scripts/get-version.sh`.

This makes it easier to build server with custom version during dev - especially when CLI/Server compatibility needs to be tested/worked upon.

### Affected components 

- Server
- Build System

### Related Issues
close #1398 

### Solution and Design
`getValFromEnv :: String -> TH.Q TH.Exp` function is added to `Hasura.Server.Utils` module and this function is called from `Hasura.Server.Version` module as `version = $(getValFromEnv "VERSION")`

### Steps to test and verify
Build the server with a custom version:
```bash
VERSION=my-custom-version make build
```
Run the server:
```bash
HASURA_GRAPHQL_DATABASE_URL=postgres://postgres:@localhost:5432/postgres make exec
```
Hit the `/v1/version` endpoint:
```bash
curl http://localhost:8080/v1/version   
{"version":"my-custom-version"}
```

NA


shahidhk:
One thing to note is that the build will throw an error if env var VERSION is not set. This needs to be added to the contributing guide and instructions should be given with make instead of stack directly.

Do not merge the PR until this is added to contributing guide.

PR #1904, opened 2019-03-29
specify better error msgs for remote schema conflicts (fix #1805)

### LINKED ISSUE #1805: give hints on remote schema type conflicts
Suppose we miss, say an `!`, in one the types, then the error is like the following:

remote-schema

This gives the impression that you can't add pre-existing types. This should hint that there is a type mismatch rather than pre-existence.

tirumaraiselvan:
### Affected components 

- Server

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Schema.hs
+++ b/server/src-lib/Hasura/GraphQL/Schema.hs
@@ -1717,10 +1717,10 @@ checkSchemaConflicts gCtx remoteCtx = do
 
     hQRName = G.NamedType "query_root"
     hMRName = G.NamedType "mutation_root"
-    tyMsg ty = "types: <" <> namesToTxt ty <>
-               "> have mismatch with current graphql schema. HINT: Types must be same."
-    nodesMsg n = "top-level nodes : <" <> namesToTxt n <>
-                 "> already exist in current graphql schema. HINT: Top-level nodes can't be same."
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -1717,10 +1717,10 @@ checkSchemaConflicts gCtx remoteCtx = do
 
     hQRName = G.NamedType "query_root"
     hMRName = G.NamedType "mutation_root"
-    tyMsg ty = "types: [" <> namesToTxt ty <>
-               "] already exist in current graphql schema"
-    nodesMsg n = "nodes : [" <> namesToTxt n <>
-                 "] already exist in current graphql schema"
+    tyMsg ty = "types: <" <> namesToTxt ty <>
+               "> have mismatch with current graphql schema. HINT: Types must be same."
+    nodesMsg n = "top-level nodes : <" <> namesToTxt n <>
```
ecthiender: use the square brackets `[`, `]` itself, because this is a list of types?

tirumaraiselvan: True. But its much clearer (IMHO) if its `< >` because with `[ ]` there is a confusion when the frontend throws error with just ONE type mismatch. 

It comes like `type [user] is conflicting`, which is confusing because I have the type user and also a return type of [user] .

PR #1924, opened 2019-04-02
update console permissions ui

rikinsk:
### Description

- Make permissions sections as collapsibles with tooltips
- Remove 'use same permission as select' type options from row permissions section
- Added clone section to apply same permission to any table-role-action 
- Disable other subsections till row permissions are set

### Affected components 

- Console


coco98:
I really like it! ❤️ 

I think the only thing I miss from the old UI is that we don't have "copy permissions from insert" type setting when I do select.

Second, do you think we should have the row-select section expanded by default when permissions are empty? And maybe we should select all the columns by default? So that if I just end up setting a row-level check but not a column level check (it was not expanded so I didn't care), my query still works...

PR #1932, opened 2019-04-02
update remote schema env var validation components

rikinsk:
### Description

- remove env var validation note from docs
- change title of remote schema create errors

### Affected components 

- Server
- Console
- Docs


rikinsk:
@tirumaraiselvan we need to update the error message returned by server in case of missing env var. It uses the phrase "custom resolver" which users dont know. Lets make int consistent with message for event triggers. The message there is generic enough. Similarly for any other error message we throw. remote schema instead of custom resolver needs to happen.
Screenshot from 2019-04-02 16-22-37

tirumaraiselvan:
@rikinsk The fix for that is in this PR: https://github.com/hasura/graphql-engine/pull/1904/

Issue #394, opened 2018-09-11
List all possible column types in the table creation UI?


0x777:
You may want to see custom types, domains etc. If I don't see the type I'll have to switch to the SQL window. The least we can do is to provide an escape hatch with a text box to enter the required type.

revskill10:
@0x777 You don't need that. This query will select all domains for a specific schema
```sql
SELECT typname FROM pg_catalog.pg_type JOIN pg_catalog.pg_namespace ON pg_namespace.oid = pg_type.typnamespace WHERE typtype = 'd' AND nspname = 'someschema'
````

rikinsk:
related to #223

PR #1933, opened 2019-04-02
add all PG column types while table creation / modification

### LINKED ISSUE #223: Add all postgres types support (e.g. lat/long) in create/modify table
Check if extension is enabled and display types accordingly.

karthikvt26:
Fix #223, #2035, #394, #2038
### Affected components 

- Console


RodolfoSilva:
I think this will solve this issue too #2038

PR #1967, opened 2019-04-09
add an api to dump postgres schema (close #1939)

### LINKED ISSUE #1939: add an api which can execute pg_dump on the database
Using this API, a user should be able to get the Postgres schema dump.

```http
POST /v1alpha1/pg_dump

{
  "schema": "public"
}
```

response will be a file download

shahidhk:
### Description
This PR adds a new admin-only API for getting schema dumps from the connected Postgres instance.

```http
POST /v1alpha1/pg_dump

{
  "opts": ["-O", "-x", "--schema", "public", "--schema-only"],
  "clean_output": true
}
```

response:
```http
HTTP/1.1 200 OK
Content-Type: application/sql

CREATE TABLE public.author (
    id integer NOT NULL,
    name text NOT NULL
);
CREATE SEQUENCE public.author_id_seq
    AS integer
    START WITH 1
    INCREMENT BY 1
    NO MINVALUE
    NO MAXVALUE
    CACHE 1;
ALTER SEQUENCE public.author_id_seq OWNED BY public.author.id;
ALTER TABLE ONLY public.author ALTER COLUMN id SET DEFAULT nextval('public.author_id_seq'::regclass);
```

If there is an error, the response will be JSON with the right error code and error messages.

Since this is an API executing a shell script on the server, an option is provided to enable/disable the API. By default, the API is enabled. To disable, use the `--enabled-apis` by excluding `pgdump` out of the default value `graphql,metadata,pgdump`.

### Affected components 

- Server
- Build System
- Tests

### Related Issues
close #1939 

### Solution and Design
A new admin-only POST endpoint which takes `schema` as a JSON key. The handler executes a shell script (https://github.com/shahidhk/graphql-engine/blob/pg-dump-on-server/server/src-rsr/run_pg_dump.sh) which executes `pg_dump` and writes the output to file. This file is then read and sent to the client in the response body.

The `clean: true` argument will clean the output to remove:

1. SQL front matter, like SET statements.
2. `CREATE SCHEMA public`.
3. `COMMENT ON SCHMEA public is 'standard public schema';`
3. Comments (`--`) and empty newlines.
4. Postgres triggers created by Hasura for event triggers.

### Steps to test and verify
```bash
curl -d'{"opts": ["-O", "-x", "--schema", "public", "--schema-only"],"clean_output": true}' http://localhost:8080/v1alpha1/pg_dump
```

~The `pg_dump` binary is not yet included in the docker image, the docker image will not work.~

~Tests are pending.~

NA


mnlbox:
@shahidhk Can you add `Console` support for this too please?

server/src-lib/Hasura/Server/PGDump.hs
```diff
@@ -0,0 +1,64 @@
...
+      return $ Right contents
+    ExitFailure _ ->  return $ Left stdErr
+
+execPGDump
+  :: (MonadError RTE.QErr m, MonadIO m)
+  => PGDumpReqBody
+  -> Q.ConnInfo
+  -> m BL.ByteString
+execPGDump b ci = do
+  output <- liftIO $ runScript dbUrl opts clean
```
rakeshkky: `runScript` is prone to IO exception when the `filename` is not found in disk or maybe other reasons. Use `try` to catch IO exception explicitly.

mnlbox:
@shahidhk Maybe we need some docs change here: https://docs.hasura.io/1.0/graphql/manual/migrations/existing-database.html#step-3-initialize-the-migrations-as-per-your-current-state

PR #1988, opened 2019-04-11
Updated to link that works for macOS

pronevich:
### Description

the previous link doesn't install cli
### Affected components 

- Docs

### DIFF
--- a/docs/graphql/manual/hasura-cli/install-hasura-cli.rst
+++ b/docs/graphql/manual/hasura-cli/install-hasura-cli.rst
@@ -45,7 +45,7 @@ Install
 
          .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


shahidhk:
Thanks for the contribution @pronevich! `https://cli.hasura.io/install` is an older version of CLI which we have deprecated, but kept alive for backward compatibility.

The correct way to install cli is by executing `https://github.com/hasura/graphql-engine/blob/master/cli/get.sh` which downloads cli from GitHub releases.

shahidhk:
If there is an issue with installation on macOS, please open an issue with repro instructions. :smile: 

Thanks again for your contribution. :pray:

PR #2016, opened 2019-04-16
schema management refactor

arvi3411301:
### Description
- fix editing a Foreign Key which is referenced to a untracked table doesn't show the reference column

### Affected components 

- Console


trickymast3r:
This fix error when using Uppercase , camelCase table name like 'Account', 'AccountReport', 'accountReport'

PR #2034, opened 2019-04-19
add ability to override server timeout for cli migrations image

Maxim-Filimonov:
### Description
Added ability to override server timeout with custom env variable as depending on weather heroku sometimes fails to spin up a server in 30 seconds.. ^_^
User now can override server timeout via HASURA_GRAPHQL_SERVER_TIMEOUT env variable.

### Affected components 

- Build System


Maxim-Filimonov:
@shahidhk just checking I'm not sure what is the best process to update the docs to mention this variable. Does it need to be specified in some README?

shahidhk:
@Maxim-Filimonov I think it's fine, the variable is pretty evident from the script. I will test this out once and merge.

Thanks for your contribution.

PR #2058, opened 2019-04-24
log inconsistent metadata objects only when there is at least one (close #2053)

### LINKED ISSUE #2053: Should not print warning if there's no inconsistent_metadata
Shouldn't this log not printed if there's no inconsistent_metadata?
This is causing our error log alarm to trigger false alarm
```
{
    "timestamp": "2019-04-23T10:20:35.225+0000",
    "level": "warn",
    "type": "startup",
    "detail": {
        "kind": "inconsistent_metadata",
        "info": {
            "objects": []
        }
    }
}
```

rakeshkky:
### Description

When inconsistent metadata objects is
`empty list` -> no log
`non empty list` -> log with warn level

### Affected components 

- Server

### Related Issues

fix #2053


server/src-lib/Hasura/Server/App.hs
```diff
@@ -103,16 +103,23 @@ data SchemaCacheRef
...
+withSCUpdate scr logger action = do
   acquireLock
   (res, newSC) <- action `catchError` onError
-  -- update schemacache in IO reference
-  liftIO $ modifyIORef' cacheRef $
-    \(_, prevVer) -> (newSC, incSchemaCacheVer prevVer)
-  liftIO onChange
+  liftIO $ do
+    -- log any inconsistent objects
+    logInconsObjs logger $ scInconsistentObjs newSC
```
0x777: log after writing schemacache to the ref

PR #2059, opened 2019-04-24
add business logic page to docs

rikinsk-zz:
### Affected components 

- Docs


rikinsk:
We should also add a page under Mutations for doing **validations** with the following solutions:

- Postgres check constraints
- Event triggers: Insert with valid=false, run event trigger and set to true if valid, filter with valid=true
- Remote schema: use like a middleware. Make call to remote schema, validate, call HGE to do action  with permissions checks.

mnlbox:
It's very useful.
@rikinsk Thanks man

tirumaraiselvan:
Not a big fan of pre-CRUD/ post-CRUD terminologies. How about sync/async logic?

hmexx:
> 
> 
> Not a big fan of pre-CRUD/ post-CRUD terminologies. How about sync/async logic?

Agree with @tirumaraiselvan . Sync/Async would be good as the top-level categorisation.

Other potential changes:

- Put Views within Synchronous instead of their own category.
- Add Postgres constraints checks within Synchronous (mention they are ideal for simple validation)
- Add Postgres native triggers to Asynchronous (this one is debatable as Event Triggers are better practice)

PR #2064, opened 2019-04-24
graphql responses are now http 200 (fix #1368)

### LINKED ISSUE #1368: Status codes for non-fatal properties
I just ran into an issue while using Apollo Client with Hasura.  Since Hasura always returns a 400 status code when there's an error in the query, Apollo Client will **always** treat that error as a `networkError` and not pull the graphql errors out of the response.

Apollo client expects a query with non-fatal errors (it seems like that means if there's a query with optional fields that fail) to return as a 200 with the `errors` property instead of the `data` property.

Can the status codes be changed in Hasura to have full compatibility with Apollo Client?

ecthiender:
All graphql clients expect responses to be HTTP 200. Non-200 responses are considered transport layer errors. Changed all graphql responses in **`/v1/graphql`** endpoint to be 200.

### Description

This change only works on the new endpoint: `/v1/graphql`. The behaviour of `/v1alpha1/graphql` remains the same.

### Affected components 

- Server
- Tests

### Related Issues

#1368 

### Steps to test and verify

1. Make any wrong request to `/v1/graphql` endpoint
2. It returns HTTP 200 status.

Console changes are pending.
Have select tests for `/v1alpha1/graphql` over websocket


0x777:
@ecthiender Can you check the error response format for subscriptions for `/v1/graphql`?
1. Are they as per the apollo-ws protocol?
2. Are they consistent with the errors in http?

shahidhk:
@0x777 You're right, subscription errors are not GraphQL compliant.

```json
{
  "type": "error",
  "id": "1",
  "payload": {
    "path": "$.selectionSet.articles.selectionSet.id1",
    "error": "field \"id1\" not found in type: 'articles'",
    "code": "validation-failed"
  }
}
```

is what the response is right now. It should be:

```json
{
  "type": "error",
  "id": "1",
  "payload": {
    "message": "field \"id1\" not found in type: 'articles'",
    "extensions": {
      "code": "validation-failed",
      "path": "$.selectionSet.articles.selectionSet.id1"
    }
  }
}
```

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -155,8 +163,17 @@ onConn (L.Logger logger) corsPolicy wsId requestHead = do
         (BL.toStrict $ J.encode $ encodeGQLErr False qErr)
 
     checkPath =
```
0x777: `checkPath` should return `ErrRespType`. You don't have to look at paths twice.

PR #2068, opened 2019-04-25
ra-data-hasura: add support for multiple schemas (close #1729)

### LINKED ISSUE #1729: add support for multiple postgres schemas in ra-data-hasura
Currently `ra-data-hasura` assumes the schema to be public. It can accept an optional parameter for schema name, just like headers/httpClient.

praveenweb:
### Description

To query schemas other than `public`, you can pass schema to resource in the format
 `<Resource name="schema.table" />`.

For example to fetch data from schema `test` and table `author`, use the following snippet:

```
  <Resource name="test.author" list={list} />
```
Install the latest version
```
npm install --save ra-data-hasura@0.0.4
```

### Affected components 

- Community Content

### Related Issues

#1729 

### Solution and Design

Support multiple schemas using the "." separator.

### DIFF
--- a/community/tools/ra-data-hasura/CHANGELOG.md
+++ b/community/tools/ra-data-hasura/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 0.0.4 (April 25, 2019)
+
+- Feature: Support multiple schemas using "." separator. 
+Example:  `<Resource name="schema.table" />`
+
 ## 0.0.3 (January 24, 2019)
 
 - Bug Fix: Fix count query to support UUID
--- a/community/tools/ra-data-hasura/README.md
+++ b/community/tools/ra-data-hasura/README.md
@@ -60,6 +60,17 @@ export default App;
 
 In case the server is configured with admin secret or auth, configure the appropriate headers and pass it to the provider.
 
+### Multiple schemas
+
+To query schemas other than `public`, you can pass it to resource in the format
+ `<Resource name="schema.table" />`.
+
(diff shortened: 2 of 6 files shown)


mnlbox:
I test latest version of react-admin data provider
My schema name is `online_exam` and it not working with this schema name. I entered resource name like this:
```
<Resource name="online_exam.difficulty" icon={ModeIcon} list={DifficultyList} show={DifficultyShow} create={DifficultyCreate} edit={DifficultyEdit} options={{ label: 'Difficulty' }}/>
```
This is my error message in browser console:
```
Warning: Missing translation for key: "{"path":"$.args[0].args.order_by[0]","error":"\"id\" does not exist","code":"not-exists"}"
```
Any idea?

praveenweb:
@mnlbox - Does the table difficulty in the online_exam schema have the column id? The error seems to say that id does not exist. 

Also it would be better if you can share the actual query that is made (check network tab?)

mnlbox:
@praveenweb It seems it's return 2 query. One with schema: public, table: difficulty` and one with `schema: online_exam, table: difficulty`. :thinking: 
Why it send 2 request?

praveenweb:
@mnlbox - May be you are calling <Resource> twice with both difficulty and online_exam.difficulty?

mnlbox:
It's workin OK, Can you merge this? I want add some new PR based on this :wink: 
@shahidhk 
@praveenweb 
@tirumaraiselvan

PR #2099, opened 2019-04-30
Skip diff for update event if listen columns is empty (fix #2087)

### LINKED ISSUE #2087: send update event even when row did not change
From sharon#1298: 
```
Hi for Hasura triggers we observed that if we set the same value then web hooks are not triggered,  is it because value is not updated? , is there any way we  can achieve ----> setting a column value always invoke trigger irrespective of database field value changes
```

Currently, we diff the old and new row and only invoke trigger if there is some change. We need a flag to disable the diffing.

tirumaraiselvan:
### Description

Currently, for update events the OLD and NEW rows are compared and event is invoked only if there is a diff. This is a sane default but you may need to trigger an update event even if no rows changed as mentioned in #2087 

User can choose a flag called (SKIP DIFF/ SKIP COMPARE) in the Listen Columns section of event trigger which would give the desired behaviour. This is equivalent to giving an empty array in the column spec of the API .

### Affected components 

- Server

### Related Issues

#2087 

### Solution and Design

We template the trigger function based on the condition of the flag.


tirumaraiselvan:
closing in favour of #2180

PR #2144, opened 2019-05-08
fix dropdown arrow inconsistency among browsers, fixes #1878

### LINKED ISSUE #1878: Minor UI glitch while browsing docs on different browsers
I dont know if this will qualify as a valid issue or not, I am just sharing what I observed while I was browsing hasura graphql-engine docs, few days back.

Follow this [link](https://drive.google.com/openid=1zWhMQhhnzwIjCrksze66VzwJ7W0lUWzg) to see the glitch.

I did some research to find a fix. Found out that this drop down arrow (of the select tag) is the most difficult thing to be customized with css. Browsers are not uniform wrt its implementation.

[Stackoverflow question (1,413,121 views)](https://stackoverflow.com/questions/1895476/how-to-style-a-select-dropdown-with-only-css)

This is minor little thing to be observed by people, usually people use only one type of browser. 
My intention is to bring about uniformity among browsing experiences. 

There is a workaround I have found out, i.e to just remove the arrow with 3-4 lines of css. (Have to figure out how to do that in sphinx)

We only have v1.0 for the docs as of now, so arrow is also not that useful, and probably not clicked so often (my thoughts).

If this is accepted as a bug, I will submit a PR.

anubhav8421:
After this commit, the dropdown arrow is displayed consistently among chromium and mozilla firefox browsers. Not checked for safari.

## Firefox
Screenshot_2019-05-08 Hasura GraphQL Engine Documentation Hasura x y documentation(1)

## Chromium
Screenshot from 2019-05-08 19-44-31


rikinsk13:
@marionschleifer I have tested this finally. Can you approve it as code owner so I can merge?

PR #2148, opened 2019-05-09
support composite unique keys from the console (close #506)

### LINKED ISSUE #506: support composite primary, and foreign keys and unique constraints through the console
This issue comes from @lukebyrne and was originally opened [here](https://github.com/hasura/graphqurl/issues/27).

As of now:
1. You can set composite primary keys from the console while creating the table
2. You cannot set composite primary keys from the console once the table is already created (can be done via SQL)
3. You cannot set composite foreign keys from the console (can be done via SQL)
4. Relationships are not suggested based on composite primary keys (can be done by adding manual relationships) *([Related issue](https://github.com/hasura/graphql-engine/issues/284))*
5. Adding a manual relationship does not support multiple columns.

@praveenweb please point out if I have missed anything.

wawhal:
#506 

### Description

Support creating and modifying unique constraints from the console

### Affected components 

- Console

### Related Issues

#506 

### Steps to test and verify

1. Try create unique constraints while creating table
2. Try modifying and deleting those constraints from the modify table section
3. Try creating new constraints from the modify table section


karthikvt26:
@wawhal Functionality looks fine. One suggestion

Should we show the constraint name too?

PR #2156, opened 2019-05-10
close websocket connection on JWT expiry (fix #578)

### LINKED ISSUE #578: server should handle JWT expiry for websocket connections
Currently, server doesn't handle JWT expiry for websocket connection.

Server should send out a connection_error message when JWT expires.

rakeshkky:
### Description

Close WebSocket connection while subscriptions on JWT expiry.

### Affected components 

- Server
- Tests

### Related Issues

close #578 

### Solution and Design

Run an async thread on websocket connection accept which terminates on JWT expiry.

### Steps to test and verify

Perform a subscription query with JWT setup. The server should close connection on JWT expiry.


server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -145,15 +147,27 @@ onConn (L.Logger logger) corsPolicy wsId requestHead = do
       sendMsg wsConn SMConnKeepAlive
       threadDelay $ 5 * 1000 * 1000
 
+    jwtExpiryHandler wsConn = do
+      currTime <- TC.getCurrentTime
```
0x777: This has to be done after the stm action. The STM action will return only after the connection is initialised so you'll sleep till after the expiry. Hence you need to do this after you get the expiry time.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -366,10 +380,10 @@ logWSEvent
   :: (MonadIO m)
   => L.Logger -> WSConn -> WSEvent -> m ()
 logWSEvent (L.Logger logger) wsConn wsEv = do
-  userInfoME <- liftIO $ IORef.readIORef userInfoR
+  userInfoME <- liftIO $ STM.readTVarIO userInfoR
   let userInfoM = case userInfoME of
-        CSInitialised userInfo _ -> return $ userVars userInfo
-        _                        -> Nothing
+        CSInitialised userInfo _ _ -> return $ userVars userInfo
```
0x777: Might as well log the expiry time here?

PR #2173, opened 2019-05-13
fix non-200 response for authorization errors on /v1/graphql

ecthiender:
### Description

Authorization checks would still return non-200 response on `/v1/graphql`.

### Affected components 

- Server


server/src-lib/Hasura/Server/App.hs
```diff
@@ -232,7 +232,9 @@ mkSpockAction qErrEncoder qErrModifier serverCtx handler = do
       manager = scManager serverCtx
 
   userInfoE <- liftIO $ runExceptT $ getUserInfo logger manager headers authMode
-  userInfo <- either (logAndThrow req reqBody False) return userInfoE
+  -- apply the error modifier
+  let modUserInfoE = fmapL qErrModifier userInfoE
+  userInfo <- either (logAndThrow req reqBody False) return modUserInfoE
```
0x777:
```haskell
userInfo <- either (logAndThrow req reqBody False . qErrModifier) return userInfoE 
```

PR #2175, opened 2019-05-13
add a haskell code style guide

ecthiender:
### Description

A first version of coding style guideline for the Haskell codebase. 

This is useful for contributors to maintain a consistent coding style throughout the codebase.

### Affected components 

- Other (list it)
Just the server contributing guide.

### Related Issues

N/A

### DIFF
--- a/server/STYLE.md
+++ b/server/STYLE.md
@@ -8,7 +8,9 @@ Use [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) to format yo
 
 #### Line Length
 
-Try to keep a line length within 100 characters. Small exceptions are okay if the line only goes over the limit by a few characters and breaking it would negatively impact readability. Use your best judgement.
+We do not enforce a hard limit on line length, but we try to keep lines under 80 characters where possible to make it easier to tile files horizontally. It’s okay if long lines spill into the 80–100 character range if wrapping them would harm readability. In particular, we don’t want to discourage descriptive variable names, nor do we want to discourage introducing local variables, so consider 80 characters a target length rather than a limit.
+
+One exception to the above is multiline comments, which should almost always be wrapped to 80 characters. Most editors support automatically reflowing blocks of text with a configurable line length; consider learning the relevant hotkey in your editor of choice.
 
 #### Whitespace
 
--- a/server/STYLE.md
+++ b/server/STYLE.md
@@ -1,30 +1,19 @@
-## Code conventions and style guide
+# Code conventions and style guide
 
-This helps enforce a uniform style for all committers.
-
-- Compiler warnings are turned on, make sure your code has no warnings.
-- Use [hlint](https://github.com/ndmitchell/hlint) to make sure your code has no
-  warnings.
-- Use [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) to format
-  your code.
-
-This is a short document describing the preferred coding style for this project.
-We've tried to cover the major areas of formatting and naming. When something
-isn't covered by this guide you should stay consistent with the code in the
-other modules.
+This is a short document describing the preferred coding style for this project. We've tried to cover the major areas of formatting and naming. When something isn't covered by this guide, err on the side of consistency with existing code.
 
 ### Formatting
 
+Use [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) to format your code.
+
 #### Line Length
 
-Try to keep a line length within 100 characters. Small exceptions are okay if
-the line only goes over the limit by a few characters and breaking it would
-negatively impact readability; use your best judgement.
+Try to keep a line length within 100 characters. Small exceptions are okay if the line only goes over the limit by a few characters and breaking it would negatively impact readability. Use your best judgement.
 
-#### Indentation
+#### Whitespace
 
-Tabs are illegal. Use spaces for indenting. Indent your code blocks with *2
-spaces*. Indent the `where` keyword two spaces to set it apart from the rest of
+Do not use tabs. Use spaces for indenting. Indent your code blocks with **2
+spaces**. Indent the `where` keyword two spaces to set it apart from the rest of
 the code and indent the definitions in a `where` clause 2 spaces. Some examples:
 
 ```haskell
@@ -42,19 +31,7 @@ filter p (x:xs)
   | otherwise = filter p xs
 ```
 
-#### Blank Lines
-
-One blank line between top-level definitions. No blank lines between type
-signatures and definitions of a function. Add one blank line between functions
-in a type class instance declaration if the function bodies are large. Use your
-judgement.
-
-#### Whitespace
-
-Surround binary operators with a single space on either side. Use your better
-judgement for the insertion of spaces around arithmetic operators but always be
-consistent about whitespace on either side of a binary operator. Don't insert a
-space after a lambda.
+Include newlines at the ends of files. Do not include other trailing whitespace on the end of lines.
 
 #### Data Declarations
 
@@ -68,15 +45,14 @@ data Tree a
 
 ```haskell
 data HttpException
-    = InvalidStatusCode Int
-    | MissingContentHeader
+  = InvalidStatusCode Int
+  | MissingContentHeader
 ```
 
 Format records as follows:
 
 ```haskell
-data Person
-  = Person
+data Person = Person
   { firstName :: !String  -- ^ First name
   , lastName  :: !String  -- ^ Last name
   , age       :: !Int     -- ^ Age
@@ -111,8 +87,7 @@ Format export lists as follows:
 
 ```haskell
 module Data.Set
-  (
-    -- * The @Set@ type
+  ( -- * The @Set@ type
     Set
   , empty
   , singleton
@@ -126,18 +101,15 @@ module Data.Set
 
 Imports should be grouped in the following order:
 
-1. standard library imports / related third party imports
-2. qualified imports of standard library / third party imports
-3. local application/library specific imports
-4. qualified imports of local application/library specific imports
+1. `Hasura.Prelude`
+2. `qualified` imports of standard library / third party imports
+3. unqualified standard library imports / related third party imports
+4. `qualified` imports of local application/library specific imports
+5. unqualified local application/library specific imports
 
 Put a blank line between each group of imports. The imports in each group should
-be sorted alphabetically, by module name (this is done by stylish-haskell
+be sorted alphabetically, by module name (this is done by `stylish-haskell`
 automatically).
-
-Always use explicit import lists or `qualified` imports for standard and third
-party libraries. This makes the code more robust against changes in these
-libraries. Exception: The Prelude.
 
 ### Naming
 
@@ -149,10 +121,6 @@ to `ColTy` and `computeAggregateFunction` to `compAggFn`. Exceptions: extremely
 common, idiomatic abbreviations, like using “err” to mean “error” or “ref” to
 mean “reference.” When in doubt about whether or not an abbreviation is
 well-known, prefer the full name.
-
-For readability reasons, don't capitalize all letters when using an
-abbreviation. For example, write `HttpServer` instead of `HTTPServer`.
-Exception: Two letter abbreviations, e.g. `IO`.
 
 #### Functions and variables
 
@@ -203,7 +171,7 @@ parseOptionalArgument userRole value >>= \case
 ```
 
 which avoids the need for the extra name entirely. If such an approach isn’t
-feasible or is hard to read, and a type-related prefix or suffix is absolutely
+practical or is hard to read, and a type-related prefix or suffix is absolutely
 necessary, prefer a prefix with the full name of the type over abbreviations:
 
 ```haskell
@@ -221,12 +189,6 @@ Use singular, not plural, module names, e.g. use `Data.Map` and
 `Data.ByteString.Internal` instead of `Data.Maps` and
 `Data.ByteString.Internals`.
 
-### Standard library usage
-
-#### Don't use `unless`
-
-Use `when` combined with `not`. Usage of `unless` can be confusing.
-
 ### Data types
 
 #### Wrap primitives in newtype (or sum types)
@@ -236,18 +198,57 @@ It is usually a good idea to introduce a custom data type (sum types or
 `Text`, etc.). This prevents confusion and mistakes of using the wrong types in
 wrong places.
 
-#### Use sum types instead `Bool` where it makes sense
+#### Prefer sum types to `Bool`
 
-To represent two different states, use sum types rather than a `Bool`.
-Specially, in the case of data types which represents enabled/disabled state.
-The problem with `Bool` is, it is not clear what the default is.
-
-Example:
+Avoid using `Bool` to represent two different states, especially if the set of possible states could potentially grow in the future. For example, instead of writing
 
 ```haskell
-data ConsoleEnabled
-  = CEDisabled
-  | CEEnabled
+-- bad
+data LiveQuery = LiveQuery
+  { ...
+  , lqIsPaused :: !Bool
+  , ...
+  }
+```
+
+prefer a sum type that has meaning even without context:
+
+```haskell
+-- good
+data LiveQuery = LiveQuery
+  { ...
+  , lqState :: !LiveQueryState
+  , ...
+  }
+
+data LiveQueryState
+  = LQActive
+  | LQPaused
+```
+
+Also, avoid defining boolean predicates that simply select distinguish between different constructors of a sum type. For example, avoid this:
+
+```haskell
+-- bad
+isPaused :: LiveQueryState -> Bool
+isPaused LQPaused = True
+isPaused _        = False
+
+...
+
+unless (isPaused queryState) do
+  handleActiveQuery query
+```
+
+This pattern can easily bite you if you later decide to extend the sum type with a new constructor. For example, suppose `LiveQueryState` was extended with an `LQStopped` constructor. Our guard using `isPaused` would still compile, but it would return `False` for stopped queries, and we’d call `handleActiveQuery` even though the query is stopped!
+
+Pattern-matching is much more robust to this kind of code evolution, and it isn’t very much more code:
+
+```haskell
+-- good
+case queryState of
+  LQActive -> handleActiveQuery query
+  LQPaused -> pure ()
 ```
 
 ### Dealing with laziness
@@ -263,8 +264,7 @@ evaluation order.
 
 ```haskell
 -- Good
-data Point
-  = Point
+data Point = Point
   { pointX :: !Double
   , pointY :: !Double
   }
@@ -272,8 +272,7 @@ data Point
 
 ```haskell
 -- Bad
-data Point
-  = Point
+data Point = Point
   { pointX :: Double
   , pointY :: Double
   }
@@ -305,9 +304,7 @@ Use line comments (comments that start with `--`) for short comments (1–3
 lines). For longer comments, use multiline comments (comments that begin with
 `{-` and end with `-}`).
 
-Use [Haddock syntax](https://haskell-haddock.readthedocs.io/en/latest/markup.html)
-for documentation comments. Running Haddock via `stack haddock` should always
-complete successfully.
+Use [Haddock syntax](https://haskell-haddock.readthedocs.io/en/latest/markup.html) for documentation comments. Running Haddock should always complete successfully.
 
 #### Module headers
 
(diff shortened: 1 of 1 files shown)


server/CONTRIBUTING.md
```diff
@@ -114,10 +113,249 @@ pytest --hge-urls http://127.0.0.1:8080 --pg-urls <database_url> -vv
...
+filter _ []   = []
+filter p (x:xs)
+  | p x       = x : filter p xs
+  | otherwise = filter p xs
+```
+
+#### Blank Lines
+
+One blank line between top-level definitions. No blank lines between type
+signatures and function definitions. Add one blank line between functions in a
```
rakeshkky: between functions' type signature and definition.

server/CONTRIBUTING.md
```diff
@@ -112,10 +111,251 @@ pytest --hge-urls http://127.0.0.1:8080 --pg-urls <database_url> -vv
...
 
+
 ## Code conventions
 
 This helps enforce a uniform style for all committers.
 
 - Compiler warnings are turned on, make sure your code has no warnings.
-- Use [hlint](https://github.com/ndmitchell/hlint) to make sure your code has no warnings.
-- Use [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) to format your code.
+- Use [hlint](https://github.com/ndmitchell/hlint) to make sure your code has no
```
jberryman: If `hlint` and `stylish-haskell` are to be part of the official project style it would be good to integrate them into the test suite. That way we avoid muddying up commit history with superficial changes when style violations inevitably get committed.

server/CONTRIBUTING.md
```diff
@@ -112,10 +111,251 @@ pytest --hge-urls http://127.0.0.1:8080 --pg-urls <database_url> -vv
...
+This is a short document describing the preferred coding style for this project.
+We've tried to cover the major areas of formatting and naming. When something
+isn't covered by this guide you should stay consistent with the code in the
+other modules.
+
+### Formatting
+
+#### Line Length
+
+Maximum line length is *80 characters*.
```
jberryman: I personally prefer 100 - 120 chars, but don't feel strongly about this.

server/CONTRIBUTING.md
```diff
@@ -112,10 +111,251 @@ pytest --hge-urls http://127.0.0.1:8080 --pg-urls <database_url> -vv
...
+    Set
+  , empty
+  , singleton
+
+    -- * Querying
+  , member
+  ) where
+```
+
+### Imports
```
jberryman: `stylish-haskell` already manages the import list right? I don't have a strong opinion on this, except that if `stylish-haskell` is to be used it needs to be idempotent for a properly formatted (according to this guide) file. Maybe there's a way to configure it to do what we want?

server/CONTRIBUTING.md
```diff
@@ -112,10 +111,251 @@ pytest --hge-urls http://127.0.0.1:8080 --pg-urls <database_url> -vv
...
+data ConsoleEnabled
+  = CEDisabled
+  | CEEnabled
+```
+
+### Dealing with laziness
+
+By default, use strict data types and lazy functions.
+
+#### Data types
```
jberryman: I think this is good advice, but why not actually enforce this by configuring `StrictData` project-wide (along with `-funbox-strict-fields`)? I've never worked on a project that did that, but have often thought it might be a good idea. Obviously a little care would need to be taken porting the project to this initially. Thoughts?

server/CONTRIBUTING.md
lexi-lambda: I agree—80 characters seems very small to me, and it seems more likely to encourage artificially short names.

server/CONTRIBUTING.md
ecthiender: Cool! Sounds like a good idea! I'll create an issue for this so we can track it.

server/CONTRIBUTING.md
ecthiender: Agreed. I'll make this ~100 chars?

server/CONTRIBUTING.md
ecthiender: Yeah, I'm not sure. stylish-haskell sorts them alphabetically if there are no newlines in between. We wanted to sort external libs first then hasura libs -> then qualified imports of external libs then hasura libs. I should look if there is stylish haskell configuration to do that. But I don't strongly feel about this, we can probably keep the stylish-haskell defaults.

server/CONTRIBUTING.md
ecthiender: I have never done this as well.

server/CONTRIBUTING.md
ecthiender: https://github.com/hasura/graphql-engine/issues/2490

lexi-lambda:
A thought I’ve had while reading through the codebase: this isn’t really Haskell-specific, but one thing I’d find really helpful is module-level comments. Nothing long—it can really just be a sentence—but essentially what I’m asking for is for each module to come with a “purpose statement.” For example, something like this:

```haskell
-- | This module provides types and functions for generating a GraphQL schema
-- from information about Postgres tables.
module Hasura.GraphQL.Context where
```

Or like this:

```haskell
-- | This module defines the top-level HTTP routes and handlers for the server
-- and provides them as a WAI application.
module Hasura.Server.App where
```

It doesn’t have to be for every module—these comments would be pretty pointless for the various `*.Extended` modules, for example—but it would have been very helpful to me as someone looking through the codebase for the first time. It also has the useful side-effect of providing some small pressure against creating modules that are just grab-bags of different things, or at least recognizing when a module has gotten to that point.

Thoughts?

ecthiender:
@lexi-lambda Yes! We lack this currently, but we have slowly started to add module-level comments whenever we work on a PR (on the modules that requires changes for that PR).

lexi-lambda:
Some other ideas I’ve had after a little more time working in the codebase:

  1. On a previous Haskell team I was on, we adopted a policy preferring newer/more generic versions of operations, e.g. preferring `pure` to `return`, `traverse` to `mapM`, `*>` to `>>`, etc. I think that policy worked out well, since it can be a bit confusing to have a mix of both, and Haskell’s history means it has a lot of that kind of redundancy.

     The only place where we usually made an exception to this rule was `map` versus `fmap`. It’s so overwhelmingly idiomatic to use `map` instead of `fmap` or `<$>` when mapping over lists that using `fmap` instead seemed to harm readability rather than helping it. I don’t have a super strong opinion about that, though, so I’d be okay with either.

  2. I’d like to propose aggressively moving helper functions into `where` clauses whenever possible (where they’re currently defined at the top-level), but keeping the type signatures. I first discovered this style when reading the PureScript code base, and it works very well: it helps make it very explicit which functions are the “entrypoints” to a module and which functions are internal.

     Taking this to its logical conclusion can sometimes result in basically an entire module being in a single function’s `where` clause, but because type signatures are preserved, it doesn’t reduce readability, and it can even help by ensuring that the top-level function appears first. For example, in `Hasura/GraphQL/Context.hs`, we might end up with

     ```haskell
     mkGCtx :: TyAgg -> RootFlds -> InsCtxMap -> GCtx
     mkGCtx tyAgg (RootFlds flds) insCtxMap =
       {- top-level implementation here ... -}
       where
         mkCompExpInp :: PGColType -> InpObjTyInfo
         mkCompExpInp = {- ... -}

         mkCompExpName :: PGColType -> G.Name
         mkCompExpName pgColTy = {- ... -}

         mkCompExpTy :: PGColType -> G.NamedType
         mkCompExpTy = {- ... -}

         {- etc. -}
     ```

     which is a very big `where` clause, but helps a lot with navigating the code.

     The one downside to this is that you can’t put datatype declarations in `where` clauses, so if internal datatypes are defined for particular helper functions, it can be helpful to still break those out into the module top level so that they can be put next to their datatype definitions.

  3. As a sort of corollary to the previous point, I think nesting `where` clauses is okay. That is, instead of writing

     ```haskell
     f :: Integer -> Integer
     f x = g x True
       where
         g :: Integer -> Bool -> Integer
         g a b = {- ... lots of code that uses `h` ... -}

         h :: String -> Integer
         h x y z = {- ... more code ... -}
     ```

     write

     ```haskell
     f :: Integer -> Integer
     f x = g x True
       where
         g :: Integer -> Bool -> Integer
         g a b = {- ... lots of code that uses `h` ... -}
           where
             h :: String -> Integer
             h x y z = {- ... lots more code ... -}
     ```

     Doing this beyond two or three levels can make the rightward drift a little ridiculous, but in moderation, it helps to make the code a lot clearer, especially since most editors support code folding `where` clauses, so it’s easy to hide parts of the code that aren’t relevant.

     Also, to be clear: I don’t think doing this is worth it for a handful of small functions, since it just adds a lot of noise. But if there are tons of little functions or just a handful of big functions, it can make a positive readability difference.

  4. This is more minor than the above three points, but I think it’d be good to agree on how we all feel about guards versus `if...then...else` versus the `bool` function. I’ve noticed a lot of places in the codebase seem to use `bool` like

     ```haskell
     someFunction = {- ... -}
       where
         foo = bool (thenExpr bar) (elseExpr baz) conditionExpr
     ```

     while I would probably write

     ```haskell
     foo = if conditionExpr then thenExpr bar else elseExpr baz
     ```

     or even

     ```haskell
     foo | conditionExpr = thenExpr bar
         | otherwise = elseExpr baz
     ```

     though I get the sense a lot of people don’t realize you can use guards on bindings that aren’t functions!

     Personally, I don’t think most of the current uses of `bool` are especially helpful for readability, since they put the boolean condition at the end instead of at the beginning. I think `bool` is most useful when partially applied to two arguments, but I prefer guards or `if...then...else` in other situations. It’s not a super strong opinion, though.

jberryman:
@lexi-lambda 

1. Sounds good to me
2. This is something I feel strongly about as well. Another benefit is that often the functions become smaller as they're already closed over the arguments you would have passed. I think this is a specific case of general principles of DRY and trying to keep the scopes of things tight.
3. Sounds good to me
4. `bool` seems unusual to me as well. I don't feel strongly but I think I would prefer guard syntax over any of the other options, mostly because it's most useful and ubiquitous. e.g. they are used in `MultiWayIf` (another option to perhaps prefer over `if` or `bool`), as well as `case`. And the latter is the most powerful way to do conditional branching because you can combine both pattern matching and predicates ( `LambdaCase` I think is also very nice):
```
    fooWithOneArg = \case
      A x _ 
          | isBlargh x -> x
      A x y -> x + y
      something 
          | isBorked something -> 0
      B x y z -> x + y + z
```

I have some similar little syntax and style things that I would like to propose as well, as long as folks think that is still in scope here. I wonder if it makes sense to have a style guide divided into two sections:
- here are rules you should follow (line length etc.)
- here are tools and strategies you can employ for code clarity, and some opinionated conclusions

...maybe those are the same thing.

Interested to hear others' thoughts as well.

tirumaraiselvan:
One comment from my side :)

Shall we use the following record syntax to populate records:

```
emptySchemaCache :: SchemaCache
emptySchemaCache =
  SchemaCache
  { scTables = M.empty
  , scFunctions = M.empty
  , scRemoteSchemas = M.empty
  , scAllowlist = HS.empty
  , scGCtxMap = M.empty
  , scDefaultRemoteGCtx = GC.emptyGCtx
  , scDepMap = M.empty
  , scInconsistentObjs = []
  , scRemoteRelInputTypes = M.empty
  }
```
instead of doing

```
emptySchemaCache :: SchemaCache
emptySchemaCache =
  SchemaCache M.empty M.empty M.empty
              HS.empty M.empty GC.emptyGCtx mempty [] M.empty
```

jberryman:
A general comment from me on the lines of the above. We should use strategies to avoid the proliferation of very short-lived, throwaway names. Good strategies I've found include:

- Taking some time to try to organize functions as compositions and using tasteful point-free style, e.g.

        parseGingerTmplt src = either parseE Right res
          where
            res = runIdentity $ TG.parseGinger' parserOptions src
            parserOptions = TG.mkParserOptions resolver
            resolver = const $ return Nothing
            parseE e = Left $ TG.formatParserError (Just "") e

    vs.

        parseGingerTmplt = left fmtErr . runIdentity . TG.parseGinger' parserOptions
          where
            parserOptions = TG.mkParserOptions resolver
            resolver = const $ return Nothing
            fmtErr = TG.formatParserError (Just "")

    `LambdaCase` can also be helpful here:

        foo <- doBlarg
        case foo of
          A -> ...

    vs

        doBlarg >>= \case
          A -> ...

- using `RecordWildCards` or `NamedFieldPuns`: these let you use only record label identifiers everywhere, and work both in patterns and in constructing. `NamedFieldPuns` is more explicit and some people prefer it so that might be better to standardize on, but I generally prefer `RecordWildCards` and haven't had problems with it,

**Constructing**:

```
  ...
  enableAL <- withEnvBool (rsoEnableAllowlist rso) $ fst enableAllowlistEnv
  enabledLogs <- Set.fromList . fromMaybe (Set.toList L.defaultEnabledLogTypes) <$>
                 withEnv (rsoEnabledLogTypes rso) (fst enabledLogsEnv)
  serverLogLevel <- fromMaybe L.LevelInfo <$> withEnv (rsoLogLevel rso) (fst logLevelEnv)
  return $ ServeOptions port host connParams txIso adminScrt authHook jwtSecret
                        unAuthRole corsCfg enableConsole consoleAssetsDir
                        enableTelemetry strfyNum enabledAPIs lqOpts enableAL
                        enabledLogs serverLogLevel
```
vs...

```
  ...
  soEnableAllowList <- withEnvBool (rsoEnableAllowlist rso) $ fst enableAllowlistEnv
  soEnableLogTypes <- Set.fromList . fromMaybe (Set.toList L.defaultEnabledLogTypes) <$>
                 withEnv (rsoEnabledLogTypes rso) (fst enabledLogsEnv)
  soLogLevel <- fromMaybe L.LevelInfo <$> withEnv (rsoLogLevel rso) (fst logLevelEnv)
  return $ ServeOptions {..}  -- RecordWildCards SYNTAX HERE
```

**binding in pattern matching**

```
mkConnInfo (RawConnInfo mHost mPort mUser pass mURL mDB opts mRetries) =
  case (mHost, mPort, mUser, mDB, mURL) of
```

vs


```
mkConnInfo RawConnInfo{..} =
  case (connHost, connPort, connUser, connDB, connUrl) of
     ...
```

Along these same lines, `DuplicateRecordFields` works well with `RecordWildCards` and works well for types where there are corresponding fields that have the same type and semantics (and so we'd like to give them the same name as well)

lexi-lambda:
On the topic of catamorphisms such as `bool`, `maybe`, and `either`, here’s an unusual Haskell pattern I “discovered” today. You can use [`Data.Function.&`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Function.html#v:-38-) to apply those kinds of functions in “control flow” order, so the thing they scrutinize comes first:

```haskell
let foo = isValid x & bool (Just x) Nothing
    bar = (y > 42) & bool
      (doSomething a b)
      (doSomethingElse b c)
```

Since `&` has a higher precedence than `$`, it still works if you want to apply some function to the result:

```haskell
fn key = pure $ lookup key dict & maybe
  (someDefault a b)
  (someFn b)
```

Is this good? Is it terrible? I don’t know, but I kind of like it, especially for simple boolean branching. I find `if` kind of unwieldy in Haskell, and sometimes I wish it were just a function, but I often find following the control flow of code that uses `bool` unintuitive. `c & bool a b` feels pretty close to having the best of both worlds, but I know some people have strong opinions about infix operators, so I won’t be offended if others disagree. :)

jberryman:
this is the sort of thing I'd like to avoid in https://github.com/hasura/graphql-engine/pull/2175#discussion_r301187947:

merge

lexi-lambda:
@jberryman You can use the `-Xignore-space-change` flag with any of the `git` commands that do merging to mostly avoid that problem.

jberryman:
Oh sweet, I didn't know you could actually use that in a merge

jberryman:
Minor proposed addition:

Avoid parenthesized expressions that cover multiple lines. Strategies when you write these:
- `LambdaCase`, or just `$` ... e.g. `forM_ xs $ \x -> -- multiline expression indented below`
- judicious use of `flip`, for instance we can avoid both parens and the difficult-to-read trailing argument with:
```
                flip Map.traverseWithKey hash $
                  \key rows ->
                    case Map.lookup key keyedMap of
                      Nothing ->
                        ...
```
- giving a lambda a name and moving to a `where` clause

jberryman:
related to above, and point-free style:

- prefer to name functions in `where` clauses, rather than naming results or values, and compose them at a higher level

This has the benefits of:
- making refactoring easier, and making it more noticeable when you have a bit of functionality that could be factored out for re-use
- naming functions provides better documentation; often when we name a value we seem to choose a name that has little meaning (or carries no semantics that's not already in the type)

lorenzo:
For future iterations, did you consider using more complete formatted such as britanny or ormolu?

server/CONTRIBUTING.md
jberryman: https://github.com/hasura/graphql-engine/issues/3941

server/STYLE.md
```diff
@@ -0,0 +1,505 @@
...
+
+This is a short document describing the preferred coding style for this project. We've tried to cover the major areas of formatting and naming. When something isn't covered by this guide, err on the side of consistency with existing code.
+
+### Formatting
+
+Use [stylish-haskell](https://github.com/jaspervdj/stylish-haskell) to format your code.
+
+#### Line Length
+
+Try to keep a line length within 100 characters. Small exceptions are okay if the line only goes over the limit by a few characters and breaking it would negatively impact readability. Use your best judgement.
```
0x777, line 11: The problem I have with 100 characters is that it doesn't work well in settings where you want to open an another window on the side (say hoogle) or view more than one file at the same time. On my 15.6" laptop with 12pt font, this is how it looks:

A single file, looks great! 

image

Now if I have two files open:

image

This is how it looks with a max line-length of 85 characters (formatted with vim's `gw`)

image

The readability is much better with line-length at at 85 characters in this case. Now the question is how common is this pattern of having something else next to the file that you are editing, be it another file or a browser window, I honestly don't know, but I think we should be accommodating to devs who prefer such a workflow if it isn't too much of an inconvenience to others. I could be wrong but on a 14" laptop with 11pt font, 80-85 character limit should fit two files without any line wrapping.

lexi-lambda: For what it’s worth, I’ve started feeling like wrapping *comments* at 100 characters is a bad idea, too. I rarely do horizontal splits, but even without them, the wide lines are just less pleasant to read! More recently, I’ve been wrapping all my comments at 70–80 characters rather than 100.

That said, I really don’t like an 80 character limit on code. I would much rather see

```haskell
      loggerCtx <- liftIO $ mkLoggerCtx (defaultLoggerSettings True logLevel) enabledLogs
```

than

```haskell
      loggerCtx <- liftIO $ mkLoggerCtx (defaultLoggerSettings True logLevel)
                                        enabledLogs
```

just because the latter was slightly too long. In my experience, setting a limit at 80 characters encourages two bad habits:

  1. It makes people less willing to create more nested binding contexts. Suddenly putting a `where` inside a `where` has a cost—instead of 80 characters, you get just 72. Now have a `let` inside a `do` block in that nested binding, and you’re working with 66 characters. Things get cramped quickly, so people tend to float things out more… but I think that often hurts readability more than it helps!

  2. It encourages shorter names. In the example I gave above, I might question whether I needed to use all those characters. Could I use `defLogSettings` and `logLvl`? Might as well write `mkPermMetaObj` instead of `mkPermissionMetadataObject`… or something even more cryptic.

     I would much rather have to soft wrap or horizontally scroll a little bit and have nice, readable variable names than stare at abbreviations that don’t make sense to me!

So I’d be happy to tighten up the length guideline for comments, but I’d rather not go below 100 on the code line length limit. (If anything, be glad I’m not lobbying for something even *longer.* 😁)

0x777:
@jberryman To avoid that formatting hell, we can start using [ImportQualifiedPost](https://downloads.haskell.org/~ghc/8.10.1/docs/html/users_guide/glasgow_exts.html#extension-ImportQualifiedPost)! Available from 8.10. 

@lexi-lambda @ecthiender This looks great! Thanks for taking the time to use appropriate examples  from our current code base! Since some parts have to change to adhere to this style guide, how do we go about doing this? Leave the camp-ground cleaner approach?

lexi-lambda:
> Since some parts have to change to adhere to this style guide, how do we go about doing this? Leave the camp-ground cleaner approach?

Yes, pretty much. Auto-formatting is something that could be worth exploring, since `stylish-haskell` is pretty limited, but all the existing auto-formatters I know of still struggle to handle certain things, so I’m not sure how much that makes sense to do right now. (And of course, some of these things are impossible to enforce with a formatter anyway.)

server/STYLE.md
0x777: When we use stop using cryptic variable names, I think we'll see the line length go beyond 100 characters more often than not (even in the above example, expanding `Ctx` to `Context` will result in a much longer line), so I think that we will need to get used to such things:

```haskell
      loggerCtx <- liftIO $ mkLoggerCtx (defaultLoggerSettings True logLevel)
                                        enabledLogs
```

In terms of readability, I prefer this to soft-wrap any day because soft wraps don't respect indentation! Horizontal scroll is equally bad I feel, you can't see everything with a quick glance.

> It makes people less willing to create more nested binding contexts. Suddenly putting a where inside a where has a cost—instead of 80 characters, you get just 72. Now have a let inside a do block in that nested binding, and you’re working with 66 characters. Things get cramped quickly, so people tend to float things out more… but I think that often hurts readability more than it helps!

I agree with all this but at the same time I think 100 characters too won't be enough with our naming convention! How about a soft limit at 85-90 characters but if the developer thinks that using more characters improves readability they can definitely do that?

server/STYLE.md
```diff
@@ -0,0 +1,505 @@
...
+```
+
+### Imports
+
+Imports should be grouped in the following order:
+
+1. `Hasura.Prelude`
+2. `qualified` imports of standard library / third party imports
+3. unqualified standard library imports / related third party imports
+4. `qualified` imports of local application/library specific imports
```
nicuveo, line 107: Do we have specific guidelines on how imports should be abbreviated?
- import qualified Data.Map as Map
- import qualified Data.Map as M

And do we have specific guidelines between qualified / unqualified imports, with or without explicit import lists?

0x777: @nicuveo We used to have `as M` previously, but currently we use more readable names like Map, Json, Set etc. 

> And do we have specific guidelines between qualified / unqualified imports, with or without explicit import lists?

I don't think we have a guideline for but the existing style is to use a qualified import for external libraries and we import frequently used internal modules directly.

lexi-lambda: An older version of this style guide recommended using either qualified imports or explicit import lists for external libraries, but I removed that sentence as part of the cleanup, since it didn’t seem like we were actually following that guideline, and I’m personally fine with using unqualified imports without an import list. ¯\_(ツ)_/¯

If someone feels strongly about it, by all means, make a case for it—I would consider adding it back! But I’d rather not codify things in the style guide unless either (a) we can automatically enforce them or (b) we all agree on them enough to enforce them ourselves. Otherwise, adding them to the style guide seems unlikely to do much.

server/STYLE.md
lexi-lambda: I have updated the section on line length to suggest a target length of 80 characters. Let me know if you’re happy with the new wording!

PR #2192, opened 2019-05-15
bundle console assets into server (close #516, close #521, close #2130)

### LINKED ISSUE #516: add a flag/env var to load console assets from a local directory path
While developing server locally, it cannot serve console since it tries to fetch assets from google cdn. 

Server should take a local directory path to load console assets. User can put the asset files in a directory and get console working on server.

### LINKED ISSUE #521: If console CDN is not available, fall back to a pre-packaged version of the console?
(via @thangngoc on discord)

Offline development is a bit of a pain otherwise.

@praveenweb @karthikvt26 What do you guys think about this? Pros/cons?

### LINKED ISSUE #2130: Hasura docker without network connection
Hi guys,
how can i create a docker image of hasura that doesn't make request for the vendor.js, main.js and main.css etc. ?

shahidhk:
This PR builds console static assets into the server docker image at `/srv/console-assets`. When env var `HASURA_GRAPHQL_CONSOLE_ASSETS_DIR=/srv/console-assets` or flag `--console-assets-dir=/srv/console-assets` is set on the server, the files in this directory are served at `/console/assets/*`.

The console html template will have a variable called `cdnAssets: false` when this flag is set and it loads assets from server itself instead of CDN.

The assets are moved to a new bucket with a new naming scheme:

```
graphql-engine-cdn.hasura.io/console/assets/
   /common/{}
   /versioned/<version/{}
   /channel/<channel>/<version>/{}
```

Console served by CLI will still load assets from CDN - will fix that in the next release.


server/stack.yaml
```diff
@@ -2,7 +2,7 @@
 
 # Specifies the GHC version and set of packages available (e.g., lts-3.5, nightly-2015-09-21, ghc-7.10.2)
 # resolver: lts-10.8
-resolver: lts-13.20
+resolver: lts-13.16
```
shahidhk: revert this change

server/src-lib/Hasura/Server/Version.hs
```diff
@@ -1,11 +1,12 @@
+{-# OPTIONS_GHC -fno-warn-type-defaults #-}
```
0x777: not needed

mnlbox:
@shahidhk Based on this PR we need download asset file or use another docker image?
If we need download so please add some download link to docs or download page :wink:

shahidhk:
@mnlbox https://docs.hasura.io/1.0/graphql/manual/deployment/graphql-engine-flags/config-examples.html#console-assets-on-server

PR #2197, opened 2019-05-16
ignore content-type header in auth webhook

ecthiender:
### Description

Ignore `content-type` header from client while sending request to auth webhook.

### Affected components 

- Server


aam1r:
Should this section of the docs be also updated?

image

ecthiender:
> Should this section of the docs be also updated?
> 
> image

Done

Issue #2060, opened 2019-04-24
foreign key constraint modification issues


0x777:
Currently, you cannot modify a foreign key constraint's cascade properties if there are relationships defined using that constraint. The underlying reason is that postgres doesn't allow altering the contraint's properties, so we resort to dropping and adding it again with the desired properties.

To not trigger the dependent objects error message we need two changes:
1. The console should issue a single statement to drop and add instead of a bulk query. Something like this:

   ```sql
   alter table public.scores
   drop constraint scores_gid_fkey,
   add constraint scores_gid_fkey
      foreign key (gid)
      references games(gid)
      on delete cascade;
   ```
2. The server side diff in `run_sql` shouldn't use oids for foreign key constraints but instead look at the definition to see if the constraint is dropped/altered.

PR #2252, opened 2019-05-22
do not consider foreign keys are dropped only if constraint oid changes (fix #2238, #2060)

### LINKED ISSUE #2238: Allow migrations to be forced, and to sync metadata after applied
I'm attempting to modify a schema where dependent tables weren't set up to cascade on delete. This seems like it should be a simple change, but it isn't, and I'm wondering if the SQL console can give me an "I know what I'm doing" escape hatch in the form of running an arbitrary query, then resyncing metadata after the query completes?

Essentially, my SQL looks like this:

```sql
BEGIN;
ALTER TABLE ONLY public.ocr_job
    DROP CONSTRAINT ocr_job_output_id_fkey,
    ADD CONSTRAINT ocr_job_output_id_fkey FOREIGN KEY (output_id) REFERENCES public.output(id) ON UPDATE CASCADE ON DELETE CASCADE;
...
COMMIT;
```

I've ensured that all foreign key names are preserved, and that they're modified in a transaction.

Unfortunately, running this migration in the console wants me to destroy and recreate relationships, permissions, etc. and there are quite a few. I'd understand the above restrictions if I was arbitrarily running SQL and not being cautious, but it seems to me like this shouldn't cause an issue.

If I run this SQL directly on the psql console, then refresh metadata in the console, my changes seem to get picked up and represented accurately. I don't know Hasura well enough to know if something is broken and I just haven't found it yet, but this seems like a case where the console is fighting me, and I hope it has a good reason to do so. :)

It also looks as if I'll have to wipe out our database, start with an initial migration with the current state, then religiously ensure that cascades are correctly set going forward.

Can we get some sort of escape hatch? :) And as an aside, is there some other way to fix this without starting over?

Thanks.

rakeshkky:
TODO:- ~~console should issue combined SQL alter statement on modifying foreign key constraints~~

### Description

Fixes:- server throws dependency error when user trying to alter a foreign key constraint to define cascade attribute when there is relationship defined on the foreign key.

### Affected components 

- Server
- Tests
- Console

### Related Issues

fix #2238, #2060 

### Solution and Design

"A foreign key is considered to be dropped only if it's `constraint oid` **and** `custom unique id` changes where a `custom unique id` is a collective reference to `column mapping` and `reference table` "

### Steps to test and verify

- Define a Postgres schema with foreign keys
- Track tables and relationship based on foreign keys
- Alter a foreign key constraint without changing the reference table and column mapping


rikinsk:
Need to fix constraint name generated. Currently it adds a suffix even if original is going to be deleted.

Issue #1637, opened 2019-02-20
ra-data-hasura add filter support for "ReferenceInput"


mnlbox:
Hi guys,
I'm using ra-data-hasura and it's working well. Really thanks for this.
I want to use filter in a "ReferenceInput".
I have two table:
1. user (id, title)
2. posts (id, title, user_id [relation called `author`] )
In post creation field I want to filter ReferenceInput to load all user except **admin** as a select input for select author (in my system admin don't need to use as a author because it's only for administration task)

Now I use below code for my create form:
```
export const PostCreate = (props) => (
    <Create {...props}>
        <SimpleForm redirect="list">
            <TextInput source="title" resettable />
            <ReferenceInput source="user_id" reference="user" label="User" filter={{ "author.title": "admin" }}>
                <SelectInput optionText="title" resettable />
            </ReferenceInput>
        </SimpleForm>
    </Create>
);
```

How we can set `filter={{ "author.title": "admin" }}` to something like `!== "admin` and is it possible to use hasura relation in filter?
Is it possible to add relation support to ra-data-hasura like `author.title`?

mnlbox:
@praveenweb Any news about this?

PR #2284, opened 2019-05-27
ra-data-hasura: add filters to GET_* types

joserocha3:
### Description
Set filters for some type. The filters were not being set, therefore they any filters added to list type components were being ignored. This was done for GET_LIST, GET_MANY and GET_MANY_REFERENCE types.

### Affected components 
- ra-data-hasura

### Solution and Design
Created `addFilter` function to add filters to `args.where`. 

### Steps to test and verify
Add a `filter` prop to a`ReferenceManyField` component. You will notice the filter has no effect. This change will apply the filter as expected.

```js
<ReferenceManyField
  filter={{ type: 'geofence' }}
  reference="event"
  target="account_id"
>
```


mnlbox:
@joserocha3 Thanks for this :wink:
Maybe related: #1637

PR #2304, opened 2019-05-30
fix: head to home console error from error page

strongSoda:
### Affected components 

- Console


strongSoda:
doing this renders the 404 page

PR #2305, opened 2019-05-30
forward cookie headers from remote servers (fix #1654)

### LINKED ISSUE #1654: Hasura doesn't forward Set-Cookie headers set by remote schemas

ecthiender:
### Description

Forward only `Set-Cookie` header received from remote schema servers, to client.

**Note**: This won't send back any remote response headers when the websocket transport is used.

### Affected components 

- Server
- Tests

### Related Issues

#1654 
### Solution and Design

Introduced new data type `HttpResponse` which encapsulates headers and some response body.  Changed corresponding handlers. It only sends `Set-Cookie` header from remote.

### Steps to test and verify

1. Add a remote schema which returns say `Set-Cookie` header
2. Make a remote schema query
3. Observe if the cookie is set in the browser


server/src-lib/Hasura/GraphQL/Execute.hs
```diff
@@ -361,10 +366,15 @@ execRemoteGQ manager userInfo reqHdrs q rsi opDef = do
...
+    filteredHeaders = filterUserVars $ filterRequestHeaders reqHdrs
 
     filterUserVars hdrs =
       let txHdrs = map (\(n, v) -> (bsToTxt $ CI.original n, bsToTxt v)) hdrs
       in map (\(k, v) -> (CI.mk $ CS.cs k, CS.cs v)) $
          filter (not . isUserVar . fst) txHdrs
+
+    getCookieHdr = maybe [] (\h -> [("Set-Cookie", h)])
+
+    mkRespHeaders hdrs =
```
0x777: We don't need this and `filterResponseHeaders`

Issue #2290, opened 2019-05-28
Add a `migration_mode` flag to the `console` CLI options


KCErb:
When using the cli `console` to modify your metadata, you can turn off migration_mode with a little toggle switch. This is great for when you are managing your own schema and don't want hasura to touch anything except its own metadata.

The trouble is that this little toggle is stateful, so if you're running out of docker, you have to remember to flip that switch each time you spin up a container.

It'd be great to just have a flag you can pass:

```
hasura console --migration-mode false
```

or

```
hasura console --no-migrations
```

The work around at the moment is to either:
* use curl (or similar) to mimic the API call that does this:

```
curl -H 'Content-Type: application/json' -X PUT -d '{name: "migration_mode", value: "false"}' http://0.0.0.0:9693/apis/migrate/settings
```
but that actually doesn't work because cookies guard against such a thing, so either use your favorite "stores cookies in memory" library and script it, or use curl and write them to a file

* execute some SQL when you spin up your postgres container:

```
CREATE TABLE IF NOT EXISTS hdb_catalog.migration_settings (setting text not null primary key, value text not null);

INSERT INTO hdb_catalog.migration_settings (setting, value) VALUES ('migration_mode', 'false') ON CONFLICT ON CONSTRAINT migration_settings_pkey 

DO NOTHING;
```

(Thank you to the awesome support team, and especially @shahidhk for the required SQL.)

shahidhk:
The best way to do this would be to have a command dedicated to doing this, instead of a flag for console command.

A flag for console command indicates that this setting is for the session of the console, but this setting is persisted on the database and is available to all consoles connected to this database.

The command could look like this:
```
hasura settings --get <key>
hasura settings --set <key>=<value>
```

Suggestions/comments are invited.

#### Implementation
The plumbing required to do this already written. The function is [`UpdateSetting`](https://github.com/hasura/graphql-engine/blob/master/cli/migrate/database/settings.go#L20) and it is implemented in the [`hasuradb` driver](https://github.com/hasura/graphql-engine/blob/master/cli/migrate/database/hasuradb/settings.go#L175).

KCErb:
Awesome, I'll see if I can't put together a PR for this this week.

KCErb:
I think I've got a good start here, but not sure how to get the right parameters [right here](https://github.com/KCErb/graphql-engine/blob/5feef8443df23eafbc17ec7413b138be5f4d4eaf/cli/commands/settings.go#L49-L50).

[...]

PR #2321, opened 2019-06-03
read/update migration settings from cli

KCErb:
### Description
This PR adds a `settings` subcommand to the `migration` command and contains two further subcommands `set` and `get`.

### Affected components 
- CLI

### Related Issues
#2290 

### Solution and Design
Mostly follows patterns of other CLI commands, nothing new in design here. Not sure if `SUCCESS` is the right thing to print when `set` runs successfully.

### Steps to test and verify
1. Spin up the test docker container
2. Build the cli and copy it into the container
  (`docker cp _output/cli-settings-de3bd50-dirty/cli-hasura-linux-amd64 a308:/bin`)
3. Init a migration folder `/bin/li-hasura-linux-amd64 init` and `cd` into it.
4. Run `cli-hasura-linux-amd64 migrate settings get migration_mode` should get "true"
5. `cli-hasura-linux-amd64 migrate settings set migration_mode=false` should print "SUCCESS"
6. Now `cli-hasura-linux-amd64 migrate settings get migration_mode` should return "false" and the actual console switch is off if you inspect it as well.

I wonder about tests. I didn't add any and I'm not sure I know enough about Go to really write them effectively. I noticed that you can set any string to any value here and if you try to get a value that doesn't exist you get an unhandled error, but those are in the underlying driver it seems.

fix issue #2290


arvi3411301:
@KCErb can you also add endpoint and admin secret flag. You can refer https://github.com/hasura/graphql-engine/blob/master/cli/commands/migrate_status.go#L43

KCErb:
@arvi3411301 Do you mean like so? (see most recent push)

I don't see `o.EC.ServerConfig.AdminSecret` getting updated anywhere so I don't see the effect this has, I'm guessing some other system is picking it up and propagating it to the EC here?

arvi3411301:
@KCErb ya correct.

You can refer https://github.com/hasura/graphql-engine/blob/master/cli/cli.go#L314 where both endpoint and admin_secret gets updated. Due to an issue with viper, we need to add these flags on all the commands.

mnlbox:
@KCErb Is it only for CLI or we also able to set this through console app?

KCErb:
@mnlbox There is already a switch labeled "Turn off migrations mode" in the Migration tab on sidebar under the Data tab. This PR adds the ability to flip the switch from the CLI.

KCErb:
anything missing on my end to get this in?

shahidhk:
No, it's pending review from our side. 😅

arvi3411301:
@KCErb Can you also add `access-key` flag on the commands and mark it as `deprecated`.

You can refer https://github.com/hasura/graphql-engine/blob/master/cli/commands/migrate_status.go#L46

KCErb:
@arvi3411301 yes, I will do that. Sorry for the slow response time. I'll probably have time for that this week. Seems very simple :)

KCErb:
@arvi3411301 nevermind, looks like opening the code and looking at it was half the battle, thanks for pointing me to an example so that I could just paste it in. All done now :)

KCErb:
@shahidhk @arvi3411301 
Ping, anything blocking this? :)

shahidhk:
@KCErb We're working on some overhauls that affects this part too. Will you be able to tell us how and why you're using migration mode off setting?

KCErb:
I think I described it in the connected issue: https://github.com/hasura/graphql-engine/issues/2290

Perhaps you could ask a clarifying question if that is unclear?

KCErb:
Hi folks, I forgot about this PR and am finally back onto my hasura project :tada:

I can see you went to v1 some time back. Is this still relevant? Is there a nice(r) way to programmatically set the migrations mode yet?

Issue #2205, opened 2019-05-16
Support JWT from both Authorization header of Cookie


campie:
Hi,

Referencing #2183, it would be great to have **--jwt-from-cookie <cookie_name>**.

This will solve the problem of having to write an auth-webhook every time we want to store the jwt client-side in a http-only cookie (secure=true http-only=true same-site=strict; more on this below). 

IMHO, we should **NEVER STORE JWT in LocalStorage** or in anything else JS has access to. For browser clients, using http-only cookie is currently the only solution to this problem (that I'm aware of).

However, the Hasura implementation should add some _protection against csrf attacks_ for old browsers that do not support same-site flag on cookies (more on this below).

I'm already using "jwt cookies" with Hasura and _this solution will allow me to ditch the auth-webhook altogether_.

I really would like to help design this solution so I wrote this.

**Proposal for Hasura in JWT mode supporting both Authorization: Bearer < JWT > header and jwt inside cookies**
(backward compatible)

Additional Hasura Graphql Engine flags:

**--jwt-from-cookie <cookie_name>** (no default value, disabled by default)
**--jwt-csrf-token <jwt_payload_csrf_token_property_name>** (default value: csrf-token)
**--csrf-token-header <csrf_token_header_name>** (default value: equals to --jwt-csrf-token)
**--csrf-token-disabled <boolean>** (dafault value: false)

To authorize an incoming request:

**1 -** Hasura looks for an _Authorization: Bearer < JWT >_ header. If it finds it, Hasura skips to step 7 (jwt verification).

(Servers and mobile clients can still request the Hasura api by sending the jwt in the Authorization header. For browser clients, an http-only cookie containing the jwt is preferable)

**2 -** If no Authorization header is found, then Hasura checks if _--jwt-from-cookie: <cookie_name>_ is set. If it's not, then the request is not authorized (skip all next steps).

**3** If the new flag --crsf-token-disabled is set to true, skip to step 7. This can be vulnerable to csrf attacks.

[...]

PR #2327, opened 2019-06-04
support reading JWT from Cookie header (fix #2183)

### LINKED ISSUE #2183: read JWT from cookie header
When Hasura decoding the JWT token (instead of passing it to webhook), Can Hasura read JWT token from secure-http-only-cookie ? If so then what should be the name of the cookie?

ecthiender:
### Description

This PR adds support to read JWT from cookies on the server. So, instead of `Authorization` Bearer token, if a cookie is sent with the JWT, Hasura will read it as well. Depending upon the configuration.

Add configuration to read JWT from Cookie header. This is an optional field in the `--jwt-secret` (`HASURA_GRAPHQL_JWT_SECRET`) JSON conf.  Defaults to `Authorization` header.

TODO: link to docs about the entire JSON conf with cookie configuration.

Storing JWT in browser localstorage is not secure and is susceptible to XSS attacks. Storing JWT in cookies with `httpOnly` flag mitigates XSS issues, but also introduces CSRF attacks. Hence to support reading JWT from cookies we should also add a way to mitigate CSRF issues. This is done by expecting a CSRF token in the JWT claims itself, and a CSRF token header from the client. Hasura then validates if they match.

The CSRF token claim name and header names can be configured in the `--jwt-secret` (`HASURA_GRAPHQL_JWT_SECRET`) JSON conf. The CSRF checking can also be disabled altogether (for example, in cases where the client is a desktop electron app).

TODO: link to docs about the entire JSON conf with cookie configuration.

### Affected components 

- Server
- Tests
- Docs

### Related Issues

#2183 
#2205 

### Solution and Design

TODO

### DIFF
--- a/docs/graphql/manual/auth/authentication/jwt.rst
+++ b/docs/graphql/manual/auth/authentication/jwt.rst
@@ -123,7 +123,8 @@ JSON object:
      "key": "<optional-key-as-string>",
      "jwk_url": "<optional-url-to-refresh-jwks>",
      "claims_namespace": "<optional-key-name-in-claims>",
-     "claims_format": "json|stringified_json"
+     "claims_format": "json|stringified_json",
+     "header": <optional-key-to-indicate-cookie-or-authorization-header>
    }
 
 ``key`` or ``jwk_url``, **one of them has to be present**.
@@ -226,6 +227,38 @@ If ``claims_format`` is ``stringified_json`` then JWT claims should look like:
     "iat": 1516239022,
     "https://hasura.io/jwt/claims": "{\"x-hasura-allowed-roles\":[\"editor\",\"user\",\"mod\"],\"x-hasura-default-role\":\"user\",\"x-hasura-user-id\":\"1234567890\",\"x-hasura-org-id\":\"123\",\"x-hasura-custom\":\"custom-value\"}"
   }
+
+``header``
+^^^^^^^^^^
+This is an optional field, which indicates which request header to read the JWT
+from. This field is an object.
+
+Following are the possible values:
+
+- ``{"type": "Authorization"}``
+- ``{"type": "Cookie", "name": "cookie_name" }``
+
+Default is ``{"type": "Authorization"}``.
+
+In the default mode, Hasura expects an ``Authorization`` header with a ``Bearer`` token.
+
+In the cookie mode, Hasura will try to parse the cookie header with the given
+cookie name. The value of the cookie should be the exact JWT.
+
+Example:-
+
+If ``header`` is ``{"type": "Authorization"}`` then JWT header should look like:
+
+.. code-block:: http
+
+   Authorization: Bearer eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWI...
+
+If ``header`` is ``{"type": "Cookie", "name": "cookie_name" }`` then JWT header should look like:
+
+
+.. code-block:: http
+
+   Cookie: cookie_name=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWI...
 
 
 Examples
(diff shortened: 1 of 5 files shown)


sanderhahn:
On Discord we had a discussion about how JWT/refresh tokens inside the cookie could help to implement a safe refresh token when the auth system is inside Hasura/PostgreSQL itself (for instance: https://github.com/sander-io/hasura-jwt-auth). Here is a link to the discussion as a reference: https://discordapp.com/channels/407792526867693568/537247473945673728/601768669646946305 :)

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -123,7 +123,8 @@ JSON object:
      "key": "<optional-key-as-string>",
      "jwk_url": "<optional-url-to-refresh-jwks>",
      "claims_namespace": "<optional-key-name-in-claims>",
-     "claims_format": "json|stringified_json"
+     "claims_format": "json|stringified_json",
+     "header": <optional-key-to-indicate-cookie-or-authorization-header>
```
rikinsk13: this will break the code highlighting as the "header" value is not valid json. will need to change the ``code-block:: json`` to ``code-block:: none``

SidhNor:
I wonder if there is any progress on this?

tirumaraiselvan:
@ecthiender @lexi-lambda Hey, what's the status on this?

lexi-lambda:
Adding some context from a slack conversation from a few months ago:

> **Alexis King  4 months ago**
> FWIW, browser support for the SameSite Set-Cookie attribute has gotten pretty good. That plus Secure; HttpOnly is IIUC enough to avoid CSRF attacks? But I am very much not a security specialist. :) https://caniuse.com/#feat=same-site-cookie-attribute
>
>
> **Tanmai Gopal  4 months ago**
> Yep yep, the samesite=strict attribute will prevent csrf over get and form-post.
The secure; httponly + CORS policies would prevent csrf type stuff over ajax.
>
> The only problem with the samesite attribute is that for static webapps deployed on a cdn which have a different API endpoint (say a different subdomain). In that case the samesite=strict would not send the cookies.
>
> **Alexis King  4 months ago**
> Yes, that makes sense. I guess it would be nice if you could whitelist a set of domains like you can with CORS.

ariadnaeleni:
@ecthiender @lexi-lambda any updates on this?

georgiemathews:
@0x777  @rikinsk are there plans to merge this?

mirzap:
This would be extremely helpful. Maybe you can tag version at which we can expect this to be merged @0x777 @rikinsk

verekia:
This really needs to go live. Only allowing the `Authorization` header encourages developers to use `localStorage` to store sensitive authentication data such as a JWT, which is vulnerable to XSS attacks from malicious scripts embedded on the page. An `HttpOnly` `SameSite` `Secure` cookie is a safer option to store and transport JWTs.

danielcolgan:
@0x777 @rikinsk @ecthiender @lexi-lambda @ecthiender How I can help to make sure we will merge it?

tsujp:
Is there any assistance I can give to get this merged? This is preventing me using Hasura in any real meaningful way for all the reasons above: JWTs for non-trivial backends I am not going to use, and therefore I am not going to use Hasura for those projects which frankly sucks because of how much Hasura provides and how good it is.

tirumaraiselvan:
Hi folks, sorry about the delay on this. We will pick up this PR from where it was left off after the next major release v1.4

valstu:
Cannot wait this gets added, it will make my application logic a lot simpler 👏

ecthiender:
We are not going to merge this PR as is. As there is a quite bit of interest, we'll consider this in our future roadmap.

Closing this PR as stale.

PR #2341, opened 2019-06-06
render 404 for nonexistent resources

strongSoda:
### Description
When an invalid path segment (eg a nonexistent table name) is part of an URL, the server responds 404, but this is treated as 'an error' and the 500 handler is invoked.

### Affected components 
- Console

### Steps to test and verify

- for tables: 
  url/data/schema/public/tables/nonexistentTable/browse
- for views:
  url/data/schema/public/views/nonexistentView/browse


strongSoda:
@rikinsk all's done.

close https://github.com/hasura/graphql-engine-internal/issues/138

PR #2343, opened 2019-06-06
support session variables in SQL functions (close #2322)

### LINKED ISSUE #2322: allow functions to use session information through an argument

rakeshkky:
TODO:- 
- [ ] console: Allow users to specify session variable argument while tracking a function in UI. 
Refer API documentation [here](https://github.com/rakeshkky/graphql-engine/blob/issue-2322-func-sess/docs/graphql/manual/api-reference/schema-metadata-api/custom-functions.rst#id4).

- [ ] cli: Update metadata export/import tests to include `version` key and support new v2 functions format which is [track_function v2 args](https://github.com/rakeshkky/graphql-engine/blob/issue-2322-func-sess/docs/graphql/manual/api-reference/schema-metadata-api/custom-functions.rst#args-syntax).

### Description

Specify a function's argument name on tracking so as to access the session variables JSON through it.

- Introduce the optional `version` field in `/v1/query` API object which is the version of the API. It defaults to `1`. 
- All existing metadata queries are of version 1
- A new version 2 `track_function` query type is defined to facilitate users to specify the session variable argument. An example request as follows
```http
   POST /v1/query HTTP/1.1
   X-Hasura-Role: admin

   {
       "type": "track_function",
       "version": 2,
       "args": {
           "function": {
               "schema": "public",
               "name": "search_articles"
           },
           "config": {
               "session_variable_argument": "hasura_user"
           }
       }
   }
```
- The server executes the SQL function with session variable JSON object.
- The export/apply metadata JSON document has a new optional `version` key. This change will export metadata with `version: 2` and accepts metadata of versions 1 and 2.

### Affected components 

- Server
- Console
- CLI
- Docs
- Tests

### Related Issues

close #2322 

### Solution and Design

- The type `RQLQuery` is updated to have sum type of `RQLQueryV1` and `RQLQueryV2`. `FromJSON` of `RQLQuery` is rewritten to support either version based on `version` field
- Each argument of a SQL function will have a additional information of argument type. It is either `input` (values supplied by user) or `session` (carries session variable JSON)
- Graphql value resolving logic  is updated to support `Session` variable value so as to substitute it with appropriate session variable JSON at different places.
- `export_metadata` and `replace_metadata` API is updated to support `version` field

### Steps to test and verify

Follow the example [here](https://github.com/rakeshkky/graphql-engine/blob/issue-2322-func-sess/docs/graphql/manual/queries/custom-functions.rst#accessing-hasura-session-variables-in-custom-functions)

- Older versions cannot import metadata exported by this.


howientc:
Is there a reason why this technique couldn't also be used for tracked views?

rakeshkky:
These changes are going through a new [PR](https://github.com/hasura/graphql-engine/pull/3143). Closing this now.

PR #2344, opened 2019-06-06
support optional parameters in database url (close #1709)

### LINKED ISSUE #1709: directly use database-url to establish connection to postgres
Currently, we're parsing given `database-url` to construct internal `ConnInfo` type and using that to create a connection to postgres. Instead, we can just use URL to create a connection.

This requires an update in [`pg-client`](https://github.com/hasura/pg-client-hs) library.

rakeshkky:
**Needs https://github.com/hasura/pg-client-hs/pull/12 to be merged**

### Description

Use database URL `String` provided via `--database-url` CLI option or `HASURA_GRAPHQL_DATABASE_URL` env variable to establish connection with Postgres directly. Before the server used to parse the URL string and resolve into respective connection options. So, the optional parameters (like `application_name` etc.) provided in the URL string are ignored.

Refer [here](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING) to learn more about PostgreSQL database URL.

Refer [here](https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-PARAMKEYWORDS) for more about connection parameters.

### Affected components 

- Server

### Related Issues

close #1709

### Solution and Design

- `pg-client-hs` library has support for creating a connection to PostgreSQL directly using database URL String in [this PR](https://github.com/hasura/pg-client-hs/pull/12). Update server to use that feature and log database URL on startup by hiding the password.

### To Reviewer
- `ConnInfo` type from `pg-client-hs` library is now a sum type which is either a URL String or `ConnOpts`. 

### Limitations
- The server has no control over database URL and it is up to PostgreSQL to validate it.


server/stack.yaml
```diff
@@ -12,8 +12,9 @@ packages:
 
 # Packages to be pulled from upstream that are not in the resolver (e.g., acme-missiles-0.3)
 extra-deps:
-- git: https://github.com/hasura/pg-client-hs.git
-  commit: 16f27a134f12c4f24fee19d57f0ca179bdb4135b
+# use https URLs so that build systems can clone these repos
+- git: https://github.com/rakeshkky/pg-client-hs.git
```
rakeshkky: TODO:- change this to `hasura` repo when related `pg-client-hs` PR is merged

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -959,15 +959,36 @@ parseLogLevel = optional $
...
+        J.object [ "host" J..= Q.connHost co
+                 , "port" J..= Q.connPort co
+                 , "user" J..= Q.connUser co
+                 , "database" J..= Q.connDatabase co
+                 , "retries" J..= Q.connRetries co
+                 ]
+
+    mkDBUriLog retries uri =
+      let mUri = parseURI uri
+          mParsedUri = flip fmap mUri $ \u -> uriToString hidePass u ""
```
ecthiender: The `Show` instance implementation of `URI` already hides password (see docs of `uriToString`) [1]. We should directly use it instead of our own implementation.


[1]: https://hackage.haskell.org/package/network-uri-2.6.1.0/docs/Network-URI.html#g:6
[2]: http://hackage.haskell.org/package/network-uri-2.6.1.0/docs/src/Network-URI.html#line-211

server/src-lib/Hasura/Server/PGDump.hs
```diff
@@ -48,7 +48,8 @@ execPGDump b ci = do
         ExitSuccess   -> Right $ CS.cs (clean stdOut)
         ExitFailure _ -> Left $ CS.cs stdErr
 
-    opts = Q.pgConnString ci : "--encoding=utf8" : prbOpts b
+    connString = T.unpack $ bsToTxt $ Q.pgConnString $ Q.ciDetails ci
```
lexi-lambda: It makes me slightly sad that this is necessary. Really, the connection URI should be of type `URL`, or at least `Text`, so this bytestring-to-text munging shouldn’t be necessary… but it probably isn’t worth the effort to change it.

Issue #2224, opened 2019-05-19
Add --docker flag to CLI for init


jasonmccallister:
When scaffolding new projects, it would be really helpful to include the Docker files as part of the init. I think something like this:

`hasura init --docker`

Which in the CLI will just run the command from the [Quickstart with Docker](https://docs.hasura.io/1.0/graphql/manual/getting-started/docker-simple.html#step-1-get-the-docker-compose-file).

`wget https://raw.githubusercontent.com/hasura/graphql-engine/master/install-manifests/docker-compose/docker-compose.yaml`

Is there any interest in adding this to the CLI?

shahidhk:
A modified version of this has been added to `1.2.0-beta.1` release.

PR #2348, opened 2019-06-06
Add flag and option for Docker on init command

jasonmccallister:
### Description
Adds a docker flag to also init a new project with a docker-compose.yaml. Addresses #2224.

### Affected components 

- CLI
- Docs
- Tests


jasonmccallister:
@shahidhk this functionality is done, I did not write any specific tests for it as I'm not sure how we would write tests for that. Happy to add given some direction, I need to add some similar type of tests for Go at work so it is a win-win. I did not update the docs, but I think we should add a quick note on the CLI installation section.

jasonmccallister:
@shahidhk ok, I looked at the docs and they were rather simplistic so I added an example in the docs

shahidhk:
@jasonmccallister Thanks a ton for this PR.

We've iterated upon this and have added `--install-manifest` flag to `init` command on the CLI with `1.20-beta.1` release. This will help you download any of the install manifests present in the directory.

I'm sorry that we delayed merging this for a long time, thank you for your patience.

PR #2356, opened 2019-06-09
Updates image widths for live-queries.md

mikeygbooth99:
### Affected components 

- Server
- Console
- CLI
- Docs
- Community Content
- Build System
- Tests
- Other (list it)


shahidhk:
Thanks for the PR @mikeygbooth99, can you add a small description on why the fix is required? Cannot see a difference in rendering on my screen.

PR #2366, opened 2019-06-11
Enable add/modify/delete for event triggers and remote schemas in migration mode, fix #1875

### LINKED ISSUE #1875: Enable event triggers & remote schemas even if migrations mode is off
v1.0.0-alpha41 running on Heroku

Accessing the console using `hasura cli console` results in the Add CT on the remote schemas tab not appearing. Accessing the app directly (via the heroku url) renders the correct / expected behaviour

AnnieAtHasura:
### Description
User will now see controls for add/modify/delete event triggers and remote schemas when in migration mode. 

### Affected components 

- Console

### Related Issues
close #1875

### Solution and Design
Removed check for migrationmode when not needed

### Steps to test and verify
set migration mode on, 
visit the events trigger page, add  a simple event trigger, modify it, then delete it
visit the remote schemas page, use the Glitch example to make a remote schema,
modify it, and delete it.
visit the Data page and see that you can't modify the DDL

You can still trick the system by entering SQL


wawhal:
@arvi3411301 this review app is not deleted even after the PR is merged. FYI.

PR #2379, opened 2019-06-14
Make catalog metadata migrations work on all schema versions (fix #2826)

### LINKED ISSUE #2826: upgrading from alpha45 to beta.6 fails
You'll need to first upgrade to `beta.5` and then upgrade to `beta.6`. The reason why this doesn't work has very little to do with `beta.6`. The root cause is the catalog migrations which add admin metadata using `yaml` files (an RQL query); these `RQL` queries require schema cache to run but building schema cache could fail because the function which does this assumes the latest catalog's schema. 

The conclusion is that we can't use RQL queries in catalog migrations and we'll need to rewrite these migrations using only SQL. 

This happened earlier too in one of the `alpha` versions.

ajeetdsouza:
### Description

As reported in #2826, catalog metadata migrations can fail when migrating from older catalog versions. This happens because catalog metadata is updated by running ordinary RQL queries like `add_existing_table_or_view` and `create_object_relationship`, but the implementations of those APIs will always use the newest version of the code, which may depend on recent schema changes.

This change fixes that by running all database migrations before changing the metadata at all, then simply dropping all metadata for the entire Hasura catalog and recreating it from scratch.

### Affected components 
- Server

### Related Issues
#2826

### Solution and Design

Changes to catalog metadata are no longer expressed as individual migrations. Instead, the `src-rsr/hdb_metadata.yaml` file is used directly to recreate all catalog metadata in one step. `src-exec/Migrate.hs` has been rewritten to both incorporate this behavioral change and to avoid so much duplication: *all* SQL migrations are now stored in separate files (now in the `src-rsr/migrations/` directory), and `Migrate.hs` automatically loads the necessary migration scripts without needing to maintain an explicit list.

A few other changes have been made to support the main goal of this PR:

  1. The catalog version is now defined in a new `Migrate.Version` module instead of directly in `Migrate`. This is unfortunately necessary to workaround the *GHC stage restriction*, which forbids using a definition from the current module in a Template Haskell splice.

  2. A new `HasSystemDefined` typeclass has been introduced to parameterize metadata-altering queries over whether or not the current metadata objects being defined should be considered “system-defined” or not. This avoids needing to explicitly specify `system_defined: true` for every query in `hdb_metadata.yaml`.

  3. The use of the `ginger` library for templating has been removed and replaced with `shakespeare`. Profiling revealed that `ginger` is remarkably slow, which caused problems due to the higher frequency at which the new approach to migrations calls `buildSchemaCache`. Since the templates are all statically known at compile-time, `shakespeare` is dramatically more efficient: it uses Template Haskell to compile the templates to Haskell code that simply uses variables from the current lexical scope.

### Steps to test and verify
Fire up an older version of `graphql-engine` using Docker, and test if migrations are working by switching to this version.

None


server/src-lib/Hasura/Server/Telemetry.hs
```diff
@@ -142,12 +142,12 @@ computeMetrics sc =
       evtTriggers = Map.size $ Map.filter (not . Map.null)
                     $ Map.map tiEventTriggerInfoMap usrTbls
       rmSchemas   = Map.size $ scRemoteResolvers sc
-      funcs = Map.size $ Map.filter (not . fiSystemDefined) $ scFunctions sc
+      funcs = Map.size $ Map.filter (not . unSystemDefined . fiSystemDefined) $ scFunctions sc
```
rakeshkky: create an alias to `unSystemDefined` as `isSystemDefined` for better readability. And use everywhere else.

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -104,26 +103,36 @@ $(deriveJSON
                  }
   ''RQLQuery)
 
+data RunCtx = RunCtx {
```
rakeshkky: Move value constructor to new line

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -155,20 +164,22 @@ peelRun
   -> HTTP.Manager
   -> SQLGenCtx
   -> PGExecCtx
+  -> SystemDefined
   -> Run a -> ExceptT QErr IO (a, SchemaCache)
-peelRun sc userInfo httMgr sqlGenCtx pgExecCtx (Run m) =
+peelRun sc userInfo httpMgr sqlGenCtx pgExecCtx systemDefined (Run m) =
```
rakeshkky:
```haskell
peelRun :: SchemaCache -> RunCtx -> Run a -> ExceptT QErr IO (a, SchemaCache)
```
The above type definition looks more elegant.

server/src-exec/Migrate.hs
```diff
@@ -1,3 +1,9 @@
+{- |
+Description: Migrations for Hasura metadata.
```
rakeshkky: Should be `Hasura catalog`

catalog: The Postgres Schema
metadata: Stuff present in the catalog

server/src-exec/Migrate.hs
```diff
@@ -1,3 +1,9 @@
...
+-}
+
+{-# LANGUAGE ConstraintKinds #-}
```
rakeshkky: Add `ConstraintKinds` to `default-extensions` in `graphql-engine.cabal` file. And remove it elsewhere in the server code.

server/src-lib/Hasura/RQL/DDL/Schema/Function.hs
```diff
@@ -148,7 +148,7 @@ trackFunctionP2 qf = do
       throw400 NotSupported $
       "function " <> qf <<> " is overloaded. Overloaded functions are not supported"
   trackFunctionP2Setup qf rawfi
-  liftTx $ saveFunctionToCatalog qf False
+  liftTx $ saveFunctionToCatalog qf $ SystemDefined False
```
rakeshkky: Are we not getting `SystemDefined` from `askSystemDefined` here?

server/src-lib/Hasura/RQL/DDL/Schema/Function.hs
```diff
@@ -83,17 +83,17 @@ mkFunctionInfo qf rawFuncInfo = do
   let funcArgsSeq = Seq.fromList funcArgs
       dep = SchemaDependency (SOTable retTable) "table"
       retTable = QualifiedObject retSn (TableName retN)
-  return $ FunctionInfo qf False funTy funcArgsSeq retTable [dep]
+  return $ FunctionInfo qf (SystemDefined False) funTy funcArgsSeq retTable [dep]
```
rakeshkky: Why we're constructing `SystemDefined` manually? This should come from `HasSystemDefined`!

shahidhk:
@ajeetdsouza Can you update the description to indicate that this PR is talking about catalog migrations that happen when Hasura versions are upgraded and not the "CLI migrations" -- which the users are familiar with?

server/src-exec/Migrate.hs
```diff
@@ -15,421 +23,101 @@ import           Hasura.Server.Query
...
     migrateFrom previousVersion
-      | previousVersion == curCatalogVer =
-          return $ "already at the latest version. current version: " <> show curCatalogVer
-      | [] <- neededMigrations =
-          throw400 NotSupported $ "unsupported version : " <> previousVersion
-      | otherwise =
-          traverse_ snd neededMigrations >> postMigrate
+      | previousVersion == latestCatalogVersionString = pure $
+          "already at the latest version. current version: " <> latestCatalogVersionString
+      | [] <- neededMigrations = throw400 NotSupported $ "unsupported version : " <> previousVersion
```
0x777: Always hated this error message. Can we change it to something which makes more sense for our user?

lexi-lambda: Good call, I’ve modified the message.

PR #2384, opened 2019-06-17
improve API error when dropping a non-existent permission (close #1623)

### LINKED ISSUE #1623: api error code is confusing when we try to drop a permission that does not exist
Error thrown when there I try to drop a permission that doesn't exist:

```
curl -d'{"type":"drop_update_permission", "args":{"table":{"name":"author", "schema":"public"}, "role":"user"}}' http://localhost:8080/v1/query

{"path":"$.args","error":"update permisison on \"author\" for role \"user\" does not exist","code":"permission-denied"}
```

Error message makes sense, but the code is kind of off - make it something like 404?

ajeetdsouza:
### Description
Trying to drop a permission that doesn't exist:
```
curl -d'{"type":"drop_update_permission", "args":{"table":{"name":"author", "schema":"public"}, "role":"user"}}' http://localhost:8080/v1/query
```
throws a `400` error, which is confusing:
```
{"path":"$.args","error":"update permisison on \"author\" for role \"user\" does not exist","code":"permission-denied"}
```
I have changed this to `404`:
```
{"path":"$.args","error":"update permission on \"author\" for role \"user\" does not exist","code":"not-exists"}
```
### Affected components 
- Server

### Related Issues
#1623 

### Solution and Design
The `askPermInfo` in `Hasura.RQL.DDL.Permission.Internal` throws this error. There is an `askPermInfo` in `Hasura.RQL.DML.Internal` as well, which is expected to throw a `400 Permission Denied` error, so I have renamed the function to `getPermInfo` and fixed the error code.

### Steps to test and verify
The curl command in the description should give a `404 Not Found` error code.

None


shahidhk:
Returning 404 was a suggestion that I gave while creating the issue. In retrospect, it should not be 404 as it can make me think that I called a wrong API or metadata APIs are disabled etc.

I think it should be `400 Bad Request` with a clear error message mentioning that this permission does not exist.

rikinsk:
This should be a bad request as the problem is in the body. 404 should be used only if the url is invalid

ajeetdsouza:
@shahidhk @rikinsk we don't seem to have a `BadRequest` error code. Would `400 NotExists` be better?

shahidhk:
@rakeshkky Can you please guide @ajeetdsouza to choose the right error code?

rakeshkky:
@ajeetdsouza  Add `BadRequest` to `Code` data type in `/Hasura/RQL/Types/Errors.hs` module

rakeshkky:
@ajeetdsouza
  Also the error message should be `update permisison on \"author\" for role \"user\" does not exist` but not `table \"author\" does not exist` as you mentioned in PR description

rakeshkky:
@ajeetdsouza  add a case to test `bad-request` error response.

PR #2400, opened 2019-06-19
make ux for destructive actions consistent

### LINKED ISSUE #1469: make ux for destructive actions consistent
Right now the hasura console displays a "Delete Table" button directly on the modify table view. Clicking this pops up a generic browser confirmation dialog which is pretty easy to instinctively agree to. Other UIs like GitHub require you to enter a "Danger Zone" and then enter the name of your project exactly in order to be able to delete anything. Considering the potential for very bad, no fun, super sad data loss something similar would be nice in hasura.

strongSoda:
fix #1469 

### Affected components 
- Console

### Related Issues
#1469

### DIFF
--- a/console/src/components/Services/CustomResolver/Edit/Edit.js
+++ b/console/src/components/Services/CustomResolver/Edit/Edit.js
@@ -20,6 +20,7 @@ import { appPrefix, pageTitle } from '../constants';
 import { NotFoundError } from '../../../Error/PageNotFound';
 
 import globals from '../../../../Globals';
+import { handleDelete } from '../../../../handleDelete';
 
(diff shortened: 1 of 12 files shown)


strongSoda:
@rikinsk  with modal passing a flag for whether or not to type doesn't make sense cause delete handlers need to be made for it. It's good without the modal. No handlers needed. Just pass true if you need to type else it will show only confirm box.

strongSoda:
Now only the table name needs to be entered on table delete. Rest have confirm boxes. But if in future they need input just pass a boolean and tweak the confirm message & it will work.  No boolean defaults to confirm box. Boolean of true renders prompt. @rikinsk

PR #2408, opened 2019-06-20
refactor AddTable validation, fix #1712

### LINKED ISSUE #1712: not specifying type of column in console allows to create table, but results in weird error
Not specifying the type of column while creating table results in a cryptic postgres error.

Ideally console should not allow me to create the table.

screenshot from 2019-03-07 15-15-10

AnnieAtHasura:
### Description
This PR is a refactor of submitValidation and it's helper, columnValidation, which had become a bit of spaghetti, and which had bug #1712  
Inspecting code found several others.
I have rewritten the code to be cleaner and more uniform in structure, in the process fixing these bugs.

### Affected components 
- Console

### Related Issues
#1712

### Solution and Design
Original code had a nest of logic to do what really is simply a series of checks.
Now simply does the checks in order.
Additionally, I have refactored the error checking to separate the error test and the
dispatch of an error so the logic is clearer.

### Steps to test and verify
See steps in #1712 
1. make a table named author with these columns
name=author  type= Integer  (auto increment) check unique
name=name  type=<leave default, none selected>

2. submit - 
observe an error message instead of the server side message

There is no code support for validating postgres type literals on the client side.
Rather than extend this fix to fixing this issue, I pass everything (which is current behavior),
and reported https://github.com/hasura/graphql-engine-internal/issues/199


console/src/components/Services/Data/Add/AddTable.js
```diff
@@ -128,144 +137,245 @@ class AddTable extends Component {
...
+    return '';
+  }
+
+  // UI has decided we're ready to submit. Validate that the
+  // props have a valid new table, and ask server for new table
+  validateAndSubmit() {
+    this.props.dispatch(resetValidation());
+    const validColumns = this.trimEmptyColumns(this.props.columns);
+
+    if (
```
rikinsk: this seems like this may get out of hand quickly with more validations. Create an array of objects with validations and notifications and loop over the array maybe?

console/src/components/Services/Data/Add/AddTable.js
```diff
@@ -128,144 +137,245 @@ class AddTable extends Component {
...
-          gqlTableErrorNotif[3]
+          notificationArray[0],
+          notificationArray[1],
+          notificationArray[2],
+          notificationArray[3]
         )
       );
       return false;
     }
+    console.log('in AddTable.js in checkAndDispatch I have an invalid type');
```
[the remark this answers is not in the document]
AnnieAtHasura: Sorry!

console/src/components/Services/Data/Common/GraphQLValidation.js
```diff
@@ -54,11 +72,53 @@ const gqlSchemaErrorNotif = [
...
+  'Error creating table!',
+  'Column type is invalid',
+  '',
+  {
+    custom: 'Column type is invalid',
+  },
+];
+
 export default gqlPattern;
 export {
```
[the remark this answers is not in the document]
AnnieAtHasura: Resolved via conversation with @rikinsk , we're not doing this here, moving to issue[ internal 206 ](https://github.com/hasura/graphql-engine-internal/issues/206).

console/src/components/Services/Data/Add/AddTable.js
```diff
@@ -124,137 +145,244 @@ class AddTable extends Component {
...
+          cols[i].name.toString() +
+          ' is Invalid. ' +
+          'Column names must be letters, numbers, and _ and start with a letter'
+        );
+      }
+    }
+    return '';
+  }
+
+  isValidType(s) {
```
rikinsk: Just noticed that this prevents creating columns of non-common types. Considering the type selector is a dropdown with all allowed values, we can probably skip this check.

All else looks good to merge.

AnnieAtHasura: fixed

PR #2476, opened 2019-07-05
allow Postgres readonly user to view hdb_primary_key (close #2429)

### LINKED ISSUE #2429: hdb_primary_key don`t view all constraints for readonly user
When using readonly user for all tables (owner for hdb_*) can not see all primary_keys. I happened because `table_constraints` view show only when current user owns or has some privilege other than SELECT on. 
See [https://www.postgresql.org/docs/11/infoschema-table-constraints.html](https://www.postgresql.org/docs/11/infoschema-table-constraints.html).
To resolve i can grant for update, but its not good for readonly.

ajeetdsouza:
### Description
Postgres users with readonly access to tables are currently unable to select from the `hdb_catalog.hdb_primary_key` view. This PR fixes this.

### Affected components 
- Server

### Related Issues
#2429 

### Solution and Design
This issue was caused by the reliance of the `hdb_catalog.hdb_primary_key` view on the builtin Postgres view `information_schema.table_constraints`, which was not accessible to readonly users. This has been fixed by making the view directly dependent on the builtin Postgres tables - `pg_namespace`, `pg_constraint`, and `pg_class`. 

### Steps to test and verify
A Postgres user with readonly access to `hdb_catalog.hdb_primary_key` should be able to see all primary key constraints with the query:
```sql
SELECT * FROM hdb_catalog.hdb_primary_key;
```

None

### To the Reviewer (server)
- No new modules have been added
- A new catalog version (version 18) has been created with the updated `hdb_catalog.hdb_primary_key` definition

**Critical files**
- `src-exec/Migrate.hs`
- `src-rsr/initialise.sql`
- `src-rsr/migrate_from_17_to_18.sql`

### DIFF
--- a/server/src-exec/Migrate.hs
+++ b/server/src-exec/Migrate.hs
@@ -19,7 +19,7 @@ import qualified Data.Yaml.TH                as Y
 import qualified Database.PG.Query           as Q
 
 curCatalogVer :: T.Text
-curCatalogVer = "17"
+curCatalogVer = "18"
 
 migrateMetadata
   :: ( MonadTx m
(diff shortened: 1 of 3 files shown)


coco98:
Should we add upgrade/downgrade instructions? This is a catalog change right?

shahidhk:
> Postgres users with readonly access to tables are currently unable to select from the hdb_catalog.hdb_primary_key view

@ajeetdsouza Can you indicate what the impact of this on the user is?

PR #2484, opened 2019-07-08
fix sql to return empty array when there are no columns in a table, close #2481

### LINKED ISSUE #2481: Can't do multiple operations on console after deleting all columns from a table through the Hasura console.
<h2>Bug</h2>
After deleting all the columns of a table through the console I:

- Can't do anything with the table which had all of its columns deleted
- Can't create new tables
- Can't modify other tables
- Can't check relationships. 

Each time I try I just get redirected to the error screen. 

<h2>Steps to Reproduce</h2>

**Hasura:** _v1.0.0-beta.2_
**PostgreSQL:** _v11.4_
**Provider:** _Heroku_

1. Create a Table through the Console
2. Create Columns for that Table through the Console
3. Delete all the Columns through the Console

arvi3411301:
### Description
When table has no columns, console gets the list of columns as [null] which breaks parsing. The columns should be []

### Affected components 

- Console

### Related Issues
#2481

### DIFF
--- a/console/src/components/Services/Data/utils.js
+++ b/console/src/components/Services/Data/utils.js
@@ -377,7 +377,7 @@ FROM
         ):: regclass, 
         'pg_class'
       ) as comment, 
-      json_agg(
+      COALESCE(json_agg(
         row_to_json(isc) :: JSONB || jsonb_build_object(
           'comment', 
           (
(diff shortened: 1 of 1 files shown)


arvi3411301:
@rikinsk can you check the browse and modify page?

PR #2526, opened 2019-07-13
Enhance Heroku update documentation

Albert-Gao:
Got me confused for some points, want to pave the way for the others.

### DIFF
--- a/docs/graphql/manual/deployment/heroku/updating.rst
+++ b/docs/graphql/manual/deployment/heroku/updating.rst
@@ -32,7 +32,7 @@ Step 2: Attach your Heroku app
 
 Let's say your Heroku app is called ``hasura-heroku`` and is running on ``https://hasura-heroku.herokuapp.com``.
 
-Go to the repo folder you just downloaded, use the `Heroku CLI <https://devcenter.heroku.com/articles/heroku-cli>`_ to configure the git repo you cloned in Step 1
+Navigate to your project directory, use the `Heroku CLI <https://devcenter.heroku.com/articles/heroku-cli>`_ to configure the git repo you cloned in Step 1
 to be able to push to this app.
 
 .. code-block:: bash
--- a/docs/graphql/manual/deployment/heroku/updating.rst
+++ b/docs/graphql/manual/deployment/heroku/updating.rst
@@ -50,7 +50,7 @@ When you ``git push`` to deploy, the Heroku app will get updated with the latest
 
    $ git push heroku master
 
-If you are getting error: ``fatal: 'heroku' does not appear to be a git repository``, you need to add your heroku repo to your local git remote via: ``git remote add heroku https://git.heroku.com/<you-app-name>.git``, you can find this URL in your Heroku - Settings - Info - Heroku Git URL. Type `git remote -v`, you will see your `heroku` app has been added as the upstream.
+If you are getting the following error: ``fatal: 'heroku' does not appear to be a git repository``, add your heroku repo to your local git remote via: ``git remote add heroku https://git.heroku.com/<you-app-name>.git``. (this URL locates in your Heroku - Settings - Info - Heroku Git URL)
 
 Deploy a specific version of Hasura GraphQL engine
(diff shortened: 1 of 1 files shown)


docs/graphql/manual/deployment/heroku/updating.rst
```diff
@@ -50,6 +50,8 @@ When you ``git push`` to deploy, the Heroku app will get updated with the latest
 
    $ git push heroku master
 
+If you are getting error: ``fatal: 'heroku' does not appear to be a git repository``, you need to add your heroku repo to your local git remote via: ``git remote add heroku https://git.heroku.com/<you-app-name>.git``, you can find this URL in your Heroku - Settings - Info - Heroku Git URL. Type `git remote -v`, you will see your `heroku` app has been added as the upstream.
```
marionschleifer: Can we write: "If you are getting _the following_ error: ..."? 

And could we make two or three sentences here instead of one long one? I think this would help for the understanding 🙂

Albert-Gao: Sounds good, thanks for reviewing! updated!

docs/graphql/manual/deployment/heroku/updating.rst
```diff
@@ -50,6 +50,8 @@ When you ``git push`` to deploy, the Heroku app will get updated with the latest
 
    $ git push heroku master
 
+If you are getting the following error: ``fatal: 'heroku' does not appear to be a git repository``, add your heroku repo to your local git remote via: ``git remote add heroku https://git.heroku.com/<you-app-name>.git``. (this URL locates in your Heroku - Settings - Info - Heroku Git URL)
```
marionschleifer: Actually, this part is already covered in the above section "Step 2: Attach your Heroku app". If the above step succeeds, there won't be an error after `git push heroku master`. But you could add the part in brackets to step 2, so that the user knows how to locate the Heroku repo.

Albert-Gao: Ah, makes sense, thanks! done. 
No idea why I still got that error back to that time, must do some wrong. :D

marionschleifer: One last thing: what do you mean by " Go to the repo folder you just downloaded...". I don't think anything is downloaded in this step. How about: "Navigate to your project directory, use the..."? 🙂

Albert-Gao: Haha, for that part, this is my first confusion, the description of this section seems to just clone the repo, and run the commands in section 2. 

But after cloning the repo, you are actually outside the repo, since I always use git cli, and I am not sure about this how does heroku wrapper `heroku git:remote` actually works, so I just wonder since the documentation didn't say `cd repo-folder`, maybe the heroku wrapper will automatically `cd` that folder for me? haha

After I figuring out there is no magic, it is already cost me some time, so I just want to emphasize this part to ease the pain for the others, most dev uses git, but not everyone uses heroku before. :D

marionschleifer: Sure, this I understand. It should be clear that you have to move inside the folder. But the `download` thing is confusing I think. So can we change it to: "Navigate to your project directory, use the..."?

Albert-Gao: makes a lot of sense to me, done! :D

marionschleifer: Thank you so much for contributing to our docs 😊

docs/graphql/manual/deployment/heroku/updating.rst
```diff
@@ -41,6 +41,8 @@ to be able to push to this app.
    $ heroku git:remote -a <hasura-heroku>
    $ heroku stack:set container -a <hasura-heroku>
 
+You can find your Heroku git repo in your Heroku - Settings - Info - Heroku Git URL
```
shahidhk:
```suggestion
```

Issue #2255, opened 2019-05-23
[COMMUNITY] Add some handy bash to community for simpler development


mnlbox:
I'm creating this bash to dump data and schema. If useful we can add such a small and handy tools or bash in community content.

```
#!/bin/bash

docker ps
echo -e "Hi, please enter postgres container id for dump schemas: \c "
read  postgres_container_id

echo -e "Please enter postgres database name: [for example: postgres] \c "
read  database_name
mkdir -p data_dump
cd data_dump

while :
do
  echo -e "Press <CTRL+C> to exit or Enter schema name for dump: \c "
  read  schema_name

  mkdir -p schema
  cd schema
  docker exec $postgres_container_id pg_dump -O -x -U $database_name --schema-only --schema $schema_name > ./$schema_name.sql
  mkdir -p clean
  curl --data-binary @./$schema_name.sql https://hasura-edit-pg-dump.now.sh > ./clean/$schema_name.sql

  cd ..
  mkdir -p data
  cd data
  docker exec $postgres_container_id pg_dump -O -x -U $database_name --data-only --schema $schema_name > ./$schema_name.sql
  mkdir -p clean
  curl --data-binary @./$schema_name.sql https://hasura-edit-pg-dump.now.sh > ./clean/$schema_name.sql

  cd ..
  mkdir -p schema_with_data
  cd schema_with_data
  docker exec $postgres_container_id pg_dump -O -x -U $database_name --schema $schema_name > ./$schema_name.sql
  mkdir -p clean
  curl --data-binary @./$schema_name.sql https://hasura-edit-pg-dump.now.sh > ./clean/$schema_name.sql

  cd ..
done
```
What is your opinion?

@shahidhk Some handy tools like your stream about migration. :wink:

PR #2547, opened 2019-07-15
Add local development swiss army knife script `scripts/dev.sh` 

jberryman:
**NOTE**: after rebasing on master two tests fail with "x-hasura-json-required-keys" header is expected but not found. @rakeshkky knows more about this

---

At the moment we can...

...run tests in isolation, generating coverage report:

    $ dev.sh test

Launch a postgres container with useful dev defaults, with PostGIS,
cleaning up afterwards:

    $ dev.sh postgres

Build and launch graphql-engine in dev mode, connecting with a
`postgres` launched above

    $ dev.sh graphql-engine


mnlbox:
Maybe related: #2255

jberryman:
TODO
[] tests crash mysteriously when no internet connection
[] 2 errors when running locally
[] incorporate pgdump functionality above
[] fix all warnings and Werror, since these make the output very annoying in terminal
[] make sure code coverage is working

jberryman:
Punting on incorporating pgdump functionality for now.
@rakeshkky  can repro the strange "x-hasura-json-required-keys" errors so punting on that for this PR (i.e. the problem doesn't seem to be the environment set up in these scripts).

This is ready for review (i.e. please try it out and let me know if it works) . @lexi-lambda if you have time to check it on mac that would be awesome as well.

jberryman:
Off topic, but: @rakeshkky I'm noticing that those two tests mentioned above seem to pass consistently when I run graphql-engine with `-N1` so perhaps a race condition of some sort

scripts/dev.sh
```diff
@@ -0,0 +1,322 @@
...
+  done
+  echo " Ok"
+
+  cd "$PROJECT_ROOT/server/tests-py"
+  command -v pytest >/dev/null
+  if [ $? -eq 0 ]; then
+      echo_pretty "pytest is installed, so assume we don't need to run 'pip3 install -r tests-py/requirements.txt'"
+      echo_pretty "you may need to run that yourself if things don't work, or improve this script"
+  else
+      pip3 install -r requirements.txt
```
lexi-lambda: Consider having the script set up a Python venv instead of installing dependencies globally. It’s easy to do: you basically just run `python3 -m venv <venv dir>`, then run `source <venv dir>/bin/activate.sh`. After that, you can run `pip3 install` and `pytest` as normal, and it will just work.

On my local setup, with `server/` as my working directory, I ran `python3 -m venv .python-venv`, and I run the tests by running `(source .python-venv/bin/activate.sh && cd tests-py && pytest <pytest args>)` (in a subshell to keep my shell’s environment clean). I have `.python-venv` in my global `.gitignore`, but we could also add it to this repo’s `.gitignore`.

scripts/dev.sh
```diff
@@ -0,0 +1,322 @@
...
+
+  echo_pretty "We will connect to postgres container '$PG_CONTAINER_NAME'"
+  echo_pretty "If you haven't yet, please launch a postgres container in a separate terminal with:"
+  echo_pretty "    $ $0 postgres"
+  echo_pretty "or press CTRL-C and invoke graphql-engine manually"
+  wait_docker_postgres
+
+  # Starts EKG, fast build without optimizations
+  BUILD_INVOCATION="stack build --fast --flag graphql-engine:developer --ghc-options=-j2"
+  RUN_INVOCATION="stack exec graphql-engine -- --host 127.0.0.1 -u postgres -p $PG_PORT -d postgres serve --enable-console --console-assets-dir \"$PROJECT_ROOT/console/static/dist\" +RTS -N -T -RTS ${PIPE_JQ-}"
```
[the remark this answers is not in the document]
jberryman: Ah I had `HASURA_GRAPHQL_DATABASE_URL` in my shell environment while testing it seems

scripts/dev.sh
jberryman: This is a great idea, but I think I'm going to leave this as a TODO unless you feel strongly.

jberryman: Nevermind :)

jberryman:
Thanks for the review @lexi-lambda ! Would you mind running both `dev.sh postgres` and `dev.sh graphql-engine` and `dev.sh test` once again on your mac? If it seems alright I'll squash and rebase again and maybe we can close this out.

I definitely understand your concerns above. The main motivators here were:
- there has been no accurately documented way to run `graphql-engine`; the script is somewhat to serve the needs of documentation
- document some tricks / existing knowledge in tooling, and produce something that works for me

re. bash: I think there's little complexity here that isn't inherent to setting up a basically functioning environment and dev workflow. But i basically agree, and certainly the script will get more complex.

re. 3: yeah I fully expect to be fixing this when stuff breaks it :P . If it gets some buy-in I would like to add it to the tests, possibly replacing some existing scripts.  

I certainly wouldn't want to force other folks to change their workflow, although I'd love to benefit from the knowledge other people have if this script can be adapted/improved to e.g. support things you like about your workflow.

jberryman:
Thanks, I guess it needs approval from those three other folks? Merge without squashing please

PR #2556, opened 2019-07-16
log when request body parsing fails (fix #2555)

### LINKED ISSUE #2555: request body parsing failure doesn't log
Since `v1.0.0-beta.3`, if JSON request body fails parsing it doesn't log.

ecthiender:
### Description

Regression: since `beta.3` whenever request body parsing failed, the error was not logged. This PR fixes the bug.

### Affected components 

- Server

### Related Issues

#2555 

### Solution and Design

Simple fix. The `parseBody` function was previously part of the handler functions. Since `beta.3` it has been moved to `mkSpockAction`, but the result of it was handled by `qErrToResp` instead of `logAndThrow` function.

**Edit**: After review from @rakeshkky , refactored `mkSpockAction` to have separate functions for [logging error and returning the response] and [logging success and returning the response]. Removed specific functions which only returned response (which was the source of the bug).

### Notes to reviewer
- ~~Simple one line fix in `Hasura.Server.App` module~~

### Steps to test and verify

1. Run graphql-engine
2. Make a request to `/v1/graphql` or `/v1/query` with wrong JSON structure.
3. Observe: The error is logged


ecthiender:
As per @rakeshkky , refactored some logging code at the server module. Need to verify if it breaks any existing logging.

ecthiender:
> These changes seem reasonable to me (though I’m not very familiar with the code and functionality in question, so it probably isn’t a super thorough review), but I do wonder: is there any way to add a test for this functionality that isn’t too difficult? It seems like logging is worth having access to in the test harness, but I don’t know if it’s set up for that right now.

@lexi-lambda there are tests for logging. Let me see if I can add tests for this quickly.

PR #1574, opened 2019-02-08
manage schema cache when horizontally scaled (closes #1182)

rakeshkky:
### Description

What component does this PR affect? 

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System

Requires changes from other components? If yes, please mark the components:

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System

### Related Issue

close #1182 

### Solution and Design

1. Haskel library `pg-client-hs` has been updated to expose a function that helps listen to `postgres` notifications over a `channel` in this [PR](https://github.com/hasura/pg-client-hs/pull/5)
2. The server records an event in a table `hdb_catalog.hdb_cache_update_event` whenever any `/v1/query` (that changes metadata) is requested. A trigger notifies a `cache update` event via `hasura_cache_update` channel
3. The server runs two concurrent threads namely `listener` and `processor`. The `listener` thread listens to events on `hasura_cache_update` channel and pushed into a `Queue`. The `processor` thread fetches events from that `Queue` and processes it. Thus server rebuilds schema cache from database and updates.

### Type

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Docs update
- [ ] Community content

### Checklist:

- [x] I have read the **[contributing guide](https://github.com/hasura/graphql-engine/blob/master/CONTRIBUTING.md)** and my code conforms to the guidelines.
- [x] This change requires a change in the documentation. 
- [ ] I have updated the documentation accordingly.
- [ ] I have added required tests.

PR #2585, opened 2019-07-22
clear events related to schema cache updates (close #2542)

### LINKED ISSUE #2542: clean up the events related to schema cache updates
Currently no clean-up ever happens on `hdb_catalog.hdb_schema_update_event` and the table will accumulate an event for every schema change that is ever done.

rakeshkky:
### Description

Schema syncing on horizontal scaling introduced in https://github.com/hasura/graphql-engine/pull/1574 uses a table to record schema cache update events. The table is populated with all events and cleaning never happens. This PR addresses this issue.

### Affected components 

- Server

### Related Issues

Close #2542

### Solution and Design

- Bump the catalog version with;
  - deleting all rows from `hdb_catalog.hdb_schema_update_event` table
  - define single row constraint on the table
  - drop existing trigger and create a trigger for each `insert` and `update` events on the table
- Use same instance id for telemetry and schema syncing threads

### Notes to Reviewer
Please look for line comments while you're reviewing


server/src-lib/Hasura/Server/Query.hs
```diff
@@ -135,20 +135,19 @@ fetchLastUpdate = do
...
   liftTx $ Q.unitQE defaultTxErrorHandler [Q.sql|
-             INSERT INTO
-                  hdb_catalog.hdb_schema_update_event
-                  (instance_id, occurred_at)
-             VALUES ($1::uuid, DEFAULT)
-            |] (Identity $ getInstanceId instanceId) True
+             INSERT INTO hdb_catalog.hdb_schema_update_event
+               (instance_id, occurred_at) VALUES ($1::uuid, DEFAULT)
+             ON CONFLICT ((occurred_at IS NOT NULL))
+             DO UPDATE SET instance_id = $1::uuid, occurred_at = DEFAULT
```
rakeshkky: Using `INSERT` instead of `UPDATE` because after initialise the `hdb_catalog.hdb_schema_update_event` is empty and insert is required in that case. Using `ON CONFLICT` clause to `update` the already present single row.

server/src-rsr/initialise.sql
```diff
@@ -433,8 +435,13 @@ $function$
 $function$
 LANGUAGE plpgsql;
 
-CREATE TRIGGER hdb_schema_update_event_notifier AFTER INSERT ON hdb_catalog.hdb_schema_update_event
-  FOR EACH ROW EXECUTE PROCEDURE hdb_catalog.hdb_schema_update_event_notifier();
+CREATE TRIGGER hdb_schema_update_event_notifier_insert AFTER INSERT ON
```
rakeshkky: `INSERT` trigger is not useful when `ON CONFLICT` clause succeeds and `DO UPDATE` happens. Postgres treats it as a normal `UPDATE` statement. So, `INSERT` and `UPDATE` triggers are defined.

server/stack.yaml.lock
```diff
@@ -0,0 +1,103 @@
+# This file was autogenerated by Stack.
```
rakeshkky: `stack.yaml.lock` is introduced in `stack` build tool from `2.1.1`. It is recommended to put it in version control. Refer [here](https://docs.haskellstack.org/en/stable/lock_files/) to know more.

server/src-rsr/initialise.sql
lexi-lambda: Can you create an `AFTER INSERT OR UPDATE` trigger instead of creating separate `INSERT` and `UPDATE` triggers?

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -135,20 +135,19 @@ fetchLastUpdate = do
        ORDER BY occurred_at DESC LIMIT 1
           |] () True
   case l of
```
lexi-lambda: Could this use the `FromRes (Maybe a)` instance with `withQE` to avoid needing to do the explicit pattern matching?

PR #2589, opened 2019-07-23
change exported metadata filename (close #1772)

### LINKED ISSUE #1772: Export metadata file name is too generic
When you export the metadata, disposed file name is `metadata.json`. Although it is most of the time fine, it is too generic. A better naming would be `hasura_metadata.json`, or even something with a date-time in it. (ie. hasura_metadata_2019_03_14_11_00_00.json)

rikinsk-zz:
### Description

**[THIS CHANGES EXISTING BEHAVIOUR]**

The current metadata file exported from the console has the filename `metadata.json` which is too generic and does not help finding the latest metadata file. This PR updates the name of metadata file to the format `hasura_metadata_<YYYY_MM_DD_hh_mm_ss_s>.json`.

 eg: `hasura_metadata_2019_07_23_17_28_03_225.json`

TODO:
- this breaks `metadata apply` on the CLI, will need to add support for the new format so that users can just export it and put it in the migrations image.

### Affected components 

- Console
- Docs
- CLI

### Related Issues

#1772

### Steps to test and verify

Head to `/metadata/actions` page and hit Export Metadata. Observe downloaded file filename

### DIFF
--- a/console/src/components/Services/Metadata/MetadataOptions/ExportMetadata.js
+++ b/console/src/components/Services/Metadata/MetadataOptions/ExportMetadata.js
@@ -48,7 +48,10 @@ class ExportMetadata extends Component {
                       encodeURIComponent(JSON.stringify(data));
                     const anchorElem = document.createElement('a');
                     anchorElem.setAttribute('href', dataStr);
-                    anchorElem.setAttribute('download', 'metadata.json');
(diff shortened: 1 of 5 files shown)


shahidhk:
Needs changes for CLI

PR #2590, opened 2019-07-25
Minimalist Support for Read Replica DB URLs - fix #1847

### LINKED ISSUE #1847: Multiple PostgreSQL connection string to support AWS Aurora.
AWS Aurora has a feature where it replicates read copies of your DB across multiple availability zones in a VPC. I think the purpose is to get better performance and increase DB redundancy. 

Here is a little [documentation on Aurora DB Clusters](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Overview.html).

Could Hasura at some point support multiple DB Connections with read write options? Or, thoughts on how to design connection options to take advantage of Aurora? ([Running as PostgreSQL 10.6](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.AuroraPostgreSQL.html))

sversteeg-scwx:
### Description

This patch allows the Hasura engine to connect to a read replica database (if Hasura has already been initialised on the writable instance.)

### Affected components 

- Server

### Related Issues

#1847

### Solution and Design

The patch is a minimalist solution to allow Hasura engine to connect to a read replica.

- Have added an extra Hasura server flag: --read_only_db or the environment variable HASURA_GRAPHQL_READ_ONLY_DB. By default this flag is false.
- When the read only database flag is true then Hasura will skip some of the start up steps that require a writable database: namely schema migration and synchronising the catalogue.
- It was also necessary to change the transaction isolation level for some of the start up queries. Three queries were previously hard coded as Serializable. Have changed these to RepeatableRead when the read only flag is set, otherwise Postgres returns an error.

### Steps to test and verify

1. Initialise and configure Hasura Engine with a writable database instance
2. Start Hasura Engine pointing to a read replica, with HASURA_GRAPHQL_READ_ONLY_DB set. E.g. using container:
	docker run -p 8081:8080 -e HASURA_GRAPHQL_DATABASE_URL=postgresql://user:password@read.replica:port -e HASURA_GRAPHQL_READ_ONLY_DB=true
3. Run a read query against GraphQL API. Or start Hasura console

I am new to Haskell, so look forward to the community's feedback regarding how code or solution can be improved.

This is a minimalist fix only. Has been tested against Aurora read replicas.

Some limitations are:
- Hasura Engine can only be pointed to a read replica if the (writable) database has already been initialised and configured for Hasura. Launching Hasura Engine against a read replica is therefore a (manual) two stage process.
- The Hasura Console "works" with a Hasura Engine pointed at a read replica, but it still shows writable operations, such as mutations and configuration updates. If the user attempts any writable operations the console will show an error. Ideally the writable console operations should be hidden or shown as disabled if Hasura Engine is running with a read only database.


0x777:
Hi @sversteeg-scwx, sorry about the delay, we'll get back to you on this soon.

sversteeg-scwx:
@0x777 Thanks - I appreciate the update

henkesde:
Looking forward this feature as well. Cheers.

KaKaVip:
Ping !

ajbouh:
@sversteeg-scwx are you running this patch in prod?

haf:
I'm also interested in this. Is there a reason this PR is ignored for more than a year?

ajbouh:
I am considering solving this problem in the application layer. This will be a bit wasteful but it should work today.

The idea will be to deploy hasura into two different autoscaling clusters. One with read/write access, the other with read-only access (enforced at the postgres permissions level). The read-only cluster will connect to a cluster of postgres read replicas.

My client logic will pick which hasura cluster to send graphql operations to based on whether or not the operation is a mutation.

dvekeman:
@ajbouh we tried something similar: we had two ECS clusters connecting to an AWS Aurora database with a reader and a writer endpoint. Both ECS clusters were running hasura community edition. 

Our test failed because both clusters tried to manage the hasura metadata schema and of course the ECS cluster connected to the read-only endpoint failed.

In the mean time we are investigating Hasura Cloud where read replicas are supported. I believe (but this is just my personal thought) that this is also the reason why the hasura engineers are not picking up on this topic for the community edition as it's a supported feature by Hasura Cloud and Enterprise.

ajbouh:
When did you try this? I think there has been recent effort to have hasura properly back off on metadata edits that come from another instance.

Also, when you tried it which docker images did you use?

dvekeman:
A little more than a month ago so I guess it must have been 1.3.2. We used the standard docker images and we stopped the test as soon as we found out that it wasn’t officially supported 

I would be interested in your results though.

Stefmore02:
Hey @sversteeg-scwx thank you for this PR, we are deeply sorry we took this long to review. 

Since a lot has changed, we are unable to proceed and we'll have to close it. 

If you think we should reopen it, please do so.

Issue #2541, opened 2019-07-15
extend support for subscriptions in 'analyze'


0x777:
Currently the 'Analyze' button in GraphiQL only works with queries. We need to add support for analysing subscriptions. The relevant information would be whether a subscription can use the multiplexed backend or not and if not what are the changes that need to be done to the subscription.

0x777:
@rikinsk The server provides explain output for subscriptions now (since beta-8 if I'm not wrong). However the explain response is an object instead of an array when the operation is a subscription. Looks like the console crashes currently when I hit `explain` on a subscription.

tsaiDavid:
@0x777 - just want to point out this still happens for us, using `v1.0.0-pro.3`

rikinsk:
Implementing this was blocked on #3073 which is now finally merged. We'll be picking this issue up soon.

PR #2650, opened 2019-08-01
subscriptions can now be explained

0x777:
Currently the 'Analyze' button only works with queries. This PR adds support for subscriptions which lets one check 
1. If a subscription can be multiplexed (and why it if can't be). 
2. The underlying query 
3. The postgres plan. 

### Affected components 

- Server
- Console

### Related Issues
#2541 

### Solution and Design
This is part of the current `v1/graphql/explain` endpoint. For a subscription the response structure will be as follows:
```json
{
  "is_multiplexable": true,
  "reason": null,
  "sql": "\n        select\n          _subs.result_id, _fld_resp.root as result\n          from\n            unnest(\n              $1::uuid[], $2::json[]\n            ) _subs (result_id, result_vars)\n          left outer join lateral\n            (\n        SELECT  coalesce(json_agg(\"root\" ), '[]' ) AS \"root\" FROM  (SELECT  row_to_json((SELECT  \"_1_e\"  FROM  (SELECT  \"_0_root.base\".\"id\" AS \"id\", \"_0_root.base\".\"name\" AS \"name\"       ) AS \"_1_e\"      ) ) AS \"root\" FROM  (SELECT  *  FROM \"public\".\"artists\"  WHERE ((\"public\".\"artists\".\"id\") = (((\"_subs\".\"result_vars\"#>>ARRAY['variables', 'artist_id']))::integer))     ) AS \"_0_root.base\"      ) AS \"_2_root\"      \n            ) _fld_resp ON ('true')\n        ",
  "plan": [
    "Nested Loop Left Join  (cost=2.40..243.00 rows=100 width=48)",
    "  ->  Function Scan on _subs  (cost=0.01..1.00 rows=100 width=48)",
    "  ->  Aggregate  (cost=2.39..2.40 rows=1 width=32)",
    "        ->  Index Scan using pk_artists on artists  (cost=0.15..2.37 rows=1 width=25)",
    "              Index Cond: (id = ((_subs.result_vars #>> '{variables,artist_id}'::text[]))::integer)",
    "        SubPlan 1",
    "          ->  Result  (cost=0.00..0.01 rows=1 width=32)"
  ]
}
```
The fields are:
1. `is_multiplexable`: If `true` the subscription can be multiplexed.
2. `reason`: Why the server thinks that the subscription cannot be multiplexed (`null` if `is_multiplexed` is `true`).
3. `sql`: The sql used in fetching the results for the subscription. In case of multiplexed subscriptions, the query will be different (to accommodate fetching the results of multiple subscriptions in the same query). 
4. `plan`: Postgres's `explain analyze` for the above `sql`. 

### Steps to test and verify
Needs console changes

None

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
+++ b/server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
@@ -14,7 +14,11 @@ module Hasura.GraphQL.Execute.LiveQuery
   , initLiveQueriesState
   , dumpLiveQueriesState
 
+  , LiveQueryOpG(..)
+
   , LiveQueryOp
+  , LiveQueryOpPartial
+  , getLiveQueryOpPartial
   , LiveQueryId
   , addLiveQuery
   , removeLiveQuery
@@ -22,6 +26,7 @@ module Hasura.GraphQL.Execute.LiveQuery
   , SubsPlan
   , subsOpFromPlan
   , subsOpFromPGAST
+
   ) where
 
 import           Data.Has
@@ -30,7 +35,6 @@ import qualified Control.Concurrent.STM                       as STM
 import qualified Data.Aeson                                   as J
 import qualified Data.HashMap.Strict                          as Map
 import qualified Data.HashSet                                 as Set
-import qualified Data.Text                                    as T
 import qualified Database.PG.Query                            as Q
 import qualified Database.PG.Query.Connection                 as Q
 import qualified Language.GraphQL.Draft.Syntax                as G
@@ -48,7 +52,6 @@ import           Hasura.GraphQL.Execute.LiveQuery.Types
 import           Hasura.Prelude
 import           Hasura.RQL.DML.Select                        (asSingleRowJsonResp)
 import           Hasura.RQL.Types
-
 import           Hasura.SQL.Types
 import           Hasura.SQL.Value
 
@@ -95,9 +98,12 @@ initLiveQueriesState (LQOpts mxOpts fallbackOpts) pgExecCtx = do
         <*> LQF.initLiveQueriesState fallbackOpts
   return $ LiveQueriesState mxMap fallbackMap pgExecCtx
 
-data LiveQueryOp
-  = LQMultiplexed !LQM.MxOp
-  | LQFallback !LQF.FallbackOp
+data LiveQueryOpG f m
+  = LQFallback !f
+  | LQMultiplexed !m
+  deriving (Show, Eq)
+
+type LiveQueryOp = LiveQueryOpG LQF.FallbackOp LQM.MxOp
 
 data LiveQueryId
   = LQIMultiplexed !LQM.LiveQueryId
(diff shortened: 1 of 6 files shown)


server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -161,33 +167,56 @@ type TextEncodedVariables
...
+type LiveQueryOpPartial = LiveQueryOpG FallbackOpPartial MultiplexedOpPartial
+
+-- | Creates a partial live query operation, used in both
+-- analyze and execution of a live query
+getLiveQueryOpPartial
+  :: ( MonadError QErr m
+     , MonadIO m
+     )
+
+  -- | to validate arguments
```
lexi-lambda: I think the syntax Haddock expects for documenting arguments is the `-- ^ some description` form (which goes under the type rather than above), and in fact I believe doing it this way actually causes Haddock to barf. I personally think it’s a bit silly, but oh well.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -95,9 +98,12 @@ initLiveQueriesState (LQOpts mxOpts fallbackOpts) pgExecCtx = do
         <*> LQF.initLiveQueriesState fallbackOpts
   return $ LiveQueriesState mxMap fallbackMap pgExecCtx
 
-data LiveQueryOp
-  = LQMultiplexed !LQM.MxOp
-  | LQFallback !LQF.FallbackOp
+data LiveQueryOpG f m
```
lexi-lambda: Can you tell me what the `G` stands for in these types? “Grammar”? “General”?

server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -161,33 +167,56 @@ type TextEncodedVariables
...
-      _             -> return $ toTxtValue colTy colVal
-  GR.UVSessVar ty sessVar ->
-    return $ fromResVars ty [ "user", T.toLower sessVar]
-  GR.UVSQL sqlExp -> return sqlExp
+type FallbackOpPartial = (GR.QueryRootFldUnresolved, Set.HashSet G.Variable)
+type MultiplexedOpPartial = (GV.VarPGTypes, Q.Query, TextEncodedVariables)
+
+type LiveQueryOpPartial = LiveQueryOpG FallbackOpPartial MultiplexedOpPartial
+
+-- | Creates a partial live query operation, used in both
```
lexi-lambda: What exactly is a “partial” live query operation—which parts of a “full” live query does it include and which parts does it lack?

server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -161,33 +167,56 @@ type TextEncodedVariables
...
   where
-    fromResVars ty jPath =
-      flip S.SETyAnn (S.mkTypeAnn ty) $ S.SEOpApp (S.SQLOp "#>>")
-      [ S.SEQIden $ S.QIden (S.QualIden $ Iden "_subs")
-        (Iden "result_vars")
-      , S.SEArray $ map S.SELit jPath
-      ]
+    -- get the variables which don't conifrm to the
+    -- 'non-null scalar' rule
+    getNonConfirmingVariables usedVariables =
```
lexi-lambda: Should this be “conforming” rather than “confirming”?

server/src-lib/Hasura/GraphQL/Explain.hs
```diff
@@ -79,32 +81,49 @@ getSessVarVal userInfo sessVar =
...
+  let txtSQL = Q.getQueryText query
+      withExplain = "EXPLAIN (FORMAT TEXT) " <> txtSQL
+  planLines <- liftTx $ map runIdentity <$>
+    Q.listQE dmlTxErrorHandler (Q.fromText withExplain) args True
+  return (txtSQL, planLines)
+
+data SubscriptionExplainOutput
+  = SubscriptionExplainOutput
+  { _seoIsMultiplexable :: !Bool
+  -- only exists when subscription is not multiplexed
```
lexi-lambda: Similar to the comment about function arguments above, you can make this comment show up in the haddocks if you put it beneath the field instead of above it and use the `-- ^` syntax.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -161,33 +167,56 @@ type TextEncodedVariables
...
-        modify $ Map.insert var (colTy, colVal)
-        return $ fromResVars (PgTypeSimple colTy)
-          [ "variables"
-          , G.unName $ G.unVariable var
-          ]
-      _             -> return $ toTxtValue colTy colVal
-  GR.UVSessVar ty sessVar ->
-    return $ fromResVars ty [ "user", T.toLower sessVar]
-  GR.UVSQL sqlExp -> return sqlExp
+type FallbackOpPartial = (GR.QueryRootFldUnresolved, Set.HashSet G.Variable)
```
lexi-lambda: Minor, mostly unimportant comment, but: I wonder if there’s a good way to encode into the type system that this `HashSet` must be non-empty.

lexi-lambda:
Integrated into/subsumed by https://github.com/hasura/graphql-engine/commit/18e8fbab1b0403aa2bd10645101de4f641f48c17.

PR #2664, opened 2019-08-03
Add notes about CLI in migrate

joshuataylor:
### Description
I noticed that the docs don't list about how the CLI tool can import/export metadata. This is a much faster developer experience IMHO. 

### Affected components 

- Docs

### Solution and Design
Added notes to docs

### Steps to test and verify
Read the docs.

### DIFF
--- a/docs/graphql/manual/migrations/manage-metadata.rst
+++ b/docs/graphql/manual/migrations/manage-metadata.rst
@@ -47,6 +47,20 @@ Exporting Hasura metadata
      add ``-H 'X-Hasura-Admin-Secret: <your-admin-secret>'`` as the API is an
      admin-only API.
 
+  .. tab:: CLI
+
+     The export can be done via the :doc:`Hasura CLI
+     <../api-reference/schema-metadata-api/manage-metadata>` too.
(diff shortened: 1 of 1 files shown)


docs/graphql/manual/migrations/manage-metadata.rst
```diff
@@ -47,6 +47,20 @@ Exporting Hasura metadata
      add ``-H 'X-Hasura-Admin-Secret: <your-admin-secret>'`` as the API is an
      admin-only API.
 
+  .. tab:: CLI
+
+     The export can be done via the :doc:`Hasura CLI
+     <../api-reference/schema-metadata-api/manage-metadata>` too.
+     A file will a YAML file with the Hasura metadata, and the file name will have the timestamp and the name you set. Here is a example using ``hasura`` CLI tool:
```
marionschleifer: Something doesn't seem right with that sentence. Did you mean to say "A YAML file will be created..."? It would be great if you could correct it. Also "an" example 🙂

marionschleifer: Hey @joshuataylor, just wanted to check in and see how it's going with the changes?

rikinsk:
This was subsumed in #3403. Closing this

PR #2668, opened 2019-08-05
improve console logic for view updatability check (fix #2667)

### LINKED ISSUE #2667: console treats a view as read-only when when it is updatable
On the permissions page for a view, Console checks `information_schema.views.is_updatable` and `information_schema.views.is_insertable_into` but not `information_schema.views.is_trigger_updatable`, `information_schema.views.is_trigger_insertable_into` or `information_schema.views.is_trigger_deletable` when determining whether to show permissions for those operations.

As a side-note: It looks like this information is correctly collected by `hdb_catalog.hdb_table_info_agg.view_info`, but the console seems to make its own query to `information_schema` instead of using this view.

A toy schema that demonstrates this:
```sql
CREATE TABLE public.a (
  id   SERIAL PRIMARY KEY,
  name TEXT
);

CREATE TYPE public.A_V1_TYPE AS (id INTEGER, name TEXT, name_lower TEXT);

CREATE FUNCTION public.a_to_a_v1(
  p public.A
) RETURNS public.A_V1_TYPE
  LANGUAGE SQL
  IMMUTABLE STRICT PARALLEL SAFE AS
$function$
SELECT p.id, p.name, lower(p.name)
$function$;

CREATE CAST (public.A AS public.A_V1_TYPE) WITH FUNCTION public.a_to_a_v1(public.A) AS IMPLICIT;

CREATE OR REPLACE VIEW public.a_v1 AS
  SELECT (a_::public.A_V1_TYPE).* FROM public.a a_;

COMMENT ON VIEW public.a_v1 IS 'Example interface a v1';
COMMENT ON COLUMN public.a_v1.name IS 'a name column';
COMMENT ON COLUMN public.a_v1.name_lower IS 'lowercase version of name';

CREATE OR REPLACE FUNCTION public.a_v1_insert(
  new_ public.A_V1
) RETURNS public.A_V1
  LANGUAGE plpgsql AS
$function$
DECLARE
  a_ public.A;
BEGIN
  INSERT
  INTO public.a(name) VALUES (new_.name) RETURNING * INTO STRICT a_;
  RETURN a_::public.A_V1_TYPE;
END;
$function$;

CREATE OR REPLACE FUNCTION public.a_v1_update(
  old_ public.A_V1,
  new_ public.A_V1
) RETURNS public.A_V1
  LANGUAGE plpgsql AS
$function$
DECLARE
  a_ public.A;
BEGIN
  UPDATE public.a SET name = new_.name WHERE id = old_.id RETURNING * INTO STRICT a_;
  RETURN a_::public.A_V1_TYPE;
END ;
$function$;

CREATE OR REPLACE FUNCTION public.a_v1_delete(
  old_ public.A_V1
) RETURNS public.A_V1
  LANGUAGE plpgsql AS
$function$
BEGIN
  DELETE FROM public.a WHERE id = old_.id;
  RETURN old_;
END;
$function$;

CREATE OR REPLACE FUNCTION public.a_v1_ops(
) RETURNS TRIGGER
  LANGUAGE plpgsql AS
$function$
BEGIN
  IF tg_op = 'INSERT'
    THEN
      RETURN public.a_v1_insert(new);
    ELSEIF tg_op = 'UPDATE'
      THEN
        RETURN public.a_v1_update(old, new);
    ELSEIF tg_op = 'DELETE'
      THEN
        RETURN public.a_v1_delete(old);
  END IF;
END
$function$;

CREATE TRIGGER a_v1_ops_trigger
  INSTEAD OF INSERT OR DELETE OR UPDATE
  ON public.a_v1
  FOR EACH ROW
EXECUTE PROCEDURE public.a_v1_ops();
```
With the above schema, the permissions page looks like this:
Screenshot from 2019-08-05 12-29-16

Where I'd expect it to look like this:
Screenshot from 2019-08-05 12-37-19

joshainglis:
### Description
Currently, the console checks `is_updatable`, and `is_insertable_into` to check whether to display permissions for insert, update and delete. However, for slightly more complex view with an associated `INSTEAD OF` trigger, the `is_updatable`, and `is_insertable_into` might be `'NO'`, while the is_trigger_insertable_into, is_trigger_updatable, is_trigger_deletable columns might be `'YES'`. In this case, we would still like to be able to set permissions for these operations.

### Affected components 
- Console

### Related Issues
#2667 

### Solution and Design
This PR adds in checks for the is_trigger_* columns when determining whether to display insert, update, and delete permissions.

### Steps to test and verify
See #2667 for an example schema for testing and verifying the fix.

This PR is fairly minimal in terms of adding in use of the necessary columns. However, given that `hdb_catalog.hdb_table_info_agg.view_info` also exposes required information more concisely, it might be worth looking into replacing the entire `information_schema` query with one to `hdb_catalog.hdb_table_info_agg`, for the purposes of centralising logic, but that's well out of scope for this PR.

### DIFF
--- a/console/src/components/Services/Data/TablePermissions/Permissions.js
+++ b/console/src/components/Services/Data/TablePermissions/Permissions.js
@@ -239,8 +239,10 @@ class Permissions extends Component {
           tableType === 'view' &&
           !(
             tableSchema.view_info &&
-            tableSchema.view_info.is_insertable_into === 'YES' &&
-            tableSchema.view_info.is_updatable === 'YES'
+            (tableSchema.view_info.is_insertable_into === 'YES' ||
+              tableSchema.view_info.is_trigger_insertable_into === 'YES') &&
+            (tableSchema.view_info.is_updatable === 'YES' ||
+              tableSchema.view_info.is_trigger_updatable === 'YES')
           )
         ) {
           hasPermissions = false;
@@ -1770,7 +1772,10 @@ class Permissions extends Component {
 
       // Add insert/update permission if it is insertable/updatable as returned by pg
       if (tSchema.view_info) {
-        if (tSchema.view_info.is_insertable_into === 'YES') {
+        if (
+          tSchema.view_info.is_insertable_into === 'YES' ||
+          tSchema.view_info.is_trigger_insertable_into === 'YES'
+        ) {
           qTypes.push('insert');
         }
 
@@ -1779,6 +1784,14 @@ class Permissions extends Component {
         if (tSchema.view_info.is_updatable === 'YES') {
           qTypes.push('update');
           qTypes.push('delete');
+        } else {
+          if (tSchema.view_info.is_trigger_updatable === 'YES') {
+            qTypes.push('update');
+          }
+
+          if (tSchema.view_info.is_trigger_deletable === 'YES') {
+            qTypes.push('delete');
+          }
         }
       } else {
         qTypes.push('select');


joshainglis:
Failed test seems unrelated? Looks like a JWT timeout test.

PR #2672, opened 2019-08-05
Support representing single-column Postgres tables as GraphQL enums (close #982)

### LINKED ISSUE #982: support GraphQL enum types
Why Hasura is not utilizing Postgres ENUM types and not translate it to GraphQL Enums?

lexi-lambda:
### Description

These changes support representing single-column (or two-column, where the second column is a documentation comment) Postgres tables in the GraphQL schema so that enum values can be explicitly represented and typechecked. They do not add any support for native Postgres enum types.

### Affected components 

- Server
- Docs

### Related Issues

#982

### Solution and Design

- [x] Track in metadata which tables represent enums
- [x] Store enum values in schema cache, sync on schema refresh
- [x] Generate GraphQL enum types in schema based on enum values
- [x] Allow using those enum types in query operations
- [x] API to change whether or not a table is an enum table
- [x] Appropriately update/reload schema cache when a new enum table is created or removed
- [x] Update documentation

### For reviewers

This is a big change. Here’s my advice if you want to review this PR:

  1. Start by looking at `Hasura.RQL.Types.Column`, followed by `Hasura.RQL.Schema.Enum`. Those two modules contain the core changes that inevitably infect everything else. Most importantly, `Hasura.RQL.Types.Column` defines a new type, `PGColumnType`, which is a wrapper around the old `PGColType`, which is now called `PGScalarType`. Read the Haddock comment attached to it, and also take a look at the newly-split `PGColumnInfo` and `PGRawColumnInfo` types.

  2. At first glance, this PR looks like it touches tons and tons of files, but a large majority of them are only changed due to renaming or minor refactoring. The things you should actually focus on are:

       1. The revised metadata and schema cache code, which now uses a two-pass strategy for processing `TableInfo` in order to resolve column types. This is by far the biggest change in actual functionality. There are some comments in places that might help.

       2. Changes to the GraphQL schema generation and query validation, which are not actually that significant in terms of lines of code but are essentially all of the actual value added by this change.

     Basically everything else is just renaming, refactoring, or moving things around to cope with the `PGColumnType` / `PGScalarType` split.

  3. These changes bundle in a couple refactorings that I made while working on this. (I have a couple others on other branches that didn’t pan out as well, but that’s probably not worth talking about for now.) Here are the two you should probably be aware of:

     1. The most significant refactoring is a split of `Hasura.RQL.DDL.Schema.Table` into a couple of smaller modules. Now, `Hasura.RQL.DDL.Schema.Table` *only* contains functions specific to tables, so the higher-level `buildSchemaCache` and `processSchemaChanges` functions have been moved into a separate module, `Hasura.RQL.DDL.Schema.Cache`. All of these modules are now exported together via the top-level `Hasura.RQL.DDL.Schema` module.

        This change can make it a little bit tricky to see the precise diff against the original version of `Hasura.RQL.DDL.Schema.Table`, since git thinks *everything* changed even when things were just moved around. Fortunately, the refactoring is in its own commit way at the end, so you might want to use GitHub’s support for looking at a diff of [only the commits before the refactoring happened](https://github.com/hasura/graphql-engine/pull/2672/files/4d78b624711ac7d982426cd2cc040ed39bc901ed) to see what the “real” changes were.

     2. The other refactoring is smaller and doesn’t really relate to any of the other changes: I added a new `Hasura.SQL.Error` module, which originally played a bigger part in this feature but ended up not really being important.

        In any case, I decided to salvage it by repurposing it as a nicer way to implement `Hasura.RQL.DML.Internal`’s `simplifyError` function. I ended up going ahead and rolling that change into `defaultTxErrorHandler`, since I don’t see why it shouldn’t be used everywhere, but if you think that’s a bad idea, let me know!

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
+++ b/server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
@@ -166,18 +166,18 @@ toMultiplexedQueryVar
   => GR.UnresolvedVal -> m S.SQLExp
 toMultiplexedQueryVar = \case
   GR.UVPG annPGVal ->
-    let GR.AnnPGVal varM isNullable colTy colVal = annPGVal
+    let GR.AnnPGVal varM isNullable _ colVal = annPGVal
     in case (varM, isNullable) of
       -- we don't check for nullability as
       -- this is only used for reusable plans
       -- the check has to be made before this
       (Just var, _) -> do
-        modify $ Map.insert var (colTy, colVal)
-        return $ fromResVars (PgTypeSimple colTy)
+        modify $ Map.insert var colVal
+        return $ fromResVars (PGTypeSimple $ pstType colVal)
           [ "variables"
           , G.unName $ G.unVariable var
           ]
-      _             -> return $ toTxtValue colTy colVal
+      _             -> return $ toTxtValue colVal
   GR.UVSessVar ty sessVar ->
     return $ fromResVars ty [ "user", T.toLower sessVar]
   GR.UVSQL sqlExp -> return sqlExp
@@ -198,19 +198,19 @@ subsOpFromPGAST
      , MonadIO m
      )
 
-  -- | to validate arguments
   => PGExecCtx
+  -- ^ to validate arguments
 
-  -- | used as part of an identifier in the underlying live query systems
-  -- to avoid unnecessary load on Postgres where possible
   -> GH.GQLReqUnparsed
+  -- ^ used as part of an identifier in the underlying live query systems
+  -- to avoid unnecessary load on Postgres where possible
 
-  -- | variable definitions as seen in the subscription, needed in
-  -- checking whether the subscription can be multiplexed or not
   -> [G.VariableDefinition]
+  -- ^ variable definitions as seen in the subscription, needed in
+  -- checking whether the subscription can be multiplexed or not
 
-  -- | The alias and the partially processed live query field
   -> (G.Alias, GR.QueryRootFldUnresolved)
+  -- ^ The alias and the partially processed live query field
 
   -> m (LiveQueryOp, Maybe SubsPlan)
 subsOpFromPGAST pgExecCtx reqUnparsed varDefs (fldAls, astUnresolved) = do
@@ -273,10 +273,10 @@ validateAnnVarValsOnPg pgExecCtx annVarVals = do
 
   Q.Discard _ <- runTx' $ liftTx $
     Q.rawQE valPgErrHandler (Q.fromBuilder $ toSQL valSel) [] False
-  return $ fmap (txtEncodedPGVal . snd) annVarVals
+  return $ fmap (txtEncodedPGVal . pstValue) annVarVals
 
   where
-    mkExtrs = map (flip S.Extractor Nothing . uncurry toTxtValue)
+    mkExtrs = map (flip S.Extractor Nothing . toTxtValue)
     mkValidationSel vars =
       S.mkSelect { S.selExtr = mkExtrs vars }
     runTx' tx = do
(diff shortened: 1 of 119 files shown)


dsandip:
@lexi-lambda Removing the `ok-to-merge` label because this needs console work too.

server/src-lib/Hasura/Server/Telemetry.hs
```diff
@@ -128,8 +128,8 @@ runTelemetry (Logger logger) manager cacheRef dbId instanceId = do
 
 computeMetrics :: SchemaCache -> Metrics
 computeMetrics sc =
-  let nTables = Map.size $ Map.filter (isNothing . tiViewInfo) usrTbls
-      nViews  = Map.size $ Map.filter (isJust . tiViewInfo) usrTbls
+  let nTables = Map.size $ Map.filter (isNothing . _tiViewInfo) usrTbls
```
0x777: Maybe we can track the number of enum tables being used?

server/src-lib/Hasura/RQL/DDL/Schema/Cache.hs-boot
```diff
@@ -0,0 +1,20 @@
+module Hasura.RQL.DDL.Schema.Cache where
```
0x777: There is no other way to get rid of cyclic dependencies?

0x777: `SQL.DML` module was originally written as mutually recursive modules. I remember `hs-boot` files causing issues though I can't recollect exactly what the issues were. If it is easy to break the import cycle without resorting to `hs-boot` files we should give it try.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -272,39 +271,17 @@ validateAnnVarValsOnPg pgExecCtx annVarVals = do
...
     mkValidationSel vars =
       S.mkSelect { S.selExtr = mkExtrs vars }
     runTx' tx = do
       res <- liftIO $ runExceptT (runLazyTx' pgExecCtx tx)
       liftEither res
 
--- | The error handler that is used to errors in the validation SQL.
--- It tries to specifically read few PG error codes which indicate
--- that the format of the value provided for a type is incorrect
-valPgErrHandler :: Q.PGTxErr -> QErr
```
0x777: We'll need this exact handler because these are the only error codes that we can expect anything else should be treated as an internal error.

server/src-lib/Hasura/RQL/DDL/Schema/Cache.hs-boot
lexi-lambda: I was wondering if this would come up. :)

There is an alternative strategy I had in mind that would avoid using `hs-boot` files by using typeclasses. Basically you use the typeclass as an interface and break the dependency by programming to that interface, then define an instance to effectively explicitly link the mutually recursive modules. I decided to go with `hs-boot` files instead, since to me it seems like a slightly easier solution (it’s basically doing the same thing but asking GHC to do the linking), but I understand why people like to avoid them. I’ll change it to use the typeclass approach if you’d prefer that.

lexi-lambda:
@0x777 I have scaled back the changes to Postgres error handling. The new changes still use `Hasura.SQL.Error` to parse the errors, and I’ve centralized the error message generation in `Hasura.Db` so it can be shared between the DML and LiveQuery modules, but all actual behavior should now remain unchanged.

server/src-lib/Hasura/Server/Telemetry.hs
lexi-lambda: Added to the metrics.

Issue #2635, opened 2019-07-30
Pretty print autogenerated SQL in YAML


dariocravero:
Hi!

While using the migrations more actively on a new project, I realised the output was coming out  in different formats and sometimes with escaped quotes which made it harder to follow.

How about we go from a one liner of SQL like:
```
- args:
    sql: CREATE TABLE "public"."users"("id" uuid NOT NULL DEFAULT gen_random_uuid(),
      "first_name" text NOT NULL, "last_name" text NOT NULL, "email" text NOT NULL,
      "created_at" timestamptz NOT NULL DEFAULT now(), "updated_at" timestamptz NOT
      NULL DEFAULT now(), PRIMARY KEY ("id") , UNIQUE ("id"));
  type: run_sql
```

to a pretty printed output like:
```
- args:
    sql: |
      CREATE TABLE "public"."users"(
        "id" uuid NOT NULL DEFAULT gen_random_uuid(),
        "first_name" text NOT NULL,
        "last_name" text NOT NULL,
        "email" text NOT NULL,
        "created_at" timestamptz NOT NULL DEFAULT now(),
        "updated_at" timestamptz NOT NULL DEFAULT now(),
        PRIMARY KEY ("id"),
        UNIQUE ("id")
      );
  type: run_sql
```

What do you think?
Thanks!
Darío

/cc @shahidhk

dariocravero:
For reference, https://github.com/domasx2/hasura-squasher is using https://www.npmjs.com/package/sql-formatter

coco98:
@dariocravero This is a very good idea. Hasura also has the option of generating a SQL migration alongside the YAML. 

So instead of:
- `migration_123_up.yaml`: contains a `run_sql` and a `track_table` say

You can also have:
- `migration_123_ up.sql`: contains the raw SQL
- `migration_123_ up.yaml`: contains the `track_table`

And then furthermore, we prettify the raw SQL. That way the SQL shows up nicely in code-editors as well. SQL inside a YAML file won't show nice syntax highlighting.

What do you think? Would you prefer a yaml+sql setup or just a sql setup. The only con is that the number of migration files will double whenever there's sql + yaml.

rikinsk:
[...]

PR #2677, opened 2019-08-06
console: format SQL before generating migrations

rikinsk-zz:
### Description

Use https://www.npmjs.com/package/sql-formatter to format SQL statements generated by console in migrations

**Problem:**
It seems like `sql-formatter` doesn't have support for `plpgsql`. Hence complex Postgres queries for creating functions, triggers, etc. get incorrectly formatted causing Postgres errors on execution. We'll either need to find another library with better support or put in PG specific hacks around this.

### Affected components 

- Console

### Related Issues

#2635


coco98:
@rikinsk I'm guessing this will also be useful if we switch to using `.sql` files for the sql migrations right?

rikinsk:
@coco98 Yeah. This just prettifies the not-so-pretty SQL strings generated by the console code. Should work for wherever we are storing SQL statements.

PR #2683, opened 2019-08-07
allow altering type of a column whose permissions defined only with session variables (close #2070)

### LINKED ISSUE #2070: altering type of a column is excessively restricted when permissions are defined
For example, when the `filter` condition is `{"user-id": "x-hasura-user-id"}`, it shouldn't matter what the type of the column is as it is determined by the session variable. So altering type should be allowed. However, if it is something like `{"is_public": true}`, then it should be denied.

rakeshkky:
### Description

Allow altering type of a column even if it is being used in permissions with session variable values.

### Affected components 

- Server
- Tests

### Related Issues

close #2070 

### Solution and Design

- Add column dependencies with reason `sess_var` if its permissions are defined only with session variables
- In `run_sql` query flow, if column types are altered and `sess_var` dependant permissions are found then rebuild those permission infos and update the schema cache.

### Steps to test and verify

- Create any permission having a column with session variable
- Try to alter the type of that column, it should go through.


server/src-lib/Hasura/RQL/GBoolExp.hs
```diff
@@ -436,19 +436,62 @@ mkColCompExp qual lhsCol = mkCompExp (mkQCol lhsCol)
...
+  ASTEquals val     -> isStaticValue val
+  ASTIntersects val -> isStaticValue val
+  ASTOverlaps val   -> isStaticValue val
+  ASTTouches val    -> isStaticValue val
+  ASTWithin val     -> isStaticValue val
+  ASTDWithinGeom (DWithinGeomOp r val)     -> any isStaticValue [r, val]
+  ASTDWithinGeog (DWithinGeogOp r val sph) -> any isStaticValue [r, val, sph]
+
+  _                -> False
+
```
rakeshkky: Is there a better way to achieve this?

lexi-lambda: It’s possible that it could be done more concisely with something like [`Data.Data.Lens`](http://hackage.haskell.org/package/lens-4.17.1/docs/Data-Data-Lens.html). I’m not an expert, so take this with a grain of salt, but if you derive `Data` instances for the necessary types, it could be as simple as writing

```haskell
hasStaticExp :: OpExpG PartialSQLExp -> Bool
hasStaticExp = has (template.filtered isStaticValue)
```

which would be pretty neat.

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -189,18 +189,35 @@ processTableChanges ti tableDiff = do
         if | oColName /= nColName -> do
                renameColInCatalog oColName nColName tn ti
                return True
+
            | oColTy /= nColTy -> do
                let colId   = SOTableObj tn $ TOCol oColName
-                   depObjs = getDependentObjsWith (== "on_type") sc colId
-               unless (null depObjs) $ throw400 DependencyError $
+                   typeDepObjs = getDependentObjsWith (== "on_type") sc colId
```
lexi-lambda: Assigning a particular meaning to a string value like this makes me nervous. Can we use a sum type here, instead? Something like

```haskell
data ColumnDependencyReason = DependentOnType | DependentOnSessionVariable
```

except maybe with better names than I can come up with based on my understanding. :)

PR #2685, opened 2019-08-07
Use SIGINT as the stopsignal in the dockerfile

lorenzo:
### Description
Haskell expects the signal to be `SIGINT`, but the docker default is
`SIGTERM`.

Correcting this.
### Affected components 

- Server
- Build System


0x777:
Looks like the last test always fails after this change. A quick glance (could be wrong) tells me that the problem is in `stop_services` in `test-server.sh`. Somehow `WH_PID` and `WHC_PID` are empty. Given that the tests don't run against a docker instance this is very surprising.

lorenzo:
@0x777 seems like all tests are passing now?

Issue #2663, opened 2019-08-02
issue in login-logout between beta.3 and beta.4


MarkusFaatz:
Hey there!

First of all: Thanks a lot, I love exploring this project!
I am running it using your provided docker-compose file and connecting it to my platform (based on nginx-proxy-companion).

I think I found a mini-bug and it could help to clean up some more data from local storage on logout!
I started with version beta.3 and switched to beta.4 (of course I did not logout before :smiley: ). After logging in, I had full access to my database but could not use graphiql (unauthorized). 

Logging out didn't help me out, but to clear the local storage "by hand" worked fine!

[docker-compose-copy.txt](https://github.com/hasura/graphql-engine/files/3462469/docker-compose-copy.txt)

rikinsk:
Should be resolved in #2686

rikinsk:
Closedc via #2686

Issue #2660, opened 2019-08-02
After Hasura login the credentials remain visible?


scheung38:
After login to Hasura with the Hasura admin secret, this now becomes visible inside the Request Headers Value field?

marionschleifer:
@scheung38 thank you for reporting this issue. 

I was able to reproduce this on Chrome with one of my projects:

![image](https://user-images.githubusercontent.com/5722022/62413081-cd7e1b00-b60b-11e9-8cf2-564db8105149.png)

However, if I open the console in another browser or in incognito mode, the admin secret is not visible.

wawhal:
Hey people! Sorry for the delay in response.
[This PR](https://github.com/hasura/graphql-engine/pull/2686) fixes all these issues related to admin-secret, login and graphiql headers. It set to be merged soon.

Thanks for reporting the issue @scheung38 :)

rikinsk:
closed via #2686

PR #2686, opened 2019-08-07
better key persistence in console

wawhal:
### Description

- in CLI mode show error message on login page if no/invalid admin secret 
- allow not persisting admin secret in browser local storage on login page
- admin secret value in graphiql headers is masked in local storage, its value is set to current admin secret when reading back from local storage
- if admin secret is set and graphiql headers does not have the admin secret header set (typically happens when admin secret is just added) add a new admin secret header and set in local storage that this has been done. Now if user removes the admin secret header it will not be added back.
- make graphiql introspection query only after headers have been set.
- change routes from /metadata -> /settings
- add separate logout page

### Affected components 

- Console

### Related Issues

#2660, #2663

### Solution and Design

- Rewrite login page and its associated actions. Add a checkbox to choose whether to persist the admin secret.
  - Persist: admin secret set to LS. admin secret is set in globals
  - Do not persist: admin secret is set in globals
- implement getAdminSecret fn which fetches from localstorage or globals
- On GraphiQL page load, dispatch header init:
  - read headers from LS. set actual admin secret value if admin secret header present
  - if admin secret is set and admin secret header is not present, add admin secret header and persist to LS that admin secret header wasAdded. Do not add admin secret header if admin secret header was already added once. ie. wasAdded value is set in LS.
  - persist headers to LS. mask admin secret value if header is present.
  - push headers to redux and mark headers are initialised as a flag
- Introspection happens only after headers are initialised

### Steps to test and verify

Cases:
- With vs Without admin secret
- Server vs CLI mode
- Persisted vs Non-persisted admin secret
- Prod vs Dev mode

Test:
- Login behaviour
- Logout behaviour
- Graphiql headers
- Data API headers


tirumaraiselvan:
Also, please make sure that when a role header is set then two queries are NOT made for introspection (first as admin, then as the role).

console/src/components/Login/Login.js
```diff
@@ -1,96 +1,115 @@
...
+                  type="checkbox"
+                  checked={shouldPersist}
+                  onChange={toggleShouldPersist}
+                  className={`${styles.add_mar_right_small} ${
+                    styles.remove_margin_top
+                  }`}
+                />
+                <label className={styles.add_mar_right_mid}>Remember in this browser</label>
+                <div style={{ marginTop: '2px' }} className={styles.cursorPointer}>
+                  <i className={'fa fa-question-circle'} data-toggle="tooltip" data-placement="bottom" title="Do not check this in production" />
```
rikinsk: Tooltip not required here. This message raises questions to why should I not check it in production. Users are smart enough to understand what 'remember me' means and its risks.

console/src/components/Login/Login.js
```diff
@@ -1,96 +1,115 @@
...
+              <div className={`${styles.display_flex} ${styles.add_pad_left}`}>
+                <input
+                  type="checkbox"
+                  checked={shouldPersist}
+                  onChange={toggleShouldPersist}
+                  className={`${styles.add_mar_right_small} ${
+                    styles.remove_margin_top
+                  }`}
+                />
+                <label className={styles.add_mar_right_mid}>Remember in this browser</label>
```
rikinsk: make the label clickable. it is irritating to have to point exactly at the checkbox

console/src/components/Login/Actions.js
```diff
@@ -0,0 +1,150 @@
...
+const adminSecretKeyString = `x-hasura-${globals.adminSecretLabel}`;
+
+export const verifyLogin = ({
+  adminSecret,
+  shouldPersist,
+  successCallback,
+  errorCallback,
+  dispatch,
+}) => {
+  console.log(adminSecret);
```
rikinsk: remove console.log

console/src/components/Login/Login.js
```diff
@@ -1,96 +1,115 @@
...
-                    <input
-                      onChange={this.handleAdminSecret}
-                      className={styles.form_input + ' form-control'}
-                      type="password"
-                      placeholder={`Enter ${globals.adminSecretLabel}`}
-                      name="password"
-                    />
-                  </div>
+  // form submit handler
+  const onSubmit = e => {
```
rikinsk: Cannot read this code block. There are 5 distinct things happening here which I need to read line by line to figure out. Please add spacing after logical blocks

```
 const onSubmit = e => {
    e.preventDefault();

    const successCallback = () => {
      setLoading(false);
      setError(null);
    };

    const errorCallback = err => {
      setAdminSecretInput('');
      setLoading(false);
      setError(err);
    };

    setLoading(true);

    verifyLogin({
      adminSecret: adminSecretInput,
      shouldPersist,
      successCallback,
      errorCallback,
      dispatch,
    });
  };
```

Issue #2461, opened 2019-07-02
Permissions for Remote Schemas


tirumaraiselvan:
v1 (wip #2690):

Schema masking: Role-based types for `query_root`, `mutation_root` and `subscription_root` fields
Deep schema masking: Role-based type for anything (object type, input object type, scalars?). 

v2:

Input argument presets.

leoalves:
Hi @tirumaraiselvan,

One idea, that I think would be easier, is to let the Remote Schema generate different schemas, for each role.

If Hasura sends the role, in the headers, when doing the introspection query.  The Remote Schema could return different schemas, based on each role.

I created an issue about that a while ago.
https://github.com/hasura/graphql-engine/issues/2139

tirumaraiselvan:
@leoalves Cool, 1) is actually the exact duplicate of #2139 . 

I will let both these issues open and use this as an umbrella issue for remote schema permissions.

bobberkarl:
Anything new regarding that?

tirumaraiselvan:
@bobberkarl This is WIP currently. It would be great if you could explain some more details of what kind of permissions you expect for Remote Schemas. Does the description of this issue satisfy your use-cases?

cjonestcgsi:
@tirumaraiselvan  Answering for a different company/use case.. In the case of an integration with Stripe - we would want to let a user see their own transactions and balance and certain status related details - such as the 'requirements' that Stripe may have for them, so we would want to query the remote schema using the stripe-customer-id, associated with the X-Hasura-User-Id, and to allow queries to certain tables and columns, where we force it to return results only for that particular stripe-customer-id.

leoalves:
@cjonestcgsi 
I think the use case you explained will be solved with Remote Joins, when that gets released.
https://hasura.io/remote-joins

tirumaraiselvan:
@cjonestcgsi leoalves is right. Remote joins allows you to join tables with remote data. So if you have access to only `stripe-customer-id` (which is `x-hasura-user-id`) then you will get joined data corresponding to only that. The strip use-case is also presented here: https://hasura.io/blog/remote-joins-a-graphql-api-to-join-database-and-other-data-sources/

jacdx:
[...]

PR #2690, opened 2019-08-08
Add permissions for Remote Schema v1 (server only)

tirumaraiselvan:
### Description

See: #2461 

### Affected components 

- Server
- Tests

### Todo

- [x] Migrations for existing users
- [x] Tests
- [x] Build schema cache only if there are remote schemas
- [x] hdb_permission_agg should also have remote roles
- [x] metadata (export, replace, clear) 
- [x] inconsistent metadata
- [x] introspection endpoint for remote schemas from schema cache
- [x] change migration logic
- [x] change api definition

### Steps to test and verify

```
type: add_remote_schema_permissions
args:
  remote_schema: comm
  role: user
  definition:
    allowed_objects:
    - type: Query
      fields: [hello, user]
    - type: User
      fields: [a,b,c]
```

```
type: introspect_remote_schema
args:
  name: hello
```

Console and docs to be done in a separate PR.

Note that the API only authorizes top-level fields in the root types. Deep nested fields will be supported after RJ merge.


tirumaraiselvan:
@rakeshkky Do we need to consider any schema dependencies for this?

tirumaraiselvan:
Alternate possible API:

type: add_remote_schema_permissions
args:
  remote_schema: comm
  role: user
  definition:
  - type: Query
    fields: [hello, user]
  - type: User
    fields: [a,b,c]

rakeshkky:
> @rakeshkky Do we need to consider any schema dependencies for this?

No. There are no Postgres dependencies for remote schema permissions.

tirumaraiselvan:
> > @rakeshkky Do we need to consider any schema dependencies for this?
> 
> No. There are no Postgres dependencies for remote schema permissions.

Right. But what about remote schema dependencies?

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
```diff
@@ -156,3 +167,130 @@ fetchRemoteSchemas =
...
+        checkUniqueTypes (rsPermDefinition remoteSchemaPerm)
+        validateTypes (rsPermDefinition remoteSchemaPerm) (GC._rgTypes $ rscGCtx remoteSchemaCtx)
+      newGCtx <- updateRemoteGCtxFromTypes newTypes (rscGCtx remoteSchemaCtx)
+      pure $ remoteSchemaCtx { rscGCtx = newGCtx }
+
+checkUniqueTypes :: (QErrM m) => [RemoteTypePerm] -> m ()
+checkUniqueTypes typePerms = do
+  let types = map rtpType typePerms
+      nubbedTypes = List.nub types
+  if types /= nubbedTypes
```
rakeshkky: There's `duplicates` function in `Hasura.Server.Utils` module. Does it fit here?

tirumaraiselvan: It probably does. But why use a function that is already available in a core lib?

server/src-exec/Migrate.hs
```diff
@@ -360,6 +375,86 @@ from21To22 = do
...
+    Q.multiQE
+      defaultTxErrorHandler
+      $(Q.sqlFromFile "src-rsr/migrate_from_22_to_23.sql")
+  allRoles <- liftTx $ getAllRoles
+  buildSchemaCacheStrict
+  sc <- askSchemaCache
+  let remoteSchemas = Map.toList (scRemoteSchemas sc)
+      remoteSchemaPermissions =
+        concat $
+        map
```
rakeshkky: Use `concatMap`

server/src-exec/Migrate.hs
```diff
@@ -360,6 +375,86 @@ from21To22 = do
...
+          (\(name, ctx) ->
+             map (uncurryPerm . (, name, getTypePerms ctx)) allRoles)
+          remoteSchemas
+  liftTx $ mapM_ addRemoteSchemaPermissionsToCatalog remoteSchemaPermissions
+  migrateMetadata False migrateMetadataFor23
+  setAsSystemDefinedFor23
+  return ()
```
rakeshkky: `return ()` is required?

server/src-exec/Migrate.hs
```diff
@@ -360,6 +375,86 @@ from21To22 = do
...
+                     _                    -> Nothing)
+                types
+        convObjTypesToPermType =
+          map
+            (\objTy ->
+               let name = (VT._otiName objTy)
+                   fields = Map.keys (VT._otiFields objTy)
+                in RemoteTypePerm name fields)
+        convInpObjTypesToPermType =
+          map
```
rakeshkky: use `flip map` to avoid surrounding lambda functions in parenthesis.

server/src-lib/Hasura/GraphQL/RemoteServer.hs
```diff
@@ -99,35 +99,48 @@ fetchRemoteSchema manager name def@(RemoteSchemaInfo url headerConf _ timeout) =
...
-mergeSchemas rmSchemaMap gCtxMap = do
-  def <- mkDefaultRemoteGCtx remoteSchemas
-  merged <- mergeRemoteSchema gCtxMap def
-  return (merged, def)
+  -> m GS.GCtxMap
+mergeSchemas rmSchemaMap rmSchemaMapWithRole initGCtxMap = do
+  merged <- mergeRoleRemoteSchemas initGCtxMap rmSchemaMapWithRole
+  remotes <- combineRemoteGCtx remoteSchemas
+  finalMerged <- addRemoteSchemaToAdminRole merged remotes
+  return finalMerged
```
rakeshkky: `addRemoteSchemaToAdminRole merged remotes` in last line will work fine.

```haskell
= do
  b <- f a
  return b
```
can be rewritten as
```haskell
= f a
```

server/src-lib/Hasura/GraphQL/RemoteServer.hs
```diff
@@ -99,35 +99,48 @@ fetchRemoteSchema manager name def@(RemoteSchemaInfo url headerConf _ timeout) =
...
-  -> GS.GCtx
+  -> RemoteSchemasWithRole
   -> m GS.GCtxMap
-mergeRemoteSchema ctxMap mergedRemoteGCtx = do
-  res <- forM (Map.toList ctxMap) $ \(role, gCtx) -> do
-    updatedGCtx <- mergeGCtx gCtx mergedRemoteGCtx
-    return (role, updatedGCtx)
-  return $ Map.fromList res
+mergeRoleRemoteSchemas initGCtxMap roleRemoteSchemas = do
+  Map.traverseWithKey
```
rakeshkky: use `flip` to avoid parenthesis in lambda functions

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -583,12 +583,12 @@ buildGCtxMapPG
...
-  writeSchemaCache sc {scGCtxMap = gCtxMap}
+  writeSchemaCache sc { scGCtxMap = gCtxMap }
 
-getGCtx :: (CacheRM m) => RoleName -> GCtxMap -> m GCtx
+getGCtx :: (QErrM m) => RoleName -> GCtxMap -> m GCtx
 getGCtx rn ctxMap = do
-  sc <- askSchemaCache
-  return $ fromMaybe (scDefaultRemoteGCtx sc) $ Map.lookup rn ctxMap
+  -- lookup role or return empty
+  pure $ fromMaybe emptyGCtx $ Map.lookup rn ctxMap
```
rakeshkky: This should throw `500 Internal error` if  `GCtx` for a `RoleName` is not found!

tirumaraiselvan: That would be a breaking change. This function never threw error. `scDefaultRemoteGCtx sc` can also be an empty value.

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
```diff
@@ -94,10 +107,38 @@ removeRemoteSchemaP2
...
+dropRemoteSchemaDirectDeps rsn = do
+  sc <- askSchemaCache
+  let depObjs = getDependentObjs sc objId
+      permDeps = catMaybes $ map getPermObj depObjs
+  mapM_ dropRemoteSchemaPermDep permDeps
+  where
+    objId = SORemoteSchema rsn
+    getPermObj = \case
+      obj@(SORemoteSchemaObj _(RSOPerm _)) -> Just obj
+      _ -> Nothing
```
rakeshkky: I'd prefer to throw `500 Internal error` with `unexpected dependent object`

rakeshkky: Oh! It's there in `dropRemoteSchemaPermDep`. It's better to `mapM_ dropRemoteSchemaPermDep` on `depObjs` directly and void `getPermObj`.

server/src-lib/Hasura/GraphQL/Schema.hs
tirumaraiselvan: Another way to think about this being that since `role` is a runtime value, it would seem logical to give an empty GCtx (and not make any asserts) which can throw validation failures later.

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
```diff
@@ -156,3 +195,176 @@ fetchRemoteSchemas =
...
+runAddRemoteSchemaPermissions q = do
+  adminOnly
+  newRSCtx <- runAddRemoteSchemaPermissionsP1 q
+  runAddRemoteSchemaPermissionsP2 q newRSCtx
+  pure successMsg
+
+runAddRemoteSchemaPermissionsP1
+  :: (QErrM m, CacheRM m) => RemoteSchemaPermissions -> m RemoteSchemaCtx
+runAddRemoteSchemaPermissionsP1 remoteSchemaPermission = do
+  validateRemoteSchemaPermissions remoteSchemaPermission
```
rakeshkky: Why another function?

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
```diff
@@ -778,3 +792,8 @@ getDependentObjsWith f sc objId =
     induces (SOTable tn1) (SOTableObj tn2 _) = tn1 == tn2
     induces objId1 objId2                    = objId1 == objId2
     -- allDeps = toList $ fromMaybe HS.empty $ M.lookup objId $ scDepMap sc
+
+getAdminGCtx :: SchemaCache -> GC.GCtx
+getAdminGCtx sc = do
+  maybe GC.emptyGCtx id $
```
rakeshkky: `fromMaybe` ?

server/src-lib/Hasura/RQL/Types/RemoteSchema.hs
```diff
@@ -94,3 +113,11 @@ validateRemoteSchemaDef (RemoteSchemaDef mUrl mUrlEnv hdrC fwdHdrs mTimeout) =
     hdrs = fromMaybe [] hdrC
 
     timeout = fromMaybe 60 mTimeout
+
+data DropRemoteSchemaPermissions
+  = DropRemoteSchemaPermissions
+  { drsPermRemoteSchema :: RemoteSchemaName
+  , drsPermRole         :: RoleName
```
rakeshkky: Use `permission` instead of `perm` as it reflects in JSON interface.

server/src-lib/Hasura/GraphQL/Schema.hs
rakeshkky: I agree! Should be this function operate in a monadic context? I don't think we're throwing an exception here.

tirumaraiselvan: Right. I should remove this from monadic context, the QErr is a remnant from before where I actually threwError and realized it was breaking change :)

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
tirumaraiselvan: because it's clear that P1 only does validate.

server/src-lib/Hasura/RQL/Types/RemoteSchema.hs
tirumaraiselvan: `perm` is dropped in the json

server/src-lib/Hasura/RQL/Types/Metadata.hs
```diff
@@ -39,11 +41,17 @@ data TableMetadataObjId
   deriving (Show, Eq, Generic)
 instance Hashable TableMetadataObjId
 
+data RemoteMetadataObjId
+  = RMORole !RoleName
```
rakeshkky: `RMOPermission`

server/src-lib/Hasura/GraphQL/RemoteServer.hs
```diff
@@ -99,35 +99,44 @@ fetchRemoteSchema manager name def@(RemoteSchemaInfo url headerConf _ timeout) =
...
-  res <- forM (Map.toList ctxMap) $ \(role, gCtx) -> do
-    updatedGCtx <- mergeGCtx gCtx mergedRemoteGCtx
-    return (role, updatedGCtx)
-  return $ Map.fromList res
+mergeRoleRemoteSchemas initGCtxMap roleRemoteSchemas = do
+  foldM
+    (\acc ((role, _rsName), remoteCtx) -> do
+       let roleGCtx = convRemoteGCtx (rscGCtx remoteCtx)
+       merged <- mergeGCtx (GS.getGCtx role acc) roleGCtx
+       pure $ Map.insert role merged acc)
```
rakeshkky: reduce `()`

server/src-lib/Hasura/RQL/DDL/Metadata.hs
```diff
@@ -419,6 +453,19 @@ fetchMetadata = do
...
+        RemoteSchemaMeta
+          name
+          def
+          comment
+          (Just
+             (mapMaybe
+                (\(RemoteSchemaPermissions name' role permDef) ->
+                   if name == name'
+                     then Just (RemoteSchemaPermMeta role permDef)
+                     else Nothing)
```
rakeshkky: We can totally avoid this if do joining in SQL itself

server/src-lib/Hasura/RQL/DDL/Metadata.hs
```diff
@@ -352,18 +381,23 @@ fetchMetadata = do
   functions <- map (uncurry QualifiedObject) <$>
     Q.catchE defaultTxErrorHandler fetchFunctions
 
-  -- fetch all custom resolvers
-  schemas <- DRS.fetchRemoteSchemas
-
+  -- fetch all remote schemas
+  remoteSchemas <- DRS.fetchRemoteSchemas
+    -- fetch all remote schema perms
+  remoteSchemaPerms <- DRS.fetchRemoteSchemaPerms
```
rakeshkky: fetch remote schemas along with their permission meta in `SQL`? Since `RemoteSchemaPermMeta` has `FromJSON` instance we can write `SQL` in such a way that it emits an array of json objects instead of merging them in Haskell code

tirumaraiselvan: I considered this but it was getting ugly to write a query which conforms to RemoteSchemaMeta mostly because it's not a simple JOIN alone, you also need to aggregate on the permissions so that one remote schema -> one row. This seemed much simpler. Do you see any problem with this approach?

rakeshkky: You can make use of `json_agg` with `group by`. It is much safer to do this in SQL in one shot :) Also fast too.

tirumaraiselvan: I agree in general that is true. But here, I think, it is much safer to make 2 simple queries vs 1 complex query. Also, there is no need to be concerned about performance here.

This is the SQL btw, seems like we have just transferred the complexity from the code to the SQL which seems to me to be a worse bargain:

```
SELECT rs.name, rs.definition::json, rs.comment, rsp.perms::json
FROM hdb_catalog.remote_schemas rs
LEFT JOIN (
     SELECT remote_schema, json_agg(json_build_object ('role', role, 'definition', definition)) as perms
     FROM hdb_catalog.remote_schema_permissions rsp
     GROUP BY remote_schema
      ) rsp
ON rs.name = rsp.remote_schema
 ```

Comment (tirumaraiselvan) [reply server/src-lib/Hasura/RQL/DDL/Metadata.hs]: If you still prefer this, I can make this change.

File: server/src-lib/Hasura/RQL/DDL/Metadata.hs (hunk 14 above)
Comment (tirumaraiselvan) [reply server/src-lib/Hasura/RQL/DDL/Metadata.hs]: See discussion here: https://github.com/hasura/graphql-engine/pull/2690#discussion_r321581083

File: server/src-lib/Hasura/RQL/Types/SchemaCache.hs [hunk 16]
```diff
@@ -774,7 +788,12 @@ getDependentObjsWith f sc objId =
     isDependency deps = not $ HS.null $ flip HS.filter deps $
       \(SchemaDependency depId reason) -> objId `induces` depId && f reason
     -- induces a b : is b dependent on a
-    induces (SOTable tn1) (SOTable tn2)      = tn1 == tn2
-    induces (SOTable tn1) (SOTableObj tn2 _) = tn1 == tn2
-    induces objId1 objId2                    = objId1 == objId2
+    induces (SOTable tn1) (SOTableObj tn2 _)               = tn1 == tn2
```
tirumaraiselvan: pls confirm that `induces (SOTable tn1) (SOTable tn2)      = tn1 == tn2` is actually redundant and hence removed here.

server/src-lib/Hasura/RQL/DDL/Metadata.hs
rakeshkky: If Postgres can able to give me correct `JSON`s then I'm happy to parse them as Haskell values and use them directly. Preparing the SQL query makes it fast. We cannot "prepare" Haskell code ;)

server/src-exec/Main.hs
```diff
@@ -215,12 +217,13 @@ main =  do
...
       pool <- getMinimalPool pgLogger ci
-      res <- runAsAdmin pool sqlGenCtx httpManager $ execQuery queryBs
+      featureFlags <- getFeatureFlags
+      res <- runAsAdmin pool sqlGenCtx httpManager featureFlags $ execQuery queryBs
       either printErrJExit BLC.putStrLn res
 
     HCVersion -> putStrLn $ "Hasura GraphQL Engine: " ++ T.unpack currentVersion
   where
-
+    getFeatureFlags = undefined
```
rakeshkky: `undefined` :fearful:

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
```diff
@@ -380,17 +386,22 @@ $(deriveToJSON (aesonDrop 2 snakeCase) ''FunctionInfo)
 type TableCache columnInfo = M.HashMap QualifiedTable (TableInfo columnInfo) -- info of all tables
 type FunctionCache = M.HashMap QualifiedFunction FunctionInfo -- info of all functions
 
+data RemoteSchemaPermsDisabled = RemoteSchemaPermsDisabled
```
tirumaraiselvan: this should be removed

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -99,25 +100,28 @@ $(deriveJSON
...
 instance HasHttpManager Run where
-  askHttpManager = asks _2
+  askHttpManager = asks (\(_, httpManager, _, _) -> httpManager)
 
 instance HasSQLGenCtx Run where
-  askSQLGenCtx = asks _3
+  askSQLGenCtx = asks (\(_, _, sqlGenCtx, _) -> sqlGenCtx)
+
+instance HasFeatureFlags Run where
+  askFeatureFlags = asks (\(_, _, _, featureFlags) -> featureFlags)
```
rakeshkky: Use `(^. _4)` (from Control.Lens module) instead of `\(_, _, _, featureFlags) -> featureFlags`. Similarly for `askSQLGenCtx` etc.

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
```diff
@@ -123,12 +169,26 @@ buildGCtxMap = do
...
+  mergedGCtxMap <- mergeRemoteSchemas (scRemoteSchemas sc) gCtxMap
+  allRemotesGCtx <- mkAllRemoteGCtx (scRemoteSchemas sc)
   writeSchemaCache sc { scGCtxMap = mergedGCtxMap
-                      , scDefaultRemoteGCtx = defGCtx
+                      , scAllRemoteGCtx = allRemotesGCtx
                       }
 
+mkCacheForRemoteSchema
+  :: (QErrM m) => SchemaCache -> RemoteSchemaName -> m SchemaCache
+mkCacheForRemoteSchema sc rsName = do
```
rakeshkky: Can we have this as `RemoteSchemaMap -> RemoteSchemaName -> m SchemaCache`, which has less confusion and make sense for the function name it has.

server/src-rsr/initialise.sql
```diff
@@ -607,3 +596,35 @@ CREATE TABLE hdb_catalog.hdb_allowlist
...
+CREATE TABLE hdb_catalog.remote_schema_permissions
+(
+  remote_schema TEXT,
+  role TEXT,
+  definition JSONB NOT NULL,
+  PRIMARY KEY (remote_schema, role),
+  FOREIGN KEY (remote_schema) REFERENCES hdb_catalog.remote_schemas(name) ON UPDATE CASCADE
+);
+
+CREATE VIEW hdb_catalog.hdb_permission_agg AS
```
0x777: We can leave this definition as is and then create a view for only roles as follows:
```sql
create view hdb_roles as 
(select distinct role_name from hdb_permission),
UNION ALL
(select distinct role_name from hdb_remote_schema_permission)
```
and then you can have array relationships to `hdb_permission_agg` and `hdb_remote_schema_permission` on `hdb_roles` table.

server/src-rsr/initialise.sql
```diff
@@ -607,3 +596,35 @@ CREATE TABLE hdb_catalog.hdb_allowlist
   collection_name TEXT UNIQUE
     REFERENCES hdb_catalog.hdb_query_collection(collection_name)
 );
+
+CREATE TABLE hdb_catalog.remote_schema_permissions
```
0x777: Our naming convention for catalog tables is singular. Change it to `remote_schema_permission`?

server/graphql-engine.cabal
```diff
@@ -90,6 +90,7 @@ library
 
                      -- URL parser related
                      , network-uri
+                     , http-api-data >= 0.4
```
0x777: maybe slightly overkill for using an instance defined there.

server/src-lib/Hasura/Server/App.hs
```diff
@@ -557,13 +573,17 @@ httpApp corsCfg serverCtx enableConsole consoleAssetsDir enableTelemetry = do
...
       post "v1alpha1/graphql" $ mkSpockAction GH.encodeGQErr id serverCtx $
-        mkPostHandler $ mkAPIRespHandler v1Alpha1GQHandler
+        mkPostHandler $ mkAPIRespHandler $ v1Alpha1GQHandler pure
 
       post "v1/graphql/explain" gqlExplainAction
 
       post "v1/graphql" $ mkSpockAction GH.encodeGQErr allMod200 serverCtx $
         mkPostHandler $ mkAPIRespHandler v1GQHandler
 
+      post ("v1/graphql/remote" <//> var) $ \(RemoteSchemaName -> remoteSchemaName) ->
```
0x777: This is needed currently for the console to show the remote schema types in the 'remote schema permissions' window. So to fetch the types, console will hit the `graphql/remote/<remote-name>` endpoint with an introspection query. 

I'm wondering regardless of this remote schema proxy endpoint, if it is good to have something like `execute_graphql_query` in `v1/query` or even scope it to something like `fetch_remote_schema_types`?

server/graphql-engine.cabal
tirumaraiselvan: yeah, i thought this might be too much.

server/src-lib/Hasura/RQL/Types/SchemaCacheTypes.hs
```diff
@@ -45,6 +54,8 @@ reportSchemaObj (SOTableObj tn (TOPerm rn pt)) =
   <> "." <> permTypeToCode pt
 reportSchemaObj (SOTableObj tn (TOTrigger trn )) =
   "event-trigger " <> qualObjectToText tn <> "." <> triggerNameToTxt trn
+reportSchemaObj (SORemoteSchema remoteSchemaName) = "remote schema " <> unNonEmptyText (unRemoteSchemaName remoteSchemaName)
```
tirumaraiselvan: use `remoteSchemaNameToTxt` here

server/src-lib/Hasura/Server/App.hs
tirumaraiselvan: I guess its fine to create a metadata api like `introspect_remote_schema` but I don't think an API like `execute_graphql_query` is much usable as it won't benefit from any graphql client side tooling.

server/src-rsr/initialise.sql
tirumaraiselvan: Except we already have `hdb_catalog.remote_schemas` but sure.

server/src-rsr/initialise.sql
tirumaraiselvan: Cool

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -116,8 +116,8 @@ validateTableConfig
 validateTableConfig tableInfo (TableConfig rootFlds colFlds) = do
     withPathK "custom_root_fields" $ do
       sc <- askSchemaCache
-      let defRemoteGCtx = scDefaultRemoteGCtx sc
-      validateCustomRootFlds defRemoteGCtx rootFlds
+      let allRemoteGCtx = scAllRemoteGCtx sc
```
tirumaraiselvan: I am not 100% sure if this should be `scAllRemoteGCtx` or simply `adminGCtx`

rakeshkky: I think it is `GCtx` of all remote schemas, which doesn't have any Postgres related `GCtx`. Hence `scAllRemoteGCtx`. 

Am I right @tirumaraiselvan ?

tirumaraiselvan: Yes, that's what scAllRemoteGCtx means but I am not sure if it should be that or if it should be GCtx of all (including PG)

server/tests-py/test_schema_stitching.py
```diff
@@ -453,12 +455,90 @@ def test_remote_query_timeout(self, hge_ctx):
...
+        assert st_code == 200, resp
+        yield
+        st_code, resp = hge_ctx.v1q(self.teardown)
+        assert st_code == 200, resp
+
+    def test_no_roles(self, hge_ctx):
+        check_query_f(hge_ctx, self.rs_dir + 'basic_fail.yaml')
+
+    def test_basic(self, hge_ctx):
+        st_code, resp = hge_ctx.v1q_f(self.rs_dir + 'setup.yaml')
```
nizar-m: The `setup` and `teardown` here can be moved to a fixture like this 
```python
# Do not use autouse=True here as it will then be applied to every test
@pytest.mark.fixture(scope='function') 
 def remote_schema_perms_basic(sef, hge_ctx):
        st_code, resp = hge_ctx.v1q_f(self.rs_dir + 'setup.yaml')
        assert st_code == 200, resp
        yield 
        st_code, resp = hge_ctx.v1q_f(self.rs_dir + 'teardown.yaml')
        assert st_code == 200, resp
```

And on the test, add the following marker

```python
@pytest.mark.usefixtures('remote_schema_perms_basic')
def test_basic(self, hge_ctx):
         check_query_f(hge_ctx, self.rs_dir + 'basic_success.yaml')
         check_query_f(hge_ctx, self.rs_dir + 'basic_fail_2.yaml')
```

This way the teardown will happen no matter whether the tests succeed or not.

server/tests-py/test_schema_stitching.py
```diff
@@ -453,12 +455,90 @@ def test_remote_query_timeout(self, hge_ctx):
...
+        assert st_code == 200, resp
+        # make good query
+        check_query_f(hge_ctx, self.rs_dir + 'basic_success.yaml')
+        # make bad query
+        check_query_f(hge_ctx, self.rs_dir + 'basic_fail_2.yaml')
+        st_code, resp = hge_ctx.v1q_f(self.rs_dir + 'teardown.yaml')
+        assert st_code == 200, resp
+        check_query_f(hge_ctx, self.rs_dir + 'basic_fail.yaml')
```
nizar-m: This test may be ignored, as it has already been done above.

tirumaraiselvan: This is to check whether the query returns the same thing after removing the perms.

tirumaraiselvan:
@lexi-lambda Over to you now

0x62:
Is there any ETA for landing this?

bobberkarl:
Hello @lexi-lambda, thanks for the great work. Any news/plans to merge it?

bobberkarl:
@tirumaraiselvan  did you modify the ui to provide remote schemas permissions?

tirumaraiselvan:
New version is available in #6079

PR #2696, opened 2019-08-08
Add optional arguments import

Maxim-Filimonov:
### Description
Without it Present, and Abscent types, used in *Construct GraphQL Mutation*, are not defined.

### Affected components 
- Community Content


Maxim-Filimonov:
Oh my bad, I haven't realised that those imports were added on the previous step. Jumped straight into update

PR #2701, opened 2019-08-09
allow creating permissions with conditions spanning tables (close #2512)

### LINKED ISSUE #2512: [RFC] Permissions: allow conditions spanning tables without relationships
Let's say reading from a table `accounts` should be allowed if there exists a row where `id = x-hasura-user-id AND is_admin = 'true'` in table `users`. Such a condition cannot be expressed as a boolean expression on `accounts` table.

Note that it is possible to workaround this by having an array relationship called `account_admins` to a view which cross joins `accounts` and `users` as follows:

```sql
create view account_admins as
select accounts.id as account_id, users.id as admin_id
from accounts, users
where users.is_admin = 'true';
```
However it is cumbersome to define such views and even the performance suffers. Let's say a user with `id=2` is reading from table `accounts`:

```sql
explain analyze
select
  *
from
  accounts
where
  exists (
    select
      1
    from
      users
    where
      is_admin = true
      AND id = 2
  );
```
has this query plan
```
Result  (cost=2.37..25.07 rows=1270 width=36) (actual time=0.025..0.025 rows=1 loops=1)
  One-Time Filter: $0
  InitPlan 1 (returns $0)
    ->  Index Scan using users_pkey on users  (cost=0.15..2.37 rows=1 width=0) (actual time=0.017..0.018 rows=1 loops=1)
          Index Cond: (id = 2)
          Filter: is_admin
  ->  Seq Scan on accounts  (cost=2.37..25.07 rows=1270 width=36) (actual time=0.005..0.006 rows=1 loops=1)
Planning time: 0.144 ms
Execution time: 0.055 ms
```

while

```sql
explain analyze
select
  *
from
  accounts a
where
  exists (
    select
      1
    from
      account_admins aa
    where
      aa.account_id = a.id
      AND aa.admin_id = 2
  );
```

has
```
Hash Semi Join  (cost=53.65..93.81 rows=1270 width=36) (actual time=0.126..0.128 rows=1 loops=1)
  Hash Cond: (a.id = accounts.id)
  ->  Seq Scan on accounts a  (cost=0.00..22.70 rows=1270 width=36) (actual time=0.010..0.010 rows=1 loops=1)
  ->  Hash  (cost=37.77..37.77 rows=1270 width=4) (actual time=0.041..0.041 rows=1 loops=1)
        Buckets: 2048  Batches: 1  Memory Usage: 17kB
        ->  Nested Loop  (cost=0.15..37.77 rows=1270 width=4) (actual time=0.032..0.036 rows=1 loops=1)
              ->  Index Scan using users_pkey on users  (cost=0.15..2.37 rows=1 width=0) (actual time=0.027..0.028 rows=1 loops=1)
                    Index Cond: (id = 2)
                    Filter: is_admin
              ->  Seq Scan on accounts  (cost=0.00..22.70 rows=1270 width=4) (actual time=0.003..0.004 rows=1 loops=1)
Planning time: 0.322 ms
Execution time: 0.179 ms
```

## Proposed solution

Extend RQL's boolean expression syntax with a new operator, say `_exists` (operates at the same level as `_and`, `_or`, `_not`) which has a schema as follows:

```
{
  "_table" : QualifiedTable,
  "_where" : BoolExp(_table)
}
```

For example, the above condition can now be represented as:

```json
{
  "_exists": {
    "_table" : "users",
    "_where" : {
      "id": "x-hasura-user-id",
      "is_admin": true
    }
  }
}
```

Implementing this as a operator also allows composing expressions across tables. For example:

```json
{
  "_or": [
    { "is_public": true},
    {
      "_exists": {
        "_table" : "users",
        "_where" : {
          "id": "x-hasura-user-id",
          "is_admin": true
        }
      }
    }
  ]
}
```
This says that a row is allowed to be read if `is_public` column is set to `true` or if it is being requested by an admin user.

See #2297 for the original issue which prompted this RFC.

rakeshkky:
### Description

Allow defining conditions based on other tables using `_exists` operator in the boolean expression of permissions. Refer RFC issue https://github.com/hasura/graphql-engine/issues/2512 for more details.

### Affected components 

- Server
- Console
- Docs
- Tests

### Related Issues

close #2512 

### Solution and Design

- Introduce a new `_exists` operator in boolean expressions.
- Validate `_where` field in `_exists` expression with fields of the table given as `_table`
- Use `BEExists` constructor in `BoolExp` (see `Hasura.SQL.DML` module) to generate `EXISTS` SQL expression

### Steps to test and verify

- Follow any test case added in this PR.


server/src-lib/Hasura/RQL/DDL/Schema/Rename.hs
```diff
@@ -50,8 +55,8 @@ renameTableInCatalog newQT oldQT = do
   forM_ allDeps $ \case
     SOTableObj refQT (TORel rn) ->
       updateRelDefs refQT rn (oldQT, newQT)
-    -- table names are not specified in permission definitions
-    SOTableObj _ (TOPerm _ _)   -> return ()
+    SOTableObj refQT (TOPerm rn pt)   ->
+      updatePermFlds refQT rn pt $ RTable (oldQT, newQT)
```
rakeshkky: Now table names are present in permission definitions.

server/src-lib/Hasura/RQL/DDL/Schema/Rename.hs
```diff
@@ -269,11 +287,38 @@ updateCols qt rf permSpec =
       PCCols c -> PCCols $ flip map c $
         \col -> if col == oCol then nCol else col
 
-updateBoolExp
+updateTableInBoolExp :: RenameTable -> BoolExp -> BoolExp
+updateTableInBoolExp rt@(oldQT, newQT) (BoolExp bExp) = BoolExp $
+  case bExp of
```
rakeshkky: Can't able to use `traverse` since we've to deal `BEExists` value differently. This is followed everywhere in this PR. Suggest any other better way of implementing this.

server/src-lib/Hasura/RQL/Types/BoolExp.hs
```diff
@@ -77,34 +100,21 @@ parseGBoolExp
...
+      | k == "$exists" -> BoolExists <$> parseGExists f v <?> Key k
+      | k == "_exists" -> BoolExists <$> parseGExists f v <?> Key k
+      | otherwise      -> BoolFld <$> f (k, v)
     return $ BoolAnd boolExps
   _ -> fail "expecting an Object for boolean exp"
   where
     parseGBoolExpL v =
       parseJSON v >>= mapM (parseGBoolExp f)
 
-foldBoolExp :: (Monad m)
```
rakeshkky: Moved to GBoolExp.hs

server/src-lib/Hasura/RQL/DDL/Schema/Rename.hs
lexi-lambda: It took me a little bit of researching to figure out how to do this, but yes, `lens` can help here.

There are a handful of different ways to do this, but the easiest way is to use [`Control.Lens.Plated`](https://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Plated.html). That module provides a variety of functions for operating on *self-similar* data structures. [`Data.Data.Lens`](https://hackage.haskell.org/package/lens-4.17.1/docs/Data-Data-Lens.html) includes related functionality, via the [`template`](https://hackage.haskell.org/package/lens-4.17.1/docs/Data-Data-Lens.html#v:template) traversal, which finds all occurrences of a particular type in a data structure. However, `template` does not handle the *self-similar* case, where the type you’re looking for is inside itself—it only stops on the outermost occurrence of the type you’re looking for.

In contrast, `Control.Lens.Plated` provides an extremely powerful operation, [`transform`](https://hackage.haskell.org/package/lens-4.17.1/docs/Control-Lens-Plated.html#v:transform), which applies a function to every element of a self-similar tree, bottom-up. We can use this operation to implement these functions: we just find all the `GBoolExp`s inside each `BoolExp`, recursively, then apply an update function to each of them.

To do this, we’ll need to define `Plated` instances for `GBoolExp` and `GExists`, since those are the types we care about recursively traversing. However, if a `Data` instance exists already, then the body of a `Plated` instance can be empty, as it will default to using the [`uniplate`](https://hackage.haskell.org/package/lens-4.17.1/docs/Data-Data-Lens.html#v:uniplate) operation from `Data.Data.Lens`. Therefore, writing these instances is trivial:

```haskell
instance (Data a) => Plated (GExists a)
instance (Data a) => Plated (GBoolExp a)
```

We’ll also want to derive some lenses, prisms, and isomorphisms:

```haskell
$(makeLenses ''GExists)
$(makePrisms ''GBoolExp)
$(makeWrapped ''BoolExp)
```

Now we can write `updateTableInBoolExp` in just a couple lines of code:

```haskell
updateTableInBoolExp :: RenameTable -> BoolExp -> BoolExp
updateTableInBoolExp (oldQT, newQT) =
  over _Wrapped . transform $ _BoolExists.geTable %~ \rqfQT ->
    if rqfQT == oldQT then newQT else rqfQT
```

Here’s how it works:

  1. We start by using `over _Wrapped` to move “under” the `BoolExp` wrapper. (We derived a `Wrapped` instance for `BoolExp` using `makeWrapped`, which is what allows us to use `_Wrapped`.)

  2. We call `transform` to apply a function to all the self-similar components of `GBoolExp ColExp`.

  3. We use `_BoolExists.geTable` to get a `Traversal` that targets each of the tables inside the `GBoolExp ColExp`.

  4. We use `%~` (which is actually just an infix alias for `over`) to apply a function to each of the tables targeted by the traversal.

That’s it! `transform` is doing the heavy lifting here, working out how to do all the recursion for us.

---

Unfortunately, we can’t use `Control.Lens.Plated` to implement `updateFieldInBoolExp`, since the values we’re updating aren’t actually self-similar at the type level. However, that’s not actually a problem, since we don’t actually do anything special in the `BoolExists` case. Therefore, we can just keep using the implementation that uses `traverse` and explicit mutual recursion, and everything should continue to work:

```haskell
updateFieldInBoolExp
  :: (QErrM m, CacheRM m)
  => QualifiedTable -> RenameField -> BoolExp -> m BoolExp
updateFieldInBoolExp qt rf = fmap BoolExp . traverse (updateColExp qt rf) . unBoolExp
```

rakeshkky: In `updateFieldInBoolExp` function:
In the `BoolExists` case, the reference table changes. So, we need to specially handle that case.

rakeshkky: Refer to `updateColExp` in `BoolFld` case. It takes the reference table and updates the columns.

Issue #2698, opened 2019-08-08
Graceful shutdown


lorenzo:
Currently Hasura, once sent the `SIGINT` signal, terminates all client connections immediately instead of gracefully shutting them down. This means that clients receive a 500 error instead of an actual response.

This is usually a problem as it will look like from the client perspective that the server is having problems, when it may just mean that it is being updated or an instance being replaced.

My suggestion would be to stop responding for new requests, serve all current requests in flight and wait at most 30 seconds for the requests to finish before actually shutting the server down.

If there is any interest in this, I also volunteer to implement it, unless it turns out to be more difficult that I think it is.

dsandip:
@0x777 @lexi-lambda Thoughts?

lexi-lambda:
I think I discussed this with @jberryman very, very briefly when we were in Bangalore. Here’s my thoughts.

Generally, it would be nice to make Hasura shutdown gracefully, waiting until existing connections have finished and ensuring all transactions are committed. However, I don’t know how much effort that would be to implement, and I wonder if it isn’t strictly worth it. Most approaches to scaling services horizontally have this functionality built-in to their gateway/proxy/router, where once an instance has been marked for termination, the gateway will automatically stop routing traffic to it, wait for existing connections to finish, and only then terminate it. I know Heroku’s router does this automatically, and Kubernetes, Docker Swarm, and Amazon ECS all have functionality designed to support it. Generally, the keywords to look for are “connection draining.”

However, as a caveat, my experience with that sort of thing only involves HTTP connections, which are designed to finish quickly. I am not sure what those services do with open websocket connections, nor am I even certain what the appropriate way to handle open websocket connections when shutting down actually is.

tl;dr: I don’t think it makes sense to go to much effort to implement graceful shutdown in Hasura itself if it would just be duplicating functionality present in virtually all gateways, but I am also not certain it would be. @lorenzo, did you have a particular use case in mind that wouldn’t be handled well by draining connections in a gateway server?

lorenzo:
@lexi-lambda currently on mobile so I cannot elaborate on my answer. Just wanted to ask, would you be interested in a pull request for this, provided that the solution is not more difficult than I think it would be?

The appropriate way to close a WebSocket gracefully can only be done with a defined protocol, where a close packet is sent, and the client reconnects immediately, but is n general it is just ok to close the connection whenever. 

I’m more concerned about regular queries and mutations, not about subscriptions since websockets require reconnecting logic anyways

[...]

PR #2717, opened 2019-08-12
Implemented graceful shutdown for HTTP requests

lorenzo:
### Description

* Listens for SIGINT as the termination signal
* Stops accepting new connections once the signal is received
* Waits for all connections to be drained, before shutting down
* Forcefully kills all pending connections after 30 seconds

Currently this does not send a close message to websocket clients, I'd
like to submit that change as a separate pull request, but at least this
solve my biggest concern which is not getting confirmation for mutations
while restarting the server.

### Affected components 

- Server

### Related Issues
refs #2698 

### Steps to test and verify

I haven't codified this as a test, although I'm happy to provide it if you guide me in running the test suite and your preferred solution. I did this:

Save this to a `test.sh` file
```sh
curl 'http://localhost:8080/v1/graphql' \
-XPOST \
-H 'Content-Type: application/json' \
-H 'Pragma: no-cache' \
-H 'Connection: keep-alive' \
--data-binary $VERY_SLOW_QUERY -v -s > /dev/null
```

Then watched it:

```sh
watch -n 0.8 ./test.sh
```
 
Waited for a few requests to go through then killed the server:

```sh
kill -s INT $PID
```

Verified that no broken socket errors occur, only the final `connection refused`

This currently does not deal with gracefully shutting down websocket connections


jberryman:
Thanks for the contribution! The easiest way to run the integration tests is to do 

    $ scripts/dev.sh test

You can also run specific tests with `-k` , e.g.

    $ scripts/dev.sh test -k "test_jsonb_has_all[http]"

The script is very new so do let us know if it works on your machine.

lorenzo:
@jberryman It works on my machine, although I consistently get 2 failures in the master branch.

The test I would need to write requires that I kill the Hasura server, which is handled in the shell file and not in the python test harness. How do you suggest I implement this? The natural way would be to have the start/stop function in the python code and be able to start or stop the server per test if required.

jberryman:
@lorenzo I appreciate you trying out the script.

I think it's not terribly important to have an automated test for this behavior right now, but I'll let @lexi-lambda chime in with her review. The script you've posted is super helpful though thank you! 

Tested locally and change LGTM, didn't look closely at the failure we're getting on CI ^

lexi-lambda:
@lorenzo This looks good to me, but I have one question: given you’re listening on an arbitrary signal, anyway, why not just listen for `SIGTERM` instead of `SIGINT` and remove the Dockerfile change?

lorenzo:
@lexi-lambda I’m fine with that too, just did not want to break backwards compatibility, since the only thing that actually stops the docker container, even before my change, is a SiGINT. I’m fine we removing that line in this PR if you think we should go that way.

lorenzo:
@lexi-lambda I change the signal to `SIGTERM` I could not see a real benefit in using a different one, after all

lorenzo:
@lexi-lambda interesting, what would the advantage of your approach be?

lexi-lambda:
> what would the advantage of your approach be?

Mostly just consistency with what GHC already does. Originally I was thinking it might make it easier to implement graceful shutdown for things running on several different threads, but that isn’t actually true, since GHC only throws `UserInterrupt` to the main thread, anyway. It’s also very slightly simpler to handle both SIGTERM and SIGINT together, since you can just catch either exception. It’s not a big difference, though.

Anyway, it seems like, interestingly enough, these changes make the CLI tests fail! It looks like, for some reason or another, the graceful shutdown on SIGTERM causes the working tree to become dirty when it didn’t before. I am not immediately sure why that is the case.

lorenzo:
@lexi-lambda I think it is trivial with this code to install the same signal handler for SIGINT as well. Do you want me to add thet

lexi-lambda:
If you would like, but there’s no need to.

lorenzo:
@lexi-lambda I used my very limited knowledge on what I thought could be wrong in the tests, but I truly don't know what it is :(

lorenzo:
@lexi-lambda I finally managed to make the test script pass

PR #2728, opened 2019-08-13
Add query example to firebase JWT sample

gabrieljoelc:
### Description

Adding this very simple query example to vanilla JavaScript sample app was very helpful for me in getting Firebase JWT working with Hasura. I am learning GraphQL, Firebase, and React Native at the same time, so simplifying to things I know helped me cut through the complexities.

### Affected components 

- Community sample apps

### Steps to test and verify

Follow the `community/sample-apps/firebase-jwt/README.md`.

### DIFF
--- a/community/sample-apps/firebase-jwt/README.md
+++ b/community/sample-apps/firebase-jwt/README.md
@@ -21,7 +21,9 @@ This example is adapted from [this guide](https://firebase.google.com/docs/auth/
 
 ### Pre-requisites
 
-This example assumes that you already have Firebase Auth setup for your app.
+This example assumes that you already have completed:
+- Firebase Auth setup for your app
+- The backend section of Hasura Firebase authentication [tutorial](https://blog.hasura.io/authentication-and-authorization-using-hasura-and-firebase#step-one-hasura)
 
 ### Add the cloud function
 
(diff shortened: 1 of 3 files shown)


marionschleifer:
@gabrieljoelc first of all, thank you very much for your contribution! 

We try to keep our sample apps as simple as possible (in this case to learn about Firebase auth + JWT). Introducing another requirement like implementing a specific Hasura backend makes the process somewhat harder.

However, we would appreciate if you could make a PR with the whole sample app from this blogpost: https://blog.hasura.io/authentication-and-authorization-using-hasura-and-firebase/#step-one-hasura. Basically a sample solution of what we're building in the post.

Let me know what you think 🙂

marionschleifer:
@gabrieljoelc I'm closing this PR. Feel free to re-open anytime if you'd like to add something to this 🙂

gabrieljoelc:
@marionschleifer meant to get back to this but had a crazy last half of 2019.

I considered doing that before I opened this PR, but it was worth a shot.

I will say that adding a single button and it's handler doesn't add too much complexity to the original sample and it doesn't break any of the existing functionality (only adds to it).

I'll see if I have time to do a separate sample like you suggested during my holiday break.

Issue #2688, opened 2019-08-07
Support multiple Set-Cookie headers in remote schemas


macalinao:
If my remote schema returns multiple Set-Cookie headers (as in the case of `cookie` and `cookie.sig` when using https://github.com/pillarjs/cookies) only one of the headers goes through.

Is there a way to fix this?

tirumaraiselvan:
Sure, we will add this ASAP.

macalinao:
is there an ETA on when this is going to be released? :smile:

tirumaraiselvan:
@macalinao This has been fixed here https://github.com/hasura/graphql-engine/pull/2739 and should be out in the next release

PR #2739, opened 2019-08-16
get all Set-Cookie headers (fix # 2688)

tirumaraiselvan:
### Description

Read all `Set-Cookie` headers from Remote Schema response and fwd them to the client. 

### Affected components 

- Server

### Related Issues

#2688 

### Solution and Design

Read all `Set-Cookie` headers.


tirumaraiselvan:
@lexi-lambda I thought about the test-case, but then got put off because I will have to add a python graphql resolver which returns multiple "Set-Cookie" headers. Just seemed too much work for an obvious fix (I tested manually ofc). But in case you insist, I can do this as well.

lexi-lambda:
If it’s a lot of work, I agree that it probably isn’t worth the effort; just figured I’d ask. I’ll merge this once CI passes.

PR #2740, opened 2019-08-16
add configurable timeout to remote schema (close #2501)

### LINKED ISSUE #2501: Increase response timeout for remote schemas
We currently use the hasura graphql-engine with a custom remote schema that implements certain special graphql mutations. Some of these mutations can run longer than 30 seconds, which seems to be the default http request timeout of hasura. After the 30 seconds we receive the following error:

```
{
  "errors": [
    {
      "extensions": {
        "path": "$",
        "code": "unexpected"
      },
      "message": "HttpExceptionRequest Request {\n  host                 = \"our-host\"\n  port                 = 80\n  secure               = False\n  requestHeaders       = [(\"Content-Type\",\"application/json\"),(\"User-Agent\",\"hasura-graphql-engine/v1.0.0-beta.2\"),(\"x-hasura-role\",\"admin\")]\n  path                 = \"/graphql\"\n  queryString          = \"\"\n  method               = \"POST\"\n  proxy                = Nothing\n  rawBody              = False\n  redirectCount        = 10\n  responseTimeout      = ResponseTimeoutDefault\n  requestVersion       = HTTP/1.1\n}\n ResponseTimeout"
    }
  ]
}
```

It would be nice if the response timeout for remote schemas would be configurable or at least could be increased to something higher than 30 seconds, which seems pretty short to us.

tirumaraiselvan:
### Description

Adds an optional `timeout_seconds` argument to `add_remote_schema` API

### Affected components 

- Server
- Console
- Docs

### Related Issues

#2501

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Execute.hs
+++ b/server/src-lib/Hasura/GraphQL/Execute.hs
@@ -357,46 +357,41 @@ execRemoteGQ
   -> m (HttpResponse EncJSON)
 execRemoteGQ reqId userInfo reqHdrs q rsi opDef = do
   execCtx <- ask
-  let logger  = _ecxLogger execCtx
+  let logger = _ecxLogger execCtx
       manager = _ecxHttpManager execCtx
-      opTy    = G._todType opDef
+      opTy = G._todType opDef
   when (opTy == G.OperationTypeSubscription) $
     throw400 NotSupported "subscription to remote server is not supported"
   hdrs <- getHeadersFromConf hdrConf
-  let confHdrs   = map (\(k, v) -> (CI.mk $ CS.cs k, CS.cs v)) hdrs
+  let confHdrs = map (\(k, v) -> (CI.mk $ CS.cs k, CS.cs v)) hdrs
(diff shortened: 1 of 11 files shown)


server/src-lib/Hasura/GraphQL/Execute.hs
```diff
@@ -357,46 +357,41 @@ execRemoteGQ
...
-      finalHdrs  = foldr Map.union Map.empty hdrMaps
-      options    = wreqOptions manager (Map.toList finalHdrs)
-
+      hdrMaps =
+        [ Map.fromList confHdrs
+        , Map.fromList userInfoToHdrs
+        , Map.fromList clientHdrs
+        ]
+      finalHdrs = foldr Map.union Map.empty hdrMaps
+      options = wreqOptions manager (Map.toList finalHdrs) (Just $ timeout * 1000000)
```
ecthiender: Why not pass seconds to `wreqOptions` and do the conversion there?

tirumaraiselvan: because `wreqOptions` is a library function and you may need more granularity than seconds for that.

PR #2751, opened 2019-08-19
add gzip brotli compression to http responses (close #2674)

### LINKED ISSUE #2674: Allow compressing all responses from Hasura via a server flag
The console, and I am sure some client apps, struggles with the quantum of data in `/v1/query` and `v1/graphql` responses with any decently sized schema. Allowing compression via a server flag/env var is required to smoothen this experience a bit.

To be researched: What's the impact on client-side tools like Apollo, urql, etc.? Do they handle this without any additional configuration and with what compression? Also, how useful is the compression?

rakeshkky:
### Description

The server compresses the responses (success only) from `/v1/query` and `/v1/graphql` endpoints. The server chooses either [brotli](https://en.wikipedia.org/wiki/Brotli) or [gzip](https://en.wikipedia.org/wiki/Gzip) based on `Accept-Encoding` client header. If both are requested then the server picks `brotli`. 

### Affected components 

- Server
- Docs
- Build System (tests)
- Tests

### Related Issues

close #2674 

### Solution and Design

- Add a new `Hasura.Server.Compression` module which exposes the compression type and function to compress. 
- Compress the end response (if required) and log the details of compression.

### New things
Compression details are added to the logs. In the start-up log, the new field `enable_compression` is introduced which is a boolean value.

A new `content_encoding` field is added to `http_info` of `http-log`. It is either `brotli` or `gzip` and `null` if the encoding is not enabled or not requested by the client. See the example below.

```json
{
  "timestamp": "2019-08-23T02:05:28.059-05:00",
  "level": "info",
  "type": "http-log",
  "detail": {
    "operation": {
      "query_execution_time": 0.018795933,
      "user_vars": {
        "x-hasura-role": "admin"
      },
      "error": null,
      "request_id": "6283add2-db7b-4eec-9b1e-d14781893b2f",
      "response_size": 45,
      "query": null
    },
    "http_info": {
      "status": 200,
      "http_version": "HTTP/1.1",
      "url": "/v1/query",
      "ip": "127.0.0.1",
      "method": "POST",
      "content_encoding": "brotli"
    }
  }
}
```

### Steps to test and verify

In the console, make any query with the `network` tab opened in developer options. Check the size of response received when compression enabled and disabled.


ecthiender:
There is a wai middleware which does gzip https://hackage.haskell.org/package/wai-extra-3.0.28/docs/Network-Wai-Middleware-Gzip.html . Shouldn't we use this directly?

server/src-lib/Hasura/Server/App.hs
```diff
@@ -288,17 +290,29 @@ mkSpockAction qErrEncoder qErrModifier serverCtx apiHandler = do
       json $ qErrEncoder includeInternal qErr
 
     logSuccessAndResp userInfo reqId req result qTime = do
-      logSuccess logger userInfo reqId req (apiRespToLBS result) qTime
+      let reqIdHeader = (requestIdHeader, unRequestId reqId)
```
ecthiender: Does it make sense to make this a Wai Middleware and use it? So it is abstracted out, and to not introduce noise in business logic? Or am I over-thinking this?

ecthiender:
> There is a wai middleware which does gzip https://hackage.haskell.org/package/wai-extra-3.0.28/docs/Network-Wai-Middleware-Gzip.html . Shouldn't we use this directly?

@rakeshkky pointed out that we wanted to support Brotli compression as well (the wai middleware only does gzip), hence we rolled our own implementation.

server/src-lib/Hasura/Server/App.hs
rakeshkky: In `Middleware` implementation we should compress the `Response` and send it to the client. But the [library](http://hackage.haskell.org/package/wai-3.2.2.1/docs/Network-Wai.html#g:6) does not provide a function which simply does the following 
```haskell
modifyResponseBytes :: (ByteString -> ByteString) -> Response -> Response
```
We should pattern match on `Response` data type and handle all cases. That turns out to be doing more than what we required. Rather I think it's a library's job. I'm planning to stick with the current implementation.

docs/graphql/manual/deployment/graphql-engine-flags/config-examples.rst
```diff
@@ -163,3 +163,26 @@ the right content-type headers.
...
    you're using the assets on server Docker image, it might not be that latest
    version of console.
+
+.. _http-compression:
+
+HTTP Compression
+----------------
+
+To enable compressing responses from Hasura GraphQL Engine, pass the ``--enable-compression`` flag or
+set ``HASURA_GRAPHQL_ENABLE_COMPRESSION`` environment variable to ``true``.
```
marionschleifer: set *the* ``HASURA_GRAPHQL_ENABLE_COMPRESSION`` environment...

docs/graphql/manual/deployment/graphql-engine-flags/config-examples.rst
```diff
@@ -163,3 +163,26 @@ the right content-type headers.
...
+Only responses from ``/v1/query`` and ``/v1/graphql`` endpoints are compressed.
+The server looks for ``Accept-Encoding`` header in request.
```
marionschleifer: ...looks for *the* ``Accept-Encoding`` header...

docs/graphql/manual/deployment/graphql-engine-flags/config-examples.rst
```diff
@@ -163,3 +163,26 @@ the right content-type headers.
...
+If header contains ``br`` then the server uses `Brotli <https://en.wikipedia.org/wiki/Brotli>`__ compression else if header contains
```
marionschleifer: If *the* header...else if *the* header contains...

docs/graphql/manual/deployment/graphql-engine-flags/config-examples.rst
```diff
@@ -163,3 +163,26 @@ the right content-type headers.
...
+``gzip`` then the server uses `Gzip <https://en.wikipedia.org/wiki/Gzip>`__ compression.
+If both values are present then the server prefers ``Brotli`` over ``Gzip``.
+Also, the server sets ``Content-Encoding`` response header value to ``br`` if ``Brotli`` or ``gzip`` if ``Gzip``.
```
marionschleifer: ...sets *the* ``Content-Encoding`` response...

docs/graphql/manual/deployment/graphql-engine-flags/config-examples.rst
```diff
@@ -163,3 +163,26 @@ the right content-type headers.
...
+
+Run server in this mode using following docker command:
```
marionschleifer: Run *the* server...

docs/graphql/manual/deployment/graphql-engine-flags/reference.rst
```diff
@@ -196,6 +196,10 @@ For ``serve`` sub-command these are the flags and ENV variables available:
      - Set the logging level. Default: ``info``. Options: ``debug``, ``info``,
        ``warn``, ``error``.
 
+   * - ``--enable-compression``
+     - ``HASURA_GRAPHQL_ENABLE_COMPRESSION``
+     - Enable brotli/gzip compression for responses from ``/v1/query`` and ``/v1/graphql`` endpoints.
```
lorenzo: Is there a default?

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -603,6 +610,12 @@ logLevelEnv =
   , "Server log level (default: info) (all: error, warn, info, debug)"
   )
 
+enableCompressionEnv :: (String, String)
+enableCompressionEnv =
+  ( "HASURA_GRAPHQL_ENABLE_COMPRESSION"
+  , "Enable brotli/gzip compression for responses from '/v1/query' and '/v1/graphql' endpoints"
```
lorenzo: I think it should also say what the default is

docs/graphql/manual/deployment/graphql-engine-flags/reference.rst
lorenzo: I meant, is it enabled by default? Other `ENABLE_*` settings already state a default, usually `false` but at least it makes it less ambiguous.

server/src-lib/Hasura/Server/App.hs
rakeshkky: > How about adding a new function in this `where` block ...

The `logSuccess` requires the compressed response to log its size. Also, it requires the compression type. So, we need to compress beforehand, log and send it to client.

docs/graphql/manual/deployment/graphql-engine-flags/config-examples.rst
```diff
@@ -170,15 +170,15 @@ HTTP Compression
...
+set the ``HASURA_GRAPHQL_ENABLE_COMPRESSION`` environment variable to ``true``.
 Only responses from ``/v1/query`` and ``/v1/graphql`` endpoints are compressed.
-The server looks for ``Accept-Encoding`` header in request.
-If header contains ``br`` then the server uses `Brotli <https://en.wikipedia.org/wiki/Brotli>`__ compression else if header contains
+The server looks for the ``Accept-Encoding`` header in request.
+If the header contains ``br`` then the server uses `Brotli <https://en.wikipedia.org/wiki/Brotli>`__ compression else if the header contains
 ``gzip`` then the server uses `Gzip <https://en.wikipedia.org/wiki/Gzip>`__ compression.
 If both values are present then the server prefers ``Brotli`` over ``Gzip``.
-Also, the server sets ``Content-Encoding`` response header value to ``br`` if ``Brotli`` or ``gzip`` if ``Gzip``.
+Also, the server sets the ``Content-Encoding`` response header value to ``br`` if ``Brotli`` or ``gzip`` if ``Gzip``.
```
marionschleifer: Hey @rakeshkky the changes look good 🙂 just one thing: the end of this sentence `Also, the server sets the ``Content-Encoding`` response header value to ``br`` if ``Brotli`` or ``gzip`` if ``Gzip``.` is not quite clear to me. Could we rephrase this a bit?

shahidhk:
@rakeshkky It looks like these changes make `brotli` a system dependency to build, test and run GraphQL Engine?

If that is the case, shouldn't we update the contributing guides as well?

lexi-lambda:
I don’t have strong feelings about it either way, but how about `ENABLE_HTTP_COMPRESSION`? That seems both terse and unambiguous.

shahidhk:
Also, why do we need a flag to enable this? The response is only compressed if the right `Accept-Encoding` header is present in the request, right? Are there any issues if this is enabled silently?

lexi-lambda:
> Are there any issues if this is enabled silently?

Not that I know of. I’d have no problem enabling it by default, but I’ll defer to @rakeshkky’s judgement on that.

rakeshkky:
@shahidhk @lexi-lambda I've removed the `--enable-compression` flag in the latest commit. The server shouldn't have control over compression. It is the client which requests the compression based on `Accept-Encoding` header.

Issue #2572, opened 2019-07-19
Remote schema forward original user-agent


danghung-dev:
Hi everyone, 

I have to save all request ip address and user-agent.
Because all request is behind aws load balancer, so i can see original ip address in "x-forwarded-for" in headers. But user-agent is always "hasura-graphql-engine/v1.0.0-alpha44"

How can I forward original user-agent. Thank you very much.

tirumaraiselvan:
Thanks, we will add this as soon as possible.

danghung-dev:
Fantastic!
That's great!

RodolfoSilva:
@tirumaraiselvan @0x777 this is a great enhancement to use with auth systems.

tirumaraiselvan:
Thanks for the suggestions everyone. This is going to be part of the upcoming release: https://github.com/hasura/graphql-engine/pull/2753

danghung-dev:
Do we have any update? @tirumaraiselvan @RodolfoSilva 
Thanks.

RodolfoSilva:
@danghung-dev No, I don't. Maybe @tirumaraiselvan.

tirumaraiselvan:
@danghung-dev Sorry for the delay. We will definitely make it part of the next release.

RodolfoSilva:
@tirumaraiselvan thanks. :P

danghung-dev:
Do we have any update? @tirumaraiselvan

tirumaraiselvan:
@lexi-lambda Can we call this `X-Hasura-Original-User-Agent` and close this case? Not sure if this needs to be thought more than fixing a convention.

lexi-lambda:
Personally, I think we should just use `X-Forwarded-User-Agent`. It’s consistent with the meaning of the other de-facto standard headers like `X-Forwarded-For`, `X-Forwarded-Host`, and `X-Forwarded-Proto`, and it’s extremely unlikely to cause any trouble.

On a semi-tangential note, until very recently, I thought proprietary headers were supposed to be prefixed by `X-`, but I recently found out that convention was deprecated in 2012 in [RFC 6648](https://tools.ietf.org/html/rfc6648). So if we introduce Hasura-specific headers in the future, we should probably just prefix them with `Hasura-`, not `X-Hasura-`… but in this case I think `X-Forwarded-` is still the natural prefix to use here by convention.

danghung-dev:
Hello @tirumaraiselvan,

Do we have any update?

[...]

PR #2753, opened 2019-08-20
add timeout to remote schema calls (close #2501)

### LINKED ISSUE #2501: Increase response timeout for remote schemas

tirumaraiselvan:
### Description

 Adds an optional `timeout_seconds` argument to `add_remote_schema` API

### Affected components 

- Server
- Console
- Docs

### Related Issues

#2501


rikinsk:
@tirumaraiselvan Can you update the API changes in docs

server/src-lib/Hasura/Server/Utils.hs
```diff
@@ -212,7 +212,7 @@ httpExceptToJSON e = case e of
 -- ignore the following request headers from the client
 commonClientHeadersIgnored :: (IsString a) => [a]
 commonClientHeadersIgnored =
-  [ "Content-Length", "Content-MD5", "User-Agent", "Host"
+  [ "Content-Length", "Content-MD5", "Host"
```
0x777: Why is this removed?

tirumaraiselvan: https://github.com/hasura/graphql-engine/issues/2572

tirumaraiselvan: I am not sure if this was the best way. Review requested.

server/src-lib/Hasura/RQL/Types/RemoteSchema.hs
```diff
@@ -30,6 +30,7 @@ data RemoteSchemaInfo
   { rsUrl              :: !N.URI
   , rsHeaders          :: ![HeaderConf]
   , rsFwdClientHeaders :: !Bool
+  , rsTimeoutSeconds   :: !Int
```
lexi-lambda: Consider maybe using [`DiffTime`](https://hackage.haskell.org/package/time-1.9.3/docs/Data-Time-Clock.html#t:DiffTime) here to add a small amount of type safety? The main advantage, imo, is that the units are unambiguous (a `DiffTime` is always treated to mean seconds when converted to/from other numeric types).

server/src-lib/Hasura/HTTP.hs
```diff
@@ -25,9 +26,12 @@ hdrsToText hdrs =
 wreqOptions :: HTTP.Manager -> [HTTP.Header] -> Wreq.Options
 wreqOptions manager hdrs =
   Wreq.defaults
-  & Wreq.headers .~  contentType : userAgent : hdrs
+  & Wreq.headers .~  defaultHeaders <> hdrs
```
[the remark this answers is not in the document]
tirumaraiselvan: Right! Thanks.

server/src-lib/Hasura/Server/Utils.hs
0x777: I don't think we can forward the original `user-agent` as is given that we are not a proxy (currently we kind of are but we won't be in the future when you have remote joins). Maybe we can add headers like `hge-client-user-agent` and `hge-client-ip`?

tirumaraiselvan: Okay, I am just going to revert this commit. And take it in a separate PR

server/src-lib/Hasura/RQL/Types/RemoteSchema.hs
tirumaraiselvan: Turns out `DiffTime` is not an instance of `Lift` which is creeping into 20 odd places.

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -345,10 +345,16 @@ decodeHeader logenv headerInfos (hdrName, hdrVal)
      decodeBS = TE.decodeUtf8With TE.lenientDecode
 
 addDefaultHeaders :: [HTTP.Header] -> [HTTP.Header]
-addDefaultHeaders hdrs = hdrs ++
-  [ (CI.mk "Content-Type", "application/json")
-  , (CI.mk "User-Agent", "hasura-graphql-engine/" <> T.encodeUtf8 currentVersion)
-  ]
+addDefaultHeaders hdrs = defaultHeaders <> rmDefaultHeaders hdrs
```
tirumaraiselvan: @lexi-lambda This change has led me to believe that we have HTTP clients which are not using the Hasura.HTTP library basically (event trigger module and remote schema module):

The only modules which seem to be using this are:

Telemetry
Auth
Auth/JWT
CheckUpdates

Is it fine for EventTriggers and RemoteSchemas to maintain their own clients because these might be specific to these modules.

server/src-lib/Hasura/RQL/Types/RemoteSchema.hs
lexi-lambda: Odd. I guess that’s alright, then, it’s not a big deal.

server/src-lib/Hasura/Events/Lib.hs
lexi-lambda: This function appears to prioritize default headers over the headers that were passed in, but the code in `GraphQL.Execute` appears to prioritize the headers that were passed in over the default headers. Is the difference intentional?

Also, can this code at least share the `defaultHeaders` declaration with the one in `Hasura.HTTP`? Having it duplicated in both places seems worth avoiding if we can help it.

tirumaraiselvan: This was my question. Is it worth sharing these tiny functions, when the bulk of the client is specified per module itself? Shouldn't HTTP.Hasura give out a client with these baked in for standard usage? Those who are not using it may do their own thing including defaultHeaders?

lexi-lambda: I don’t know, as I haven’t looked at the code in detail, but my gut instinct is that

  1. these modules should probably be using `Hasura.HTTP`,

  2. but if they aren’t, and you don’t want to do the refactoring to make them use it right now, sharing the default headers is an alright first step.

If the first assumption isn’t true, then you could be right, but I’d be surprised if it were. Is there a reason these modules shouldn’t be using the same library?

tirumaraiselvan: I don't think there is any real reason except the library didn't have proper functions to use with Event Triggers. I guess we should make an attempt to consolidate the library then.

Cool, I am just going to share the functions then. Let's do the full consolidation in a separate no-op refactor PR.

lexi-lambda: Sounds good. 👍

PR #2757, opened 2019-08-20
improve relational insert docs

### LINKED ISSUE #239: upsert queries do not work for non admin roles

marionschleifer:
### Description
Add documentation on how to do relational inserts

### Affected components 

- Docs


docs/graphql/manual/mutations/insert.rst
```diff
@@ -241,33 +245,51 @@ in the response
...
+              "id": 11
             }
           ]
         }
       }
     }
 
+**How it works**
+
+Let's say an ``author`` has an ``object relationship`` called ``address`` to the ``addresses`` table and an ``array relationship`` called ``articles`` to the ``articles`` table.
```
0x777: We can move "Let's say an ... . We want to insert an author object along with its address and one article" to before the example so that there is more context?

docs/graphql/manual/mutations/insert.rst
```diff
@@ -241,33 +245,51 @@ in the response
...
+
+We want to insert an author object along with its address and one article. The order of actions are as follows:
```
0x777: An insert mutation is processed as follows:

docs/graphql/manual/mutations/insert.rst
```diff
@@ -241,33 +247,49 @@ in the response
         affected_rows
         returning {
           id
-          title
-          author {
-            id
-            name
+          name
+          address {
```
rikinsk13: can we also get the auto-generated `id`s, `address_id` and ``author_id`` fields in the response so that its clear what values have been auto-generated and set

docs/graphql/manual/mutations/upsert.rst
```diff
@@ -167,3 +167,18 @@ You can specify ``on_conflict`` clause while inserting nested objects
   To allow upserting in these cases, set ``update_columns: [<conflict-column>]``. By doing this, in case of a
   conflict, the conflicted column will be updated with the new value (which is the same value it had before and hence
   will effectively leave it unchanged) and will allow the upsert to go through.
+
+Upsert is not a substitute for update
```
rikinsk13: I would move this section as a child of the `convert insert mutation to upsert` _(and make content tree depth to 2)_ or just move it as a top level section right after that 

In its current location, as it is after the examples, I do not think it will have the desired impact as I am not sure if enough ppl will end up discovering it.

docs/graphql/manual/mutations/insert.rst
```diff
@@ -213,26 +213,32 @@ Insert an object and get a nested object in response
       }
     }
 
+Insert related data through relationships
```
rikinsk13: How about `Insert an object along with its related objects through relationships`. Seems to fit the naming convention of the other sections a little more snugly

docs/graphql/manual/mutations/insert.rst
```diff
@@ -213,26 +213,32 @@ Insert an object and get a nested object in response
...
-            author: {
+            id: 11,
+            name: "John",
+            address: {
+              data: {
+                id: 12,
+                location: "San Francisco"
+              }
+            },
+            articles: {
```
rikinsk13: can we add 2 articles here instead of just 1. I remember some user mentioning that they thought you could just add one nested object at a time based on just seeing the example. lets not ever have to answer that again

PR #2765, opened 2019-08-21
fix issue with editing boolean fields (fix #2536)

### LINKED ISSUE #2536: bug: cannot edit boolean properties of records when using Chrome
Using Google Chrome 75, I cannot edit boolean properties of individual records (I *can* edit them if I use Firefox 68). Clicking once to edit a boolean property opens a selection menu that immediately closes again. See GIF demonstrating problem (note, in the video I am only clicking the mouse *once* in an attempt to open the menu):

hasura_issue

This problem occurs with the following setup:

1. Create a new table named `user_profile` using the Hasura web console GUI. The table should have:
   1. An `id` column of type `uuid` and default `gen_random_uuid()`. This is the primary key.
   2. An `admin` column of type `boolean` and default `false`.
2. Save the table.
3. Use the web console GUI to insert a new row for the table using default values.
4. Navigate to the "browse rows" tab for the table and click to edit the row we just inserted.
5. Attempt to edit the `admin` boolean property (you can't).

This problem exists in Hasura v1.0.0-beta.3 using Google Chrome 75 (I've also tested with extensions disabled). The problem also exists in Hasura v1.0.0-beta.2, I've just dragged my feet on reporting it. The problem does not appear to exist in Firefox 68.

A reproduction of the problem can be seen here: https://hasura-issue-dfdeax.herokuapp.com/console/data/schema/public/tables/user_profile/browse.

- Using Google Chrome, simply click to edit the only table row. I'll take this reproduction down in about a week.

alireza-ahmadi:
### Description

Clicking on the boolean field value dropdown would open a select box but it got closed automatically, therefore, users were unable to edit boolean fields values. 

### Affected components 

- Console

### Related Issues

#2536 #2798 

### Solution and Design

The new code checks the radio button by changing its `checked` property instead of simulating a click on it (which would lead to the loss of focus and automatic closure of dropdown).

### Steps to test and verify

You can follow the exact reproduction steps that have been written in the original issue.

### DIFF
--- a/console/src/components/Services/Data/TableBrowseRows/EditItem.js
+++ b/console/src/components/Services/Data/TableBrowseRows/EditItem.js
@@ -82,7 +82,9 @@ class EditItem extends Component {
         refs[colName].valueNode = node;
       };
       const clicker = e => {
-        e.target.closest('.radio-inline').click();
+        e.target
+          .closest('.radio-inline')
+          .querySelector('input[type="radio"]').checked = true;
         e.target.focus();
       };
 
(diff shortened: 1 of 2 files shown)


alireza-ahmadi:
From what I see, the CI pipeline failure caused by some configuration issues and is not related to one line of change in this PR. Is there anything that can do about that?

PR #2790, opened 2019-08-28
fix console server-build script

lexi-lambda:
### Description

This is a tiny quick fix for the `npm run server-build` script: previously, the `rm static/dist/versioned/*.gz` step would fail on first run, since the glob would match no files, and `rm` would complain. This approach just deletes the whole directory, which is probably better, anyway, since it avoids leaving behind any lingering stale files.

### Affected components 

- Console


shahidhk:
@lexi-lambda We have a better version of this at https://github.com/hasura/graphql-engine-pro/blob/master/console/Makefile#L43

shahidhk:
I am adding a better version of this script in a while.

PR #2827, opened 2019-09-04
Implemented graceful shutdown for websockets

lorenzo:
### Description
Implements graceful shutdown for websockets. When the sockets are not actively receiving data they will receive a closing message and the server will promptly shut down.

### Affected components 
- Server

### Solution and Design
The general design I followed:

* Keep the knowledge of how to close connections inside the Websocket
  module
* Add a field to `WsServer` to hold a channel, the channel signals when
  the application is stopping
* Return a callback to `Main` that can be invoked to send a message to
  the channel

### Steps to test and verify

* Start the server
* Use `gql` with any subscription
* Send a `TERM` signal to the process. It should stop immediately


server/src-lib/Hasura/Server/App.hs
```diff
@@ -435,6 +435,14 @@ initErrExit e = do
     <> T.unpack (qeError e)
   exitFailure
 
+data HasuraApp
```
lexi-lambda: 👍 for pulling this out.

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -106,13 +107,18 @@ data WSServer a
   = WSServer
   { _wssLogger  :: L.Logger
   , _wssConnMap :: ConnMap a
+  , _wssShutdown :: STM.TChan ()
```
lexi-lambda: Using a `TChan` here seems unnecessarily powerful. Why not just `_wssShouldShutdown :: TVar Bool`? You can efficiently block on that becoming `True` using `atomically (guard =<< readTVar shouldShutdownVar)`.

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -176,12 +184,16 @@ createServerApp (WSServer logger@(L.Logger writeLog) connMap) wsHandlers pending
         WS.sendTextData conn msg
         writeLog $ WSLog wsId $ EMessageSent $ TBS.fromLBS msg
 
+      closeRef <- A.async $ do
+        _ <- STM.atomically $ STM.readTChan shutdownChan
```
lexi-lambda: Please don’t use `_ <-` if the result is `()`, anyway (and if it isn’t, prefer `void`).

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
lorenzo: Makes sense, I'll change that

lorenzo: The result looks less elegant, as basically I'm replicating the internals of `readChan`, let me know if that is what you had in mind

lorenzo:
@lexi-lambda added your suggestions, thanks for the review!

lorenzo:
@lexi-lambda I figured that my implementation was not entirely correct as new WS connections were not being rejected properly. By moving the logic one layer above I was able to achieve that goal.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -219,6 +224,9 @@ onConn (L.Logger logger) corsPolicy wsId requestHead = do
       _                   ->
         throw404 "only '/v1/graphql', '/v1alpha1/graphql' are supported on websockets"
 
+    assertAcceptingConns ShuttingDown = throw500 "server is shuting down"
```
lexi-lambda: Small typo here—should be “shutting,” not “shuting.”

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -219,6 +224,9 @@ onConn (L.Logger logger) corsPolicy wsId requestHead = do
...
+    assertAcceptingConns _ = pure ()
```
lexi-lambda: I’d explicitly pattern match on `AcceptingConns` here rather than using the wildcard pattern to make it more future-proofed against changes to `ServerStatus`.

lorenzo: good call

lorenzo:
@lexi-lambda That's a great question and I was also mystified at how new connections were apparently making their way in. My working theory right now is that those connections were already accepted before closing the socket but had not been yet handled by the Websockets server code. Those still in transit connections would apparently cause a race condition where they were accepted and then served as live queries after the call to closing all connections was already made. 

I'm actually not sure how to validate my theory as true, but it seemed to solve an edge case when rapidly trying to create new WS connections as I send the `TERM` signal. Does this make any sense at all to you, sorry for not being more clear about this myself?

lexi-lambda:
> My working theory right now is that those connections were already accepted before closing the socket but had not been yet handled by the Websockets server code. Those still in transit connections would apparently cause a race condition where they were accepted and then served as live queries after the call to closing all connections was already made.

Aha, that makes sense to me, thanks for explaining. The guard does seem necessary, then, but I have a couple small requests:

  1. Could you add support for shutting the server down to `Hasura.GraphQL.Transport.WebSocket.Server` instead of `Hasura.GraphQL.Transport.WebSocket`? The former module is somewhat poorly named, as it doesn’t really have anything to do with GraphQL, but I think this functionality belongs there, since it’s really agnostic to what the server is being used for.

     The approach I have in mind is to add a `_wssIsShuttingDown :: !(TVar Bool)` field to `WSServer` (and though this is subjective, I think a `Bool` is fine here; I think introducing a different sum type is premature), then add a `shutdownWSServer` export.

  2. I think it would be nice to return an HTTP 503 Service Unavailable response here rather than HTTP 500 Internal Server Error, since it’s more descriptive.

     I would also include a `Retry-After: 0` header in the response, since in a cluster of Hasura instances, we’d like the client to retry connecting immediately. (Unfortunately, AFAIK no browsers will actually automatically retry the request even if the header is present, so it probably isn’t actually that useful, but it feels like good HTTP hygiene.)

lorenzo:
> 1. ; I think introducing a different sum type is premature)

That was actually my first attempt and I immediately confused myself by writing the function `assertAcceptingConnection True = ...`  did `True` mean accepting connections or the opposite? (it was the opposite)

In the light that it took 5 seconds for me to get boolean blindness I resolved to make a cheap sum type

> 1. The approach I have in mind is to add a `_wssIsShuttingDown :: !(TVar Bool)` field to `WSServer`

You may remember that was the first version of this PR you reviewed. I got confused by the module naming thinking that it actually implemented a server-like structure, or that at least the code handling the very beginnings of a connection happened there.

I think that those 2 modules should swap names: `Hasura.GraphQL.Transport.WebSocket` is actually implementing the server (accepts connections, rejects, checks for the right route, and actaully create a server-like struct which in the case of WAI is a curried function expecting a connection)

I will attempt to put the logic back there in the best way it makes sense to me now. Let me know if this is what you had in mind.

lorenzo:
@lexi-lambda can you please review again?

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -145,13 +161,28 @@ createServerApp
...
   writeLog $ WSLog wsId EConnectionRequest
-  let reqHead = WS.pendingRequest pendingConn
-  onConnRes <- _hOnConn wsHandlers wsId reqHead
-  either (onReject wsId) (onAccept wsId) onConnRes
+  serverStatus <- STM.readTVarIO shutdownVar
+  case serverStatus of
+    AcceptingConns -> do
+      let reqHead = WS.pendingRequest pendingConn
+      onConnRes <- _hOnConn wsHandlers wsId reqHead
+      either (onReject wsId) (onAccept wsId) onConnRes
```
lexi-lambda: Is there a lingering race here? If `_wssShutdown` is set to `ShuttingDown` after `shutdownVar` is read but before `onAccept` adds the connection to the map, then I think it still won’t be closed. I think, fundamentally, we need to arrange for the insertion itself to be guarded by a `serverStatus` check.

This is tricky to get right (I didn’t think about it when I was reviewing this yesterday, either, so we’ve both gotten it wrong!), so maybe we should change `WSServer` to the following:

```haskell
data WSServer a
  = WSServer
  { _wssLogger :: !L.Logger
  , _wssStatus :: !(TVar (ServerStatus a))
  }

data ServerStatus a
  = AcceptingConns !(ConnMap a)
  | ShuttingDown
```

I think that should make a race of this sort essentially unrepresentable, assuming `shutdown` does the obvious thing of setting the state to `ShuttingDown` and reading the current connections in a single, atomic operation.

lorenzo: Very interesting, and very tricky indeed. I will play with that idea for a bit since it hasn't yet clicked for me, but at least it is obvious now that there are more potential races in the current solution. Thanks for noticing!

lorenzo:
@lexi-lambda Alright, I tried your suggestion and really liked how much safer the code looks like now. I had to do a few changes in order to keep things atomic, but I'm feeling more confident now that this is the right design.

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -163,17 +182,15 @@ createServerApp
...
-  serverStatus <- STM.readTVarIO shutdownVar
-  case serverStatus of
-    AcceptingConns -> do
+  status <- STM.readTVarIO serverStatus
+  case status of
+    AcceptingConns connMap -> do
       let reqHead = WS.pendingRequest pendingConn
       onConnRes <- _hOnConn wsHandlers wsId reqHead
-      either (onReject wsId) (onAccept wsId) onConnRes
+      either (onReject wsId) (onAccept connMap wsId) onConnRes
```
lexi-lambda: Oh, hm. I didn’t think about the possibility that one can just pull the `connMap` out and hold onto a reference to it—I think there’s still a race here. 😅 The problem is that `serverStatus` might not be `AcceptingCons` anymore by the time `_hOnConn` returns, so the connection should not be accepted.

Fundamentally, the `STMMap.insert` into the connection map *must be atomic* with respect to a check that the server is accepting connections. Otherwise there could always be a race. I’m not sure how to guarantee that at the type level (at least without linear types), since it is, indeed, always possible to hold onto a “stale” connection map this way.

I apologize for the back-and-forth on this. This change seems useful, so I’d definitely like to get it merged! If you’d like, I can make the (small) change I have in mind myself and push it.

lorenzo: I'm happy to make the changes, but if you think you have the solution already and will be faster at it, please go ahead. Otherwise point to to the place you think should be changed and I'll do it myself :)

lexi-lambda: Sure, if you’re willing to make the change, that’d be great! I don’t have any super specific solution in mind, but the relevant spot to change is here:

https://github.com/hasura/graphql-engine/blob/768baf6a769b91ccf0e2f5d651569f1ec247799c/server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs#L215

Currently, that uses the `connMap` value that was retrieved earlier, but it needs to instead get the map out of `serverStatus` again to ensure the server hasn’t shut down since then. If it has, we’re in a bit of a tight spot, since we’ve already accepted the connection, so all we can do is immediately close it and bail. Does that make sense?

lorenzo: Makes sense!

lorenzo:
@lexi-lambda can you please review again?

lorenzo:
Thanks @lexi-lambda I’m actually glad that you were so thorough, I’m happy with his more robust solution.

PR #2829, opened 2019-09-04
WIP: In tests, pass 'transport' to check_query_f where apparently erroneously missing

jberryman:
We've been running quite a few tests twice for 'http' where it looks
like we meant to be running for http and websocket.  It looks like many
of these were just mistaken omissions in a40bf10.

The result is quite a few test failures.

FYI found with pylint, e.g.:

    $ find . -iname "*.py" -not -path "./.hasura-dev-python-venv/*"  \
        | xargs pylint | grep "Unused argument 'transport'" \
        | grep -Eo '^([^:]*)' | sort | uniq

Tried to think of ways this could not be a future footgun. Options:

1. Maybe turn check_query and check_query_f into fixtures themselves?:
  http://www.obeythetestinggoat.com/a-pytest-pattern-using-parametrize-to-customise-nested-fixtures.html
I started on this path (making `check_query/_f` a method of HGECxt), with the idea of making `transport` a fixture with default returning "http", then user could override by passing transport to `check_query`, otherwise it would take from the pytest `@parameter`. But the work didn't seem justified...

2. Maybe pragmatic route would be to have a test that runs pylint over test files looking for unused parameters named 'transport'.

3. give `check_query` and `check_query_f` a nonptional first parameter so that it must be called like `check_query(locals(), ...)` and in the body there we check for the key "transport" and use it unless an explicit transport argument was passed.

**TODO** see removal of 'transport' in a21f6cd server/tests-py/test_graphql_queries.py
      in test_select_query_user_col_change(); was that intentional?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -146,7 +146,8 @@ jobs:
   # build the server binary, and package into docker image
   build_server:
     docker:
-    - image: hasura/graphql-engine-server-builder:20190826
+    - &server_builder_image
+      image: hasura/graphql-engine-server-builder:20191031
     working_directory: ~/graphql-engine
     steps:
     - attach_workspace:
@@ -174,8 +175,8 @@ jobs:
             echo "Branch starts with dev or release, or tagged commit starts with v. Optimized build"
             make ci-binary
           else
-            echo "Non-release branch, non-optimized build with coverage"
-            BUILD_FLAGS="--fast --coverage" make ci-binary
+            echo "Non-release branch, build with coverage"
+            BUILD_FLAGS="--coverage" make ci-binary
           fi
     - run:
         name: Build the docker image
@@ -222,7 +223,7 @@ jobs:
     environment:
       PG_VERSION: "12"
     docker:
-    - image: hasura/graphql-engine-server-builder:20190826
+    - *server_builder_image
     - image: circleci/postgres:12-alpine-postgis
       <<: *test_pg_env
  
@@ -231,7 +232,7 @@ jobs:
     environment:
       PG_VERSION: "11"
     docker:
-    - image: hasura/graphql-engine-server-builder:20190826
+    - *server_builder_image
     - image: circleci/postgres:10-alpine-postgis
       <<: *test_pg_env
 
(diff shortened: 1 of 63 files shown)


jberryman:
This exposes some potential bugs and inconsistencies that still need to be investigated. The approach in the last commit above may be the wrong one.

nizar-m:
Making `check_query_f` a fixture is a good idea. That would avoid the aforementioned types of inconsistencies.

jberryman:
It looks like many or all of these errors might be from strange behavior of the python graphql-core library (e.g. `null` in a request disallowed). The new library only supports python 3.6+ which is annoying: https://github.com/graphql-python/graphql-core-next

I also don't know whether it is actually an improvement for our purposes.

jberryman:
I wonder how much coordination is required if we move to requiring python 3.7 for local development. Thoughts?

Another issue concerning python versions (dev scripts don't work on python 3.7 due to ruamel version): https://github.com/hasura/graphql-engine/issues/3312

nizar-m:
> I wonder how much coordination is required if we move to requiring python 3.7 for local development. Thoughts?
> 
> Another issue concerning python versions (dev scripts don't work on python 3.7 due to ruamel version): #3312 

[pipenv](https://github.com/pypa/pipenv) along with [pyenv](https://github.com/pyenv/pyenv)  should solve these issues. With pipenv we can pin the packages and the Python version. Pyenv helps in installing and managing mulitple versions of Python. Also pipenv uses pyenv when it is available.

The main issue is that pipenv is [too slow](https://github.com/pypa/pipenv/issues/3840). The equivalent of pip freeze, `pipenv lock`, [can sometimes take as much as 3 minutes](https://github.com/pypa/pipenv/issues/2284#issuecomment-397867839).

Maybe we can use only pyenv to handle Python version.

jberryman:
Just trying out `pyenv` for the first time and seems to work smoothly for me (thanks!). I wonder if just recommending `pyenv` somewhere at a top-level document in this repo, and then checking in `.python-version` files would be sufficient (at first? I don't totally understand what pipenv adds; it sounds like we could move to it later if need be)

Stefmore02:
Closing this PR since it's quite out of date.

PR #2831, opened 2019-09-05
Test result ordering, add `--accept` test mode to automatically accept new/changed test cases

jberryman:
We add a new pytest flag `--accept` that will automatically write back
yaml files with updated responses. This makes it much easier and less
error-prone to update test cases when we expect output to change, or
when authoring new tests.

Second we make sure to test that we actually preserve the order of the
selection set when returning results. This is a "SHOULD" part of the
spec but seems pretty important and something that users will rely on.

To support both of the above we use ruamel.yaml which preserves a
certain amount of formatting and comments (so that --accept can work in
a failry ergonomic way), as well as ordering (so that when we write yaml
the order of keys has meaning that's preserved during parsing).

Use ruamel.yaml everywhere for consistency (since both libraries have
different quirks).

Quirks of ruamel.yaml:
- trailing whitespace in multiline strings in yaml files isn't written
  back out as we'd like: https://bitbucket.org/ruamel/yaml/issues/47/multiline-strings-being-changed-if-they
- formatting is only sort of preserved; ruamel e.g. normalizes
  indentation. Normally the diff is pretty clean though, and you can
  always just check in portions of your test file after --accept

**UPDATE**: marked xfail for the failing case we found case: https://github.com/hasura/graphql-engine/issues/3271 . This isn't great since we lose the other functionality those tests were excercising.

### DIFF
--- a/server/tests-py/test_tests.py
+++ b/server/tests-py/test_tests.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+
+# This module is for tests that validate our tests or test framework, make sure
+# tests are running correctly, or test our python test helpers.
+
+import pytest
+from super_classes import DefaultTestSelectQueries
+from validate import check_query_f, collapse_order_not_selset
+from ruamel.yaml.comments import CommentedMap
+
+class TestTests1(DefaultTestSelectQueries):
+    """
+    Test various things about our test framework code. Validate that tests work
+    as we expect.
+    """
+
+    # NOTE: We don't care about this for now, but should adapt this to test
+    # that xfail detection in code that handles `--accept` works correctly.
+    @pytest.mark.xfail(reason="expected")
+    def test_tests_xfail(self, request):
+        try:
+            marker = request.node.get_closest_marker("xfail")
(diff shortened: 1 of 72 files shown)


jberryman:
This exposes some potential bugs and inconsistencies that still need to be investigated.

nizar-m:
Awesome feature `--accept`.  Would love to see this merged.

jberryman:
**TODO** I should strip trailing whitespace from the graphql query string before writing back, so ruamel doesn't do ugly formatting.

server/tests-py/test_tests.py
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python3
+
+# This module is for tests that validate our tests or test framework, make sure
+# tests are running correctly, or test our python test helpers.
```
lexi-lambda: 👍 This is a nice detail.

jberryman:
Cool, thanks! (I don't have permissions to merge to master; let me know if there's anything else you need from me)

PR #2835, opened 2019-09-05
disable auto-update checks in cli-migrations container entrypoint

abn:
### Description
This change disables auto-update checks when the cli is used in the entrypoint script for the cli-migrations container. This serves no real purpose in this context and adds additional delays.

### Affected components 
- Docker Container (cli-migrations)

### Solution and Design
The fix was to add an additional configuration `show_update_notification` into the generated temporary `config.yaml` file.


shahidhk:
Thanks a lot @abn :smile:

abn:
@shahidhk thanks for the project ;)

Issue #2665, opened 2019-08-03
Type with PostGIS `geography` can't be saved using console editor


michaelhayman:
I'm 90% sure that this is the issue; on other sites on records without the geography type everything works.

I get an error when editing such records.

marionschleifer:
@michaelhayman could you please share a screenshot showing what exactly you wanted to edit and ideally also the logs? Thank you 🙂

ManUtopiK:
I think I got the same issue with `PosGIS geoghaphy`. When I edit a row in Hasura console, the geography value is displayed as `[object Object]`, not an object like:
```json
{
    "type": "Point",
    "crs": {
        "type": "name",
        "properties": {
            "name": "urn:ogc:def:crs:EPSG::4326"
        }
    },
    "coordinates": [
        -55.162353515625,
        5.82301611438301
    ]
}
```
So, when I want to save this edited row, I got this error:

    Edit failed!
    unexpected : unexpected character (at offset 1)

And I can't save any modifications...  
This is annoying, I have to change data from graphiql 😯 

<details><summary>Full error details</summary>

```js
{
    "statement": "WITH \"public_photo__mutation_result_alias\" AS (UPDATE \"public\".\"photo\" SET \"status\" = $1,\"uploadedAt\" = $2,\"wpID\" = $3,\"path\" = $4,\"createdAt\" = $5,\"location\" = ST_GeomFromGeoJSON($6 ),\"size\" = $7,\"medium\" = $8,\"country\" = $9,\"data\" = $10,\"format\" = $11,\"alert\" = $12,\"content\" = $13,\"userId\" = $14,\"base64\" = $15,\"id\" = $16,\"price\" = $17,\"updatedAt\" = $18,\"title\" = $19,\"stock\" = $20,\"agence\" = $21,\"file\" = $22  WHERE (('true') AND ((((((\"public\".\"photo\".\"id\") = ($23)) OR (((\"public\".\"photo\".\"id\") IS NULL) AND (($23) IS NULL))) AND ('true')) AND ('true')) AND ('true'))) RETURNING * ) SELECT  json_build_object('affected_rows', (SELECT  COUNT(*)  FROM \"public_photo__mutation_result_alias\"      ) )        ",
    "prepared": true,
    "error": {
        "exec_status": "FatalError",
        "hint": null,
        "message": "unexpected character (at offset 1)",
        "status_code": "XX000",
        "description": null
    },
    "arguments": [
[...]

PR #2843, opened 2019-09-06
fix console insert/edit row glitches (close #2840, #2665)

### LINKED ISSUE #2840: 'default' radio button is disabled for identity columns
Create a table as follows:
```sql
CREATE TABLE public.page (
    id INTEGER GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
    title VARCHAR(100) NOT NULL
);
```
Head to insert row and try to insert data into it, you see that the 'default' is disabled:

image

If I try to insert data, I'll see this error:
```json
{
    "statement": "...",
    "prepared": true,
    "error": {
        "exec_status": "FatalError",
        "hint": "Use OVERRIDING SYSTEM VALUE to override.",
        "message": "cannot insert into column \"id\"",
        "status_code": "428C9",
        "description": "Column \"id\" is an identity column defined as GENERATED ALWAYS."
    },
    "arguments": [
        "(Oid 0,Just (\"\",Text))",
        "(Oid 25,Just (\"hello\",Binary))"
    ]
}
```
If you edit the radio button and enable it you can insert into this table. I'm guessing the console disables the `default` radio button if it can't find a default value for columns. With identity columns, there is no default as such. I guess we can always enable the `default` radio button in the console.

rikinsk13:
### Description

- Handle identity columns in insert / edit row
- Handle object content columns in edit row
- Create new tab for edit row
- Disable Null and Default options in edit row based on column properties
- italicize views in untracked section _(sneaking this in as not worthy of a separate PR)_

- refactor Table header

### Affected components 

- Console

### Related Issues

#2840, #2665


console/src/components/Common/utils/pgUtils.js
```diff
@@ -77,6 +77,17 @@ export const getTrackedTables = tables => {
   return tables.filter(t => t.is_table_tracked);
 };
 
+export const isColumnAutoIncrement = column => {
```
wawhal: We sure this is the foolproof way of checking autoIncrement?
Because the sequence could refer to a custom sequence too. We should be wary of that since we support existing databases.

Something like this would be safer? What do you think? Its a false-negative vs false-positive problem.

```js
column.column_default && column.column_default.includes("_seq'::regclass)");
```

console/src/components/Services/Data/TableCommon/TableHeader.js
```diff
@@ -99,7 +104,7 @@ const TableHeader = ({
                 }
                 data-test="table-insert-rows"
               >
-                Insert Row
+                {tabName === 'edit' ? 'Edit' : 'Insert'} Row
```
wawhal: Can we just have an extra tab called `Edit` "only" at the `:tableName/edit` route?
Tabs magically switching functionalities is a bit confusing.

PR #2850, opened 2019-09-09
remote relationship create api

tirumaraiselvan:
### Description

Independent PR for Remote Joins (excludes execution)

### Affected components 

- Server

### Todo

- [x] Create API
- [ ] Delete API
- [ ] Update API
- [ ] Schema generation
- [ ] Migrations
- [ ] Metadata


server/src-lib/Hasura/RQL/Types/RemoteRelationship.hs
```diff
@@ -0,0 +1,337 @@
...
+
+data RemoteField =
+  RemoteField
+    { rmfRemoteRelationship :: !RemoteRelationship
+    , rmfGType              :: !G.GType
+    , rmfParamMap           :: !(HashMap G.Name VT.InpValInfo)
+    }
+  deriving (Show, Eq, Lift)
+
+data RemoteRelationship =
```
tirumaraiselvan: This definition can/should be changed to incorporate for a more holistic remote relationship i.e. (remote-to-remote), (remote-to-db)

jberryman:
This seems fine, but it would be wonderful if we could get documentation on all exported functions and data types. 

Is there anything specifically you think could use more review/attention?

tirumaraiselvan:
> This seems fine, but it would be wonderful if we could get documentation on all exported functions and data types.
> 
> Is there anything specifically you think could use more review/attention?

1. https://github.com/hasura/graphql-engine/pull/2850#discussion_r325179923
2. Can we do some other general cleanup to reduce nesting etc?

server/graphql-engine.cabal
```diff
@@ -61,6 +61,7 @@ library
                      , wai
                      , postgresql-binary
                      , process
+                     , validation
```
tirumaraiselvan: monad-validate might remove some boilerplate

server/src-lib/Hasura/RQL/Types/RemoteRelationship.hs
jberryman: This can be changed as we add features right? Or are you concerned with extending the API while staying backwards compatible?

jberryman:
@nizar-m Would you mind giving a quick summary of how the new test framework works, and what it can do?

Also is there any part in particular that you'd like closer review on?

server/src-lib/Hasura/RQL/Types/RemoteRelationship.hs
tirumaraiselvan: Yep, keeping it backward compatible. Also, we already know what's going to come up.

tirumaraiselvan: Do you think we can make a sum type called `data SourceFields = TableSource QualifiedTable (Set FieldName)` ?

jberryman: Honestly I don't think it's worth trying to do this right now. It seems likely it will need to change again once we actually get deep into implementation, right?

nizar-m:
> @nizar-m Would you mind giving a quick summary of how the new test framework works, and what it can do?
> 
> Also is there any part in particular that you'd like closer review on?

What we were initially planning to do was running another Hasura GraphQL engine as a remote GraphQL server and attach that as remote. This way we can test with a very rich remote GraphQL schema with much lesser amount of configuration. So now, before any test, we will have to setup the remote Hasura GraphQL engine, which would be an extra configuration setup.

Then it evolved to why not create a proxy to GraphQL server which will add prefixes to the top-level fields of operations (query, mutation, subscription), and to the names of all object, input object and enum types. Then add this proxy as remote itself.  That way we do not have to create another remote Hasura GraphQL engine, and so the current test configuration would work.  

The implementation of the prefixer proxy is in [tests-py/gql_prefixer_proxy.py](https://github.com/hasura/graphql-engine/blob/64431252c9158dd6c981ebce703703065ab8a1cf/server/tests-py/gql_prefixer_proxy.py).  In [tests-py/graphql_server.py](https://github.com/hasura/graphql-engine/blob/64431252c9158dd6c981ebce703703065ab8a1cf/server/tests-py/graphql_server.py#L723) this proxy is linked to the path /graphql-prefixer-proxy. So now as remote GraphQL servers, we have those that are implemented in Graphene, plus this proxy.

nizar-m:
The last yaml templating part is something that should have a second look. The solution for now works, but it may be confusing for people who are writing tests for the first time. We may move the whole tests into the code, or we may use templating languages like jsonnet or dhall or ytt. 

If we are using something like jsonnet, ytt or dhall, we need to look at whether they support insert ordering for objects.

Looks [jsonnet](https://github.com/google/jsonnet/issues/407#issuecomment-349418427) and [dhall](https://github.com/dhall-lang/dhall-haskell/issues/1187) would not support insert order for jsons for now. [ytt](https://get-ytt.io/#example:example-variable) on the other hand seems to support it, at least on their playground, but it lacks python bindings.

server/graphql-engine.cabal
tirumaraiselvan: I tried doing this and TBH it didn't make much of a difference. So just leaving this as is atm.

server/src-lib/Hasura/RQL/Types/RemoteRelationship.hs
tirumaraiselvan: I agree but we already have an implementation to refer: https://github.com/hasura/graphql-engine/blob/985e1cd6c58757ffcd3138572d573d83ec9cded1/server/src-lib/Hasura/RQL/Types/RemoteRelationship.hs#L341

I had to create a `RemoteToRemoteRelationship` type only because the types were so different. 

Do you think we should do something like:

```
data RemoteRelationshipNew =
  RemoteRelationshipNew
  { type: TableToRemote
  , configuration: RemoteRelationship
  }

This seems to affect much less. I just want to avoid some backward incompatibility nightmare later.

File: server/tests-py/gql_prefixer_proxy.py [hunk 3]
```diff
@@ -0,0 +1,291 @@
...
+      return ty['kind'] == 'ENUM'
+
+   def _add_name_prefix(self, obj):
+      assert not obj['name'].startswith(self.prefix), obj
+      obj['name'] = self.prefix + obj['name']
+
+   def get(self, request):
+      return Response(HTTPStatus.METHOD_NOT_ALLOWED)
+
+   def _assert_prefixes(self, introspect):
```
nizar-m: Here we are ensuring that gql proxy is adding prefixes as intended.

server/tests-py/graphql_server.py
```diff
@@ -308,7 +313,7 @@ def post(self, req):
         if not req.json:
             return Response(HTTPStatus.BAD_REQUEST)
         res = character_interface_schema.execute(req.json['query'])
-        respDict = res.to_dict()
+        respDict = to_dict(res)
```
nizar-m: These changes exist because we upgraded to Graphene-3.
Some of the tests which we could not run because graphql-core was throwing error can be done now.
Also the  errors are much more descriptive with Graphene 3, which includes stuff like the field name and the location of the field in the query.

server/tests-py/queries/remote_schemas/add_remote_schema_err_missing_arg.yaml
```diff
@@ -11,6 +11,6 @@ query:
   args:
     name: err-missing-arg
     definition:
-      url: http://localhost:5000/iface-graphql-err-missing-arg
+      url: ${REMOTE_GRAPHQL_ROOT_URL}/iface-graphql-err-missing-arg
```
nizar-m: Take the root url of remote GraphQL server from the environmental variable ${REMOTE_GRAPHQL_ROOT_URL}  instead. This will be set by the remote_gql_server fixture when the remote server is running.

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -0,0 +1,365 @@
...
+    fmap
+      HM.fromList
+      (flip traverse (toList (rrHasuraFields remoteRel)) $ \fieldName ->
+         case HM.lookup fieldName (_tiFieldInfoMap tableInfo) of
+           Nothing -> Left . pure $ TableFieldNotFound tableName fieldName
+           Just fieldInfo -> pure (fieldName, fieldInfo))
+  let initFieldCalls = NE.init $ rrRemoteFields remoteRel
+      leafFieldCall = NE.last $ rrRemoteFields remoteRel
+  (leafParentTypeInfo, leafParentTypeMap) <-
+    foldl
```
tirumaraiselvan: use foldl' (see: https://github.com/hasura/graphql-engine/pull/2933#discussion_r328821960)

PR #2852, opened 2019-09-09
allow nested inserts for one-to-one relationships (fix #2576)

### LINKED ISSUE #2576: One to One relationship causing Not-Null Constraint Violation on Nested Data Insert
```js
{
  "errors": [
    {
      "extensions": {
        "path": "$.selectionSet.insert_a.args.objects[0]",
        "code": "constraint-violation"
      },
      "message": "Not-NULL violation. null value in column \"id\" violates not-null constraint"
    }
  ]
}
```

The following table definitions reproduce the above error:

# Tables

## Table a:
Screenshot 2019-07-20 at 14 58 23

## Table B:
Screenshot 2019-07-20 at 14 58 42

# Relationships

## Table A:
Screenshot 2019-07-20 at 14 59 54

## Table B:
Screenshot 2019-07-20 at 15 00 10

# Invoking The Error
That error happens when inserting data from the "has one" side (table a), but not when inserting from the "belongs to" side (I wish Hasura could label relationships like this as well in addition to what's there to make things clear quickly):

the above error happens when you run this query:
```graphql
mutation {
  insert_a(objects: {
    junk_a: "This is a's junk"
    b:{
      data: 
        {
          junk_b: "B's junk inserted when inserting a's junk"
        }
    }
  }){
    affected_rows
  }
}
```

but not when you run this query from the other side:

```graphql
mutation {
  insert_b(objects: {
    junk_b: "This is B's junk"
    a:{
      data: 
        {
          junk_a: "A's junk inserted when inserting B's junk"
        }
    }
  }){
    affected_rows
  }
}
```

> This could be related to #2508, but I am not sure cause what he is reporting on is different.

rakeshkky:
### Description

This PR extends existing `create_object_relationship` query API to support nested inserts for one-to-one object relationships.

**Console changes are must as relationship tab fails to load if any relationships are defined using new API**
Refer [here](https://deploy-preview-2852--hasura-docs.netlify.com/graphql/manual/api-reference/schema-metadata-api/relationship.html#create-object-relationship-syntax) to learn more about the new API.

### Affected components 

- Server
- Docs
- Tests

TODO:
- [ ] console

### Related Issues

fix #2576 

### Solution and Design

- Distinct manual configuration types for object and array relationships
- Define `ObjRelUsingFKeyOn` for one-to-one object relationships
- Now `RelInfo` carries the information of `insertion order`
- The `insertion order` of array relationships are always `after_parent`
- In `Hasura.GraphQL.Resolve.Insert`, while inserting a nested object, the object relationships are inserted according to their insertion order.
- Just `ConstraintName` is not enough to validate the remote table's column uniqueness. Hence, `TableInfo` carries a primary key and unique constraints along with their associated column names.

### Steps to test and verify

Follow and test the reproduction guide given in issue #2576

### DIFF
--- a/docs/graphql/manual/api-reference/schema-metadata-api/relationship.rst
+++ b/docs/graphql/manual/api-reference/schema-metadata-api/relationship.rst
@@ -27,9 +27,9 @@ create_object_relationship
 --------------------------
 
 ``create_object_relationship`` is used to create an object relationship on a
-table. There cannot be an existing column or relationship with the same name. 
+table. There cannot be an existing column or relationship with the same name.
 
-There are 2 ways in which you can create an object relationship.
+There are 3 ways in which you can create an object relationship.
 
 1. Using foreign key constraint on a column
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -54,23 +54,49 @@ Create an ``object relationship`` ``author`` on ``article`` *table*,  *using* th
        }
    }
 
+2. One-to-One object relationship using foreign key constraint
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A one-to-one object relationship can be created using foreign key constraint
+on remote table's column. The condition is that the remote table's column alone
+should have either ``primary key`` or ``unique`` constraint.
+
+.. code-block:: http
+
+   POST /v1/query HTTP/1.1
+   Content-Type: application/json
+   X-Hasura-Role: admin
+
+   {
+       "type": "create_object_relationship",
+       "args": {
+           "table": "article",
+           "name": "config",
+           "using": {
+               "foreign_key_constraint_on" : {
+                   "table": "article_config",
+                   "column": "article_id"
+               }
+           }
+       }
+   }
 
 .. _manual_obj_relationship:
 
-2. Manual configuration
+3. Manual configuration
 ^^^^^^^^^^^^^^^^^^^^^^^
 
 This is an advanced feature which is mostly used to define relationships on or
 to views. We cannot rely on foreign key constraints as they are not valid to or
 from views. So, when using manual configuration, we have to specify the remote
 table and how columns in this table are mapped to the columns of the
-remote table. 
+remote table.
 
 Let's say we have a view called ``article_detail`` which has three columns
 ``article_id`` and ``view_count`` and ``average_rating``. We can now define an
 object relationship called ``article_detail`` on the ``article`` table as
-follows: 
- 
+follows:
+
 .. code-block:: http
 
    POST /v1/query HTTP/1.1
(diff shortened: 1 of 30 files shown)


server/src-lib/Hasura/RQL/DDL/Relationship/Types.hs
```diff
@@ -93,13 +113,27 @@ type ArrRelUsing = RelUsing ArrRelUsingFKeyOn ArrRelManualConfig
 type ArrRelDef = RelDef ArrRelUsing
 type CreateArrRel = WithTable ArrRelDef
 
-newtype ObjRelManualConfig =
-  ObjRelManualConfig { getObjRelMapping :: RelManualConfig }
-  deriving (Show, Eq, FromJSON, ToJSON, Lift)
+data ObjRelUsingFKeyOn
```
lexi-lambda: This type name does not follow the proposed [Haskell style guide](https://github.com/hasura/graphql-engine/blob/950d3842e4662c2ff5d8d5b20ac64bf9f9387b80/server/CONTRIBUTING.md#naming). The name `ObjectRelationshipUsingForeignKeyOn` would be a little long, but maybe there’s a shorter name that could capture what it does without needing to abbreviate?

Taking a step back for a moment, the Ruby on Rails nomenclature for this is that there are three kinds of relationships: `has_one`, `has_many`, and `belongs_to`. The first two describe one-to-one and one-to-many relationships where the foreign table contains the reference, while `belongs_to` describes a one-to-one relationship where the “self” table contains the reference.

Could we have the same kind of explicit modeling? Instead of “array relationship” and “object relationship,” could we have

```haskell
data ForeignColumn = ForeignColumn !QualifiedTable !PGCol

data Relationship
  = HasOne !ForeignColumn
  | HasMany !ForeignColumn
  | BelongsTo !QualifiedTable !PGCol
```

To me, that modeling is a lot more natural than the “object” vs “array” split we have currently, which focuses on the GraphQL interface rather than the actual relational modeling.

docs/graphql/manual/api-reference/schema-metadata-api/relationship.rst
```diff
@@ -93,6 +119,36 @@ follows:
        }
    }
 
+If we want to define a one-to-one object relationship using manual configuration
+then we've to specify ``insertion_order`` as ``after_parent`` in ``manual_configuration``.
+``insertion_order`` is the order in which the object relationship inserted
+in a nested insert in respect of the parent. It expects either ``before_parent`` or
+``after_parent`` string value. It is optional and by default it is ``before_parent`` which
+indicates many-to-one relationship.
```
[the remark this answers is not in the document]
rakeshkky: Thank you @lexi-lambda for a detailed description. This helps me a lot.

server/src-lib/Hasura/RQL/DDL/Relationship/Types.hs
0x777: I agree, I'm not a fan of our current relationship terminology (object/array). I like the explicit nature of rails [associations](https://guides.rubyonrails.org/association_basics.html), there are six kinds as opposed to the two that we have, `has_many` and `belongs_to`  (this PR introduces the third one, `has_one`). 

I'm not sure what's the best way forward though. I would like us to get to rails like simplicity for defining relationships but without disrupting how we treat relationships across our entire codebase. We could for example easily get rid of `insertion_order` and add something like `kind` to specify either `belongs_to` or `has_one`? The end goal could be to have an api like `create_relationship` which takes an extra `kind` argument? The exported metadata could then have `relationships` instead of `object_relationships` and `array_relationships` like we currently have.

beepsoft:
@rakeshkky when is this fix planned to be merged? I'm really looking forward to it. 🙏

cooldesigns:
Waiting for it.

richbayliss:
@rakeshkky do you have an ETA on this? It would be great to get it out there, I could really use it. Is there anything I could do to assist in moving it forward at all?

vladciobanu:
A new PR which fixes this same issue should land in master soon. Closing this for now.

dohomi:
@vladciobanu will there be documentation update how this is done in the future?

Issue #2801, opened 2019-08-29
[RFC] Refactor GraphQL schema generation and query validation to move parsing out of resolvers


lexi-lambda:
## Problem

As mentioned in https://github.com/hasura/graphql-engine/pull/2509#discussion_r319111272, the way we currently parse and validate GraphQL queries leaves too much parsing logic for resolvers. This leads to a proliferation of `throw500` calls in resolvers, which could be ruled impossible by the type system with a more structured approach, but currently cause runtime errors.

### Example: Boolean expressions

To give a concrete example, let’s consider the way we currently resolve GraphQL boolean expressions. The relevant code is in `src-lib/Hasura/GraphQL/Resolve/BoolExp.hs`, which provides the following function:

```haskell
parseBoolExp
  :: (MonadError QErr m, MonadReader r m, Has FieldMap r)
  => AnnInpVal -> m (AnnBoolExp UnresolvedVal)
```

Just looking at the types of this function, it’s clear that it must be doing some kind of parsing. The result type, `AnnBoolExp`, is a type alias for a sum type with a very particular shape:

```haskell
data GBoolExp a
  = BoolAnd ![GBoolExp a]
  | BoolOr  ![GBoolExp a]
  | BoolNot !(GBoolExp a)
  | BoolFld !a
```

However, the only input to `parseBoolExp` has type `AnnInpVal`, which is little more than a small wrapper around an `AnnGValue`, which is a very general type that essentially describes an arbitrary GraphQL value:

```haskell
data AnnGValue
  = AGScalar !PGScalarType !(Maybe PGScalarValue)
  | AGEnum !G.NamedType !AnnGEnumValue
  | AGObject !G.NamedType !(Maybe AnnGObject)
  | AGArray !G.ListType !(Maybe [AnnInpVal])
  deriving (Show, Eq)
```

This is not necessarily a bad thing—`parseBoolExp` is doing *parsing*, after all—but the problem is that `parseBoolExp` should not actually ever fail. It’s a bug if it does! Why is that? Well, we’ve actually *already checked* that the `AnnInpVal` has the proper shape, since we checked it against GraphQL schema we generated.

[...]

PR #2856, opened 2019-09-11
Refactor query plan caching and improve subscription multiplexing

lexi-lambda:
### Description

Broadly speaking, this PR makes two related changes to GraphQL query execution:

  1. The way we determine which queries are eligible for plan caching is now completely different. Rather than performing a post-hoc check, the bindings in `Hasura.GraphQL.Resolve.InputValue` now track the query’s reusability in monadic state. The main purpose of this is simplifying some logic, but it also fixes an unlikely bug where queries could be incorrectly cached.

  2. *All* live queries now use the “multiplexed” implementation, which may improve the performance of subscriptions by reducing the number of concurrent Postgres queries by batching similar queries together. I’ve also refactored the live query implementation to split it across several modules.

Documentation comments in the code explain some more of the details.

### Affected components 

- Server

### Related Issues

hasura/graphql-engine-internal#260

I was originally hoping to come up with a slightly more robust solution to the query caching by making it genuinely impossible to look at a variable’s value without performing the appropriate `markNotReusable` call. Unfortunately, it was just too hard to slot into things due to the way parsing is split across the `GraphQL.Validate` and `GraphQL.Resolve` layers. The current solution is probably fine, but it’s technically possible to have a bug if someone calls `_aivValue` directly instead of using one of the helper functions like `asPGColumnValue`.


server/src-lib/Hasura/GraphQL/Execute/LiveQuery.hs
```diff
@@ -1,305 +1,118 @@
...
+  * A 'Cohort' is a set of 'Subscriber's that are all subscribed to the same query /with the exact
+    same variables/. (By batching these together, we can do better than multiplexing, since we can
+    just query the data once.)
+
+  * A 'Poller' is a worker thread for a single, multiplexed query. It fetches data for a set of
+    'Cohort's that all use the same parameterized query, but have different sets of variables.
+
+  * Finally, the 'LiveQueriesState' is the top-level container that holds all the active 'Poller's.
+
+Additional details are provided by the documentation for individual bindings.
```
0x777: This is lovely. Thanks for taking the time to document this.

server/src-lib/Hasura/GraphQL/Resolve/Types.hs
```diff
@@ -161,14 +160,64 @@ partialSQLExpToUnresolvedVal = \case
...
+  | NotReusable
+  deriving (Show, Eq)
+
+instance Semigroup QueryReusability where
+  Reusable a <> Reusable b = Reusable (a <> b)
+  _          <> _          = NotReusable
+instance Monoid QueryReusability where
+  mempty = Reusable mempty
+
+class (MonadError QErr m) => MonadResolve m where
```
0x777: I'm wondering about the use of `MonadError` constraint as both `recordVariableUse` and `markNotReusable` functions don't need this. From what I understand, the `MonadError` constraint is needed for all the `as..ValM` family of functions. Does it make sense to bring the requirement of those functions into this type class? If I'm not wrong, all of these do a `throw500`, so maybe this typeclass can have something like `throwResolutionError`?

lexi-lambda: That’s fair; I debated including this superclass. Ultimately the main reason I decided to include it is because, currently, essentially *everything* in `Resolve/` needs to be able to throw (again, due to the issue described in #2801), and having every function’s constraint say `(MonadError QErr m, MonadResolve m)` just felt redundant. I’m happy to split it back out, though, if you think that’s better; I don’t really have an opinion either way.

0x777: I think it is okay, doesn't significantly improve the status quo.

PR #2892, opened 2019-09-17
server: Don’t allow warnings when building in CI

lexi-lambda:
### Description

Currently, CI does not build the server with the `-Werror` GHC option set, so warnings are not treated as errors. This allowed https://github.com/hasura/graphql-engine/commit/99174cca9b99ebb4b7d55d02fa1dcda150a3d57c to get merged with an unfixed warning. This PR fixes that.

### Affected components 

- Server
- Build System

### DIFF
--- a/server/src-lib/Hasura/RQL/Types/SchemaCache.hs
+++ b/server/src-lib/Hasura/RQL/Types/SchemaCache.hs
@@ -16,6 +16,7 @@ module Hasura.RQL.Types.SchemaCache
 
        , TableInfo(..)
        , tiName
+       , tiDescription
        , tiSystemDefined
        , tiFieldInfoMap
        , tiRolePermInfoMap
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/RQL/Types/SchemaCache.hs
```diff
@@ -16,6 +16,7 @@ module Hasura.RQL.Types.SchemaCache
 
        , TableInfo(..)
        , tiName
+       , tiDescription
```
ecthiender: does this PR require this symbol to be exported?

lexi-lambda: Yes, the warning in https://github.com/hasura/graphql-engine/commit/99174cca9b99ebb4b7d55d02fa1dcda150a3d57c is due to it not being exported, which causes an unused identifier warning. (The definition can’t just be deleted because it’s generated by `makeLenses`, and some of the other lenses *are* used.)

PR #2900, opened 2019-09-18
Update auth0-jwt.rst (close #2655)

### LINKED ISSUE #2655: Review and verify guide on Auth0 JWT Integration
Following the instructions [here](https://docs.hasura.io/1.0/graphql/manual/guides/integrations/auth0-jwt.html), I encountered problems when using the suggested URI under the heading **Test auth0 login and generate sample JWTs for testing**.  Two things: first, if you want to use the **/login** endpoint you have to substitute 

> response_type=token

 with 

> response_type=token id_token

Second, for the **/login** endpoint to work, as given in these instructions, you have to disable **OIDC Compliance** under the Auth0 application you are targeting, found under **Applications** > **\<your-hasura-application\>** > **Settings** > **Advanced Settings** > **OAuth**.  If you don't do this Auth0 will give an error after you submit the login form: 

> access_denied: Password login via OIDC-conformant clients with externally-hosted login pages is unsupported

I am a beginner at all of this.  It may be that these instructions worked before Auth0 disabled "as many legacy features as possible".  I quote from [auth0 docs](https://auth0.com/docs/api-auth/tutorials/adoption/oidc-conformant).  I hope this is useful info.

afreix:
### Description
I needed to update this to accessToken in order to get request to work, and from what I've read online, the accessToken is what should be used for authorizing access to resources within an API

### Affected components 
- Docs

### DIFF
--- a/docs/graphql/manual/guides/integrations/auth0-jwt.rst
+++ b/docs/graphql/manual/guides/integrations/auth0-jwt.rst
@@ -32,7 +32,7 @@ In the dashboard, navigate to ``Rules``. Add the following rules to add our cust
 
     function (user, context, callback) {
       const namespace = "https://hasura.io/jwt/claims";
-      context.idToken[namespace] = 
+      context.accessToken[namespace] = 
         { 
           'x-hasura-default-role': 'user',
           // do some custom logic to decide allowed roles


phylor:
I can confirm that `accessToken` needs to be used. Additionally, with the new [auth0-spa-js library](https://github.com/auth0/auth0-spa-js) you need to create an API at auth0 and specify its identifier as the `audience` in your JS code (when configuring auth0). Only then you'll get the JWT required for Hasura.

@marionschleifer If the CLA is blocking this PR, would it be possible to apply this change in a different manner? I guess everyone will currently have problems when trying to use Hasura with auth0.

praveenweb:
For this PR to be merged, we need to document more steps. Just replacing `idToken` with `accessToken` doesn't work out of the box. As @phylor mentioned, there is a step to create an API at auth0 and configure audience parameter.

Closing this PR till we have comprehensive docs related to this migration from `idToken` to `accessToken`.

PR #2917, opened 2019-09-21
support computed fields returning scalars or set of tables (close #1387)

### LINKED ISSUE #1387: support computed columns in queries
Like Postgraphile does: https://www.graphile.org/postgraphile/computed-columns/

rakeshkky:
### Description

**Computed fields** are extra fields added to a table whose value is computed from other columns. It facilitates users to define an additional field to an existing table whose return value is computed via an SQL function.

**How it is different from Postgres generated columns**

Postgres, from version ``12``, is introducing [Generated Columns](https://www.postgresql.org/docs/12/ddl-generated-columns.html).
Postgres' generated columns come with its own limitations. Hasura's computed fields are defined via an SQL function, which allows users to define any complex business logic in a function. Generated columns will go together with computed fields where Hasura treats generated columns as normal Postgres columns.

Refer to docs [here](https://github.com/rakeshkky/graphql-engine/blob/comp-col-new/docs/graphql/manual/schema/computed-fields.rst)

Refer to API reference [here](https://github.com/rakeshkky/graphql-engine/blob/comp-col-new/docs/graphql/manual/api-reference/schema-metadata-api/computed-field.rst)

### Affected components 

- Server
- Docs
- Tests

### Related Issues

close #1387 

### Solution and Design

This PR is a significant change which touches code base for metadata, GraphQL schema generation and resolving and SQL generation. 

**Metadata**

- Added two new query types to manage computed fields. See module `Hasura.RQL.DDL.ComputedField`
  1. `add_computed_field`
  2. `drop_computed_field`
- Detect changes to functions associated with computed fields in `run_sql` query
- Add computed fields in `create_select_permission`

**GraphQL**

- `Either` is being used before to represent whether a field is a column or a relationship. Now, those types are replaced with proper sum types to represent computed fields too.
- Resolve the computed field

**SQL generation**

Modules:-
- `Hasura.RQL.DML.Select.Internal`
- `Hasura.RQL.DML.Select.Types`

Changes:-
- The `BaseNode` now contains table computed fields as sub-nodes. 
- `TableFrom` changed to `SelectFromG` to include a selection from rows returned by a function execution
- While resolving a `BaseNode` to SQL `Select` keep track of parent node prefix to execute the computed field function with the parent table row identifier.

In module `Hasura.SQL.DML` add `FunctionExp` to `SQLExp` AST.

The following function is added to the `author` table as a computed field with name
`computed_articles`

```sql
CREATE FUNCTION fetch_computed_articles(author_arg author, search text)
RETURNS SETOF article AS $$
    SELECT *
    FROM article
    WHERE
      (title ilike ('%' || search || '%')
      OR content ilike ('%' || search || '%'))
      AND article.author_id = author_arg.id
$$ LANGUAGE sql STABLE;
```

Querying data from author table along with computed field `computed_articles`

```graphql
query {
  author{
    id
    name
    computed_articles(args:{search:"search_text"}){
      id
      title
    }
  }
}
```

Generates the SQL below.

```sql
SELECT
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_5_e"
          FROM
            (
              SELECT
                "_0_root.base"."id" AS "id",
                "_0_root.base"."name" AS "name",
                "_4_root.cf.root.computed_articles"."computed_articles" AS "computed_articles"
            ) AS "_5_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public"."author"
        WHERE
          ('true')
      ) AS "_0_root.base"
      LEFT OUTER JOIN LATERAL (
        SELECT
          coalesce(json_agg("computed_articles"), '[]') AS "computed_articles"
        FROM
          (
            SELECT
              row_to_json(
                (
                  SELECT
                    "_2_e"
                  FROM
                    (
                      SELECT
                        "_1_root.cf.root.computed_articles.base"."id" AS "id",
                        "_1_root.cf.root.computed_articles.base"."title" AS "title"
                    ) AS "_2_e"
                )
              ) AS "computed_articles"
            FROM
              (
                SELECT
                  *
                FROM
                  "public"."fetch_computed_articles"(author_arg => "_0_root.base", search => 'search_text')
                WHERE
                  (("_0_root.base"."id") = ("author_id"))
              ) AS "_1_root.cf.root.computed_articles.base"
          ) AS "_3_root.cf.root.computed_articles"
      ) AS "_4_root.cf.root.computed_articles" ON ('true')
  ) AS "_6_root"
```

### Steps to test and verify

Follow any test case or example

- Aggregations are not available on computed fields which returns setof table
- Scalar computed fields are not allowed in `where` clause (bool exps)
- The SQL selection for computed fields need to be optimised (merge sub-queries for same computed fields)


server/src-lib/Hasura/Server/Telemetry.hs
```diff
@@ -159,7 +159,7 @@ computeMetrics sc =
     calcPerms fn perms = length $ catMaybes $ map fn perms
 
     relsOfTbl :: TableInfo PGColumnInfo -> [RelInfo]
-    relsOfTbl = rights . Map.elems . Map.map fieldInfoToEither . _tiFieldInfoMap
+    relsOfTbl = getRels . _tiFieldInfoMap
```
lexi-lambda: I think this function is short enough now that it’d be better to just inline it at its use-site, since it’s only used in one spot.

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
```diff
@@ -153,11 +154,17 @@ mkColDep :: DependencyReason -> QualifiedTable -> PGCol -> SchemaDependency
...
+  :: DependencyReason -> QualifiedTable -> ComputedColumnName -> SchemaDependency
+mkComputedColumnDep reason tn computedColumn =
+  flip SchemaDependency reason . SOTableObj tn $ TOComputedColumn computedColumn
+
 type WithDeps a = (a, [SchemaDependency])
 
 data FieldInfo columnInfo
   = FIColumn !columnInfo
   | FIRelationship !RelInfo
+  | FIComputedColumn !ComputedColumnInfo
```
lexi-lambda: I’m curious: do you think it would make any sense to slot computed columns into `PGColumnType` directly, as another case in addition to `PGColumnScalar` and `PGColumnEnumReference`? I can think of arguments both for and against: on one hand, computed columns are genuinely virtual *columns*, while relationships are an entirely different sort of thing, so it might make sense to model them that way, but on the other hand, they’re a lot less like a plain old column than enum references are. I’d be curious to hear your thoughts.

server/src-lib/Hasura/GraphQL/Resolve/BoolExp.hs
```diff
@@ -158,13 +158,15 @@ parseColExp
...
+    RFPGColumn pgColInfo -> do
       opExps <- parseOpExps (pgiType pgColInfo) val
       return $ AVCol pgColInfo opExps
-    Right (RelationshipField relInfo _ _ permExp _)-> do
+    RFRelationship (RelationshipField relInfo _ _ permExp _)-> do
       relBoolExp <- parseBoolExp val
       return $ AVRel relInfo $ andAnnBoolExps relBoolExp $
         fmapAnnBoolExp partialSQLExpToUnresolvedVal permExp
+    RFComputedColumn _ -> throw500
+          "computed columns are not allowed in bool_exp"
```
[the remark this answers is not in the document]
rakeshkky: This is not impossible to implement. But it is not "straight forward" to implement. 
Will add this separately.

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
rakeshkky: I'm going to rename `computed column` to `computed field` to make it distinct from Postgres column. In that way, maybe, I think `FIComputedField` (new name) is good enough to be isolated.

server/src-lib/Hasura/GraphQL/Resolve/BoolExp.hs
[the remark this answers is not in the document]
lexi-lambda: I’m curious—what is the primary obstacle here? If a computed column is “just” a laterally-joined function, can’t we filter on it the same way we filter on other columns?

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
lexi-lambda: I think that renaming is a good idea, though I’m still curious about whether you think it makes sense for them to be grouped in with columns (as, really, I think I’m a little inclined to think that maybe `PGColumnType` ought to have been called `FieldType`, instead, and we should reserve “column” for the Postgres concept 🙂).

server/src-lib/Hasura/GraphQL/Resolve/BoolExp.hs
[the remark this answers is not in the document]
rakeshkky:
```sql
-- Base root select
(
SELECT  * FROM "public"."author"
WHERE ... -- where clause goes here
)
```
Functions returning scalar values, we've to include the function call in `where` clause AND selection extractors
```sql
SELECT my_function("root") AS computed_field
FROM
(
  SELECT * FROM "public"."author" AS "table_row"
  WHERE my_function("public"."author") = 'some value'
) AS "root"
```
I'm planning to avoid two function calls do something like

```sql
SELECT my_function("root") AS computed_field
FROM
(
  SELECT * FROM "public"."author" 
) AS "root"
WHERE computed_field = 'some value'
```
This would potentially change some types in `Select/Internal.hs`. I'm planning to add this separately. Also, this PR is very huge. I thought it'd be prettier to add as a separate feature. :)

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
rakeshkky: I thought in this way. The `FieldInfo` should show the kinds of fields that a table can have. Columns (from Postgres), relationships, computed fields (this PR), remote join fields (coming soon) etc. Except for columns all other fields are virtual and only defined through Hasura metadata (though enums are defined through metadata but the columns are actually present in table schema). So, I'm thinking it's better to keep `FIComputedField` isolated and do not group with `FIColumn` :)

lexi-lambda: Okay, I think that’s fair, you’ve convinced me. :) Reading your comment made me realize the distinction here: `PGColumnType` describes the type of a value, while `FieldInfo` describes where we get it from. And the value of a computed column is not a different type of value from the value of an ordinary column! The way we retrieve it is just different.

server/src-lib/Hasura/GraphQL/Resolve/BoolExp.hs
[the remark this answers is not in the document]
lexi-lambda: > Functions returning scalar values, we've to include the function call in `where` clause AND selection extractors

Let’s say we have a query like this:

```graphql
query {
  posts(where: {is_public: {_eq: true}}) {
    id
    title
    computed_score
  }
}
```

I’d expect the SQL we generate to look sort of like this (aside from all the name-mangling we do, of course):

```sql
SELECT json_build_object(
    'id', posts.id,
    'title', posts.title,
    'computed_score', computed_score.value)
FROM posts
LEFT JOIN LATERAL
  (SELECT compute_post_score(posts.*))
  AS computed_score(value) ON true
WHERE posts.is_public = true
```

It seems to me that this really does just bring `computed_score` into exactly the same scope that `posts` is, so it should be very straightforward to handle this slightly modified query:

```diff
  query {
-   posts(where: {is_public: {_eq: true}}) {
+   posts(where: {is_public: {_eq: true}, computed_score: {_gt: 0}}) {
      id
      title
      computed_score
    }
  }
```

…since all we’d have to do is make this tiny change to the resulting SQL:

```diff
  SELECT json_build_object(
      'id', posts.id,
      'title', posts.title,
      'computed_score', computed_score.value)
  FROM posts
  LEFT JOIN LATERAL
    (SELECT compute_post_score(posts.*))
    AS computed_score(value) ON true
  WHERE posts.is_public = true
+   AND computed_score.value > 0
```

But your comment makes me think the SQL we generate isn’t like that after all, and I’m curious why not!

> Also, this PR is very huge. I thought it'd be prettier to add as a separate feature. :)

I appreciate the sentiment, but I think `throw500`s are always ugly. :)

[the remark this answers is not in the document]
rakeshkky: In my first comment

> This is not impossible to implement. But it is not "straight forward" to implement.

I didn't consider a case where things get complicated to implement computed fields in `where` expression. Let's say if the SQL function associated with a computed field has extra arguments other than table row type, then we've to pass them as input arguments. 

```sql
CREATE public.author_name_prefix(author_arg public.author, prefix text)
 RETURNS TEXT
 LANGUAGE sql
 STABLE
AS $function$
    SELECT prefix || ' ' || author_arg.name
$function$;
```
`name_with_prefix` is the computed field for `author_name_prefix` function
```graphql
query {
  author{
    name_with_prefix(args: {prefix: "Mr. "})
  }
}
```
The input for `name_with_prefix` in `where` will look like
```graphql
author(where: {name_with_prefix: {args: {prefix: "Mr. "}, _eq: ""}})
```
To achieve the above, we need to generate new set of boolean types
```graphql
input <table-name>_bool_exp {
  int_col: Int_comparison_exp
  string_col: String_comparison_exp
  computed_field: <table-name>_<computed-field-name>_comparison_exp
}
input <table-name>_<computed-field-name>_comparison_exp {
  args: <function_args>!
  _eq: <computed-field-type>
  # other comparison operators based on field type
}
```

Well, then it is possible to use different `args` values in `where` and in the actual selection
```graphql
query {
  author(where: {name_with_prefix: {args: {prefix: "Dr. "}, _eq: ""}}){
    name_with_prefix(args: {prefix: "Mr. "})
  }
}
```
Definitely we should make two function calls, one in `WHERE` clause and the other in extractors

```sql
SELECT author_name_prefix(author.*, 'Mr. ') as name_with_prefix FROM author WHERE author_name_prefix(author.*, 'Dr. ') = ' ';
```
Not over yet, we should also think about computed fields in permissions (`RQL` query). 

Boolean expression syntax in JSON may look like:-
```json
{
	"_where": {
		"computed_field": {
			"args": {"prefix": "Mr. "},
			"_eq": ""
		}
	}
}
```
`args` value can be static or values from session variables. 

I feel supporting computed fields in boolean expressions need fresh thinking and considerable code overhaul. This can go out as a separate feature/enhancement.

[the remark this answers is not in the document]
lexi-lambda: > To achieve the above, we need to generate new set of boolean types

Ah, I see! Yes, I wasn’t thinking about the complication of needing to encode the function arguments in the boolean expressions differently, since input object fields don’t have “arguments” in the same way query fields do. Annoying! The issue of having different combinations of arguments in different places makes sense, too.

Sorry for missing the point for a while, but I appreciate the explanation.

server/src-lib/Hasura/SQL/Types.hs
```diff
@@ -520,3 +526,49 @@ instance (ToSQL a) => ToSQL (PGType a) where
...
     -- typename array is an sql standard way of declaring types
     PGTypeArray ty -> toSQL ty <> " array"
+
+data PGTypeKind
+  = PGKindBASE
+  | PGKindCOMPOSITE
+  | PGKindDOMAIN
+  | PGKindENUM
+  | PGKindRANGE
+  | PGKindPSEUDO
```
lexi-lambda: I think these should still use the usual naming convention, e.g. `PGKindBase`, `PGKindComposite`, etc., with a custom `ToJSON` instance if necessary. I also think the `ToJSON` and `FromJSON` instances should roundtrip, but currently they do not.

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -182,6 +192,61 @@ instance Hashable ForeignKey
...
+$(deriveToJSON (aesonDrop 4 snakeCase) ''ComputedFieldFunction)
+
+data ComputedFieldInfo
+  = ComputedFieldInfo
+  { _cfiName       :: !ComputedFieldName
+  , _cfiFunction   :: !ComputedFieldFunction
+  , _cfiReturnType :: !ComputedFieldReturn
+  , _cfiComment    :: !(Maybe T.Text)
+  } deriving (Show, Eq)
+$(deriveToJSON (aesonDrop 4 snakeCase) ''ComputedFieldInfo)
```
lexi-lambda: Could these types perhaps go in their own module, maybe `Hasura.RQL.Types.ComputedField`? It’d be nice to avoid making `Common.hs` too big if we can help it, and it doesn’t look to me like anything else in this module depends on any of these types (but I could be missing something).

rakeshkky:
@lexi-lambda The [commit](https://github.com/hasura/graphql-engine/commit/b7d8991f68f071b24e174e3d52b81d204bb11afb) improves the logic of embedding `table argument` in resolved input function arguments of a computed field.

server/src-lib/Hasura/RQL/Types/ComputedField.hs
```diff
@@ -29,12 +29,12 @@ fromComputedField = FieldName . computedFieldNameToText
 
 data FunctionTableArgument
   = FTAFirstArgument
-  | FTAName !FunctionArgName
+  | FTAName !FunctionArgName !Int -- ^ argument name and index
```
lexi-lambda: Does this subsume the `FTAFirstArgument` case? It seems like `FTAFirstArgument` is now just a particular case of `FTAName` where the index is `0`.

server/src-lib/Hasura/GraphQL/Resolve/Select.hs
```diff
@@ -80,8 +81,16 @@ resolveComputedField computedField fld = fieldAsPath fld $ do
...
         FTAFirstArgument -> RS.FunctionArgsExp
                             (RS.AETableRow:positional) named
-        FTAName argName  -> RS.FunctionArgsExp positional $
-                            Map.insert (getFuncArgNameTxt argName) RS.AETableRow named
+        FTAName argName index ->
+          let insertAt i a = toList . Seq.insertAt i a . Seq.fromList
+          -- If table argument positional index is less than or equal to
+          -- length of 'positional' arguments then insert the table row
+          -- value in 'positional' arguments else insert the table row
+          -- value with argument name in 'named' arguments
```
lexi-lambda: This seems like it could be a useful function on `FunctionArgsExp` values generally, and it would make this code a little easier to read if it didn’t have to worry about these details. Maybe pull this out into a separate function `insertFunctionArg :: FunctionArgName -> Int -> a -> FunctionArgsExpG a -> FunctionArgsExpG a`, which can go in the same module that defines `FunctionArgsExpG`? The comment you have here can become a Haddock documentation comment on `insertFunctionArg` (and it would be even better if you gave a couple examples of the corresponding SQL in each case!).

server/src-lib/Hasura/RQL/Types/ComputedField.hs
rakeshkky: > Does this subsume the FTAFirstArgument case? It seems like FTAFirstArgument is now just a particular case of FTAName where the index is 0

Yes. But, if the index is not `0`, then `FunctionArgName` is a must. The first argument may or may not have a `FunctionArgName` and the argument name have no significance since it will be the first argument value and always positional. This type is directly inferred from the metadata API. If `table_argument` is not given then it is `FTAFirstArgument` else `FTAName`. And `FTAName` can be a first argument (if the index is `0`). My point here is to say if the table argument is first or it is named. :)

lexi-lambda: Ah, I see! And the name is needed in the other cases because the other arguments might be optional? That makes sense to me. Maybe add a short comment explaining the distinction? (Also, this is a minor nit, but I think `FTAFirst` and `FTANamed` might be more consistent names for the constructors.)

server/src-lib/Hasura/RQL/DML/Select/Types.hs
```diff
@@ -316,6 +317,24 @@ data FunctionArgsExpG a
 emptyFunctionArgsExp :: FunctionArgsExpG a
 emptyFunctionArgsExp = FunctionArgsExp [] HM.empty
 
+-- If argument positional index is less than or equal to
+-- length of 'positional' arguments then insert the value in
+-- 'positional' arguments else insert the value with
+-- argument name in 'named' arguments
```
lexi-lambda: This comment needs to start with a `|` so it can get picked up by Haddock. Also, since [our style guide currently encourages a 100-character line length](https://github.com/hasura/graphql-engine/blob/950d3842e4662c2ff5d8d5b20ac64bf9f9387b80/server/CONTRIBUTING.md#line-length), could you wrap this comment to 100 character lines?

server/src-lib/Hasura/RQL/Types/ComputedField.hs
```diff
@@ -27,13 +27,19 @@ computedFieldNameToText = unNonEmptyText . unComputedFieldName
 fromComputedField :: ComputedFieldName -> FieldName
 fromComputedField = FieldName . computedFieldNameToText
 
+-- The function table argument is either
+-- the very first argument or the named
+-- argument with an index. The index is 0
+-- if the named argument is the first.
```
lexi-lambda: Same as the above comment.

lexi-lambda:
No worries! I mostly just didn’t want to bother you with any more tiny changes; I’ve asked you to make enough of them already. 😅

rikinsk:
Just went through the docs and seems like there is a change in the select permissions api. Won't this need changes in the console for adding select permissions for computed fields?

PR #2924, opened 2019-09-23
add brotli library dependencies to graphql-engine-packager image

rakeshkky:
### Description

Add `brotli` library dependencies to the server docker image.

The compression feature introduced in https://github.com/hasura/graphql-engine/pull/2751 requires `brotli` shared libraries at runtime. In original PR, adding them to server packager image was missing.

### Affected components 

- Server

### Solution and Design

Install brotli from [source](https://github.com/google/brotli) and copy shared object files to `rootfs`.

### Steps to test and verify

Open Heroku preview app.


tirumaraiselvan:
LGTM

PR #2931, opened 2019-09-24
support aliasing root fields and column fields in console; introduce ReasonML in console codebase

wawhal:
Fixes internal issue-270.
Implements the UI for setting GraphqL aliases to your root fields and columns.
Also introduces ReasonML in the codebase.

### Affected components 

- Console


marionschleifer:
@wawhal just asking to confirm that no additional documentation is needed for this?

server/src-exec/Migrate.hs
```diff
@@ -432,4 +432,4 @@ migrateCatalog migrationTime = migrateFrom =<< getCatalogVersion
                 UPDATE "hdb_catalog"."hdb_version"
                    SET "version" = $1,
                        "upgraded_on" = $2
-                    |] (curCatalogVer, migrationTime) False
```
lexi-lambda: I’m not sure why a bunch of these files in the server code had their trailing newlines stripped, but could you revert these changes on your branch? (That way I won’t be a code owner on this PR, and you won’t have to wait for me to review it.)

wawhal: Done.

wawhal:
This PR is re-done [here](#3175).

PR #2932, opened 2019-09-24
update pr template with server breaking changes checklist

rakeshkky:
### Description

Update PR template to include server breaking changes checklist

**Suggestions are welcome**


dsandip:
@lexi-lambda, could you please include this in your review pipeline?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -26,6 +26,65 @@
...
+
+   Schema Resolve:-
+   <!-- Any change in logic of resolving input request -->
+   - [ ] Change in treatment of `null` value for input fields
+     <!-- Add if anything not listed above -->
+
+3. Logging
+
+   - [ ] Log `JSON` schema
+   - [ ] Log `type` names
```
lexi-lambda: What is our policy on the stability of the logging schema, anyway? Is it documented anywhere?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -26,6 +26,65 @@
...
+   Schema Generation:-
+   <!-- Any changes in schema auto-generation logic -->
+   <!-- All GraphQL schema names are case sensitive -->
+   - [ ] Change in any `NamedType`
+   - [ ] Change in table field names
+     <!-- Add if anything not listed above -->
+
+   Schema Resolve:-
+   <!-- Any change in logic of resolving input request -->
+   - [ ] Change in treatment of `null` value for input fields
```
lexi-lambda: This seems like an oddly specific bullet point to me, but maybe I’m just not understanding it? Is there a reason this in particular is worth calling out above other changes?

.github/PULL_REQUEST_TEMPLATE.md
rakeshkky: Our logging is documented [here](https://docs.hasura.io/1.0/graphql/manual/deployment/logging.html#). Few users are using JSON logs to monitor the activity in production. Any change in the logging system may break their assumptions.

.github/PULL_REQUEST_TEMPLATE.md
rakeshkky: [GraphQL spec](https://graphql.github.io/graphql-spec/June2018/#sec-Null-Value) says explicitly providing `null` value and omitting the input field are similar but not identical.
For example, in insert mutation, the objects are inserted with `DEFAULT` value for omitted fields and `NULL` value for `null` fields. Change in treatment of `null` value may break the behaviour. This can be a "soft" breaking change. 

I'll rephrase the sentence to `.. for any input fields`

rakeshkky:
I've few updates to add, hence `do-not-merge` label.

PR #2948, opened 2019-09-26
Revert "fix console server-build script (#2790)"

rikinsk:
This reverts commit eb0c1f66425568c5fcf32c882223a2e92503412d.

### Description

Testing if reverted commit is the cause

### Affected components 

- Tests


lexi-lambda:
This seems to have been an issue with the CircleCI `node_modules` cache: somehow, the cache built by https://github.com/hasura/graphql-engine/commit/eb0c1f66425568c5fcf32c882223a2e92503412d was poisoned in some way. I’ve rebuilt the cache and restarted the failed jobs, and they’re passing now, so reverting this is unnecessary.

PR #2949, opened 2019-09-26
Fix typo in warning message

mvdriel:
Fix typo in warning message

### Affected components 
- Server


lexi-lambda:
Thanks for the fix!

PR #2967, opened 2019-10-01
remove brotli compression

rakeshkky:
### Description

The [brotli lib](http://hackage.haskell.org/packdependeciesage/brotli) has licensed under GPL 3.0 only. Hence removing brotli compression until the library author updates the license or an alternates solution found. Gzip compression works as expected.

### Affected components 

- Server
- Tests
- Docs

### Related Issues

N/A

### Solution and Design

- Remove `brotli` from library dependencies
- Comment out the brotli compression related code


server/src-lib/Hasura/Server/Compression.hs
```diff
@@ -7,22 +7,23 @@ where
 
 import           Hasura.Prelude
 
-import           Hasura.Server.Utils       (brHeader, gzipHeader)
+-- import           Hasura.Server.Utils       (brHeader, gzipHeader)
```
lexi-lambda: I think these comments should be removed rather than just commented out—we can always get the deleted code back by reverting this commit.

server/tests-py/test_compression.py
```diff
@@ -63,27 +63,28 @@ def test_gzip_compression_v1_query(self, hge_ctx):
...
+    #     self._assert_brotli(resp, exp_resp)
 
-    def test_brotli_compression_v1_query(self, hge_ctx):
-        url, q, exp_resp = self._get_config(self.dir() + '/v1_query.yaml')
-        resp = self._make_post(hge_ctx, url, q, self.brotli_header)
-        self._assert_brotli(resp, exp_resp)
+    # def test_brotli_compression_v1_query(self, hge_ctx):
+    #     url, q, exp_resp = self._get_config(self.dir() + '/v1_query.yaml')
+    #     resp = self._make_post(hge_ctx, url, q, self.brotli_header)
+    #     self._assert_brotli(resp, exp_resp)
```
lexi-lambda: Could you also remove these commented-out tests? (Sorry for being nitpicky about this, since it’s kind of a minor thing, but it’s easy to fix, so it seems worth doing.)

PR #2974, opened 2019-10-01
Update README.md (close #2974) 

DEBSUBHRO:
### Description
I think that that the user will not find it good if there are two types of [Read more] elements. 

### Affected components 
- Docs
- Community Content

### Solution and Design
Making both the elements appear same.

--


shahidhk:
Hi @DEBSUBHRO , We appreciate the effort, but this pull request doesn't add any value. Feel free to pick up an issue labelled `hacktoberfest` from https://github.com/hasura/graphql-engine/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Ahacktoberfest :smile:

PR #2980, opened 2019-10-02
fix query is logged as string on authz errors (close #2963)

### LINKED ISSUE #2963: query is logged as string on authz errors
For errors like invalid-jwt etc, the `http-log` contains `detail.operation.query` as a string, instead of the parsed JSON. This is causing trouble in analysing logs.

```
{
  "timestamp": "2019-10-01T08:38:22.034+0000",
  "level": "error",
  "type": "http-log",
  "detail": {
    "operation": {
      "query_execution_time": null,
      "user_vars": null,
      "error": {
        "path": "$",
        "error": "Malformed Hasura-Collaborator-Token header",
        "code": "invalid-headers"
      },
      "request_id": "52faa2d5-5269-4e68-96ab-71cacf0db765",
      "response_size": 90,
      "query": "{\"query\":\"{\\n  online_users {\\n    count\\n  }\\n}\\n\",\"variables\":null}"
    },
    "http_info": {
      "status": 200,
      "http_version": "HTTP/1.1",
      "url": "/v1/graphql",
      "ip": "xxxxxx",
      "method": "POST"
    }
  }
}
```

The query key should be a JSON object.

hgiasac:
### Description

For errors like invalid-jwt etc, the http-log contains detail.operation.query as a string, instead of the parsed JSON. This is causing trouble in analysing logs.

### Affected components 

- Server
- Tests

### Related Issues

#2963 

### Solution and Design

Parse JSON `http-log.detail.operation.query` when query is JSON string

### Steps to test and verify

- Do some unauthorized (empty x-hasura-admin-secret) request 
- Run automation test

### DIFF
--- a/server/src-lib/Hasura/Server/Logging.hs
+++ b/server/src-lib/Hasura/Server/Logging.hs
@@ -14,6 +14,7 @@ module Hasura.Server.Logging
 import           Data.Aeson
 import           Data.Aeson.Casing
 import           Data.Aeson.TH
+import           Data.Aeson.Types
 import           Data.Bits                 (shift, (.&.))
 import           Data.ByteString.Char8     (ByteString)
 import           Data.Int                  (Int64)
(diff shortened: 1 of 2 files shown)


hgiasac:
According to @lexi-lambda, reading the entire payload on every request makes us more vulnerable to denial-of-service attacks, and parsing the whole thing makes it even worse (consider: bad client sends us a several GB payload, we reject their auth, but we still try to decode their whole JSON document into memory).

Therefore we have another option: define new field `raw_query` without parsing. The code implementation is different a bit. So I create another PR (#3020) and close this one

PR #2997, opened 2019-10-03
fix rolename change bug when perm edit is open (closes #2682)

### LINKED ISSUE #2682: Changing role name is not possible after expanding permissions
Steps to reproduce:

1. Add a role in the permissions tab on the console
2. Click to expand to add permissions to the role

The role name still shows as editable:

image

3. Edit the role name
4. Save permissions

The role name will not be changed but the one that was created before expanding is used.

wawhal:
### Description

Check #2682  description

### Affected components 

- Console

### DIFF
--- a/console/src/components/Services/Data/TablePermissions/Actions.js
+++ b/console/src/components/Services/Data/TablePermissions/Actions.js
@@ -638,7 +638,7 @@ const permChangePermissions = changeType => {
     const limitEnabled = permissionsState.limitEnabled;
 
     const table = permissionsState.table;
-    const role = permissionsState.role;
+    const role = permissionsState.newRole || permissionsState.role;
     const query = permissionsState.query;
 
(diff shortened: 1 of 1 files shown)


console/src/components/Services/Data/TablePermissions/Actions.js
```diff
@@ -638,7 +638,7 @@ const permChangePermissions = changeType => {
     const limitEnabled = permissionsState.limitEnabled;
 
     const table = permissionsState.table;
-    const role = permissionsState.role;
+    const role = permissionsState.newRole || permissionsState.role;
```
rikinsk: This isn't a neat solution. permissionState.role needs to be the source of truth for what role we are setting permissions for. With this solution the display still shows the old role on the UI but applies to the updated value which is confusing.

Better solution is to update value of permissionState.role along when newRole is updated if edit section is open

PR #3002, opened 2019-10-03
support where clause in on_conflict of insert mutation (close #2795)

### LINKED ISSUE #2795: Support where clause parameter on insert mutation on_conflict
We have a use case where we want to upsert entries but in the case of an update, the row is only affected if it meets a certain condition, e.g. the timestamp of the new data is greater than the timestamp of the old.

An example query might look like:

```graphql
mutation {
  insert_item(
    objects: {itemId:"1234", creationEpochSeconds: 1567026834},
    on_conflict: {
    	constraint: item_pkey,
        update_columns: [itemId, creationEpochSeconds]
        where: {creationEpochSeconds: {_lt: 1567026834 }}
    }) {
    affected_rows
  }
}
```

rakeshkky:
### Description

Support `where` clause in `on_conflict` input object. It is optional. This `where` condition applies to old data.

For more details, refer to issue https://github.com/hasura/graphql-engine/issues/2795

### Affected components 

- Server
- Docs
- Tests

### Related Issues

closes #2795 

### Solution and Design

- Generate nullable `where` field in `on_conflict` input object
```graphql
    on_conflict: {
      constraint: table_constraint!
      update_columns: [table_update_column!]!
      where: table_bool_exp
    }
```
- Resolve the `where` field using text encoding of input values.
- Join the obtained boolean expression from `where` with update filter using `AND` operator

### Steps to test and verify

Follow any test case added in this PR

### DIFF
--- a/docs/graphql/manual/mutations/upsert.rst
+++ b/docs/graphql/manual/mutations/upsert.rst
@@ -107,7 +107,7 @@ The ``published_on`` column is left unchanged as it wasn't present in ``update_c
 Update selected columns on conflict using a filter
 --------------------------------------------------
 Insert a new object in the ``article`` table, or if the primary key constraint ``article_pkey`` is violated, update
-the columns specified in ``update_columns`` only if provided the ``where`` condition is met:
+the columns specified in ``update_columns`` only if the provided ``where`` condition is met:
 
 
 .. graphiql::
--- a/docs/graphql/manual/mutations/upsert.rst
+++ b/docs/graphql/manual/mutations/upsert.rst
@@ -44,11 +44,11 @@ The upsert functionality is sometimes confused with the update functionality. Ho
 differently. An upsert mutation is used in the case when it's not clear if the respective row is already present
 in the database. If it's known that the row is present in the database, ``update`` is the functionality to use.
 
-For an upsert, **all columns need to be passed**. 
+For an upsert, **all columns need to be passed**.
 
 **How it works**
 
-1. Postgres tries to insert a row (hence all the columns need to be present) 
+1. Postgres tries to insert a row (hence all the columns need to be present)
 
 2. If this fails because of some constraint, it updates the specified columns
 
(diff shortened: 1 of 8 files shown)


docs/graphql/manual/mutations/upsert.rst
```diff
@@ -104,6 +104,53 @@ the columns specified in ``update_columns``:
 
 The ``published_on`` column is left unchanged as it wasn't present in ``update_columns``.
 
+Update selected columns on conflict using a filter
+--------------------------------------------------
+Insert a new object in the ``article`` table or, if the primary key constraint ``article_pkey`` is violated, update
+the columns specified in ``update_columns`` only if provided ``where`` condition is met:
```
lexi-lambda: There should be a comma before the “or” in this sentence, and a “the” between “provided” and “`where` condition.” Generally this sentence reads very nicely, though. 👍

docs/graphql/manual/mutations/upsert.rst
```diff
@@ -104,6 +104,53 @@ the columns specified in ``update_columns``:
...
+            {
+              "id": 2,
+              "published_on": "2018-10-12"
+            }
+          ]
+        }
+      }
+    }
+
+The ``published_on`` column is updated only if the new value is greater than old value.
```
lexi-lambda: There should be a “the” before “old value” in this sentence.

server/src-lib/Hasura/GraphQL/Resolve/Insert.hs
```diff
@@ -193,6 +196,12 @@ parseOnConflict tn updFiltrM allColMap val = withPathK "on_conflict" $
       (_, enumVal) <- asEnumVal v
       return $ ConstraintName $ G.unName $ G.unEnumValue enumVal
 
+    parseWhereExp o = do
+      whereExpM <- forM (OMap.lookup "where" o) $ \v -> do
+        unresolved <- parseBoolExp v
+        traverse (traverse resolveValTxt) unresolved
```
lexi-lambda: I’d probably write this as `traverse (traverse resolveValTxt) =<< parseBoolExp v` so you don’t have to come up with an intermediate name (since `unresolved` isn’t very meaningful, anyway). In fact, you could even get rid of the lambda entirely and write

```haskell
for (OMap.lookup "where" o) $
  traverse (traverse resolveValTxt) <=< parseBoolExp
```

to eliminate the equally-unhelpful `v` name… and if you wanted to take it *really* far, you could rewrite the whole thing point-free in a sort of “pipeline” style, like

```haskell
parseWhereExp =
      OMap.lookup "where"
  >>> traverse (parseBoolExp >=> traverse (traverse resolveValTxt))
  >>> fmap (maybe (S.BELit True) (toSQLBoolExp (S.mkQual tn)))
```

but I’m not sure that’s actually better. :) I’ll leave it up to you if you want to incorporate any of these changes, or how far you want to go with them.

docs/graphql/manual/mutations/upsert.rst
```diff
@@ -104,6 +104,53 @@ the columns specified in ``update_columns``:
 
 The ``published_on`` column is left unchanged as it wasn't present in ``update_columns``.
 
+Update selected columns on conflict using a filter
+--------------------------------------------------
+Insert a new object in the ``article`` table, or if the primary key constraint ``article_pkey`` is violated, update
+the columns specified in ``update_columns`` only if provided the ``where`` condition is met:
```
lexi-lambda: Sorry, this was my fault—I told you the wrong thing. The “the” in this sentence should come before the word “provided,” so it should read “update the columns specified in `update_columns` only if the provided `where` condition is met.” I’ve made the change myself, though, so you don’t have to worry about it.

Issue #1740, opened 2019-03-12
Call for Boilerplates


tirumaraiselvan:
# Intro
Hasura GraphQL Engine is an open-source realtime GraphQL API layer which can be put over any Postgres database (new or existing). Along with instant GraphQL APIs, the Hasura GraphQL Engine also provides two features for writing custom logic for your app:

**Remote Schemas:** Integrate any external GraphQL service with Hasura to provide a merged GraphQL API.
**Event Triggers:** Trigger a webhook on any action (Insert, Update, Delete) on the database.

With Remote Schemas, you can extend Hasura with additional APIs for performing sync actions for e.g. `checkAndAddUser` to make custom validations before inserting a user. With Event Triggers, you can perform aysnc actions for e.g. `sendEmail` after a user is inserted. Combining both these features allows you to build complex applications easily.

To get started with Remote Schemas or Event Triggers, developers can use one of our boilerplates from here: https://github.com/hasura/graphql-engine/tree/master/community/boilerplates

Currently, the boilerplates are available for major cloud providers (AWS, GCP, Azure) in NodeJS. We want to add other infrastructure providers and languages to this repository. Check the next section for more detailed requirements.

# Requirements

You can write any of two types of boilerplates:

- Base boilerplates
- Use-case boilerplates

**Base boilerplates**

Base boilerplates are “Hello World” type boilerplates which has the scaffolding to get started with a minimal example. 

For Remote Schemas, it is a GraphQL server with a “hello” query, ref: https://github.com/hasura/graphql-engine/tree/master/community/boilerplates/remote-schemas/aws-lambda/nodejs

For Event Triggers, it is a function which simply “echoes” the input, ref:
https://github.com/hasura/graphql-engine/tree/master/community/boilerplates/event-triggers/aws-lambda/nodejs8/echo

**Use-case boilerplates**

Use-case boilerplates are common examples of real-world tasks.

For Remote Schemas, for e.g., it could be “Validate User Before Insert” or “Check Inventory Before Placing Order” GraphQL APIs. This would typically involve modeling the database and then using the types generated by Hasura GraphQL Engine in your remote schema.

For Event Triggers, for e.g., it could be “Send Email” on user creation or “Creating a data pipeline” on ingest of data. This would typically involve modeling the database.

**Infra Providers**

[...]

PR #3007, opened 2019-10-03
Add ruby examples for triggers on lambda (address #1740)

grzegorz-wcislo:
### Affected components 

- Community Content

### Related Issues

#1740 

### Steps to test and verify

Follow the READMEs


grzegorz-wcislo:
Thank you for your feedback! :smile:

I think this env variable is called `HASURA_GRAPHQL_ADMIN_SECRET` on the server side. Could you try that?

If it doesn't work replace `rescue` with

```ruby
rescue StandardError => e
  puts e.message
  puts e.backTrace
```

And please paste the AWS request logs here.

I have only tested this using local server with ngrok, not heroku,

marionschleifer:
Hey, yes, I'm using the correct env variable on the server side. There are no errors in the AWS request logs as far as I can see.

Would you mind testing this with Heroku and see if it works for you? Otherwise, you can ping me on our Discord channel (@marion#9554) and we can debug together 🙂

grzegorz-wcislo:
Hi @marionschleifer, I believe I have found a solution.

Turns out it was a ssl issue (I was using http on localhost). It should work fine now :smile:

marionschleifer:
@grzegorz-wcislo have you tried this with Heroku? I still get the same error 🤔

grzegorz-wcislo:
Yes, I did. I used this guide to create the dyno: https://docs.hasura.io/1.0/graphql/manual/getting-started/heroku-simple.html.

Could you try to print the error message (I have provided a snippet above) and send me AWS request logs? :)

I don't want to clutter example code with a lot of error handling, that's why there is only one error message :O

marionschleifer:
Sure, here they are:

image

grzegorz-wcislo:
Thanks for your help :grin:

Now I'm pretty sure the issue is solved :smile: I have made a mistake in the docs, it should say `update trigger` instead of `insert trigger`. Could you verify that the update trigger works for you?

I will fix the readme right away :smile:

community/boilerplates/event-triggers/aws-lambda/ruby/mutation/lambda_function.rb
```diff
@@ -0,0 +1,44 @@
+require 'json'
+require 'net/http'
+
+def lambda_handler(event:, context:)
+  access_key = ENV['ACCESS_KEY']
+  hge_endpoint = ENV['HGE_ENDPOINT']
+  query = <<~QUERY
+    mutation updateNoteRevision ($noteId: Int!, $data: String!) {
+      insert_note_revisions (objects: [
```
marionschleifer: This should be `insert_note_revision`. It worked for me after changing it to that 🙂

grzegorz-wcislo:
All done! :)

marionschleifer:
I don't see any changes since my last review 🤔

grzegorz-wcislo:
@marionschleifer Oh, I didn't notice my push was rejected. Now it's all done :smile:

PR #3012, opened 2019-10-04
allow specifying an upper limit on the query plan cache size

0x777:
This PR allows limiting the size of the query plan cache that we use. To keep the cache size constant, the older entries are discarded using LRU strategy. 

This can be configured through the command line argument `--query-plan-cache-size` and the env variable `QUERY_PLAN_CACHE_SIZE`. The accepted values are from `0-65535` (`0` disables the cache). When this value isn't specified, an unbounded cache is used which is the current behaviour. Ideally we should get rid of unbounded cache, maybe in the next release? We can then default to a fixed cache size. 

### Affected components 

- Server

### Solution and Design
1. A striped LRU cache is implemented using a priority queue in `Bounded.hs`. 
2. Current implementation in `Cache.hs` is moved to `Unbounded.hs`


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -926,13 +934,23 @@ enableAllowlistEnv =
...
-fallbackRefetchDelayEnv =
-  ( "HASURA_GRAPHQL_LIVE_QUERIES_FALLBACK_REFETCH_INTERVAL"
-  , "results will only be sent once in this interval (in milliseconds) for "
-  <> "live queries which cannot be multiplexed. Default: 1000 (1sec)"
+planCacheSizeEnv :: (String, String)
+planCacheSizeEnv =
+  ( "HASURA_GRAPHQL_QUERY_PLAN_CACHE_SIZE"
+  , "The maximum number of query plans that can be cached, allowed values: 0-65535, " <>
+    "0 disables the cache. If this value is not set, there is no limit on the number " <>
+    "of plans that are cached"
```
lexi-lambda: Personally, I don’t think we should default to an unlimited cache, and in fact I’m not even really sure we should allow an unbounded cache at all (since, in my opinion, an unbounded cache used in this way is more or less indistinguishable from a memory leak). I know that making informed decisions about this kind of thing is hard right now given our lack of benchmarks, but if anything, I think that’s a good reason to opt for a safe default.

In your ad-hoc tests, how significant was the performance overhead of the LRU eviction? And is it a cost imposed on every insertion, or only on evictions?

server/src-lib/Hasura/Cache/Unbounded.hs
```diff
@@ -0,0 +1,90 @@
...
+initialise :: IO (UnboundedCache k v)
+initialise = do
+  capabilities <- getNumCapabilities
+  UnboundedCache <$> V.replicateM capabilities initialiseLocal
+
+clear :: UnboundedCache k v -> IO ()
+clear (UnboundedCache caches) =
+  V.mapM_ clearLocal caches
+
+{-# INLINE getLocal #-}
```
lexi-lambda: What is the performance impact of this `INLINE` pragma? It seems unlikely to do much to me, but maybe I’m missing something.

server/src-lib/Hasura/Server/Init.hs
0x777: @lexi-lambda I agree. The plan is to get rid of the unbounded cache as soon as I can spend time running benchmarks to see if there is any noticeable performance difference (will get to it as soon as my work on actions wraps up). 

> In your ad-hoc tests, how significant was the performance overhead of the LRU eviction? And is it a cost imposed on every insertion, or only on evictions?

I don't remember the exact numbers to be honest. I don't think it has anything to do with the `HashPSQ` data structure, the overhead was most likely because of the contention on the ioref which holds the cache structure as every read modifies the cache structure and it translates to a write on the ioref.

jberryman:
@lexi-lambda writes on slack:

> I agree. I have zero idea how #3012 actually impacts performance in the wild, but I think it’s a conservative change for two reasons: (1) it is disabled by default, and (2) it replaces an unbounded cache with a bounded one, which is fundamentally the right thing to do. An unbounded cache is sort of necessarily a memory leak by definition. 
> I think it’s highly unlikely to reduce performance meaningfully, but I have no evidence (for or against) whether it’s related to the user-reported memory leak problems in any way. So I’m happy to merge it, but I don’t think we should consider it “the fix” for those issues in any sense without further investigation.

I'm comfortable with that as well. But I'll try to do some very quick measurements today and report back.

Unfortunately I'm not aware of a great way to inspect and debug space leaks in production, but some initial ideas:
- compile _without_ `-prof` but run using `-h` with a very low snapshot interval (like once every 5 minutes or something; so `+RTS -h -i5` iirc). This might provide a clue but is just as likely to be useless imo.
- longer term: see if we can use or make useable https://github.com/bgamari/ghc-debug/blob/master/call-for-participation.md 
- see also https://simonmar.github.io/posts/2018-06-20-Finding-fixing-space-leaks.html , though I need to re-read this; I think it's more of a hack for asserting that objects don't leak by doing a lot of manual work. But it also presents inspecting the heap with gdb. see also: https://gitlab.haskell.org/ghc/ghc/wikis/debugging/compiled-code

It would be really wonderful to just be able to run a script and get a live heap snapshot of some sort for a production (optimized, not profiling) instance.

Finally, it's orthogonal but one note I remember making re. caching back in Bangalore was: I wonder if we ought to only cache plans for requests that make use of variables? It seems possible there are some queries that are completely static and very common, but my understanding is the cache design is based on the hypothesis that users are going to use parameterized graphql queries. Maybe it makes sense to at least treat the two with different policies eventually.

jberryman:
I ran a simple benchmark blasting the same query over and over, as a way to check bookkeeping overhead. I wasn't able to see any difference in latency between the bounded and unbounded version or master.

jberryman:
https://github.com/hasura/graphql-engine-internal/issues/317

PR #3020, opened 2019-10-04
add raw query field for error http log (close #2963)

### LINKED ISSUE #2963: query is logged as string on authz errors

hgiasac:
### Description

For errors like invalid-jwt etc, the http-log contains detail.operation.query as a string, instead of the parsed JSON. This is causing trouble in analysing logs.

### Affected components 

- Server
- Tests

### Related Issues

#2963 

### Solution and Design

We can parse JSON query data. However, according to @lexi-lambda, reading the entire payload on every request makes us more vulnerable to denial-of-service attacks, and parsing the whole thing makes it even worse (consider: bad client sends us a several GB payload, we reject their auth, but we still try to decode their whole JSON document into memory).

Therefore, there is an option that add `raw_query` field for raw query string

### Steps to test and verify

- Do some unauthorized (empty x-hasura-admin-secret) request
- Or run automation test

None

### DIFF
--- a/server/src-lib/Hasura/Server/Logging.hs
+++ b/server/src-lib/Hasura/Server/Logging.hs
@@ -146,8 +146,8 @@ data OperationLog
   , olResponseSize       :: !(Maybe Int64)
   , olQueryExecutionTime :: !(Maybe Double)
   , olQuery              :: !(Maybe Value)
-  , olRawQuery           :: !(Maybe Value)
-  , olError              :: !(Maybe Value)
+  , olRawQuery           :: !(Maybe Text)
+  , olError              :: !(Maybe QErr)
   } deriving (Show, Eq)
 $(deriveToJSON (aesonDrop 2 snakeCase) ''OperationLog)
 
@@ -205,7 +205,7 @@ mkHttpErrorLog
   -> RequestId
   -> Wai.Request
   -> QErr
-  -> Maybe Value
+  -> Either ByteString Value
   -> Maybe (UTCTime, UTCTime)
   -> Maybe CompressionType
   -> HttpLog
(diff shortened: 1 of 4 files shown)


server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -146,6 +146,7 @@ data OperationLog
   , olResponseSize       :: !(Maybe Int64)
   , olQueryExecutionTime :: !(Maybe Double)
   , olQuery              :: !(Maybe Value)
+  , olRawQuery           :: !(Maybe Value)
   , olError              :: !(Maybe Value)
```
lexi-lambda: I think these probably shouldn’t all just be stored as `Value`, since the derived `ToJSON` instance will handle the conversion to JSON for us, so we should store more specific types if we can. I think the reason we’re using `Value` for `olQuery` is that queries have many different types, so we convert them to `Value` beforehand to normalize them, but `olRawQuery` and `olError` can probably have the more specific types `Maybe ByteString` and `Maybe QErr`, respectively.

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -146,6 +146,7 @@ data OperationLog
...
   } deriving (Show, Eq)
 $(deriveToJSON (aesonDrop 2 snakeCase) ''OperationLog)
```
lexi-lambda: It probably makes sense to enable the [`omitNothingFields`](https://hackage.haskell.org/package/aeson-1.4.5.0/docs/Data-Aeson.html#v:omitNothingFields) option here to suppress generating explicit `null`s in the output, since it isn’t very helpful to get things like `"raw_query": null` in log messages.

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -221,14 +223,20 @@ mkHttpErrorLog userInfoM reqId req err query mTimeT compressTypeM =
            , olUserVars     = userVars <$> userInfoM
            , olResponseSize = respSize
            , olQueryExecutionTime = respTime
-           , olQuery = toJSON <$> query
+           , olQuery = toJSON <$> query'
+           , olRawQuery = toJSON <$> rawQuery
            , olError = Just $ toJSON err
```
lexi-lambda: There should be no need to call `toJSON` explicitly here if the types are updated as mentioned above, since the values should already have the appropriate types.

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -221,14 +223,20 @@ mkHttpErrorLog userInfoM reqId req err query mTimeT compressTypeM =
...
            }
   in HttpLog L.LevelError $ HttpAccessLog http op
   where
     status = qeStatus err
     respSize = Just $ BL.length $ encode err
     respTime = computeTimeDiff mTimeT
+    query' = query >>= \q -> case q of
+      Object _ -> Just q
+      Array _  -> Just q
+      _        -> Nothing
```
lexi-lambda: I think this is probably not the right approach, since it assumes all valid queries will always be objects or arrays, and it mixes concerns. I think the right solution is probably to change the type of the `query` argument to `Either ByteString Value`, or alternately, to make a new datatype `data SomeQuery = UnparsedQuery !ByteString | ParsedQuery !Value` (either is fine with me). Then the call site in `Hasura/RQL/Types/Column.hs` should be updated to pass the raw query using `Left`/`UnparsedQuery` without converting it to `Value` explicitly.

server/src-lib/Hasura/Server/Logging.hs
hgiasac: Can `ByteString` be `Text`, because type `ByteString` doesn't have ToJSON instance?

hgiasac:
Cool. Thank you for review

PR #3039, opened 2019-10-06
Punctuation and typo fix

OrginalS:


marionschleifer:
@OrginalS could you please send an email to marion@hasura.io, so that we can arrange the shipping of your swag? 🙂

PR #3041, opened 2019-10-06
Accept query params for filter/sort in browse table page (close #833)

### LINKED ISSUE #833: accept query params for filter/sort in browse table page
Accept URL query params for filter/sort in console's browse table page so that it can be like the default query when the page is opened.

(via comfortablydumb on Discord)

feloy:
### Affected components 
- Console

### Related Issues
#833


feloy:
Hi @rikinsk , any  thought on this PR?

rikinsk:
@feloy Sorry for the delay in this. Will get around to reviewing this in a day or two.

console/src/components/Services/Data/TableBrowseRows/FilterQuery.js
```diff
@@ -171,9 +174,59 @@ const renderSorts = (orderBy, tableSchema, dispatch) => {
...
+    } else if (Array.isArray(this.props.urlQuery.sort)) {
+      urlSorts = this.props.urlQuery.sort;
+    }
+    this.props.curQuery.order_by = urlSorts.map(sort => {
+      const parts = sort.split(';');
+      const column = parts[0];
+      const type = parts[1];
+      const nulls = 'last';
+      return { column, type, nulls };
+    });
```
beerose: Here, you are mutating the props, which is an anti-pattern in react. A component should not manage its props. `props` are literally  **the state that is managed by the component owner.** That's why they should be treated as immutable data. 

My proposition here would be to use local variables:
```js
    const where = {
      $and: urlFilters.map(filter => {
        const parts = filter.split(';');
        const col = parts[0];
        const op = parts[1];
        const value = parts[2];
        return { [col]: { [op]: value } };
      }),
    };

    const order_by = urlSorts.map(sort => {
      const parts = sort.split(';');
      const column = parts[0];
      const type = parts[1];
      const nulls = 'last';
      return { column, type, nulls };
    });
```

And then use them in `setDefaultQuery` function:
```js
dispatch(setDefaultQuery({ where, order_by }));
```

console/src/components/Services/Data/TableBrowseRows/ViewTable.js
```diff
@@ -201,6 +201,7 @@ class ViewTable extends Component {
         updateInvocationFunction={this.updateInvocationFunction.bind(this)}
         triggeredRow={triggeredRow}
         triggeredFunction={triggeredFunction}
+        location={this.props.location}
```
beerose: Could you destruct this variable from props in here: https://github.com/hasura/graphql-engine/blob/9afe08ce094fbe9d2d009f826c008ce44d9c8c09/console/src/components/Services/Data/TableBrowseRows/ViewTable.js#L142
Just for the consistency.

console/package.json
```diff
@@ -97,7 +97,7 @@
     "react-tabs": "^2.1.0",
     "react-toggle": "^4.0.2",
     "redux": "^4.0.0",
-    "redux-logger": "^3.0.6",
+    "redux-logger": "3.0.1",
```
beerose: What was the reason for this change?

PR #3072, opened 2019-10-09
cli(migrations): new folder structure and squash

### LINKED ISSUE #2254: Squash migrations
We are just starting to use hasura and I'm trying to modify schema via console UI to leverage automatic migrations. Because it trial and error process I ended up with tens of migration files. These files are meant to be kept in repository and subject to code review. Right now its not possible because all errors in the process are also expressed as separate migrations.

Is it possible to squash the migrations somehow or generate them only after pressing some "save" button?

### LINKED ISSUE #2724: Let's make resetting(squash) migrations EASY!
Migration files can easily be piled up. So "squash" feature would be required in some projects. According to <https://blog.hasura.io/resetting-hasura-migrations/>, in order to squash, we should do 

1. Remove every migration files (`rm migrations/*`).
2. Remove whole data from the table `hdb_catalog.schema_migrations` (`TRUNCATE hdb_catalog.schema_migrations;`)
3. Pull the metadata (and optionally schema) from hasura.
4. Apply the pulled metadata (and optionally schema) just back to hasura (with `--skip-execution` option as hasura obviously already have the metadata, but only needs "version" to be applied.)

## Problems

In my humble opinion, there are these problems.

### Inconsistency (high-level vs low-level migration tasks)

Users handle migration through cli. The high-level approach should be consistent. But suddenly having to take care about `schema_migrations` table is not enough abstract. So it does not give 'smooth' and 'expected' experience. I guess many even don't know `schema_migrations`, as docs(<https://docs.hasura.io/1.0/graphql/manual/how-it-works/metadata-schema.html>) only explicitly explains `hdb_table`, `hdb_relationship`, and `hdb_permission`, not `schema_migrations` and many others.
 
### Somewhat hacky? (at least to me)

Users should specify `--skip-execution`. After they just pulled migrations files from hasura, they have to "pretend" applying migration only for writing a new row on `schema_migrations`. It makes me feel I am doing a "workaround" rather than a straight official way.

### Too manual

There are total 4 manual steps for one "squash". These are cumbersome and error-prone. 

"squash"ing does happen as projects grow. People obviously just don't want hundreds of migration files to be stacked, because it'd be inconvenient to open migration directory and it just simply looks not "neat". So the "squash" can't but happen and it should be automatic for convenience.

## Solution

I suggest a command like 

```bash
hasura migrate squash
```

which would safely execute the 4 steps automatically.

arvi3411301:
### Description
This PR introduces three new features:

- Support for a new migrations folder structure.
- Add `squash` command in preview.
- ~List of migrations on the Console and ability to squash them from console.~

#### New migrations folder structure

Starting with this commit, Hasura CLI supports a new directory structure for migrations folder and defaults to that for all new migrations created. 

Each migration will get a new directory with the name format `timestamp_name` and inside the directory, there will be four files:

```bash
└── migrations
    ├── 1572237730898_squashed
    │   ├── up.sql
    │   ├── up.yaml
    │   ├── down.yaml
    │   └── down.sql
```

Existing files old migration format `timestamp_name.up|down.yaml|sql` will continue to work alongside new migration files.

#### Squash command

Lots of users have expressed their interest in squashing migrations (see #2724 and #2254) and some even built [their own tools](https://github.com/domasx2/hasura-squasher) to do squash. In this PR, we take a systematic approach to squash migrations.

A new command called `migrate squash` is introduced. Note that this command is in **PREVIEW** and the correctness of squashed migration is not guaranteed (especially for down migrations). From our tests, **it works for most use cases**, but we have found some issues with squashing all the down migrations, partly because the console doesn't generate down migrations for all actions.

Hence, until we add an extensive test suite for squashing, we'll keep the command in preview. We recommend you to confirm the correctness yourself by diffing the SQL and Metadata before and after applying the squashed migrations (we're also thinking about embedding some checks into the command itself).

```bash
$ hasura migrate squash --help
(PREVIEW) Squash multiple migrations leading upto the latest one into a single migration file

Usage:
  hasura migrate squash [flags]

Examples:
  # NOTE: This command is in PREVIEW, correctness is not guaranteed and the usage may change.

  # squash all migrations from version 1572238297262 to the latest one:
  hasura migrate squash --from 1572238297262

Flags:
      --from uint             start squashing form this version
      --name string           name for the new squashed migration (default "squashed")
      --delete-source         delete the source files after squashing without any confirmation
```

### Affected components 

- CLI

### Related Issues

Close #2724, Close #2254, 

### Solution and Design

For the squash command, a state machine is implemented to track changes to Hasura metadata. After applying each action on the metadata state, a list of incremental changes is created.

### Steps to test and verify
1. Open console via cli and create some migrations.
2. Run `hasura migrate squash --from <version>`

- The `squash` command is in preview
- Support for squashing from the console is WIP
- Support for squashing migrations that are not committed yet is planned.
- Un-tracking or dropping a table will cause inconsistent squashed down migration since console doesn't generate correct down migration.
- If cascade setting is set to `true` on any of the metadata action, generated migration may be wrong


arvi3411301:
TODO:

- [ ]  Add support for track_table v2
- [ ]  Add support for set_table_custom_fields v2

PR #3084, opened 2019-10-10
fix hdb_catalog.hdb_column view  (fix #3083)

### LINKED ISSUE #3083: Upgrade to v1.0.0-beta.7 causes problems on tables set as enum
Enum tables that comply with the criteria and work perfectly fine in beta.6 cause the beta.7 to fail when starting up. An example message from the logs (schema and table name obfuscated):

{"definition":{"schema":"xxx","name":"yyy"},"reason":"the table \"xxx.yyy\" cannot be used as an enum for the following reasons:\n  • the table’s primary key must not span multiple columns (id, id)\n  • the table must have exactly one primary key and optionally one comment column, not 4 columns (id, id, value, value)\n","type":"table"}

As you can see the server seems to think there are 2 id and 2 value fields but in reality there is only 1 id field and only 1 value field.

This inconsistency in the metadata (we have multiple of these messages) prevents the startup process from continuing. We currently reverted back to beta.6 where everything works just fine.

rakeshkky:
### Description

Avoid duplicate rows in hdb_catalog.hdb_column view if two tables with same name present in different schemas.

### Affected components 

- Server
- Tests

### Related Issues

fix #3083 

### Solution and Design

Use normal `JOIN` instead of `LEFT JOIN` for pg_class and pg_namespace tables

### DIFF
--- a/server/src-rsr/initialise.sql
+++ b/server/src-rsr/initialise.sql
@@ -450,13 +450,13 @@ CREATE VIEW hdb_catalog.hdb_column AS
         , coalesce(pkey_refs.ref_tables, '[]') AS primary_key_references
         , col_description(pg_class.oid, columns.ordinal_position) AS description
      FROM information_schema.columns
+JOIN pg_class ON pg_class.relname = columns.table_name
+JOIN pg_namespace ON pg_namespace.oid = pg_class.relnamespace
+     AND pg_namespace.nspname = columns.table_schema
 LEFT JOIN primary_key_references AS pkey_refs
            ON columns.table_schema = pkey_refs.src_table_schema
           AND columns.table_name   = pkey_refs.src_table_name
-          AND columns.column_name  = pkey_refs.src_column_name
-LEFT JOIN pg_class ON pg_class.relname = columns.table_name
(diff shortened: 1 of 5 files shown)


server/src-rsr/initialise.sql
```diff
@@ -448,15 +448,22 @@ CREATE VIEW hdb_catalog.hdb_column AS
...
+LEFT JOIN
+     (
+       SELECT pg_class.oid
+            , pg_class.relname AS table_name
+            , pg_namespace.nspname AS table_schema
+         FROM pg_class
+         LEFT JOIN pg_namespace ON pg_namespace.oid = pg_class.relnamespace
+     ) table_oid ON
+       table_oid.table_name = columns.table_name
+       AND table_oid.table_schema = columns.table_schema;
```
lexi-lambda: Why not just change the original `LEFT JOIN`s to ordinary (`INNER`) `JOIN`s, and move them above the `LEFT JOIN` to `primary_key_references`? That would avoid the issue with a much simpler query, and it’s not like those joins can ever fail, right?

rakeshkky: Yes, That's much simpler

Issue #1210, opened 2018-12-13
enable up down arrows in dropdown+textbox inputs (event trigger, remote schemas)


wawhal:

pradeepgangwar:
@wawhal Let me fix this up.

pradeepgangwar:
@wawhal Dropdowns in event trigger form are working with up-down arrow keys. Please let me know the area where this issue exists.

pradeepgangwar:
@wawhal I went through dropdown code and found that there is no default option in react-bootstrap that enables arrow keys. So I guess I have to add it manually by adding a function that acts on arrow key. This is my first issue on this repo so pls let me know if I am thinking in right direction. 😅

wawhal:
@pradeepgangwar yes you'll have to write a key handler.

thangngoc89:
https://github.com/paypal/downshift is a good library for handling these kind of things

jeremie1112:
Hi, is someone working on this? @wawhal 
else let me know if I can work on this issue.

wawhal:
Go for it @jeremie1112 :+1:

wawhal:
@jeremie1112 yes @codingdefined has opened [a PR](#3105)

jeremie1112:
> @jeremie1112 yes @codingdefined has opened [a PR](#3105)

@wawhal Can you recommend me any "good first issue" tag please?

marionschleifer:
@jeremie1112 here is a list of good first issues: https://github.com/hasura/graphql-engine/issues?q=is%3Aopen+is%3Aissue+label%3Ap%2Fmedium+sort%3Aupdated-desc. Here are some guidelines for contributing: https://github.com/hasura/graphql-engine/wiki/Hacktoberfest-2019-%F0%9F%8E%89 🙂

PR #3105, opened 2019-10-12
enabled up down arrows in dropdown+textbox inputs

codingdefined:
Fixed https://github.com/hasura/graphql-engine/issues/1210

### Description
Enabled the up and down arrows in dropdown + textbox inputs. Added class 'active' based on the up and down arrow.  

### Affected components 
- Console

### Solution and Design
Created an event handler and based on the up and down key, changing the position of the cursor.

We need to use active class because of the default behavior and there is no way we can stop the default behavior.

### DIFF
--- a/console/src/components/Common/DropdownButton/DropdownButton.js
+++ b/console/src/components/Common/DropdownButton/DropdownButton.js
@@ -6,6 +6,29 @@ import DropdownButton from 'react-bootstrap/lib/DropdownButton';
 import MenuItem from 'react-bootstrap/lib/MenuItem';
 
 class DropButton extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.state = {
+      cursor: 0,
+    };
+  }
+
+  handleKeyDown(e) {
+    const { cursor } = this.state;
+    const { dropdownOptions } = this.props;
+    // arrow up/down button should select next/previous list element
+    if (e.keyCode === 38 && cursor > 0) {
+      this.setState(prevState => ({
+        cursor: prevState.cursor - 1,
+      }));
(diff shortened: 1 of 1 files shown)


codingdefined:
@wawhal It looks like I forgot to check that condition. After spending some time, I think react-bootstrap dropdown has some issues, is it ok if we add another component react-select.com?

wawhal:
@codingdefined not really. We don't want to add another library just for this task.
Can you explain what issues does react-bootstrap have?

codingdefined:
@wawhal The problem is the default behavior of up and down arrow in DropdownButton with MenuItem is actually working opposite. That means if you click down arrow the for the first time nothing happens, but if you click up arrow the second element gets highlighted. 

To have a workaround, I have added 'active' class which makes that change invisible. Now the problem is when you add a click handler i.e. have 'Enter' the one selected by default by the DropdonButton is sent to the parent component. 

Tried changing from onClick to onSelect in MenuItem, but did not solve the problem. Do you have any insights to it?

karthikvt26:
@codingdefined thanks for attempting it. Did you come across this issue (https://github.com/react-bootstrap/react-bootstrap/issues/2105) where some ideas are discussed on how to achieve this. 

Check out the complete list of issues related to this at react-bootstrap: https://github.com/react-bootstrap/react-bootstrap/issues?q=is%3Aissue+keyboard+navigation+dropdown+is%3Aclosed

Will you check them out and let me know where we are at?

marionschleifer:
@codingdefined just wanted to check in to see if you're interested in implementing the proposed changes 🙂

Stefmore02:
Hey @codingdefined, thank you for submitting this PR.

At this time, we'll close it, and feel free to re-open it if you consider the changes are still relevant.

PR #3143, opened 2019-10-15
functions can access session info via input arg (close #2322)

### LINKED ISSUE #2322: allow functions to use session information through an argument

rakeshkky:
### Description

We can access session information (in JSON) in a transaction using `current_setting('hasura.user')` when queries are run in a transaction. But subscriptions are not run in a transaction any more. This PR supports tracking SQL functions by configuring an input session argument so as to access the session variables JSON through it.

Metadata exported by the server (through `export_metadata`) is versioned via a top-level field `version` whose value is either `1` or `2`. If `version` field is absent (metadata exported by previous server versions) then it is considered as `version: 1`. In version `2`, the exported metadata functions are a list of JSON objects with `function` and `configuration` field.

A sample metadata yaml exported.
```yaml
version: 2
functions:
- function: search_posts
  configuration: {}
- function: get_test
  configuration: {}
- function: my_add
  configuration: {}
- function: get_users
  configuration: {}
- function: get_session_var
  configuration:
    session_argument: hasura_session
# Other metadata objects come below
```

See [preview docs](https://deploy-preview-3143--hasura-docs.netlify.com/graphql/manual/queries/custom-functions.html#accessing-hasura-session-variables-in-custom-functions) to learn more about accessing session variables in SQL functions.

### Affected components 

- Server
- Docs
- Tests

### Related Issues

close #2322

### Solution and Design

- A new version 2 `track_function` query type is defined to facilitate users to specify the configuration with session argument. An example request as follows
```http
   POST /v1/query HTTP/1.1
   X-Hasura-Role: admin

   {
       "type": "track_function",
       "version": 2,
       "args": {
           "function": {
               "schema": "public",
               "name": "search_articles"
           },
           "configuration": {
               "session_argument": "hasura_session"
           }
       }
   }
```
- The server executes the SQL function with session variable JSON object.
- The export/apply metadata JSON document has a new optional `version` key.  The server exports metadata with version `2` and accepts metadata of versions `1` and `2`.

### Steps to test and verify

Follow the example [here](https://github.com/rakeshkky/graphql-engine/blob/issue-2322-func-sess-new/docs/graphql/manual/queries/custom-functions.rst#accessing-hasura-session-variables-in-custom-functions)

- Older versions cannot import metadata exported by this.


lexi-lambda:
Cross-posting a comment I made on slack, since that discussion never really went anywhere:

> It’s unfortunate that Postgres has no notion of dynamically scoped variables or something like that so that we could preserve the current API. I do think this could be an area where convention over configuration feels better to me, though—why not just automatically use an argument named `hasura_session`? That seems pretty unlikely to ever conflict with anything else, and it seems silly to track this information for every function.

@rakeshkky, do you have any thoughts on whether or not doing something like that would work/is a good idea? The overhead of configuring this for every function individually seems a little high to me, but maybe I’m overreacting.

rakeshkky:
> Cross-posting a comment I made on slack, since that discussion never really went anywhere:
> 
> > It’s unfortunate that Postgres has no notion of dynamically scoped variables or something like that so that we could preserve the current API. I do think this could be an area where convention over configuration feels better to me, though—why not just automatically use an argument named `hasura_session`? That seems pretty unlikely to ever conflict with anything else, and it seems silly to track this information for every function.
> 
> @rakeshkky, do you have any thoughts on whether or not doing something like that would work/is a good idea? The overhead of configuring this for every function individually seems a little high to me, but maybe I’m overreacting.

@lexi-lambda Now, we may avoid configuring the SQL functions and auto derive the session argument using `hasura_session` argument name, but, in future, we may need to configure the SQL function definitely if a feature does not allow us to auto derive things. Already, we are configuring the tracked tables/views. I don't feel it is not good to introduce configuration for SQL functions too. 

As a developer, IMO, I'd like to define things explicitly. While tracking a SQL functions I'd always like to specify what argument in my function accepts the session information JSON :slightly_smiling_face:. Auto deriving may affect the functions which are already tracked (if a function already has `hasura_session` as an argument).

lexi-lambda:
Adding a comment from Rikin on slack to this thread for posterity’s sake:

> Even I feel a little ambivalent about this. There does seem to be some scope for usability improvements.
>
> On one hand, the current interface is as unambiguous as it can get as each tracked function definition is complete on it own and doesn't depend on any external factors. Hence it is very straightforward to understand.
>
> On the other hand it does seem like its a little inconvenient and would cause a lot of repeated common config considering most projects would probably use the same argument to denote the session variable argument across all functions.
>
> I would probably let the current interface be as it is and release the feature. We can then later add a "new feature" where one can configure a global variable to represent the default value of the session variable argument across all the functions and then the `session_argument` configuration for the API can be made optional.
>
> My only concern here is how we would accept this global variable and persist it. It seems like it should be a part of the metadata as other metadata entities (i.e. fn defs) depend on it. A new API perhaps that persists it in the DB? A special API does seem a little odd for this as it is literally just 1 constant but thats what I can think of currently.
>
> This seems to achieve the best of both worlds and leave the choice to the users.

Makes sense to me. Sorry for the ambivalence about this!

server/src-lib/Hasura/RQL/Types/Function.hs
```diff
@@ -0,0 +1,62 @@
...
+  } deriving (Show, Eq)
+$(deriveToJSON (aesonDrop 2 snakeCase) ''SessionArgument)
+
+data FunctionInfo
+  = FunctionInfo
+  { fiName          :: !QualifiedFunction
+  , fiSystemDefined :: !SystemDefined
+  , fiType          :: !FunctionType
+  , fiInputArgs     :: !(Seq.Seq FunctionArg)
+  , fiSessionVarArg :: !(Maybe SessionArgument)
```
lexi-lambda: Keeping the session argument “on the side” like this seems like an overly indirect encoding to me, since it’s easy to just (accidentally) ignore it, and the code will still typecheck (but do the wrong thing). Ideally, when adding things like this, we should try to make it hard to ignore things that should not be ignored.

Generally, in Haskell, the simplest way to ensure something *cannot* be ignored is not to use a product type, but to use a sum. You can always just take one field out of a product type and ignore the rest, and often that happens unintentionally—you have code that uses a couple fields of a product, then you add a new field, and the old code still typechecks so you don’t think to update it—but you can’t do that with sums. Sums require you handle them with exhaustive case analysis, so when you add a new case, the compiler will tell you the places that need to be updated.

In this case, I think we can improve the representation in two ways with one change: we can make it hard to accidentally ignore the session argument, *and* we can avoid having to deal with explicit indexes (which are usually a pain in Haskell, since indexed datatypes are not inductive, and functional programming really likes inductive data structures). Here’s my proposal:

```haskell
newtype HasDefault = HasDefault { unHasDefault :: Bool }
  deriving (Show, Eq)

data FunctionArg
  = FAUserProvided !(Maybe FunctionArgName) !QualifiedPGType !HasDefault
  | FASessionVariables !FunctionArgName
  deriving (Show, Eq)

data FunctionInfo
  = FunctionInfo
  { ...
  , fiInputArgs :: !(Seq.Seq FunctionArg)
  , ...
  } deriving (Show, Eq)
```

This representation can drop the index because it’s just a property of an argument’s position in the sequence. It does technically allow `FASessionVariables` to appear more than once in the argument list, but that isn’t really a problem—we can handle it fine, even if we don’t actually provide a way for users to configure a function that way (since it wouldn’t be very useful), so it’s not as though we’ve made a truly illegal state representable.

server/src-lib/Hasura/RQL/DDL/Metadata.hs
```diff
@@ -147,8 +164,9 @@ runClearMetadata _ = do
 
 data ReplaceMetadata
   = ReplaceMetadata
-  { aqTables           :: ![TableMeta]
-  , aqFunctions        :: !(Maybe [QualifiedFunction])
+  { aqVersion          :: !(Maybe MetadataVersion)
+  , aqTables           :: ![TableMeta]
+  , aqFunctions        :: !(Maybe [Value])
```
lexi-lambda: This means we’ll insert absolutely any arbitrary JSON value into the database, right? That seems like it’s probably too permissive, since we don’t allow loading arbitrary JSON values out of the database, right? It seems like a user could end up with their database in an invalid state that way.

server/src-lib/Hasura/GraphQL/Resolve/Mutation.hs
```diff
@@ -51,6 +51,7 @@ convertMutResp ty selSet =
       UVPG annPGVal -> UVSQL <$> txtConverter annPGVal
       UVSessVar colTy sessVar -> pure $ UVSessVar colTy sessVar
       UVSQL sqlExp -> pure $ UVSQL sqlExp
+      UVSession -> pure UVSession
```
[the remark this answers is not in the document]
rakeshkky: Here we're intentionally encoding input values in `returning` field to Postgres literals but not prepared values (as like `$<Int>`) since mutations can be performed in two separate queries (See https://github.com/hasura/graphql-engine/commit/5f274b552717d107553368fec440eb7137e830fb) if `returning` has nested fields selected. It is hard to maintain the sequence for prepared values for `returning` field. Though the function is not informative, I'll make it so.

server/src-lib/Hasura/RQL/Types/Function.hs
rakeshkky: My initial type model is similar to what you mentioned above. But I removed it because
> It does technically allow FASessionVariables to appear more than once in the argument list

I want my type should convey that it has no session argument or only one session argument. Here's my proposal:
```haskell
newtype HasDefault = HasDefault { unHasDefault :: Bool }
  deriving (Show, Eq)

data FunctionArg
  = FunctionArg
  { faName       :: !(Maybe FunctionArgName)
  , faType       :: !QualifiedPGType
  , faHasDefault :: !HasDefault
  } deriving (Show, Eq)

newtype FunctionArgIndex = FunctionArgIndex {unFunctionArgIndex :: Int}
  deriving (Show, Eq)

data SessionArgument
  = SessionArgument
  { saName  :: !FunctionArgName
  , saIndex :: !FunctionArgIndex
  } deriving (Show, Eq)
$(deriveToJSON (aesonDrop 2 snakeCase) ''SessionArgument)

data InputArguments
  = IAWithSessionArgument !SessionArgument !(Seq.Seq FunctionArg)
  | IAWithoutSessionArgument !(Seq.Seq FunctionArg)
  deriving (Show, Eq)
```

Please give suggestions if any :slightly_smiling_face:

server/src-lib/Hasura/RQL/Types/Catalog.hs
```diff
@@ -89,7 +89,8 @@ data CatalogFunction
   = CatalogFunction
   { _cfFunction        :: !QualifiedFunction
   , _cfIsSystemDefined :: !SystemDefined
-  , _cfInfo            :: ![RawFunctionInfo] -- ^ multiple functions with same name
+  , _cfConfiguration   :: !FunctionConfig
+  , _cfInfo            :: ![RawFunctionInfo] -- multiple functions with same name
```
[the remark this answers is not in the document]
rakeshkky: I didn't recognize it while resolving a merge conflict. It came unintentionally.

server/src-lib/Hasura/RQL/DDL/Metadata.hs
rakeshkky: It is really odd to have `Value` here. Just now, I realised `parse, don't validate!` :smile: . As in I use `Value` here to validate it later. I'll make the necessary change.

server/src-lib/Hasura/RQL/Types/Function.hs
lexi-lambda: I’m not very fond of dealing with indexes if it isn’t absolutely necessary. The reason is that indexes are a very indirect way to access data: if you have a list of elements and an index into that list, looking up the element is a partial operation. Put another way, representations that use indexes can represent a *lot* of invalid states, since most of the valid values of type `Int` do not correspond to any element of the list. Furthermore, if you manipulate the list to rearrange it in some way, you may have to keep the indexes in sync.

Sometimes indexes are necessary for some reason or another, but I don’t think they’re necessary here. To me, the illegal state of having multiple `FASessionVariables` values in the argument list is much less unpleasant than the illegal state of having an out-of-range index, since the former still has a totally reasonable interpretation for the downstream code and isn’t really “illegal” per se (it corresponds to an entirely valid function application), while the latter is unambiguously incorrect.

My other point of reasoning is that I was thinking about what Vamshi was talking about with respect to “argument presets,” which is a reasonable idea. Even if we don’t implement those right now, the approach I proposed models it well: it makes `FASessionVariables` into just a particular kind of argument preset. Would it ever be useful for a user to preset two different arguments of a function to be session variables? No, almost certainly not! But allowing it doesn’t really have any downside, either, which is why it doesn’t really bother me.

rakeshkky: Thank you @lexi-lambda for an exhaustive explanation.

> most of the valid values of type `Int` do not correspond to any element of the list

This point very convincing :). I've never thought in this way while using an `Int` before. I'll implement your proposal. However, the discussion is still open. We've to make use of the type system to represent our requirements correctly. `the types should speak` :). If we introduce "argument presets", then it demands significant refactor in types which also speaks about the change being made to the reviewer without being written explicitly in PR description.

docs/graphql/manual/queries/custom-functions.rst
```diff
@@ -395,6 +395,51 @@ Search nearby landmarks with ``distance_kms`` default value which is 2 kms:
...
 
 
+Accessing Hasura session variables in custom functions
+******************************************************
+
+Use the v2 :ref:`track_function <track_function_v2>` to add a function by defining a session argument.
+The session argument will be a JSON object where keys are session variable names (in lower case) and values are strings.
+Use the ``->>`` JSON operator to fetch the value of a session variable as shown in the following example.
+
+.. note::
```
marionschleifer: Can we have the note in the end of this section? When we say `in the following example`, then the example should follow immediately.

PR #3154, opened 2019-10-16
allow identical fields in custom column names configuration (fix #3137) & improve root fields validation

### LINKED ISSUE #3137: New custom field API odd behaviour
The new API added in #2509 adds `track_table` (version 2) and `set_table_custom_fields` which both allow setting `custom_root_fields` and `custom_column_names`.

Attempting to set my fields to exposed as `camelCase` today revealed some odd behaviours.

I am aware this is unreleased functionality! Just want to raise the bug and suggest my opinions with regards it :) 

Tested on docker tag `hasura/graphql-engine:pull3135-9926e789` and `postgres:12`

#### Firstly,

when using `track_table`, including some of the one-word columns without changes to their name threw an error. (e.g. this request errors until you remove the fields with the <--)

```
{
  "type": "track_table",
  "version": 2,
  "args": {
    "table": "users",
    "configuration": {
        "custom_root_fields": {
          ...
        },
        "custom_column_names": {
          "created_at": "createdAt",
          "id": "id", <--
          "email": "email", <--
          "verified_at": "verifiedAt",
        }
    }
  }
}
```

`"column/relationship "id" of table "users" already exists"`

It's understandable if this is the intended behaviour. It was confusing/annoying to deal with  when wanting to update lots of fields to `camelCase` via modifying the metadata file. It could make sense to instead write the identical settings or just ignore them without throwing an error.

When the console has functionality for managing these custom names, it will obviously need to deal with excluding non-changing names.

#### Secondly,

(and more importantly) the `set_table_custom_fields` type seems to always return

```
{
  "message": "success"
}
```

whether or not I include the identical fields and it *removes* all of the `custom_root_fields` and `custom_column_names` back to defaults if it's a _"success"_ (rather than giving some sort of error).

#### Current workflow

Given what I described above, to add a new snake columns that are exposed as `camel` and have two or more words, I currently have to do the following:

- Add the col via the console (e.g. `created_at`)
- Do an `untrack_table` via an external gql console on `http://localhost:8080/v1/query`
- And then do `track_table` including the `{"created_at": "createdAt"}`

**Suggestion**

(this section should likely be a new issue but unsure what I or hasura might want)

My use case is here is that I want my JS/TS clients to be completely `camelCase` based. I like `snake_case` coming from a Rails/PG background, however, I hate mixing them! JS + graphql land seems to be in favour of `camelCase`.

One suggestion I'd like to make for a feature is to be able to pass a header (or some other per-client option) which specifies the field names it wants. This might seem a bit crazy but if I begin to write data processing service in a Python (for example), it would be cool to let it consume the API using the _real_ `snake_case` fields!

A few ideas for implementation of this might be:

- _Automagically_ convert the schema fields and types based on `X-Hasura-Graphql-Casing` passed as either `snake` or `camel`. This would mean not having to deal with setting up custom column and root fields which would be useful for me (and I assume, many others). It also would allow avoiding creating complex UI in the console for something that is quite simple.
- Keep the current system introduced in #2509 but allow a header such as `X-Hasura-Ignore-Custom-Fields` to disable/enable use of them when the client loads the schema and queries.
- Rather than custom field names being a single object associated with a table, a concept of "transformers"? could be introduced.
  - One way to think of them may simply just be **named** `custom_ root_field/column_name` configurations. However, they could possibly be far more extensible.
  - This sort of functionality is found in some gql clients already. I was looking use it as a solution to my problem here initially. It got too messy when needing also to do it consistently for codegen too...
  - I believe other people deal with this by placing a transform "proxy" of sorts in front of hasura. Although, for hasura to be the main GraphQL joiner of remote schemas that it seems to want to be, I bet some form of _graph transformation_ functionality will eventually be required!

rakeshkky:
### Description

Do not throw an error if a custom column name is identical to the Postgres name of the column.

`track_table` v2 API throws an error if the request is
```
{
  "type": "track_table",
  "version": 2,
  "args": {
    "table": "users",
    "configuration": {
        "custom_column_names": {
          "created_at": "createdAt",
          "id": "id",
          "email": "email",
          "verified_at": "verifiedAt",
        }
    }
  }
}
```
Error: `"column/relationship "id" of table "users" already exists"`

This PR improves the logic of validating custom column names thus fixes the above. Also, this PR improves the check of duplicate root fields generated for all tables. 

### Affected components 

- Server
- Tests

### Related Issues

Fix a part of #3137 

### Solution and Design

- Modify logic of validating custom column names. Create unified Hashmap of column and custom name and check for duplicate custom names. 
- Validate custom column names against existing not column fields for `set_table_custom_fields` query type.
- Store root fields in `GCtx` as lists instead of hashmaps. Check for duplicate root field names using the lists.

### Steps to test and verify

Try defining the `custom_column_names` with identical fields. The server shouldn't throw an error.


server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -180,11 +183,26 @@ instance FromJSON SetTableCustomFields where
...
   adminOnly
-  void $ askTabInfo tableName
+  fieldInfoMap <- _tiFieldInfoMap <$> askTabInfo tableName
+  validateWithExistingRelationships fieldInfoMap
   let tableConfig = TableConfig rootFields columnNames
   updateTableConfig tableName tableConfig
   buildSchemaCacheFor (MOTable tableName)
   return successMsg
+  where
+    validateWithExistingRelationships fields = do
```
lexi-lambda: I don’t think I totally understand all the subtleties of this issue. Could you explain why `validateWithExistingRelationships` is different from/not a part of `validateTableConfig`?

rakeshkky: Sure. `validateTableConfig` happens when building `TableInfo`. At that time no relationships were added. `set_table_custom_fields` is applied for only tracked tables. So we should validate with existing relationships.

rakeshkky: Well, I also stuck with the thought while writing the above reply. What about validating the new relationships being added with existing custom column names? We should do that definitely. Will push a commit with that change soon.

rakeshkky: Putting my all thoughts below:

I have a simple schema

```plpgsql
CREATE TABLE "user"(
  id SERIAL PRIMARY KEY,
  name TEXT NOT NULL,
  age INTEGER
)
```
I tracked the above table with the following custom column names configuration.
```json
{
    "id": "userId",
    "name": "userName"
}
```
What could be the best way to validate a relationship with name `X` that is being defined? Currently, we only check with existing column names and relationships. We should definitely validate with custom column names (`userId` and `userName`). 
Also, I'm thinking about new class fields coming soon (computed fields and remote joins). I think we should improve the logic of function `checkForFieldConflict` and should change `validateWithExistingRelationships` to something like `validateWithFieldsOtherThanColumns`. 

I'll answer all these in the next commit.

lexi-lambda: > Sure. `validateTableConfig` happens when initialising `TableInfo` of a table. At that time no relationships were added. Hence, we cannot do validation with relationships.

Got it, that makes sense! As a followup question: do you think we could consolidate this logic somehow? I know that each code path has slightly different inputs, and you have a firmer grasp on the details than I do, but it would be nice if we could share the logic between `validateTableConfig`, `validateWithNonColumnFields`, and `checkForFieldConflict` somehow so that it can’t get out of sync.

rakeshkky: Prior to [this](https://github.com/hasura/graphql-engine/commit/ed26da59a6d00b34e67707b754a9622848064e75) commit the server used to build the `TableInfo` on `track_table` query and add it to the cache later. Now `buildSchemaCacheFor` is used to construct `TableInfo` from scratch and add update it with other table metadata objects. Here we don't have luxury of having `TableConfig` before we call `buildSchemaCacheFor` to validate the input custom fields configuration. This forces the validation of custom fields configuration with existing column names to occur only in `buildSchemaCacheFor` code path.

In the `set_table_custom_fields` query API, `buildSchemaCacheFor` is being used to rebuild the `TableConfig`. Here, the validation for custom fields occur in two steps
1. With existing un-updated `TableInfo`, validate the custom column fields with existing non-columns (relationships, computed fields as of now).
2. We can validate the input custom column fields with existing columns here. But, it doesn't make sense since the validation of custom column fields occurs in `buildSchemaCacheFor` code path.

After a bit of thinking,
1. Scrap `validateTableConfig` function and break it into two isolated logics; `validateCustomRootFields` and `validateWithColumns`.
2. Use `validateCustomRootFields` directly in `trackExistingTableOrViewP2` and `runSetTableCustomFields`
3. Use `validateWithColumns` only in `buildTableCache`

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -98,41 +98,29 @@ trackExistingTableOrViewP1 qt = do
...
-      forM_ (M.toList colFlds) $ \(col, customName) -> do
-        void $ askPGColInfo (_tiFieldInfoMap tableInfo) col ""
-        withPathK (getPGColTxt col) $
-          checkForFieldConflict tableInfo $ FieldName $ G.unName customName
-        when (not $ null duplicateNames) $ throw400 NotSupported $
-          "the following names are duplicated: " <> showNames duplicateNames
-  where
-    duplicateNames = duplicates $ M.elems colFlds
+  => GC.TableCustomRootFields -> m ()
+validateCustomRootFields customRootFields = do
```
lexi-lambda: In the vein of “parse, don’t validate,” is there some way we could make this return something other than `()`? 🙂 Maybe it could build a map from root field names to something? I don’t know if that makes sense, but just throwing it out there.

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -766,19 +773,19 @@ instance Monoid TyAgg where
 -- | A role-specific mapping from root field names to allowed operations.
 data RootFields
   = RootFields
-  { rootQueryFields    :: !(Map.HashMap G.Name (QueryCtx, ObjFldInfo))
-  , rootMutationFields :: !(Map.HashMap G.Name (MutationCtx, ObjFldInfo))
+  { rootQueryFields    :: ![(QueryCtx, ObjFldInfo)]
+  , rootMutationFields :: ![(MutationCtx, ObjFldInfo)]
```
lexi-lambda: Related to the above comment: maybe we could keep these as `HashMap`s, and we could make the validation function produce them from the lists? A la the example at the end of my blog post.

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
rakeshkky: There is no type that `TableCustomRootFields` resolve to. Hence, this function cannot return other than `()`. But partly we can achieve "parse, don't validate," :). This function is performing two tasks:

1. check for duplicate custom root field names within `TableCustomRootFields` type.
    Happily, we can move this part to `FromJSON` deriving of the type.
2. check for any conflicting root field in the remote schema.   
    This is something we can't avoid right now.

lexi-lambda: Sorry for taking so long to respond to this—I’ve had a few too many things going on, and in order to understand all the details here, I needed to actually take the time to checkout this branch and look at the code.

I’ve now done that, and I see what you mean about it not being able to return anything more meaningful. However, that’s actually kind of why I added the `buildSchemaCacheFor` function in the first place—my idea was that `buildSchemaCache` could be where all the “canonical” parsing/validation logic happens, and we could just optimistically assume the data is correct, update the catalog, and run `buildSchemaCache` to see if it builds successfully. If it doesn’t, the process will fail, and the database transaction will get rolled back.

I don’t love that solution, since it’s fairly indirect, and it duplicates a lot of work. But it does have *some* things going for it. Here was my reasoning:

  - The idea of “parse, don’t validate” is that we want to use the type system to ensure the necessary parsing happens as a necessary consequence of the data processing pipeline.

  - The problem is that part of our processing pipeline goes out into Postgres, then back into graphql-engine, which is tricky, since obviously the Haskell type system cannot guarantee anything about what goes on inside Postgres.

  - However, at least with Postgres we get database transactions, so we can always muck up the state of the catalog as much as we want as long as we always ensure it is consistent before we eventually *commit*.

  - Since `buildSchemaCache` fundamentally *has* to check integrity of the catalog (since it has to build a `SchemaCache` value), it makes sense as a canonical code path for ensuring invariants. Therefore, if we can successfully build the schema cache, the catalog must be well-formed, right?

There are, however, some problems with this approach, as you’ve probably seen for yourself. For one, it’s slow. For another, it can make doing error reporting tricky, since the error that gets generated might not have anything to do with the thing we’re changing. The latter issue was what `buildSchemaCacheFor` was supposed to be about—it’s a way of specifying that the schema cache is being rebuilt with the intent to check the integrity of a particular object.

The former issue is more complicated to deal with (though I don’t think it’s impossible), but it’s not actually a huge issue for this change, since `trackExistingTableOrViewP2` already calls `buildSchemaCacheFor`, anyway. Therefore, we can’t be slowing it down any more by taking that code path, right?

Therefore, what if we skipped the call to `validateCustomRootFields` entirely and did the integrity-checking via the canonical `buildSchemaCacheFor` code path? Is that viable, do you think?

rakeshkky: I want to give you a bit of history :). Initially, the GraphQL Engine has only Postgres tables with relationships and permissions. Later on, in many iterations, it has evolved to what it is now with remote schemas, event triggers, computed fields etc. If we see, the `buildSchemaCache` function is quite sequential. The sequence partially (except computed fields) implies the evolution of the GraphQL Engine (interesting!). The `TableInfo` for all tables are built first, relationships are added next, then computed fields, permissions etc. The validation checks are inter-dependant. For example, there shouldn't be a computed field with a relationship name or vice versa. The permissions are defined only after all "fields" for a table are resolved. Any change in the sequence may result in permanent inconsistency (a potential bug). Well, the `buildSchemaCacheFor` is useful in `trackExistingTableOrViewP2` since the server is defining the metadata of the table for the first time. But, the validation of its custom root fields configuration dependant on the existing remote schemas. Here, building the `TableInfo` dependant on the remote schemas which are added much later in the `buildSchemaCache`. Hence, It is hard to achieve `validateCustomRootFields` in `buildSchemaCacheFor`.  So I'm calling it well before `buildSchemaCacheFor` to validate with existing remote schemas and optimistic about the remote schemas are not being altered in the current code path.  

The same happening with `set_table_custom_fields` metadata API. We've to validate the custom column names with existing fields other than columns, which are added to the `TableInfo` later in the `buildSchemaCache`. Also, we should definitely do `validateCustomRootFields` before `buildSchemaCacheFor`.

Apart from two problems you mentioned with `buildSchemaCacheFor`, the third problem aforementioned (with the peculiar sequential dependency of `buildSchemaCache`) in the above two paragraphs is making the function, `buildSchemaCacheFor`, not a one-stop solution in achieving "parse, don't validate". We should definitely do `validate :: m ()` in the metadata APIs which rely existing metadata objects which are being added later in the `buildSchemaCache`.

lexi-lambda: > The validation checks are inter-dependant. For example, there shouldn't be a computed field with a relationship name or vice versa. The permissions are defined only after all "fields" for a table are resolved. Any change in the sequence may result in permanent inconsistency (a potential bug).

This makes sense, but surely `buildSchemaCache` *must* validate *everything* (including custom column names!), or else `buildSchemaCache` is buggy, right? If the metadata were to end up in a state where the custom column name configuration is invalid, then `buildSchemaCache` *must* to fail. To me, this implies that, regardless of the order in which it does the validation, eventually it *will* get validated.

Perhaps what you’re telling me, however, is that isn’t the case? If I were to, with this PR, start up an instance of graphql-engine against a database where a custom column name configuration is somehow invalid, what would happen? Would it report an error on startup about an invalid custom column name configuration, or would it report some arcane internal error message about something else entirely (or worse, not report an error at all until something else went wrong during execution)?

I do understand that some validation (in the `m ()` sense) must happen somewhere, and it’s highly unlikely that we can encode *all* of our invariants into the type system, but I don’t understand why that validation can’t happen despite the sequential dependencies inside `buildSchemaCache`. In the absolute worst-case scenario, we add some validation checks at the end of `buildSchemaCache`, after the whole schema cache has been built, and we have all the information we need.

(As a related aside, I have started working on a fix for #3354 today, which involves making the performance of `buildSchemaCache` better, so I’ve been implementing the “automatic incrementalization” idea I’ve had for a little while now. That should (eventually) make the performance of `buildSchemaCache` much better, since it can avoid rebuilding everything every time it is called. I don’t think the performance problem of `buildSchemaCache` is fundamental, and I think there are significant advantages to consolidating the logic around the catalog!)

rakeshkky: > In the absolute worst-case scenario, we add some validation checks at the end of buildSchemaCache, after the whole schema cache has been built, and we have all the information we need.

This seems good to me. For now, I'll move validation of custom configuration of tables which dependant on the objects added later to end of `buildSchemaCache`.

PR #3169, opened 2019-10-17
Fix potentially inconsistent state update

pedroprado010:
### Description

`this.state` may asynchronously update. 

Using callback as parameter  will  calculate the  state when the update is about to be performed


shahidhk:
@pedroprado010 Thank you for the PR, but we have moved the project to a new repo: https://github.com/hasura/graphiql-online

Would you mind making the same PR over there?

PR #3174, opened 2019-10-18
(wip) read only run sql api

tirumaraiselvan:
### Description

Add a `read_only` parameter to `run_sql` metadata API. This sets the postgres transaction mode to "ReadOnly". This can be used by the console, etc to perform select operations without forcing cache/gctx recomputation.

### Affected components 

- Server

### Solution and Design

This is just an attempt. Opening for review.


server/src-lib/Hasura/Db.hs
```diff
@@ -56,28 +58,28 @@ instance (MonadTx m) => MonadTx (ValidateT e m) where
...
+  | LTTx !(Q.TxE e a) !(Maybe Q.TxAccess)
 
 lazyTxToQTx :: LazyTx e a -> Q.TxE e a
 lazyTxToQTx = \case
   LTErr e  -> throwError e
   LTNoTx r -> return r
-  LTTx tx  -> tx
+  LTTx tx _txaM -> tx
 
 runLazyTx
```
tirumaraiselvan: The other idea (and simpler IMO) was to provide the access mode as a parameter to `runLazyTx` but metadata queries are all peeled outside the main handlers and hence don't use `runLazyTx` at all, if I understand correctly. See: https://github.com/tirumaraiselvan/graphql-engine/blob/e6c83ac646ed16a31667a78a3ba98bccf4e7a70e/server/src-lib/Hasura/Server/Query.hs#L201

server/src-lib/Hasura/Db.hs
```diff
@@ -56,28 +58,28 @@ instance (MonadTx m) => MonadTx (ValidateT e m) where
...
+  LTTx tx txaM -> Q.runTx pgPool (txIso, txaM) tx
 
 runLazyTx'
   :: PGExecCtx -> LazyTx QErr a -> ExceptT QErr IO a
-runLazyTx' (PGExecCtx pgPool _) = \case
+runLazyTx' (PGExecCtx pgPool txIso) = \case
   LTErr e  -> throwError e
   LTNoTx a -> return a
-  LTTx tx  -> Q.runTx' pgPool tx
+  LTTx tx txaM  -> Q.runTx pgPool (txIso, txaM) tx
```
rakeshkky: We should be using `Q.runTx'` here?
`runLazyTx'` are non-transactional queries (Without `BEGIN`/`COMMIT`/`ABORT`). It's being used to run subscription queries.

server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -58,18 +58,20 @@ data RunSQL
...
   , rCheckMetadataConsistency :: !(Maybe Bool)
+  , rReadOnly                 :: !(Maybe Bool)
   } deriving (Show, Eq, Lift)
 $(deriveJSON (aesonDrop 1 snakeCase){omitNothingFields=True} ''RunSQL)
 
 runRunSQL :: (CacheBuildM m, UserInfoM m) => RunSQL -> m EncJSON
-runRunSQL (RunSQL t cascade mChkMDCnstcy) = do
+runRunSQL (RunSQL t cascade mChkMDCnstcy mReadOnly) = do
   adminOnly
+  let txAccessMode = fmap boolToAccessMode mReadOnly
```
rakeshkky: If `read_only` mode is `true` we shouldn't check for `isAltDropReplace`. AFAIK, Postgres doesn't allow to execute DDL statements in `READY ONLY` mode transaction. It is safe to directly execute SQL without metadata check. Postgres will throw exceptions if the query contains any DDL statements.

`withMetadataCheck` function shouldn't account for `txAccessMode` :wink: . It is always `Q.ReadWrite`.

tirumaraiselvan: Oh yeah. I didn't want too much diff here so haven't touched this bit.

server/src-lib/Hasura/Db.hs
tirumaraiselvan: Ah interesting. Another break.

server/src-lib/Hasura/Db.hs
```diff
@@ -36,14 +36,16 @@ data PGExecCtx
   }
 
 class (MonadError QErr m) => MonadTx m where
+  liftTxWith :: Q.TxE QErr a -> Maybe Q.TxAccess -> m a
```
lexi-lambda: I think this probably isn’t the right thing to do here since `MonadTx` describes a class of computations that are already in a transaction, so `liftTxWith` can’t do anything meaningful if the transaction has already begun. However, I think Postgres actually allows changing a transaction’s access mode in the middle of a transaction via [`SET TRANSACTION READ ONLY`](https://www.postgresql.org/docs/12/sql-set-transaction.html). Therefore, `run_sql` can probably just execute that directly to switch into read-only mode, then switch back to read/write mode afterwards.

tirumaraiselvan:
closing this for better alternative: https://github.com/hasura/graphql-engine/pull/3174#discussion_r336716568

PR #3191, opened 2019-10-21
add read_only to run_sql metadata api

tirumaraiselvan:
### Description

Add a `read_only` parameter to `run_sql` metadata API. This sets the postgres transaction mode to "ReadOnly". This can be used by the console, etc to perform select operations without forcing cache/gctx recomputation.

### Affected components 

- Server

### Solution and Design

Change API to consider transaction mode while executing.


server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -58,22 +61,26 @@ data RunSQL
...
     execRawSQL =
-      fmap (encJFromJValue @RunSQLRes) . liftTx . Q.multiQE rawSqlErrHandler . Q.fromText
+      fmap (encJFromJValue @RunSQLRes) . liftTx . setTransactionAccess . Q.multiQE rawSqlErrHandler . Q.fromText
       where
         rawSqlErrHandler txe =
           let e = err400 PostgresError "query execution failed"
           in e {qeInternal = Just $ toJSON txe}
+    setTransactionAccess tx = if isReadOnly then setReadOnly >> tx else tx
+
+    setReadOnly =  Q.unitQE defaultTxErrorHandler "SET TRANSACTION READ ONLY" () False
```
tirumaraiselvan: Does this override the isolation level by any chance? 

For e.g. if server is initialized with isolation `SERIALIZABLE` but if we `SET TRANSACTION READ ONLY` in a transactio, does that "downgrade" the transaction to default isolation i.e. `REPEATABLE READ`?

tirumaraiselvan: > Does this override the isolation level by any chance?
Turns out it doesn't affect the isolation level.

lexi-lambda: One issue I foresee with this is using read-only `run_sql` queries in `bulk` queries, since we run entire `bulk` queries in one transaction, so setting the transaction to read-only will cause problems for other queries that come later if they made modifications.

Personally, I think it would be nice to add some sort of scoped `withAccessMode` abstraction somewhere in  `Hasura/SQL/` with a type like

```haskell
withAccessMode :: (MonadTx m) => TxAccess -> m a -> m a
```

that gets the current access mode, sets it if necessary, and then resets it to the old value when control leaves the function. (Using some form of lifted `bracket` operation is probably the right thing to implement `withAccessMode`, which would additionally require something like `MonadMask` or `MonadBaseControl`.)

server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -58,22 +61,26 @@ data RunSQL
...
   where
     execRawSQL :: (MonadTx m) => Text -> m EncJSON
     execRawSQL =
-      fmap (encJFromJValue @RunSQLRes) . liftTx . Q.multiQE rawSqlErrHandler . Q.fromText
+      fmap (encJFromJValue @RunSQLRes) . liftTx . setTransactionAccess . Q.multiQE rawSqlErrHandler . Q.fromText
       where
         rawSqlErrHandler txe =
           let e = err400 PostgresError "query execution failed"
           in e {qeInternal = Just $ toJSON txe}
+    setTransactionAccess tx = if isReadOnly then setReadOnly >> tx else tx
```
lexi-lambda: I think this would be clearer as `when isReadOnly setReadOnly` followed by `tx`.

server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -58,22 +61,26 @@ data RunSQL
...
 $(deriveJSON (aesonDrop 1 snakeCase){omitNothingFields=True} ''RunSQL)
 
 runRunSQL :: (CacheBuildM m, UserInfoM m) => RunSQL -> m EncJSON
-runRunSQL (RunSQL t cascade mChkMDCnstcy) = do
+runRunSQL (RunSQL sql cascade mChkMDCnstcy (fromMaybe False -> isReadOnly)) = do
   adminOnly
-  isMDChkNeeded <- maybe (isAltrDropReplace t) return mChkMDCnstcy
-  bool (execRawSQL t) (withMetadataCheck (or cascade) $ execRawSQL t) isMDChkNeeded
+  isMDChkNeeded <- maybe (if isReadOnly then pure False else isAltrDropReplace sql) return mChkMDCnstcy
+  bool (execRawSQL sql) (withMetadataCheck (or cascade) $ execRawSQL sql) isMDChkNeeded
```
lexi-lambda: Maybe it’s just me, but I feel like all the different ways of handling boolean logic here are getting a little confusing. Personally, I think a lot of this code would read a lot more clearly if we wrote a manual `FromJSON RunSQL` instance that handles the default values there instead of having `Maybe`s. Then maybe we could avoid so much of the twisty logic in this function and just do something like this:

```haskell
isMetadataCheckNeeded <- maybeCheckMetadataConsistency `onNothing` if
  | isReadOnly -> pure False
  | otherwise  -> isAlterDropReplace sql
```

But then I start wondering: is the current logic even right? Does it make sense to not check metadata consistency if the user specifies `check_metadata_consistency: false` in the query, but they also specify `cascade: true`? That seems nonsensical to me, but maybe I’m missing something.

server/src-lib/Hasura/RQL/DDL/Schema.hs
tirumaraiselvan: ah right! Will incorporate this.

server/src-lib/Hasura/RQL/DDL/Schema.hs
tirumaraiselvan: what do you think if we use `ViewPatterns` to peel the maybe with default? (Just like mReadOnly). Writing `(fromMaybe False -> isReadOnly)` is not very clean but we can wrap that like `(setDefaultFalse -> isReadOnly)` for more cleanliness.

lexi-lambda: I’d prefer doing it in the `FromJSON` instance so that we normalize things at the same time we parse them; it’s really a parsing concern and it avoids the kind of mistake where the logic is duplicated in two different places in different ways (admittedly pretty unlikely here, but a good principle to follow in general, imo).

server/src-lib/Hasura/RQL/DDL/Schema.hs
tirumaraiselvan: TBH, I have no idea how to unset a transaction. Do you mean we query for the previous transaction access mode first, store it and then set it back after the run?

tirumaraiselvan: How about we set it right back to `READ WRITE` after we are done with this query?

server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -56,24 +59,42 @@ import           Hasura.Server.Utils            (matchRegex)
...
 
 runRunSQL :: (CacheBuildM m, UserInfoM m) => RunSQL -> m EncJSON
-runRunSQL (RunSQL t cascade mChkMDCnstcy) = do
+runRunSQL RunSQL{..} = do
   adminOnly
-  isMDChkNeeded <- maybe (isAltrDropReplace t) return mChkMDCnstcy
-  bool (execRawSQL t) (withMetadataCheck (or cascade) $ execRawSQL t) isMDChkNeeded
+  metadataCheckRequired <- case rReadOnly of
+    True -> pure False
+    False -> onNothing rCheckMetadataConsistency $ isAltrDropReplace rSql
```
tirumaraiselvan: could not set default for `rCheckMetadataConsistency` during parse because if the user explicitly says `false` then doesn't make sense to override that.

server/src-lib/Hasura/RQL/DDL/Schema.hs
lexi-lambda: I like the `bracket`-like abstraction because it’s less stateful (it just parameterizes a region with dynamic extent, a la environment variables), but you might be right that in this case we really only want to set the access mode for a whole query.

Do you know if there’s any performance cost to issuing `SET TRANSACTION READ WRITE` in a query that’s already in read/write mode? (I feel like there must not be, since it doesn’t actually do anything, but I’m always a little wary of anything that modifies the transaction state.)

tirumaraiselvan: There is some cost to `SET TRANSACTION READ WRITE` even if it's already in `READ WRITE` mode but I think its practically nothing especially for a metadata API.

I ran a quick test using this function once with `SET TRANSACTION READ WRITE` and next without it

```
CREATE FUNCTION check_transaction(batchsize INTEGER)
RETURNS INTEGER AS $$
DECLARE
   counter INTEGER := 0 ; 
BEGIN
  LOOP
     EXIT WHEN counter = batchsize ;
     counter := counter + 1;
     SET TRANSACTION READ WRITE; // toggle this
  END LOOP;
RETURN counter;
END;
$$  LANGUAGE plpgsql;
```

Results on Heroku hobby-dev:

```
dbuiv0sje6a6v1=> explain analyze select check_transaction(1000);
                                     QUERY PLAN                                     
------------------------------------------------------------------------------------
 Result  (cost=0.00..0.05 rows=1 width=4) (actual time=4.117..4.117 rows=1 loops=1)
 Planning Time: 0.028 ms
 Execution Time: 4.184 ms
dbuiv0sje6a6v1=> explain analyze select check_transaction2(1000);
                                     QUERY PLAN                                     
------------------------------------------------------------------------------------
 Result  (cost=0.00..0.05 rows=1 width=4) (actual time=3.917..3.918 rows=1 loops=1)
 Planning Time: 0.014 ms
 Execution Time: 3.932 ms
(3 rows)

dbuiv0sje6a6v1=> explain analyze select check_transaction(10000);
                                      QUERY PLAN                                      
--------------------------------------------------------------------------------------
 Result  (cost=0.00..0.05 rows=1 width=4) (actual time=49.254..49.254 rows=1 loops=1)
 Planning Time: 0.010 ms
 Execution Time: 49.266 ms
(3 rows)

dbuiv0sje6a6v1=> explain analyze select check_transaction2(10000);
                                       QUERY PLAN                                       
----------------------------------------------------------------------------------------
 Result  (cost=0.00..0.05 rows=1 width=4) (actual time=107.129..107.129 rows=1 loops=1)
 Planning Time: 0.014 ms
 Execution Time: 107.196 ms
(3 rows)

dbuiv0sje6a6v1=> explain analyze select check_transaction(100000);
                                       QUERY PLAN                                       
----------------------------------------------------------------------------------------
 Result  (cost=0.00..0.05 rows=1 width=4) (actual time=508.105..508.106 rows=1 loops=1)
 Planning Time: 0.035 ms
 Execution Time: 508.121 ms
(3 rows)

dbuiv0sje6a6v1=> explain analyze select check_transaction2(100000);
                                       QUERY PLAN                                       
----------------------------------------------------------------------------------------
 Result  (cost=0.00..0.05 rows=1 width=4) (actual time=773.483..773.483 rows=1 loops=1)
 Planning Time: 0.013 ms
 Execution Time: 773.500 ms
(3 rows)

Comment (lexi-lambda) [reply server/src-lib/Hasura/RQL/DDL/Schema.hs]: Sounds good, thanks for checking!

File: server/src-lib/Hasura/RQL/DDL/Schema.hs [hunk 5]
```diff
@@ -56,25 +59,46 @@ import           Hasura.Server.Utils            (matchRegex)
...
-      fmap (encJFromJValue @RunSQLRes) . liftTx . Q.multiQE rawSqlErrHandler . Q.fromText
+      fmap (encJFromJValue @RunSQLRes) . liftTx . setTransactionAccess . Q.multiQE rawSqlErrHandler . Q.fromText
       where
         rawSqlErrHandler txe =
           let e = err400 PostgresError "query execution failed"
           in e {qeInternal = Just $ toJSON txe}
 
+    setTransactionAccess tx
+      | rReadOnly = setReadOnly >> tx
+      | otherwise = setReadWrite >> tx
```
lexi-lambda: This seems insufficient, right? Since if we have a `bulk` query with a read-only `run_sql` query followed by a non-`run_sql` query that makes modifications, the transaction access state won’t be restored to read/write.

tirumaraiselvan: ufff my bad! So I will reset to `READ WRITE` at the end again?

```
    setTransactionAccess tx
      | rReadOnly = setReadOnly >> tx *> setReadWrite
      | otherwise =  tx
```

Now, I can see why you might need a nice pattern here. I don't know much about this bracket-thing you mentioned. Do you want me to look into it?

lexi-lambda: [`Control.Exception.bracket_`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Exception.html#v:bracket_) is a function that implements an exception-safe region with setup/teardown actions, kind of like `try...finally` in other languages (but Haskell is a little more subtle here than other languages because Haskell has asynchronous exceptions, so it’s important to use `bracket_` to deal with everything properly). To implement this, you’d do something like

```haskell
withTransactionAccess newTxAccess m = do
  oldTxAccess <- getCurrentTxAccess
  bracket_ (setTxAccess newTxAccess) (setTxAccess oldTxAccess) m
```

The problem with `bracket_` is that it’s a non-algebraic operation, i.e. it has `IO`-typed arguments as well as an `IO` result type, so it can’t be easily lifted using `liftIO`. My fancy new [`eff`](https://github.com/hasura/eff) library has support for a safe version of `bracket_` that is not coupled to `IO`, but in the meantime, you’ll probably want to use [the version from `Control.Monad.Catch`](https://hackage.haskell.org/package/exceptions-0.10.3/docs/Control-Monad-Catch.html#v:bracket_), which provides `bracket_` via the mtl-style typeclass `MonadMask`.

tirumaraiselvan: I thought a bit about this and TBH, I am not sure how to implement this or where. Don't we need something like a Monad State to keep track of the current transaction access mode?

Either way, not sure if all these changes are worth it atm. What do you think of the current hack?

server/src-lib/Hasura/Db.hs
```diff
@@ -66,11 +66,12 @@ lazyTxToQTx = \case
 
 runLazyTx
   :: PGExecCtx
+  -> Maybe Q.TxAccess
```
lexi-lambda: I get that the underlying API allows `Nothing`, which presumably means “use the value of the `default_transaction_read_only` configuration parameter, but it seems like it would be clearer to accept a `Q.TxAccess` directly here and just pass `Q.ReadWrite` instead of `Nothing`. That seems more correct anyway, since the places where we use `Nothing` currently need `Q.ReadWrite` access.

server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -56,24 +59,35 @@ import           Hasura.Server.Utils            (matchRegex)
 data RunSQL
   = RunSQL
   { rSql                      :: Text
-  , rCascade                  :: !(Maybe Bool)
+  , rCascade                  :: !Bool
   , rCheckMetadataConsistency :: !(Maybe Bool)
+  , rReadOnly                 :: !Bool
```
lexi-lambda: Since we have the custom `FromJSON` instance now, anyway, why not make the type of this field `Q.TxAccess` instead of `Bool`? 😄 Or if you’d rather not need to write the `ToJSON` instance by hand, you could add a `newtype IsReadOnly = IsReadOnly (Q.TxAccess)` with custom `ToJSON`/`FromJSON` instances, like we already have with the `SystemDefined` type.

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -369,7 +381,15 @@ runQueryM rq =
...
-      RQBulk qs                    -> encJFromList <$> indexedMapM runQueryM qs
+      RQBulk qs                    -> unless (allSameTxAccess qs) throwTxMismatch >>
+                                      encJFromList <$> indexedMapM  runQueryM qs
+      where
+        allSameTxAccess qs = allReadWrite qs || allReadOnly qs
+        allReadWrite = all isReadWrite
+        allReadOnly = all isReadOnly
+        isReadWrite q = getQueryAccessMode q == Just Q.ReadWrite || getQueryAccessMode q == Nothing
+        isReadOnly q = getQueryAccessMode q == Just Q.ReadOnly
+        throwTxMismatch = throw400 BadRequest "bulk query has transaction access mode mismatch"
```
lexi-lambda: How hard would it be to provide slightly more information in this error message? I’m envisioning something like:

```
incompatible access mode requirements in bulk query:
  run_sql at $.query[0] requires read-only access
  track_table at $.query[1] requires read-write access
```

It wouldn’t have to list all the contradictions, just the first one.

server/src-test/Main.hs
```diff
@@ -39,7 +39,7 @@ main = do
       pgContext = PGExecCtx pgPool Q.Serializable
 
       runAsAdmin :: Run a -> IO (Either QErr a)
-      runAsAdmin = runExceptT . fmap fst . peelRun emptySchemaCache runContext pgContext
+      runAsAdmin = runExceptT . fmap fst . peelRun emptySchemaCache runContext pgContext Q.ReadWrite
```
tirumaraiselvan: Not sure what is this file used for!

lexi-lambda: It’s a regression test for #3165.

server/src-lib/Hasura/RQL/DDL/Schema.hs
lexi-lambda: I think this comment is still relevant?

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -290,6 +295,17 @@ queryNeedsReload (RQV2 qi) = case qi of
   RQV2TrackTable _           -> True
   RQV2SetTableCustomFields _ -> True
 
+-- TODO: add RQSelect after console changes
```
lexi-lambda: What does this `TODO` comment mean?

tirumaraiselvan: The RQL select statement should also be(ideally) run in READ ONLY mode. But this could be part of the current console's bulk statement and hence would be breaking if we do it now itself.

server/src-lib/Hasura/RQL/DDL/Schema.hs
tirumaraiselvan: Oops. Missed it, yes.

tirumaraiselvan: I tried to make this change but it throws that `Lift` is not defined for `Q.TxAccess`, here is the patch for your reference: https://github.com/tirumaraiselvan/graphql-engine/pull/74/files

lexi-lambda: I’ve pushed https://github.com/hasura/pg-client-hs/commit/bd21e66d8197af381a6c0b493e22d611ed1fa386, which derives a `Lift` instance for `TxAccess`.

server/src-lib/Hasura/RQL/DDL/Schema.hs
```diff
@@ -55,25 +58,47 @@ import           Hasura.Server.Utils            (matchRegex)
...
+
+instance FromJSON RunSQL where
+  parseJSON (Object o) = do
+    rSql <- o .: "sql"
+    rCascade <- o .:? "cascade" .!= False
+    rCheckMetadataConsistency <- o .:? "check_metadata_consistency"
+    isReadOnly <- o .:? "read_only" .!= False
+    let rTxAccessMode = if isReadOnly then Q.ReadOnly else Q.ReadWrite
+    pure RunSQL{..}
+  parseJSON _ = fail "expecting an object for args"
```
lexi-lambda: This is what [`Data.Aeson.withObject`](https://hackage.haskell.org/package/aeson-1.4.6.0/docs/Data-Aeson.html#v:withObject) is for. :)

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -369,7 +384,18 @@ runQueryM rq =
...
-      RQBulk qs                    -> encJFromList <$> indexedMapM runQueryM qs
+      RQBulk qs                    -> assertAllTxAccess (zip [0::Integer ..] qs) >>
+                                      encJFromList <$> indexedMapM runQueryM qs
+      where
+        assertAllTxAccess = \case
+          [] -> pure ()
+          (_q:[]) -> pure ()
+          (q1:q2:qs) -> assertSameTxAccess q1 q2 >> assertAllTxAccess (q2:qs)
+
+        assertSameTxAccess (i1, q1) (i2, q2) = unless (getQueryAccessMode q1 == getQueryAccessMode q2) $
```
lexi-lambda: This could be altered to be more of a “parser” than a “validator” (using the terminology from [Parse, Don’t Validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)) by combining it with `getQueryAccessMode` and giving that function this type:

```haskell
getQueryAccessMode :: (MonadError QErr m) => RQLQuery -> m Q.TxAccess
```

That way, it’s impossible (or at least much harder) to accidentally omit the consistency check.

Issue #2520, opened 2019-07-12
array columns and relationships in returning don't play well together


0x777:
Schema:
```sql
CREATE TABLE public.author (
    id integer primary key,
    name text NOT NULL
);
insert into public.author values (1, 'foo');
CREATE TABLE public.article (
    id integer primary key,
    title text,
    tags text[],
    author_id integer references author(id)
);
insert into public.article values (1, 'hello world', '{bar,baz}', 1);
```
Metadata:

```json
{
  "functions": [],
  "remote_schemas": [],
  "query_collections": [],
  "allowlist": [],
  "tables": [
    {
      "table": "author",
      "object_relationships": [],
      "array_relationships": [
        {
          "using": {
            "foreign_key_constraint_on": {
              "column": "author_id",
              "table": "article"
            }
          },
          "name": "articles",
          "comment": null
        }
      ],
      "insert_permissions": [],
      "select_permissions": [],
      "update_permissions": [],
      "delete_permissions": [],
      "event_triggers": []
    },
    {
      "table": "article",
      "object_relationships": [
        {
          "using": {
            "foreign_key_constraint_on": "author_id"
          },
          "name": "author",
          "comment": null
        }
      ],
      "array_relationships": [],
      "insert_permissions": [],
      "select_permissions": [],
      "update_permissions": [],
      "delete_permissions": [],
      "event_triggers": []
    }
  ],
  "query_templates": []
}
```
Now try any mutation on `article` table with relationships in `returning`. For example 
```graphql
mutation {
  update_article(where: {id: {_eq: 1}}, _set: {tags: "{tada}"}) {
    affected_rows
    returning {
      author {
        id
      }
    }
  }
}
```
This will fail with this response:
```json
{
  "errors": [
    {
      "extensions": {
        "path": "$",
        "code": "parse-failed"
      },
      "message": "A string is expected for type : _text"
    }
  ]
}
```
The reason is that when we are trying to parse the mutated rows we expect string representation of array types but we instead they are json encoded. So I'm not sure what we can do about this till https://github.com/hasura/graphql-engine/pull/2243 is landed.

0x777:
Fixed in `v1.1.0-beta.3`.

PR #2243, opened 2019-05-21
native support for postgres array types (close #348)

rakeshkky:
### Description

TODO:- [console] support new input types format (json) for functions. 

Allows following types of queries 
```
query q {
  bar: foo (where: { a : { _contained_in : [1,2,3,4,5,6,7,8]  }} ) {
    a
  }

  baz: foo (where: { a : { _contains : [1,2,3,4] }} ) {
    a
  } 
}
```
And the following mutations with array
```
mutation m {
  insert_foo ( objects: [ {a: [1,3,5,7,9] } ]){
    affected_rows
    returning {
      a 
      id
    }
  }
}
```

Support `_preprend` and `_append` operators in updating array columns

### Affected components 

- Server
- Tests
- Console

### Related Issues

#348 
#349

### Solution and Design

- `PGColType` is modified to support postgres types
- `PGColValue` is modified to parse json values according to new `PGColType`
-  `hdb_type_info` view is added to `hdb_catalog` schema to list all types with details
- catalog version bump

PR #3198, opened 2019-10-22
fix insert with nested returning clause fails alongside text[] column (fix #3148 & #2520)

### LINKED ISSUE #3148: Insert with nested returning clause fails alongside text[] column
Attempting to add a nested query inside the `returning` clause of an insert mutation fails when the table being inserted into has a `text[]` column. I've created a minimal reproduction at https://returning-bug-repro.herokuapp.com/. This bug is present on at least `v1.0.0-beta.2` and `v1.0.0-beta.6`.

Running an insert mutation without a nested returning clause succeeds. For example:
```
mutation MyMutation {
  insert_site(objects: {name: "My Site"}, on_conflict: {constraint: site_name_key, update_columns: name}) {
    returning {
      id
      name
    }
  }
}
```

However, when I add a nested query to the returning clause, the query fails. For example:
```
mutation MyMutation {
  insert_site(objects: {name: "My Site"}, on_conflict: {constraint: site_name_key, update_columns: name}) {
    returning {
      id
      name
      site_trials {
        id
      }
    }
  }
}
```
results in the error:
```
{
  "errors": [
    {
      "extensions": {
        "path": "$.selectionSet.insert_site.args.objects",
        "code": "parse-failed"
      },
      "message": "A string is expected for type : _text"
    }
  ]
}
```

The same query with a nested returning clause runs successfully when the `site` table does not include the `emails` column. The `email` column was added with the statement
```sql
alter table site add column emails text[] not null default '{}'::text[];
```

### LINKED ISSUE #3248: Enable permissions for computed fields

rakeshkky:
### Description

If a table has a column with the array type then insert mutation to that table with returning its relationships throws an unexpected error.  
As given in #3148:
For the following mutation:
```graphql
mutation MyMutation {
  insert_site(objects: {name: "My Site"}, on_conflict: {constraint: site_name_key, update_columns: name}) {
    returning {
      id
      name
      site_trials {
        id
      }
    }
  }
}
```
Gives the following unexpected error:
```json
{
  "errors": [
    {
      "extensions": {
        "path": "$.selectionSet.insert_site.args.objects",
        "code": "parse-failed"
      },
      "message": "A string is expected for type : _text"
    }
  ]
}
```
When relationship fields are found in `returning` fields, the server performs the mutation in two steps:
1. Mutate with returning all columns
2. Parse the returned column rows and query the `returning` fields.

Refer to commits https://github.com/hasura/graphql-engine/commit/5f274b552717d107553368fec440eb7137e830fb and https://github.com/hasura/graphql-engine/commit/5bafdce9a3edf7ebb2198a26b24d535134bfd458 to learn more about the above steps.

In step 2, parsing the array column values fails since our parser expects `String` representation of Postgres arrays. There are two ways to solve this problem
1. The PR https://github.com/hasura/graphql-engine/pull/2243 auto-fixes the problem since it has support for arrays. But seems like it is not going out any soon.
2. Perform the insert with returning the Postgres column values as Strings (except for geojson). It is kind of a hack and implemented in this PR. 

### Affected components 

- Server
- Tests

### Related Issues

Fix #3148 & #2520

### Solution and Design

As mentioned in the above description, the mutated column values are returned as text.

### Steps to test and verify

Follow the reproduction guide given in #3148 or #2520 and error shouldn't happen.


server/src-lib/Hasura/RQL/DML/Internal.hs
```diff
@@ -273,18 +273,19 @@ dmlTxErrorHandler = mkTxErrorHandler $ \case
...
       , S.SEUnsafe "4"  -- to print out crs
       ] Nothing
       `S.SETyAnn` S.jsonTypeAnn
-  | isScalarColumnWhere isBigNum colTy && strfyNum =
-      expn `S.SETyAnn` S.textTypeAnn
-  | otherwise = expn
+  | isScalarColumnWhere isBigNum colTy && strfyNum = asTextExpn
+  | otherwise = if asText then asTextExpn else expn
+  where
+    asTextExpn = expn `S.SETyAnn` S.textTypeAnn
```
lexi-lambda: I think it would probably be slightly clearer to write this:
```haskell
| asText || (isScalarColumnWhere isBigNum colTy && strfyNum) =
    expn `S.SETyAnn` S.textTypeAnn
| otherwise = expn
```

server/src-lib/Hasura/RQL/DML/Select/Types.hs
```diff
@@ -142,19 +142,34 @@ data ColOp
   , _colExp :: S.SQLExp
   } deriving (Show, Eq)
 
+data AnnColField
+  = AnnColField
+  { _acfInfo   :: !PGColumnInfo
+  , _acfAsText :: !Bool
```
lexi-lambda: The context in the PR description is super helpful, and it would be nice if some of it could make it into the code, since otherwise it’s kind of hard to tell what this field is for if you’re just reading the code. Could you maybe add a short blurb about what this is for, with a note like “see #3198 for more details” at the end? Just to keep everything tied together.

PR #3211, opened 2019-10-23
computed fields in export/import metadata, other fixes

### LINKED ISSUE #3232: After adding computed columns, GraphQL schema returns empty
This could easily be me not knowing how to create the computed column, or trying to create it in a way that is not supported.  
But after I created it the GraphQL schema is blank.  

Here are the steps to reproduce it.

```sql
CREATE TABLE public.calistings
(
    id uuid NOT NULL DEFAULT gen_random_uuid(),    
    location geometry(Point,4326)
)
```

```sql
CREATE FUNCTION calc_calisting_distance(calisting_row calistings,location json)
 RETURNS INTEGER AS $$
   SELECT ST_DISTANCE(calisting_row.location::geography, ST_GeomFromGeoJSON(location::text)::geography)::integer
 $$ LANGUAGE sql STABLE;
```

```javascript
{
    "type":"add_computed_field",
    "args":{
        "table":{
            "name":"calistings",
            "schema":"public"
        },
        "name":"distance",
        "definition":{
            "function":{
                "name":"calc_calisting_distance",
                "schema":"public"
            },
            "table_argument":"calisting_row"
        }
    }
}
```

After I create the computed column, using the query endpoint.  The GraphQL schema, from the intropsection query, is returned blank.  If I remove the computed column, the introspection query returns the schema.

Tried checking the logs in the graphql-engine but no error is shown.  I can see the introspection query hitting the server and responding with a 200 status OK.

I think the issue might be in the way I am creating the computed column.  I didn't see a way to pass the function arguments to the api.  Only the row name and type.

Thanks

### LINKED ISSUE #3290: computed fields should be persisted via metadata
Computed fields can be added via the metadata api but they are not exported or persisted via the metadata.json file.

This is critical for reproducible dockerized deployments.

rakeshkky:
### Description

- Support computed fields for each of item in `tables` in export/import metadata
- Add computed fields to cache before permissions, since permissions depend on computed fields availability
- Add (explicitly) computed fields' function input arguments scalars to GraphQL schema
- Remove the traces of `computedCol..` names in the codebase.

### Affected components 

- Server
- Tests

### Related Issues

Fix #3232, fix #3290

### Solution and Design

- Add `computed_fields` to each `table` in `export_metadata` and `replace_metadata` query API.
- In `buildSchemaCacheWithOptions` function, move computed fields related code before permissions.
- While generating the input object of `args` input field of computed fields, generate scalars for input argument fields and add them explicitly to scalar set.

### Steps to test and verify

- After adding computed fields to metadata, the `export_metadata` should comprise computed fields for a table
- For issue #3232, follow the reproduction guide and make sure the introspection query passes through.


server/src-lib/Hasura/RQL/DDL/Permission.hs
```diff
@@ -236,16 +235,7 @@ buildSelPermInfo role tabInfo sp = do
...
-    case _cfiReturnType computedFieldInfo of
-      CFRScalar _ -> pure ()
-      CFRSetofTable returnTable -> do
-        returnTableInfo <- askTabInfo returnTable
-        let function = _cffName $ _cfiFunction $ computedFieldInfo
-            errModifier e = "computed field " <> name <<> " executes function "
-                             <> function <<> " which returns set of table "
-                             <> returnTable <<> "; " <> e
-        void $ modifyErr errModifier $ askPermInfo returnTableInfo role PASelect
+    void $ askComputedFieldInfo fieldInfoMap name
```
rakeshkky: Why these checks are removed?

Defining select permission of a table depends upon the existence of select permission on another table. It complicates the adding permissions to metadata in `replace_metadata` query API. The order of permissions in query payload matters now. The `replace_metadata` should pass whatever the order the metadata items are given. Hence the check for select permission are being removed.

While generating the schema for a role, the select permission for computed field returning table is taken into account and omitted if select permission is not defined. However, this logic is been there.

lexi-lambda: > The order of permissions in query payload matters now.

Does this have any potential to make `export_metadata` + `replace_metadata` non-idempotent? That is, do we take care to export the metadata in the right order so that it will never cause any problems when passed directly to `replace_metadata`?

rakeshkky: > Does this have any potential to make export_metadata + replace_metadata non-idempotent?

Yes. 

> do we take care to export the metadata in the right order so that it will never cause any problems when passed directly to replace_metadata?

No, neither the export metadata nor replace metadata should care about the order. However, in https://github.com/hasura/graphql-engine/pull/3230 we’re making the order deterministic to help CLI to compare and show “diff” of exported metadata.

lexi-lambda: > No, neither the export metadata nor replace metadata should care about the order.

I’m confused. You said

> The order of permissions in query payload matters now. The `replace_metadata` should pass whatever the order the metadata items are given.

which sounds like `replace_metadata` caring about the order to me… but maybe I misunderstood your comment. I interpreted it to mean “we used to do some complicated check to ensure the consistency between a recursive group of things, but now we do a simpler check by assuming we’re given things in the right order so that we can check them top to bottom.” Was my interpretation wrong?

rakeshkky: Apologies for the unclear comment. 

> The order of permissions in query payload matters now. 

I mean the above sentence if checking for select permissions on returning table. I updated the comment. Hope this is clear to you.

rakeshkky: Trying to remove all confusions.

We used to check for select permissions on returning tables of computed fields. This complicates the `replace_metadata` logic since the addition of select permission depends on the existence of another select permission. Thus the `replace_metadata` expects the permissions to be in "order". 

Now, the validation of computed fields in permission is simplified by avoiding the select permission check on returning tables.

rakeshkky: In commit https://github.com/hasura/graphql-engine/pull/3211/commits/c776fe6a2fca490dd8b978e1438e8142a8121639, I've made change to only accept the computed fields returning scalar types in select permissions.

PR #3226, opened 2019-10-24
[wip] improve validation of root fields

### LINKED ISSUE #3220: Why is Hasura erring on startup with “following root fields are duplicated” in beta8?
This question is for a Hasura application running against a Postgres database. When running beta6, everything is working. However, after updating to beta8, Hasura does not start.

Here is the last log message when starting with beta6:

```
{
    "timestamp": "2019-10-23T19:07:55.521+0000",
    "level": "info",
    "type": "startup",
    "detail": {
        "kind": "server",
        "info": {
            "time_taken": 0.486567746,
            "message": "starting API server"
        }
    }
}
```

And here is the last log message starting with beta8:

```
{
    "path": "$.args[0].args",
    "error": "following root fields are duplicated: ******",
    "code": "unexpected"
}
```

We are carrying on with beta6 for now, if anyone has an open communications channel with Hasura, it might be good to let them know they introduced a breaking change with their latest release. It would be good to get on the upgrade pathway again, hopefully we are not stuck on beta6 for whatever reason.

Cross posted from: https://stackoverflow.com/questions/58529871/why-is-hasura-erring-on-startup-with-following-root-fields-are-duplicated-in-b

rakeshkky:
### Affected components 

- Server
- Console
- CLI
- Docs
- Community Content
- Build System
- Tests
- Other (list it)

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Schema.hs
+++ b/server/src-lib/Hasura/GraphQL/Schema.hs
@@ -331,15 +331,15 @@ getRootFldsRole'
 getRootFldsRole' tn primCols constraints fields funcs insM
                  selM updM delM viM tableConfig =
   RootFields
-    { rootQueryFields = makeFieldMap
-        $  funcQueries
+    { rootQueryFields =
+        funcQueries
         <> funcAggQueries
         <> catMaybes
           [ getSelDet <$> selM
           , getSelAggDet selM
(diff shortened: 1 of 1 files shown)


rakeshkky:
This improvement is going through https://github.com/hasura/graphql-engine/pull/3154, hence closing this.

PR #3236, opened 2019-10-25
improve event fetch query

tirumaraiselvan:
### Description

Improve the fetch events query 

### Affected components 

- Server
- Console

### Solution and Design
1. When an event trigger is deleted, set `archived = 't'` for all its corresponding events. Instead of joining to find active event triggers (like it happens currently), check for archived.  

2. Add index to `delivered` column. 

### Steps to test and verify
```
postgres=# explain  SELECT l.id
                        FROM hdb_catalog.event_log l
                        JOIN hdb_catalog.event_triggers e
                        ON (l.trigger_name = e.name)
                        WHERE l.delivered ='f' and l.error = 'f' and l.locked = 'f'
                              and (l.next_retry_at is NULL or l.next_retry_at <= now())
                        FOR UPDATE SKIP LOCKED
                        LIMIT 100 ;
                                                                QUERY PLAN                                                                
------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.57..30.06 rows=100 width=49)
   ->  LockRows  (cost=0.57..64027.16 rows=217110 width=49)
         ->  Merge Join  (cost=0.57..61856.06 rows=217110 width=49)
               Merge Cond: (l.trigger_name = e.name)
               ->  Index Scan using event_log_trigger_name_idx on event_log l  (cost=0.42..59087.56 rows=217110 width=56)
                     Filter: ((NOT delivered) AND (NOT error) AND (NOT locked) AND ((next_retry_at IS NULL) OR (next_retry_at <= now())))
               ->  Index Scan using event_triggers_pkey on event_triggers e  (cost=0.15..53.70 rows=370 width=38)
(7 rows)

postgres=# explain  SELECT l.id
                        FROM (select id, trigger_name from hdb_catalog.event_log 
                                      WHERE delivered ='f' and error = 'f' and locked = 'f'
                              and (next_retry_at is NULL or next_retry_at <= now()) and archived = 'f') l
                        FOR UPDATE SKIP LOCKED
                        LIMIT 100;
                                                                      QUERY PLAN                                                                       
-------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.42..16.54 rows=100 width=43)
   ->  LockRows  (cost=0.42..34995.02 rows=217110 width=43)
         ->  Index Scan using event_log_delivered_idx on event_log  (cost=0.42..32823.92 rows=217110 width=43)
               Index Cond: (delivered = false)
               Filter: ((NOT delivered) AND (NOT error) AND (NOT locked) AND (NOT archived) AND ((next_retry_at IS NULL) OR (next_retry_at <= now())))
(5 rows)

```

It's not clear if the index would help because sometimes postgres prefers seq scan (during experiments), in which case it is just costly to maintain the index.


server/src-rsr/initialise.sql
```diff
@@ -280,11 +280,13 @@ CREATE TABLE hdb_catalog.event_log
   tries INTEGER NOT NULL DEFAULT 0,
   created_at TIMESTAMP DEFAULT NOW(),
   locked BOOLEAN NOT NULL DEFAULT FALSE,
-  next_retry_at TIMESTAMP
+  next_retry_at TIMESTAMP,
+  archived BOOLEAN NOT NULL DEFAULT FALSE
```
lexi-lambda: Instead of doing this, would it be possible to add a foreign key reference to `event_triggers` with `ON DELETE SET NULL`? That seems like it would be essentially the same as adding this `archived` field, but with slightly more support from the database for data integrity.

tirumaraiselvan: Hmm... it would remove the information of the trigger `name` though. I guess that is acceptable, except it would be a breaking change!

lexi-lambda: > what if someone recreates an event trigger and expects the older events to be delivered as well.

To be honest, as a user, I think I would find this behavior extremely surprising… but I also don’t know that much about the use cases for event triggers. Are there other ways that users might currently observe the trigger name of events that have had their associated event trigger deleted? I guess I’m not 100% sure what the purpose of keeping the events around after the trigger has been deleted is in the first place, honestly.

tirumaraiselvan: This was kinda an unfinished story. We had not finalized what to do with the event log of triggers that were deleted. We left it at the point of asking the user what they want explicitly: 

1) keep the events
2) delete the events
3) keep undelivered events till they are delivered

We don't have a statistic of how people are using event triggers and I don't mind doing 2) but it would be a breaking change which would need to be called out.

lexi-lambda: Understood, that makes sense. I don’t know enough about the details here, so I guess the current approach seems alright for now. It does seem like it would be nice to have a more cohesive story for this in general, though.

tirumaraiselvan: Agreed. Will open an issue for making this coherent.

tirumaraiselvan:
@wawhal @rikinsk Can we accommodate `archived` column in the events console page?

tirumaraiselvan:
@rikinsk I tested this and seems fine.

PR #3240, opened 2019-10-25
[RFC] Build server in CI with optimizations, even in pull requests

lexi-lambda:
This PR modifies the CI configuration to always build `graphql-engine` with optimizations enabled, even in pull requests. I think we should do this for the following reasons:

  1. Haskell performance without optimizations can be very poor. Doing any performance-related testing against a non-optimized build is not very meaningful, and enabling optimizations could theoretically even make certain tests run faster (even if the compile times are slower).

  2. We push a Docker image with the built server that we encourage users to try, and it’s not good to encourage anyone to ever use non-optimized builds, even if they’re development versions.

  3. If we start doing automated benchmarking in CI, which we plan to, we’ll have to build an optimized version at some point anyway.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,8 +174,8 @@ jobs:
             echo "Branch starts with dev or release, or tagged commit starts with v. Optimized build"
             make ci-binary
           else
-            echo "Non-release branch, non-optimized build with coverage"
-            BUILD_FLAGS="--fast --coverage" make ci-binary
+            echo "Non-release branch, build with coverage"
+            BUILD_FLAGS="--coverage" make ci-binary
           fi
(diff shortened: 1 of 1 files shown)


jberryman:
I wonder if it would be worthwhile to run the build first with `-fno-code` (or whatever it is), just to get a fast failure when things don't type-check?

lexi-lambda:
Don’t you think it’s pretty unlikely that people are going to be pushing code that doesn’t typecheck, anyway? Most of the value of CI is running the full test suite against all the different configurations in parallel, which is difficult to do on one’s workstation. The typechecking shouldn’t be any different in CI than it is on your machine.

PR #3257, opened 2019-10-27
fix server dev script (close #3246)

### LINKED ISSUE #3246: scripts/dev.sh test fails due to missing bdist_wheel
dev.sh does not install wheel, and CONTRIBUTING.md does not tell you to install it either. It is unclear which is the best way to solve this issue: some form of installing wheel into the venv, or listing wheel as a dependency in the .md file, for the user to install manually?

cheater:
…g missing. (closes #3246)

Signed-off-by: cheater <195390+cheater@users.noreply.github.com>

### Affected components

- [x] Tests

### Related Issues

#3246 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

(not that i know of. tests now execute, so could this break something? maybe?)

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -305,6 +305,7 @@ elif [ "$MODE" = "test" ]; then
   if [ ! -d "$PY_VENV" ]; then
     python3 -m venv "$PY_VENV"
     source "$PY_VENV/bin/activate"
+    pip3 install wheel
     pip3 install -r requirements.txt
   else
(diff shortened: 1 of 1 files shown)


marionschleifer:
@cheater thank you for your contribution 🙂 could you please send an email to marion@hasura.io, so that we can arrange the shipping of your swag?

PR #3278, opened 2019-10-30
docs: fix typos

sameshl:
`can deployed` -> `can be deployed`
Remove tautology: `exact same` -> `same`

- [x] Docs


marionschleifer:
@sameshl could you please send an email to marion@hasura.io, so that we can arrange the shipping of your swag? 🙂

PR #3310, opened 2019-11-06
Regression benchmarks setup

nizar-m:
### Description

This PR tries to address the issue of performance regressions (as explained in hasura/graphql-engine-internal#297)

[wrk](https://github.com/wg/wrk) is used for running benchmarks. The script `hge_wrk_bench.py` sets up a database based on [sportsdb schema](http://sportsdb.org/sd/samples), and runs benchmarks using wrk with queries defined in file `queries.graphql`, one query at a time. The results are also stored in the *results GraphQL engine*. 

The defaults are such that the scripts can be run without any extra configuration on the local machine. However with the provided options, it should be possible to run the benchmarks with an already running Postgresql, a given Hasura GraphQL engine docker image, and also push results to a different database running a Hasura GraphQL Engine.

The script collects as much info as possible before pushing the results. This includes a bunch of CPU info (like model, number of cores etc which can be collected from /proc/cpuinfo), Postgres version used, wrk parameters used, GraphQL query used, version info of GraphQL engine (or Docker image tag) etc.

### Affected components

- [x] Benchmarks

### Related Issues
hasura/graphql-engine-internal#297

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### Steps to test and verify

```sh
cd server/bench-wrk
# Setup virtualenv
python3 -m venv venv
source venv/bin/activate
pip3 install -r requirements.txt
# Run benchmarks
python3 hge_wrk_bench.py
```

The benchmarking of subscriptions are not covered


server/bench-wrk/hge_wrk_bench.py
```diff
@@ -0,0 +1,292 @@
...
+            hge_docker_image = self.hge_docker_image,
+            hge_admin_secret = self.hge_admin_secret,
+            skip_stack_build = self.skip_stack_build,
+            graphql_queries_file = self.graphql_queries_file,
+            connections = self.connections,
+            duration = self.duration
+        )
+        # TODO initialize values
+
+    # TODO Any extra argument passed after -- should go to GraphQL Engine
```
jberryman: Are all these `TODOs` still TODO?

server/bench-wrk/queries.graphql
```diff
@@ -0,0 +1,128 @@
+
+# A query which carries a lot of data
```
jberryman: Thanks for documenting the queries. It would be even better going forward when we add a query to answer the question: what do we expect to learn from measuring this? i.e. what's our hypothesis, for instance "a large payload returned from postgres to exercise our parsers and the GC (parameters such as nursery size might affect this)". This sort of documentation can be added to later as we learn more about the system as well.

server/bench-wrk/Readme.md
```diff
@@ -0,0 +1,70 @@
...
+### Setup ###
+
+The setup includes two Postgres databases with [sportsdb](https://www.thesportsdb.com/) schema and data, and two GraphQL engines running on the Postgres databases. Then one of the GraphQL engines is added as a remote schema to another GraphQL engine.
+
+The data will be same in both the databases. The tables reside in different database schema in-order to avoid GraphQL schema conflicts.
+
+The methods in script `sportsdb_setup.py` helps in setting up the databases, starting the Hasura GraphQL engines, and setting up relationships. This script can either take urls of already running Postgres databases as input, or it can start the databases as Docker instances. The GraphQL engines can be run either with `stack exec` or as Docker containers.
+
+### Run benchmark ###
+- Install dependencies for the Python script in a virtual environment (requires Python>=3.6).
```
jberryman: Using `pyenv` to handle python versions worked well for me (thanks for the tip!)

jberryman:
This seems really thoughtfully done!

# A few thoughts as I try this out, and a brain/note-dump:

- there seems to be about a 1minute period at the start of a `wrk` run where nothing much is happening (no CPU usage by hge instances or wrk), do you know why that is?

- As I run the tests with default settings all 4 cores (2 real cores) on my laptop are maxed out, while `wrk` itself is using between 6% and 32% CPU (depending on the benchmark, it seems); this is probably okay, but I wonder if performance of wrk itself could be improved? Haven't looked at what we're doing with Lua.

- Although I've used wrk in the past I could find no detailed documentation: what does the wrk throughput result mean exactly? e.g. Is throughput limited by the parameters to wrk itself (or does it attempt to find maximum supported throughput by looking at timeouts/errors?)? Are the req/s numbers _per wrk thread_? What does it mean to talk about mean and percentiles of throughput, as in...

      "requests": {
        "max": 242,
        "mean": 120.75356755403,
        "dist": {
          "98": 190,
          "95": 171,
          "99": 200
        },


- In general I'm deeply distrustful of `wrk` and all these tools; it seems very few people take the time to look at whether they work correctly, and when they do the results are concerning or mysterious. e.g. there is the "coordinated omission" problem that motivated the wrk2 project; it's not clear to me what the latency numbers from wrk mean and if they can be trusted

- Graph not very useful (I've been playing and thinking quite a bit on this lately; will ping you)

- what happens when there are many errors or timeouts? It's possible for those to manifest as _improved_ performance if we look at the latency metrics alone, say.

- The numbers raise a lot of questions (and this is okay/good): is postgres the bottleneck? (If so can it be configured better for performance?) It would be a useful validation to be able to turn on a cache in hge that turns a PG query into a table lookup; a benchmark where PG is the bottleneck is still super useful, but ultimately we want benchmarks to help us _ deeply understand_ performance of our application deeply, not only notice regressions.

- there's a whole range of more realistic load tests we could do of course

- it's much more useful for engineers to be able to see performance regressions in microbenchmarks (since they are more targeted, obviously, and usually less noisy); this sort of end-to-end test can catch regressions we haven't anticipated or in components we don't have criterion benchmarks (TBD) for.

---

## Load testing tools notes (FYI)

Deep analysis of **wrk2** vs **fortio**
https://github.com/envoyproxy/envoy/issues/5536
https://docs.google.com/document/d/10xeQuEjUjdmfFq36kGrxI6v78GKHspqn_Nb-pohKfOo/edit 

Review of different load tools (from load testing company):
https://blog.loadimpact.com/open-source-load-testing-tool-benchmarks 
    "Performance-wise, **Wrk** and **Apachebench** are in a class of their ownh3" 
    "**Gatling**, **Jmeter**, **Grinder**, **Tsung** and **Boom** all offer good performance, accuracy and reliability" 
...and comparing with **K6** (their tool): https://blog.loadimpact.com/open-source-load-testing-tool-benchmarks-v2 
**K6** is scriptable, constant-rps, slick looking from a company dedicated to load testing:
https://github.com/loadimpact/k6 
Control in realtime, with:
https://blog.loadimpact.com/building-a-ui-for-the-k6-load-testing-tool 

### Also interesting (maybe for more interactive exploration)

Get throughput over time:
https://github.com/BuoyantIO/slow_cooker

Real-time histogram:
https://github.com/ikruglov/slapper 

Constant rps, can spit out nice graphs:
https://github.com/tsenart/vegeta 

---

## A Nice write-up on load testing and analysis

I was reminded of this blog / talk summary: http://highscalability.com/blog/2015/10/5/your-load-generator-is-probably-lying-to-you-take-the-red-pi.html . A few points:

- never throw away the max latency
- When thinking about UX of someone on the web (where each visit might produce dozens of requests), 99.9 percentile latencies are experienced by the _vast majority of users_
- Typical metric collection tools are often inadequate ("lie"). An algorithm like https://github.com/HdrHistogram/HdrHistogram can be used to accurately collect long-tail stats (maybe useful to implement the kind of reporting we care about within hasura itself)
- The entire last "_Sustainable Throughput_" section suggests, I think, a good approach towards thinking about load and the various ways to e.g. compare two systems or present performance information about a system

jberryman:
## Some more focused/actionable comments about this PR specifically:

After experimenting with your other branch that used wrk2 (super helpful!) and thinking about this stuff: 

1. I think **wrk2** (maybe K6 too, although it's not clear to me whether their constant-RPS functionality handles the CO problem properly...) might be more appropriate here: we care about finding and knowing throughput _limits_ (e.g. we can support X-RPS while maintaining reasonable latency Y), and explore latency distributions at various RPS, but I'm not sure that will help to find performance regressions in our code; i.e. I think "latency at some reasonable throughput" is a better thing to measure in CI

2. It would be really useful to e.g. log timings of `runGQ` and plot those, both as a way of validating the wrk2 measurements, as well as differentiating Response Time from Service Time (I'm a little hazy on this distinction tbh...); I've been doing this in a throwaway branch.

3. If we use wrk2 I think we should be plotting the HdrHistogram; but we also need to validate the high percentile measurements if we're trying to make decisions about them (they might be just noise if we have few samples...); the `max` is always useful to know (...or is it when latency always has a long tail...?)

I think points 1 and 3 should probably be resolved before merge, and other improvements or changes can happen incrementally.

What are your thoughts?

nizar-m:
>     * there seems to be about a 1minute period at the start of a `wrk` run where nothing much is happening (no CPU usage by hge instances or wrk), do you know why that is?

This is probably because wrk is run inside a docker, and docker pull took some time. I could not see the same in the tests I am running.

nizar-m:
>     * Although I've used wrk in the past I could find no detailed documentation: what does the wrk throughput result mean exactly? e.g. Is throughput limited by the parameters to wrk itself (or does it attempt to find maximum supported throughput by looking at timeouts/errors?)? Are the req/s numbers _per wrk thread_? What does it mean to talk about mean and percentiles of throughput, as in...

It seems wrk is trying to push as much requests as possible with given number of open connections. Most probably the mean corresponds to the maximum possible requests per second. Which also means the latency results are useless. (The average queue size (and hence latency) should be close to inifinity. It is only limited by the number of connections.)

We probably should be running wrk2 at something like 5%, 10% etc upto 50% of this average requests per second given by wrk. 

wrk2 provides results detailed enough to generate hdrHistograms of latency. Should we use a lua script and try to get each latency measure? This could increase the CPU usage of wrk2.

jberryman:
> Should we use a lua script and try to get each latency measure? This could increase the CPU usage of wrk2.

I was wondering that myself. Was trying to think if it would be useful to be able to plot timing from wrk2 on the same graph as the latency timings measured from instrumented code (the example plot script I showed you was measuring logged timings for each `runGQ`). I'm not sure.

jberryman:
It looks like `boto3` was added to `requirements-top-level.txt` but didn't make it into frozen `requirements.txt`. I did a `pip3 install boto3` then it worked for me.

Here is my `pip3 freeze` after doing that:

```
aniso8601==7.0.0
boto3==1.10.45
botocore==1.13.45
certifi==2019.11.28
chardet==3.0.4
Click==7.0
colorama==0.4.3
cycler==0.10.0
dash==1.7.0
dash-bootstrap-components==0.7.2
dash-core-components==1.6.0
dash-html-components==1.0.2
dash-renderer==1.2.2
dash-table==4.5.1
docker==4.1.0
docutils==0.15.2
Flask==1.1.1
Flask-Compress==1.4.0
future==0.18.2
graphene==3.0.dev20190817210753
graphql-core==3.0.1
graphql-relay==3.0.0a1
idna==2.8
inflection==0.3.1
itsdangerous==1.1.0
Jinja2==2.10.3
jmespath==0.9.4
kiwisolver==1.1.0
MarkupSafe==1.1.1
matplotlib==3.1.2
numpy==1.17.4
pandas==0.25.3
plotly==4.3.0
psycopg2==2.8.4
py-cpuinfo==5.0.0
pyparsing==2.4.5
python-dateutil==2.8.1
pytz==2019.3
requests==2.22.0
requests-cache==0.5.2
retrying==1.3.3
ruamel.yaml==0.16.5
ruamel.yaml.clib==0.2.0
s3transfer==0.2.1
scipy==1.3.3
seaborn==0.9.0
six==1.13.0
urllib3==1.25.7
websocket-client==0.56.0
Werkzeug==0.16.0
```
**EDIT**: although after trying again from another clean venv I get:

> 
>   Could not find a version that satisfies the requirement graphql-core==3.0.1 (from -r requirements.txt (line 18)) (from versions: 0.1a0, 0.1a1, 0.1a2, 0.1a3, 0.1a4, 0.4.7b0, 0.4.7b1, 0.4.7b2, 0.4.9, 0.4.11, 0.4.12, 0.4.12.1, 0.4.13, 0.4.14, 0.4.15, 0.4.16, 0.4.17, 0.4.18, 0.5b1, 0.5b2, 0.5b3, 0.5, 0.5.1, 0.5.2, 0.5.3, 1.0.dev20160814231515, 1.0.dev20160822075425, 1.0.dev20160823054952, 1.0.dev20160909030348, 1.0.dev20160909040033, 1.0.dev20160920065529, 1.0, 1.0.1, 1.1, 1.2.dev20170724044604, 2.0.dev20170801041408, 2.0.dev20170801051721, 2.0.dev20171009101843, 2.0, 2.1rc0, 2.1rc1, 2.1rc2, 2.1rc3, 2.1, 2.2, 2.2.1)
> No matching distribution found for graphql-core==3.0.1 (from -r requirements.txt (line 18))

Minor observations which shouldn't hold up a merge IMO:
- It might be nice to print something like "Benchmarking queries at different RPS rates up to maximum: [10, 20, ... "; I was confused for a moment about what was happening when I saw 10 RPS.
- We get `CPU KEY: ...` printed several times which is a little ugly.

Some questions / possible small improvements:
- Is the 5 minute runtime for the benchmarks arbitrary? What if we scaled the benchmark runtime so that each RPS had approximately the same number of samples (but ran for at least a minute... or something)? This is something we might want to iterate on as we use the benchmarks more.

lexi-lambda:
@jberryman Thanks for trying this out. I actually tried cloning it myself just a little earlier, and I bumped into this error when trying to run `pip3 install -r requirements.txt`:

```
Collecting graphql-relay==3.0.0a1 (from -r requirements.txt (line 19))
  Downloading https://files.pythonhosted.org/packages/b3/e9/f746793c405695c6c3d2fe023dd19738d1eef20ecdd50ebb9e240b2f6cfc/graphql-relay-3.0.0a1.tar.gz
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  Complete output from command /Users/alexis/gits/hasura/graphql-engine/hge-bench-wrk/server/bench-wrk/.python-venv/bin/python3 /Users/alexis/gits/hasura/graphql-engine/hge-bench-wrk/server/bench-wrk/.python-venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py get_requires_for_build_wheel /var/folders/tz/11j773ts1qb5q8xzjbr0292m0000gn/T/tmprwfs5ydw:
  Traceback (most recent call last):
    File "/Users/alexis/gits/hasura/graphql-engine/hge-bench-wrk/server/bench-wrk/.python-venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 207, in <module>
      main()
    File "/Users/alexis/gits/hasura/graphql-engine/hge-bench-wrk/server/bench-wrk/.python-venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 197, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/Users/alexis/gits/hasura/graphql-engine/hge-bench-wrk/server/bench-wrk/.python-venv/lib/python3.7/site-packages/pip/_vendor/pep517/_in_process.py", line 54, in get_requires_for_build_wheel
      return hook(config_settings)
    File "/private/var/folders/tz/11j773ts1qb5q8xzjbr0292m0000gn/T/pip-build-env-0_bw21ba/overlay/lib/python3.7/site-packages/poetry/masonry/api.py", line 25, in get_requires_for_build_wheel
      poetry = Factory().create_poetry(Path("."))
    File "/private/var/folders/tz/11j773ts1qb5q8xzjbr0292m0000gn/T/pip-build-env-0_bw21ba/overlay/lib/python3.7/site-packages/poetry/factory.py", line 54, in create_poetry
      raise RuntimeError("The Poetry configuration is invalid:\n" + message)
  RuntimeError: The Poetry configuration is invalid:
    - Additional properties are not allowed ('licese' was unexpected)
    - [keywords] 'graphql relay api' is not of type 'array'
```

I haven’t the foggiest idea what that means or what went wrong, but I’d like to be able to run this on my machine! I started looking into it, but I haven’t figured it out yet.

lexi-lambda:
I reinstalled from `requirements-top-level.txt`, and it seemed to work fine. My freeze file is this:

```
aniso8601==7.0.0
boto3==1.10.45
botocore==1.13.45
certifi==2019.11.28
chardet==3.0.4
Click==7.0
colorama==0.4.3
cycler==0.10.0
dash==1.7.0
dash-bootstrap-components==0.8.0
dash-core-components==1.6.0
dash-html-components==1.0.2
dash-renderer==1.2.2
dash-table==4.5.1
docker==4.1.0
docutils==0.15.2
Flask==1.1.1
Flask-Compress==1.4.0
future==0.18.2
graphene==3.0.dev20190817210753
graphql-core==3.0.1
graphql-relay==3.0.0
idna==2.8
inflection==0.3.1
itsdangerous==1.1.0
Jinja2==2.10.3
jmespath==0.9.4
kiwisolver==1.1.0
MarkupSafe==1.1.1
matplotlib==3.1.2
numpy==1.18.0
pandas==0.25.3
plotly==4.4.1
psycopg2==2.8.4
py-cpuinfo==5.0.0
pyparsing==2.4.6
python-dateutil==2.8.1
pytz==2019.3
requests==2.22.0
requests-cache==0.5.2
retrying==1.3.3
ruamel.yaml==0.16.5
ruamel.yaml.clib==0.2.0
s3transfer==0.2.1
scipy==1.4.1
seaborn==0.9.0
six==1.13.0
urllib3==1.25.7
websocket-client==0.57.0
Werkzeug==0.16.0
```

Reinstalling from that freeze file in a fresh venv works fine for me, so I’m not sure what issue you were bumping into with `graphql-core`, @jberryman. Maybe some caching somewhere? I don’t know very much about `pip`.

jberryman:
Thanks @lexi-lambda , I realized the issue was the first time I tested was in a dirty directory that had a `.python-version` file with "3.7.5", while my system has 3.5.3 (which was causing the problems).

@nizar-m What do you think about adding an appropriate `.python-version` file to that directory, and documenting it? I've found pyenv to be really easy to use, and this seems like a lightweight solution (vs. that other build tool you told me about).

nizar-m:
The srcipt can run tests in one virtual machine, push their results in to a remote Hasura GraphQL Engine, and push the large files like latencies into s3. 

After each iteration of script, it should keep pushing the data to the database (and to s3). Using the data across benchmarks on multiple versions/commits, we should be able to use another script to plot the regression in performance (and then push it into a Github repo or even to slack).

What is missing of course is a Github hook to listen for changes in code and run the script on a new commit to master. I am planning to add that in a separate repo.

With the current script, a developer himself should be able to run regression benchmarks across various commits in his PR.

server/bench-wrk/.gitignore
```diff
@@ -0,0 +1,4 @@
+__pycache__
```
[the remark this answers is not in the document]
nizar-m: We should I guess.

jberryman:
My last comments/questions are:

- In the "plot type" dropdown, can we change the labels to e.g. "latency histogram", and "latency violinplot"? So we're clear.
- When the plot page first opens to the throughput plot I don't see all the graphs and need to add some manually. Is it easy to make them all appear by default, or is that undesirable for some reason?
- similarly, it might be nice if the violinplot view showed e.g. both the lowest and highest RPS runs side by side (the user could add additional by hand), or maybe include up to three or something.

Again I don't think these are blockers, just things we can iterate on.

The boto3 thing and adding a `.python-version` were my only real issues.

lexi-lambda:
> The srcipt can run tests in one virtual machine, push their results in to a remote Hasura GraphQL Engine, and push the large files like latencies into s3.
> 
> After each iteration of script, it should keep pushing the data to the database (and to s3). Using the data across benchmarks on multiple versions/commits, we should be able to use another script to plot the regression in performance (and then push it into a Github repo or even to slack).

Cool, that sounds great! Is this process described/documented anywhere? I don’t see anything about it in the README, but I think the information you described in your comment is helpful, and I’d like to see it make it into the repo somewhere.

tirumaraiselvan:
@nizar-m Sorry to jump out-of-nowhere. But where is the VM? How does it run tests when a PR is made? Also, Can we add an event trigger to the regression Hasura instance and make it send a message on the PR itself?

We should definitely have this indicate in CI somehow. I am not sure if I missed anybody's comment on how that would happen.

nizar-m:
Will the following work?
- When `/perf-test` comment is sent on a PR, a bot would trigger running of the benchmarks for that PR. 
- Once the benchmarks are  done, the bot will send the results/errors as comment(s) in the same PR.

lexi-lambda:
I would like to get this merged. But some of the comments/questions I had are still outstanding:

  1. Can the README please be updated to include information about pushing results to S3?

  2. Can the README include some information about how to read/evaluate the results?

I think, with those changes, I would be happy to merge this PR.

nizar-m:
@0x777  and I had a discussion on this pull request. 
He wanted some changes so that 
- It can be both run by a developer (on both Mac and linux) and on CI.
- on CI, the components, that is wrk2, GraphQL Engine and Postgres should be on separate machines.
- Also probably a script which on AWS/GCE, will create these components, start the benchmarks, and teardown the system.

This needs a bit more modular approach. I am working on those changes right now.

jberryman:
@nizar-m after getting this working with cabal new-build I'm finding a lot of little errors that make the scripts crash with errors at various points. I'm wondering if the commit history got screwed up perhaps? e.g. these are the local changes I'm working with, and I'm still hitting new issues:

```
diff --git server/bench-wrk/hge_wrk_bench.py server/bench-wrk/hge_wrk_bench.py
index dd32e9eb..75afdf1c 100644
--- server/bench-wrk/hge_wrk_bench.py
+++ server/bench-wrk/hge_wrk_bench.py
@@ -31,7 +31,8 @@ class HGEWrkBench(HGETestSetup):
 
     wrk_docker_image = 'hasura/wrk:v0.3'
 
-    lua_dir = '/tmp/bench_scripts'
+    base_path = pathlib.Path(__file__).parent 
+    lua_dir = str((base_path / "bench_scripts").resolve())
 
     rps_steps = [10, 20, 50, 100, 200, 500, 1000, 2000, 5000]
 
@@ -227,10 +228,12 @@ class HGEWrkBench(HGETestSetup):
         query_str = graphql.print_ast(query)
         print(Fore.GREEN + "(Compute maximum Request per second) Running wrk benchmark for query\n", query_str + Style.RESET_ALL)
         self.hge.graphql_q(query_str) # Test query once for errors
-        bench_script = os.path.join(self.lua_dir + '/bench.lua')
+        bench_script = os.path.join(self.lua_dir + '/bench-wrk.lua')
         graphql_url = self.hge.url + '/v1/graphql'
         params = self.get_wrk2_params()
-        duration = 30
+        # TODO This seems to take much longer than 30 sec...
+        duration = 2
+        # duration = 30
         wrk_command = [
             'wrk',
             '-t', str(params['threads']),
@@ -255,6 +258,7 @@ class HGEWrkBench(HGETestSetup):
             user = self.get_current_user()
         )
         summary = json.loads(result)['summary']
+        # TODO explain this calculation. Why aren't we using wrk's reported 'max'? Should we call this avg_sustained_rps or something?
         max_rps = round(summary['requests']/float(duration))
         self.insert_max_rps_result(query, max_rps)
         print("Max RPS", max_rps)
@@ -444,7 +448,7 @@ mutation insertMaxRps($result: hge_bench_query_max_rps_insert_input!) {
                 max_rps = self.max_rps_test(query)
                 # The tests should definitely not be running very close to or higher than maximum requests per second
                 rps_steps = [ r for r in self.rps_steps if r < 0.6*max_rps]
-                print("Benchmarking queries with wrk2 for the following requests/sec", ", ".join(rps_steps))
+                print("Benchmarking queries with wrk2 for the following requests/sec", ", ".join(map(str, rps_steps)))
                 for rps in rps_steps:
                     if rps < int(0.6*max_rps):
                         self.wrk2_test(query, rps)
```

jberryman:
self-assigned since I've been working on this. What I care about is mostly being able to easily run benchmarks on different branches and compare them locally, so that means:

- [ ] get working with `cabal install`, a few other broken bits to fix
- [ ]  figure out how weird several minutes hang in lua `require` so that benchmarks don't take forever
- [ ] make plot script work on directories of data in `test_output` rather than requiring a graphql-engine
- [ ]  figure out if the low RPS tests are giving real latency numbers even though they report erroneous timeout (I think this is okay, but worth double-checking): https://github.com/giltene/wrk2/issues/56

jberryman:
There's been more of an interest in this. Let's merge this ASAP and continue iterating on it (integrating with CI at a later time).

@codingkarthik do you want to add information here about how you are running the benchmarks since mac doesn't work ATM? If it doesn't seem useful we can just make a note that it doesn't work on mac. Hopefully the situation can be resolved soon in new docker and in the meantime folks could spin up a linux VM or something

codingkarthik:
@jberryman I had used a hosted linux VM instance to run the benchmarks. The existing instructions are accurate and I followed those, however, I have added a couple of small sections in the `bench-wrk/Readme.md` file for further clarity. It'll be great if you could go over it once.

PR #3327, opened 2019-11-10
Add scripts/plot.py for comparing latencies, etc.

jberryman:
This can be used to visualize performance changes between two or more
branches, across several categories (e.g. requests-per-second load).

While not perfect I think the output goes a long way to helping the user
absorb a lot of information about the data, while avoiding drawing wrong
inferences, e.g. paying too much attention to mean, or forgetting how
noisey the data is.

This will hopefully be integrated into some benchmarking scripts, and improved, probably in collaboration with Nizar's work in e.g.

https://github.com/tirumaraiselvan/graphql-engine/pull/70  , and...
https://github.com/hasura/graphql-engine/pull/3310#pullrequestreview-314320463

Wanted to share this ASAP though.

### Steps to test and verify
See README.md

### Example

The graph offers:
- easy comparison, across categories
- very granular view of actual samples, while the violinplot (like a histogram) shows us where points are most dense (more expressive than just mean or median); the raw points also help validate whether comparing percentiles is meaningful (although, we should do this better)
- useful percentiles

Tentative plan is to use this to plot internal service timing, and combine it with some hdr histogram timing from wrk2 (the client), so we can see where throughput is actually making things fall over.

plot


jberryman:
@nizar-m let me know what you think of this;

nizar-m:
Definitely very interesting plots. Why the Indian oil lamp like plots for latencies, I have no idea.

jberryman:
> Why the Indian oil lamp like plots for latencies, I have no idea.

These are "violin plots" (maybe they have other names too), and technically speaking graph https://en.wikipedia.org/wiki/Kernel_density_estimation but you can think of them like a continuous histogram; so places where the violin is very thick are areas that are very dense with samples (above these are unsurprisingly right around the median represented by the thin black line).

jberryman:
Closing this for now, as I think it's going to live in its own repo.

jberryman:
Development continues here for this: https://github.com/hasura/latency-plotting

PR #3347, opened 2019-11-13
add few x-forwarded- headers (close #2572)

### LINKED ISSUE #2572: Remote schema forward original user-agent

tirumaraiselvan:
### Description

Adds `X-Forwarded-Host` and `X-Forwarded-User-Agent` headers to the remote schema requests.

### Affected components

- [x] Server

### Related Issues
#2572 

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

This PR does not forward other headers like `X-Forwarded-For` which is more complicated.


lexi-lambda:
@tirumaraiselvan The tests you just pushed seem to be failing.

tirumaraiselvan:
Bit of a conundrum. The server in CI seems to be running in localhost:8081.

tirumaraiselvan:
@lexi-lambda I am thinking of dropping the test for `X-Forwarded-Host` as it is dynamic ( but I will keep `X-Forwarded-User-Agent`) ?

lexi-lambda:
Ah, yeah, the tests in CI spin up two instances of `graphql-engine` in parallel. There must be a way to access the current host and port we’re running the tests against from inside the test suite code, so we should use that for the expected value of `X-Forwarded-Host`.

tirumaraiselvan:
We can get the host:port in pytest but the expected value will then somehow will have to be fed to the graphql-server.py (which so far was independent server). Worth the fight?

lexi-lambda:
Ah, I see, right of course. I think it’s probably worth doing—`graphql-server.py` will need to check to see if the host is either of the expected GraphQL servers, so it should probably be fed a list of URLs the same way the core test suite is.

tirumaraiselvan:
Wow awkwardly stumbled on fixing this till I realized that there was only one pytest session but parallelized internally.

Issue #3350, opened 2019-11-13
Make cli-migrations Heroku Compatible/Add URL dependency Injection


simonireilly:
Running the container for `hasura/graphql-engine:v1.0.0-beta.9.cli-migrations` works great in an environment where I can control the ENV used at ENTRYPOINT.

This issue arises when a managed service, such as Heroku, specifies the environment variables to be used in ENTRYPOINT.

### Proposed Solution

The solution would be to allow the configuration of the database URL when the server boot to perform migrations.

This solution would fall back to the default $HASURA_GRAPHQL_DATABASE_URL to maintain backward compatibility.

PR #3351, opened 2019-11-13
cli(migrations-docker): take a specific env var for database url

simonireilly:
### Description

This change will enable a user to perform migrations using the `cli-migrations` images on any:

- Database URL configured as: 
`HASURA_GRAPHQL_MIGRATIONS_DATABASE_URL:postgres://postgres:@postgres:5432/database`
- Database ENV variable pointer:
`HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR: DATABASE_URL`
- The default fallback will always be the default ENV used in the graphql-engine server
`HASURA_GRAPHQL_MIGRATIONS_DATABASE_URL=$HASURA_GRAPHQL_DATABASE_URL`

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)
  - images tagged as `hasura/graphql-engine:v1.0.0-beta.< greater than 9 >.cli-migrations` though full backward compatibility should be maintained.

### Related Issues

closes: #3350 

### Solution and Design

This solution solves an issue I experience myself were attempting to use the docker image with `cli-migration` on Heroku I can't inject the  HASURA_GRAPHQL_DATABASE_URL.

It is not suitable to assign this directly as an ENV var as heroku rotates password credentials.

It would also not be possible to migrate a secondary database such as `HEROKU_POSTGRES_RED_URL` which is the default heroku behaviour for multiple attached databases.

This change allows for both multiple databases with distinct URL's to be supported through multiple release commands, and for a single database to be supported by assigning `HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR: DATABASE_URL` which will be evaluated in the ENTRYPOINT to give the current value of the ENV  `DATABASE_URL`.

All evaluated ENV VAR's are fetched from printenv to prevent script injection.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### Steps to test and verify

This implementation is fully deployed from my Heroku sample fork as:

https://github.com/simonireilly/graphql-engine-heroku/pull/1/files
https://heroku-migration-tester-002.herokuapp.com/console


shahidhk:
@simonireilly I am going through this PR and I am not sure if I fully understand what the use case is. Will you be able to explain in a more detailed manner?

scripts/cli-migrations/docker-entrypoint.sh
```diff
@@ -11,6 +11,16 @@ log() {
 DEFAULT_MIGRATIONS_DIR="/hasura-migrations"
 TEMP_MIGRATIONS_DIR="/tmp/hasura-migrations"
 
+# configure the target database for migrations
+if [ ${HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR} ]; then
```
simonireilly: @shahidhk This is my implementation.

I am able to migrate a database, that is not named as `HASURA_GRAPHQL_DATABASE_URL` without needing to have a custom `entrypoint.sh`.

image

Since heroku will always name the first database as `DATABASE_URL`

simonireilly: Here is an example of how this would look when using a Heroku manifest (heroku.yml) to manage the application entirely from a committed code base.

https://github.com/simonireilly/graphql-engine-heroku/blob/7b5e9181a759df50f10424226ff5f63a25278b5b/heroku.yml#L1-L9

simonireilly: My understanding of the current script is that it will fall back to `HASURA_GRAPHQL_DATABASE_URL` which can't be provisioned at deploy time, so you will need to set up the app, go to the admin panel, and then change the name of the database primary URL.

shahidhk: Got it.

scripts/cli-migrations/docker-entrypoint.sh
```diff
@@ -11,6 +11,16 @@ log() {
...
+    log "database url is pointed at env $HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR"
+    HASURA_GRAPHQL_MIGRATIONS_DATABASE_URL=$(printenv $HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR)
+elif [ -z ${HASURA_GRAPHQL_MIGRATIONS_DATABASE_URL+x} ]; then
+    HASURA_GRAPHQL_MIGRATIONS_DATABASE_URL=$HASURA_GRAPHQL_DATABASE_URL
+fi
+
+log "database url has been set to $HASURA_GRAPHQL_MIGRATIONS_DATABASE_URL for migrations"
```
shahidhk: It is not recommended to log database urls as it may contain the database password.

shahidhk:
```suggestion
log "database url for migrations is set by HASURA_GRAPHQL_DATABASE_URL"
```

scripts/cli-migrations/docker-entrypoint.sh
```diff
@@ -11,6 +11,16 @@ log() {
 DEFAULT_MIGRATIONS_DIR="/hasura-migrations"
 TEMP_MIGRATIONS_DIR="/tmp/hasura-migrations"
 
+# configure the target database for migrations
+if [ ${HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR} ]; then
+    log "database url is pointed at env $HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR"
```
shahidhk:
```suggestion
    log "database url for migrations is set by $HASURA_GRAPHQL_MIGRATIONS_DATABASE_ENV_VAR"
```

PR #3355, opened 2019-11-14
add sync-users section in auth0 integration guide

tirumaraiselvan:
### Description

Auth0 users need to be synced with the postgres database so that they can be used effectively in any application. This PR adds a section for this purpose.

### Affected components

- [x] Docs

### DIFF
--- a/docs/graphql/manual/guides/integrations/auth0-jwt.rst
+++ b/docs/graphql/manual/guides/integrations/auth0-jwt.rst
@@ -167,15 +167,10 @@ just like any other role.
 Sync Users from Auth0
 ^^^^^^^^^^^^^^^^^^^^^
 
-Now, that you have created an Auth0 application and can signup/login, you will need a way to sync your users in Postgres as well.
-All you really need is the Auth0 ``user_id`` in something like a "users" table.
+Now that you have created an Auth0 application and can signup/login, you will need a way to sync your users in Postgres as well.
+All you really need is the Auth0 ``user_id`` in something like a ``users`` table.
 
-Using Auth0 Rules again, add the following rule which will insert a new user everytime someone signs-up.
-
-.. note::
-
-   We need to use an ``upsert`` operation here because social logins do not distinguish between sign-up and login. Hence, we need to run this rule everytime a successful login is made and do nothing if the user already exists.
-
+Using Auth0 Rules again, add the following rule which will insert a new user every time someone signs up.
 
 .. code-block:: javascript
 
@@ -202,3 +197,7 @@ Using Auth0 Rules again, add the following rule which will insert a new user eve
    }
 
 That’s it! This rule will be triggered on every successful signup/login and sync your Auth0 user into your postgres database.
+
+.. note::
+
+   We need to use an ``upsert`` operation here because social logins do not distinguish between sign-up and login. Hence, we need to run this rule every time a successful login is made and do nothing if the user already exists.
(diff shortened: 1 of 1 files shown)


docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -163,3 +163,42 @@ The configured unauthorized role will be used whenever an access token is not pr
 
 This can be useful for data that you would like anyone to be able to access and can be configured and restricted
 just like any other role.
+
+Sync Users from Auth0
+^^^^^^^^^^^^^^^^^^^^^
+
+Now, that you have created an Auth0 application and can signup/login, you will need a way to sync your users in Postgres as well.
```
marionschleifer:
```suggestion
Now that you have created an Auth0 application and can signup/login, you will need a way to sync your users in Postgres as well.
```

docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -163,3 +163,42 @@ The configured unauthorized role will be used whenever an access token is not pr
...
+All you really need is the Auth0 ``user_id`` in something like a "users" table.
```
marionschleifer:
```suggestion
All you really need is the Auth0 ``user_id`` in something like a ``users`` table.
```

docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -163,3 +163,42 @@ The configured unauthorized role will be used whenever an access token is not pr
...
+
+Using Auth0 Rules again, add the following rule which will insert a new user everytime someone signs-up.
```
marionschleifer:
```suggestion
Using Auth0 rules again, add the following rule which will insert a new user every time someone signs up.
```

docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -163,3 +163,42 @@ The configured unauthorized role will be used whenever an access token is not pr
...
+
+.. note::
+
+   We need to use an ``upsert`` operation here because social logins do not distinguish between sign-up and login. Hence, we need to run this rule everytime a successful login is made and do nothing if the user already exists.
```
marionschleifer:
```suggestion
   We need to use an ``upsert`` operation here because social logins do not distinguish between sign-up and login. Hence, we need to run this rule every time a successful login is made and do nothing if the user already exists.
```

docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -163,3 +163,42 @@ The configured unauthorized role will be used whenever an access token is not pr
...
+
+Sync Users from Auth0
+^^^^^^^^^^^^^^^^^^^^^
+
+Now, that you have created an Auth0 application and can signup/login, you will need a way to sync your users in Postgres as well.
+All you really need is the Auth0 ``user_id`` in something like a "users" table.
+
+Using Auth0 Rules again, add the following rule which will insert a new user everytime someone signs-up.
+
+.. note::
```
marionschleifer: Can we have the note after the example. When we say "...add the following rule..." then the example should come right after that.

docs/graphql/manual/guides/integrations/auth0-jwt.rst
tirumaraiselvan: I feel `Rules` should be capitalized here to emphasize that it's a feature and not used as a common noun.

PR #3370, opened 2019-11-18
add read_only arg to docs

tirumaraiselvan:
### Description

Documentation for #3191 

### Affected components

- [x] Docs

### DIFF
--- a/docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
+++ b/docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
@@ -208,7 +208,7 @@ Args syntax
    * - read_only
      - false
      - Boolean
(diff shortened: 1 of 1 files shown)


docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
```diff
@@ -205,6 +205,10 @@ Args syntax
      - false
      - Boolean
      - When set to ``false``, the sql is executed without checking metadata dependencies.
+   * - read_only
+     - false
+     - Boolean
+     - When set to ``true``, the query will be run in ``READ ONLY`` transaction access mode which means only ``select`` queries will be successful. This flag ensures that the graphql schema is not refreshed and is hence highly performant.
```
marionschleifer:
```suggestion
     - When set to ``true``, the query will be run in ``READ ONLY`` transaction access mode which means only ``select`` queries will be successful. This flag ensures that the GraphQL schema is not modified and is hence highly performant.
```

PR #3372, opened 2019-11-18
add local dev note for auth0 via ngrok

tirumaraiselvan:
### Description

Add a note on how to sync users while developing locally.

### Affected components

- [x] Docs

### DIFF
--- a/docs/graphql/manual/guides/integrations/auth0-jwt.rst
+++ b/docs/graphql/manual/guides/integrations/auth0-jwt.rst
@@ -201,3 +201,8 @@ That’s it! This rule will be triggered on every successful signup/login and sy
 .. note::
 
    We need to use an ``upsert`` operation here because social logins do not distinguish between sign-up and login. Hence, we need to run this rule every time a successful login is made and do nothing if the user already exists.
+
+
+.. admonition:: Local dev with Auth0 rules
+
+   The sync step will require a reachable endpoint to Hasura and this is not possible in localhost. You can use ``ngrok`` or similar services to expose your endpoint with a public endpoint temporarily.
--- a/docs/graphql/manual/guides/integrations/auth0-jwt.rst
+++ b/docs/graphql/manual/guides/integrations/auth0-jwt.rst
@@ -205,4 +205,4 @@ That’s it! This rule will be triggered on every successful signup/login and sy
 
 .. admonition:: Local dev with Auth0 rules
 
-   The sync step will require a reachable endpoint to Hasura and this is not possible in localhost. You can use ``ngrok`` or similar services to expose your endpoint with a public endpoint temporarily.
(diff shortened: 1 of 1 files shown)


docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -201,3 +201,8 @@ That’s it! This rule will be triggered on every successful signup/login and sy
 .. note::
 
    We need to use an ``upsert`` operation here because social logins do not distinguish between sign-up and login. Hence, we need to run this rule every time a successful login is made and do nothing if the user already exists.
+
+
+.. admonition:: Local dev with Auth0 rules
```
marionschleifer:
```suggestion
.. admonition:: Note
```

docs/graphql/manual/guides/integrations/auth0-jwt.rst
```diff
@@ -201,3 +201,8 @@ That’s it! This rule will be triggered on every successful signup/login and sy
...
+
+   The sync step will require a reachable endpoint to Hasura and this is not possible in localhost. You can use ``ngrok`` or similar services to expose your endpoint with a public endpoint temporarily.
```
marionschleifer:
```suggestion
   When developing locally with Auth0 rules, the sync step will require a reachable endpoint to Hasura which is not possible in localhost. You can use ``ngrok`` or similar services to expose your endpoint with a public endpoint temporarily.
```

docs/graphql/manual/guides/integrations/auth0-jwt.rst
tirumaraiselvan: Wouldn't it be clearer to notice this if it had a bold title?

docs/graphql/manual/guides/integrations/auth0-jwt.rst
tirumaraiselvan: This is actually ambiguous because you can't develop the *Auth0 rule* locally.

docs/graphql/manual/guides/integrations/auth0-jwt.rst
marionschleifer: Yes, but then it's not consistent with the rest of the docs. It always says `Note`.

tirumaraiselvan: This should go into a subsection then.

marionschleifer: But you are right, it would be clearer.

marionschleifer: Let's keep it the way you proposed.

tirumaraiselvan: BTW I used `admonition:: <title>` here because it allows you to supply a title. Otherwise, the rest of the doc has `::note`

docs/graphql/manual/guides/integrations/auth0-jwt.rst
marionschleifer: Okay then let's keep the bold title and leave this as it was before 🙂

PR #3375, opened 2019-11-18
fix json/jsonb columns as String values in nested returning of a mutation (fix #3365)

### LINKED ISSUE #3365: Append Mutation Returning JSONB as String
I am finding an issue in v1.0.0-beta.10
Problem: Update mutation returns a jsonb field as string type.
The Json data is inserted into the database successfully.
Other (non json) fields in the reponse are ok.
The json field in the reponse is a proper string representation of the json and can be successfully parsed into an object with JSON.parse()

```
mutation update_something($id: Int!, $value: jsonb!) {
  update_something(where: { id: { _eq: $id } }, _append: { something_jsonb: $value }) {
    affected_rows
    returning {
      ...fields
    }
  }
}

response:
{"data":{"update_something":{"affected_rows" : 1, "returning" : [{"id":2,"something_jsonb":"{\"key1\": [{\"type\": \"id\", \"field\": \"id\", \"headerName\": \"ID\",   ...
```

rakeshkky:
### Description

PR #3198 caused the JSON/JSONB values returning as Strings when a mutation is performed with returning nested fields.
query:-
```graphql
mutation {
  delete_author(where: {id: {_eq: 3}}){
    affected_rows
    returning{
      id
      name
      info
      articles{
        id
        title
        content
      }
    }
  }
}
```
Response:-
```json
{
  "data": {
    "delete_author": {
      "affected_rows": 1,
      "returning": [
        {
          "id": 3,
          "name": "Author 3",
          "info": "{\"age\": 23}",
          "articles": []
        }
      ]
    }
  }
}
```
 
This PR fixes the same.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

Fix #3365 

### Solution and Design

The `parsePGScalarValue` is used to parse the Postgres column values correctly. The change in #3198 is made to select the mutated columns as Text values except for geometry/geography types. JSON/JSONB column values are parsed as JSON Strings and not as Postgres literals, which gives those values as Strings. A special `parseJSONStringAsLiteral` function is introduced, which forces all String values as Postgres literals. 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### Steps to test and verify

Reproduce #3365, the json/jsonb columns shouldn't be returned as strings.


server/src-lib/Hasura/RQL/DML/Mutation.hs
```diff
@@ -90,8 +98,9 @@ mutateAndFetchCols qt cols (cte, p) strfyNum =
 
 mkSelCTEFromColVals
   :: (MonadError QErr m)
-  => QualifiedTable -> [PGColumnInfo] -> [ColVals] -> m S.CTE
-mkSelCTEFromColVals qt allCols colVals =
+  => (PGColumnType -> Value -> m (WithScalarType PGScalarValue))
+  -> QualifiedTable -> [PGColumnInfo] -> [ColVals] -> m S.CTE
+mkSelCTEFromColVals parseFn qt allCols colVals =
```
lexi-lambda: Just for my own understanding: could you explain to me from a high-level what this function is doing? When I was first looking through this diff, I was slightly confused, because it seems like we’re parsing the values (using `parseFn`) but then immediately re-encoding them (using `toTxtValue`). It seems like `mutateAndSel` is doing that as part of some kind of two-pass query strategy, but it’s not quite clicking for me what the reason for doing that is.

rakeshkky: To be honest, this diff is not so informative. I'll give you a bit of context. 

`toTxtValue` will encode any `PGScalarValue` to literal values. I thought it is enough and went for #3198. In case of json/jsonb values, the json string we're getting, let's say some `String t` where `t` is json string, is parsed into `PGScalarValue` as `PGValJSON (String t)`.  The `toTxtValue` converts `PGValJSON (String t)` to json encoding of a string value. The `{"age": 23}` is converted to something like `"{\"age\": 23}"`. In `mutateAndSel` function, I'm enforcing all `String t` to `PGValText t`, hence `toTxtValue` encodes correctly for json/jsonb values. 

I hope you understand the above explanation. I'm happy to know any improvements to the current fix.

lexi-lambda: That definitely *is* helpful, so thank you for that explanation, but it actually doesn’t answer my original question (which is only semi-related to this pull request). Specifically, I’m trying to understand what `mkSelCTEFromColVals` is doing in the first place, from a high level. 🙂 Sorry if I didn’t make that very clear!

rakeshkky: Okay :slightly_smiling_face: . The `mkSelCTEFromColVals` will generate a `SELECT` query, that is used inside a `WITH ....` statement. The `SELECT` statement look like
```sql
SELECT * FROM VALUES ((col1-val, col2-val, col3-val), .....) AS "table_name" (col1, col2, col3)
```
The `parseFn` converts JSON `value` to `col-val`.

lexi-lambda: I probably should have been more clear: I can mostly make out *what* it does, but what I’m less sure about is *why*. If `mutateAndFetchCols` does the actual mutation query, and it returns a bunch of values, why don’t we parse them directly from that result? Why do we send them back through Postgres another time?

rakeshkky: >  Why do we send them back through Postgres another time?

To join those values and fetch associated relationships in `returning`. Note that, this code path is executed only when relationships are selected in `returning` (See `hasNestedFld $ _mFields mut`) 

I'm happy to refactor this part of code to make it much clear and self-descriptive. :)

lexi-lambda: Ahh, I see—the reason we have to do an extra `SELECT` is the information can be in related tables that aren’t actually part of the `INSERT`/`UPDATE`. Is that right?

In that case, it seems like we either (1) need to make sure our parsing logic for SQL values is round-trippable, or (2) we should avoid parsing the values we get from Postgres the first time and just store them in the textual format Postgres already knows about. Is that feasible? The approach in this PR seems a little indirect to me.

rakeshkky: The (1) is possible only if we have the context of array types which #2243 is meant to solve. I think this PR is solving (2). We're fetching column values as Strings (except for geometry/geography types) and storing them as `PGValText`. The `parseJSONStringAsLiteral` is doing the job.

lexi-lambda: > I think this PR is solving (2). We're fetching column values as Strings (except for geometry/geography types) and storing them as `PGValText`. The `parseJSONStringAsLiteral` is doing the job.

Okay, I think that helps me to understand better what’s going on. A followup question: why are we treating the geometry/geography types differently? They have a textual representation as well, right? So it seems as though we ought to be able to treat everything uniformly.

rakeshkky: Yes, we can use the textual representation for geometry/geography columns. Ideally, the returned columns are `HashMap PGCol Text`. I'll make the necessary change in next commit.

rakeshkky: Now geometry/geography values are stored as Strings via https://github.com/hasura/graphql-engine/pull/3375/commits/0f484df408336bb9827ad5bae845d98e8ce24594. Please approve if everything is OK. :slightly_smiling_face:

rakeshkky: @lexi-lambda Updated the PR with the latest master. Please review. :slightly_smiling_face:

PR #3393, opened 2019-11-21
export metadata without nulls, empty arrays & default values

rakeshkky:
### Description

The response of `export_metadata` query API is simplified by omitting fields with an empty array, nulls and default values.

### Affected components

- [x] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [x] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

N/A

### Solution and Design

- A brand new module `Data.Aeson.Ordered` to enable encoding JSON with order
- Separate functions to encode the metadata types to ordered JSON.
- Move types from `Hasura.RQL.DDL.Metadata` to `Hasura.RQL.DDL.Metadata.Types`.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL
 
#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/RQL/DDL/Metadata/Types.hs
```diff
@@ -0,0 +1,434 @@
...
+    remoteSchemasPair = remoteSchemas >>=
+                        listToMaybeOrdPair "remote_schemas" remoteSchemaQToOrdJSON
+
+    queryCollectionsPair = queryCollections >>=
+                           listToMaybeOrdPair "query_collections" createCollectionToOrdJSON
+
+    allowlistPair = allowlist >>= listToMaybeOrdPair "allowlist" AO.toOrdered
+
+tableMetaToOrdJSON :: TableMeta -> AO.Value
+tableMetaToOrdJSON ( TableMeta
```
lexi-lambda: For all these little functions, I’d consider bumping them two indentation stops to be in the `where` clause of `replaceMetadataToOrdJSON`, just to emphasize that they’re really only to implement `replaceMetadataToOrdJSON`, as is mentioned in https://github.com/hasura/graphql-engine/pull/2175#issuecomment-514081273.

server/src-lib/Hasura/RQL/DDL/Metadata/Types.hs
```diff
@@ -0,0 +1,434 @@
...
+ deriving(Show, Eq, Lift)
+$(deriveToJSON defaultOptions ''DropInconsistentMetadata)
+
+instance FromJSON DropInconsistentMetadata where
+  parseJSON _ = return DropInconsistentMetadata
+
+-- Functions enable encoding Metadata types and their dependants to ordered JSON
+
+replaceMetadataToOrdJSON :: ReplaceMetadata -> AO.Value
+replaceMetadataToOrdJSON ( ReplaceMetadata
```
lexi-lambda: Could you maybe add a short comment here about what invariants we maintain about order and why? It isn’t immediately clear to me just from reading the code what matters and what doesn’t.

My understanding is that we’re doing this so that metadata export is deterministic, so if you export the same thing twice, you get the same results. That makes complete sense, but it raises questions like: do we guarantee any consistency about ordering across versions of `graphql-engine`? If so, which invariants? And if not, it would be helpful to explicitly say so just so that people reading/modifying the code don’t worry about it.

rakeshkky: The deterministic export of metadata is already achieved via #3230. This PR intends to remove fields with `[]`/`null`/defaults in exported metadata. And, the order of keys in an object is "hardcoded" using `Data.Aeson.Ordered` module since there's no way to determine the order of keys in `Data.Aeson`.

The CLI based developer experience needs the exported metadata should be precise and clean. The developer should be able to edit the exported YAML file and apply them. In an offline discussion between me, @coco98 and @arvi3411301 we decided the order of the keys in `ReplaceMetadata`, `TableMeta` etc. and the same followed here. 

> do we guarantee any consistency about ordering across versions of graphql-engine?

We do guarantee the order of keys in the exported metadata. I'll add the necessary description as comments.

I think there are 2 kinds of "order" we could consider. 
1. The order of items in an array. (solved in #3230)
2. The order of keys in an object. (In this PR)

lexi-lambda: Okay, thanks, that helps! I’ve tweaked the comment slightly to explicitly mention the fact that the ordering should be maintained across versions; feel free to change it if you think it should be worded differently.

Your note that the encoding should be roundtrippable got me wondering, though: do you think it would make sense to add some quickcheck tests to test that property?

rakeshkky: Adding a quick check to test the property makes sense. I'll add them in next commit.

lexi-lambda: > To add a quick check, we need to defined generator `Gen a` for all `ReplaceMetadata` dependent types. A lot of code to write. Is there a way to generate `Gen a` with a simple template haskell or type class deriving?. I'm looking at the `Hedgehog` property testing package.

Good question. Neither QuickCheck nor hedgehog provide anything like that out of the box, since they (somewhat reasonably) point out that the choice of generator for each type is important for getting value out of the tests. That said, I think in many cases, you really do just want to use a generic generator for each field of a record type, for example.

The QuickCheck documentation links to two packages that can do that sort of thing: [testing-feat](http://hackage.haskell.org/package/testing-feat) and [generic-random](http://hackage.haskell.org/package/generic-random); I also found the [generic-arbitrary](http://hackage.haskell.org/package/generic-arbitrary) package. In the hedgehog realm, there’s [hedgehog-generic](http://hackage.haskell.org/package/hedgehog-generic), but I’m skeptical of that package’s utility: hedgehog doesn’t have an `Aribitrary` class, so a generic generator is “deeply” generic. I can’t imagine that being what you want very often.

I don’t actually have a ton of experience using property-based testing for anything non-trivial, so I’m not sure what the right thing to do is here. Maybe @paf31 has some thoughts?

paf31: I don't have a lot of useful input here, but I think it would be nice to avoid adding many `Arbitrary` instances in the `src-lib` area of the code, especially when they'll probably end up being specific to this sort test. Instead, we can use `genericArbitrary` or some other method of getting values of type `Gen a`, entirely within the test suite, and then use the `Testable a => Testable (Gen a)` instance to pass the composed generator itself to `quickCheck`.

I'm not sure to what extent that is common knowledge or not, but it seemed worth mentioning.

lexi-lambda:
Since the tests still seem to be running out of memory even with such a small number of cases (which is a bit strange!), I think it probably does make sense to do what I suggested: let’s move the tests that don’t require Postgres into a separate CI job, and we can bump up the resource limits on that job slightly if necessary.

lexi-lambda:
@rakeshkky Do you know why the CLI tests seem to think the working tree is dirty?

rakeshkky:
> @rakeshkky Do you know why the CLI tests seem to think the working tree is dirty?

PR #3042, opened 2019-10-06
allow custom mutations through actions

0x777:
**Needs https://github.com/hasura/graphql-parser-hs/pull/19 to be merged**

I'll be pushing changes to this PR over the next couple of days. A high level overview of actions can be found in this PR's [docs](https://deploy-preview-3042--hasura-docs.netlify.com/graphql/manual/actions/index.html). 

### Updated: 23/10/2019
Tasks done:
- [x] defining custom types
- [x] sync actions
- [x] async actions
- [x] metadata export/import
- [x] http endpoint as action handlers
- [x] relationships

Remaining:
- [ ] pg functions as action handlers
- [ ] async action relationships

`update_action`:-
```yaml
type: update_action
args:
  name: action-name
  definition: action-definition # will replaces existing one
```

The webhook is called over `POST`. The payload looks like:-
```json
{
  "session_variables": {
    "x-hasura-user-id": "1234"
  },
  "input": {
    "arg1": "value1",
    "arg2": "value2"
  },
  "action": {
    "name": "action-name"
  }
}
```

**The webhook response:**
- On success, should return a `2xx` code with JSON body whose value is of the defined output type
- Should return a `4xx` code for an error that should be propagated back to the client. The response JSON body is an object with a compulsory `message` field (text) and optional `code` field (text)
- All other status codes are treated as internal errors and nothing is propagated to the client. The details should be logged as `internal` in `error` field and part of extensions in response for the admin role. 

**Single row mutations:-** Implemented the RFC https://github.com/hasura/graphql-engine/issues/3731

PR #3394, opened 2019-11-21
Support incremental builds of the schema cache (fix #3354)

### LINKED ISSUE #3354: Migrations drop and recreate hdb_views dozens of times, causing max_locks_per_transaction to be exceeded
Updating from v1.0.0-beta.6 to v1.0.0-beta.7 results in this error when running `sudo docker-compose up -d`: `ERROR: manifest for hasura/graphql-engine:v1.0.0-beta.7 not found: manifest unknown: manifest unknown`

Updating from v1.0.0-beta.6 to v1.0.0-beta.8 results in `502 Bad Gateway` in the browser when accessing the console. The ui fails too because all graphql calls get a 502 response.

Updating from v1.0.0-beta.6 to v1.0.0-beta.9 or v1.0.0-beta.10 results in same error.

Reverting to v1.0.0-beta.6 instantly works after `sudo docker-compose up -d`.

What am I doing wrong?

lexi-lambda:
### Description

As of #2379, we drop and recreate all system metadata using a bulk query during schema migrations. This is satisfyingly decisive, but it can lead to a performance problem because it can cause the schema cache to be rebuilt many times in a single transaction, which created the problem described in #3354.

A simple fix would be to tweak the migrations to avoid rebuilding the schema cache so many times, but I’ve been wanting to avoid doing that. Currently, we go to a lot of effort to avoid rebuilding the schema cache in places, but this duplicates a lot of logic, and as that logic becomes increasingly complicated, it becomes extremely difficult to keep in sync. Ideally, we’d like to have a single source of truth for how to build the schema cache and what rules its contents must follow, but we need to do it without incurring such a huge performance cost.

The idea I’ve been thinking about for a while is to *incrementalize* the process of building the schema cache, in the [incremental compilation](https://en.wikipedia.org/wiki/Incremental_compiler) sense. Essentially, the idea is to add dependency tracking to the build process, so if some part of the metadata doesn’t change, we can avoid rebuilding it. Tracking all those dependencies by hand would be tricky and probably about as error-prone as our existing solution, so this PR adds some scaffolding for *automatic incrementalization* that does a lot of the work automatically. More details below.

### Affected components

- Server

### Related Issues
#2379, #3354

### Solution and Design

The good news about this change is that it actually *removes* a lot of logic, so the resulting implementation is simpler. However, the bad news is that it is a fairly invasive refactor, and it touches all parts of the code that modify the catalog or schema cache. There’s also a lot of new supporting code to help with the incremental build process, especially since arrows have almost no standard library to speak of.

For some more details, see the “For reviewers” section below.

#### TODO

- [x] Implement incrementalization abstraction
- [x] Refactor `buildSchemaCache` to remove imperative updates
- [x] Update metadata APIs to use `buildSchemaCache` code path
  - [x] tables
  - [x] relationships
  - [x] computed fields
  - [x] permissions
  - [x] event triggers
  - [x] remote schemas
- [x] Reintegrate proper `SchemaDependency` tracking
- [x] Update Diff.hs logic
- [x] Fix broken tests
- [x] Insert calls to `cache` in the right places
- [x] Use a non-leaky `WriterT` implementation
- [x] Improve performance of catalog fetch query
- [x] Figure out how to handle foreign key dependencies properly
- [x] ~~Improve performance of `bulk` queries~~ Performance is good enough, so punting this to future work
- [x] Handle conflicting GraphQL field names
- [x] Track dependencies from permissions to tables in a more fine-grained way
- [x] ~~Track error codes in inconsistent metadata objects~~ Decided not to do this.
- [x] Final cleanup of loose ends

#### Catalog upgrade

- [x] **Yes**

#### Metadata
- [x] No

#### GraphQL

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [x] Change in error code
     

  3. Logging

### For reviewers

This is a big change, so here are some tips on where to start and what to focus on if you’d like to review it:

1. First, familiarize yourself very briefly with arrows.

   1. A good introduction to the high-level concepts is [Understanding arrows](https://en.wikibooks.org/wiki/Haskell/Understanding_arrows) from the Haskell wikibook, but it doesn’t cover `proc` notation.

   2. For a primer on `proc` notation, the [Arrow syntax](https://www.haskell.org/arrows/syntax.html) page on the Haskell website includes some basic information, and [the relevant section of the GHC User’s Guide](https://downloads.haskell.org/ghc/8.8.1/docs/html/users_guide/glasgow_exts.html#arrow-notation) has more details.

   3. If you want something really thorough, you can read Paterson’s paper introducing arrow notation, [A New Notation for Arrows](http://www.staff.city.ac.uk/~ross/papers/notation.html).

   You don’t need to thoroughly understand arrows to understand most of the code in this pull request, and in fact if you squint `proc` notation looks quite a bit like plain `do` notation.

2. Having done that, take a quick look at [`Hasura.Incremental`](https://github.com/lexi-lambda/graphql-engine/blob/3354-faster-metadata-migrations/server/src-lib/Hasura/Incremental.hs) and read the documentation comments. You don’t need to worry about the implementation to understand the rest of the code. You can also look at [`Hasura.IncrementalSpec`](https://github.com/lexi-lambda/graphql-engine/blob/3354-faster-metadata-migrations/server/src-test/Hasura/IncrementalSpec.hs) for a couple small examples.

3. Now you can start reading through the bulk of the changes. Start with [`Hasura.RQL.DDL.Schema.Cache`](https://github.com/lexi-lambda/graphql-engine/blob/3354-faster-metadata-migrations/server/src-lib/Hasura/RQL/DDL/Schema/Cache.hs), focusing on `buildSchemaCacheRule`, which defines how to build a `SchemaCache` from catalog metadata.

4. Look at [`Hasura.RQL.DDL.Schema.Cache.Dependencies`](https://github.com/lexi-lambda/graphql-engine/blob/3354-faster-metadata-migrations/server/src-lib/Hasura/RQL/DDL/Schema/Cache/Dependencies.hs), which implements dependency resolution to ensure integrity of the schema cache.

5. Read [`Hasura.RQL.Types.SchemaCache.Build`](https://github.com/lexi-lambda/graphql-engine/blob/3354-faster-metadata-migrations/server/src-lib/Hasura/RQL/Types/SchemaCache/Build.hs), which defines the basic interface for triggering cache rebuilds.

6. The rest of the changes are mostly just adjustments to the rest of the code to accommodate the modified schema cache construction process.


rakeshkky:
Hi @lexi-lambda, 

I tried writing code for [actions](https://github.com/hasura/graphql-engine/pull/3042) metadata API with new arrow syntax by merging this branch into actions. I'm happy that now we can concentrate only on building schema cache, exporting metadata and handling dependencies on `run_sql` while adding any new metadata feature. 

The following are the bottlenecks for anyone newly working on metadata DDL queries (/v1/query):-
1. Arrow syntax - banana brackets, `proc` notation etc.
2. IDE support for arrow syntax (As of now I'm using spacemacs with Haskell layer which is not auto-indenting inside arrow syntax)

I think It'd be good if we keep the internal code for `Incremental` (and few similar modules) as a separate package which is not bound to change frequently. Probably in a subdirectory in `server` folder which will have a `make` rule to build and host haddock documentation at localhost. Interested ones may refer to the documentation (to know the magic!).

lexi-lambda:
> I think It'd be good if we keep the internal code for `Incremental` (and few similar modules) as a separate package which is not bound to change frequently. Probably in a subdirectory in `server` folder which will have a `make` rule to build and host haddock documentation at localhost. Interested ones may refer to the documentation (to know the magic!).

@rakeshkky Do you think there’s an advantage to keeping those things in a separate package rather than just in a separate directory in the same package?

PR #3440, opened 2019-11-28
check db connection in healthz endpoint (close #2645)

### LINKED ISSUE #2645: implement readiness and liveness proble endpoints
I'm using `hasura/graphql-engine:v1.0.0-beta.3` with Kubernetes.

In the Kubernetes deployment for a Hasura instance, I see that the `/healthz` endpoint is returning `500` when there are inconsistencies in the metadata.

Logs:

```
{"timestamp":"2019-07-31T05:43:21.049+0000","level":"warn","type":"metadata","detail":{"message":"Inconsistent Metadata!","info":{"objects":[{"definition":{"using":{"foreign_key_constraint_on":{"column":"city_id","table":"offers"}},"name":"offers","comment":null,"table":"cities"},"reason":"column \"city_id\" does not exist","type":"array_relation"},{"definition":{"using":{"foreign_key_constraint_on":{"column":"store_id","table":"offers"}},"name":"offers","comment":null,"table":"stores"},"reason":"column \"store_id\" does not exist","type":"array_relation"},{"definition":{"using":{"foreign_key_constraint_on":"store_id"},"name":"store","comment":null,"table":"offers"},"reason":"column \"store_id\" does not exist","type":"object_relation"},{"definition":{"using":{"foreign_key_constraint_on":"city_id"},"name":"city","comment":null,"table":"offers"},"reason":"column \"city_id\" does not exist","type":"object_relation"}]}}}
```

Quoting @shahidhk from the Discord discussion:

```
The pod is starting, but the health endpoint is more like a liveness probe
```

tirumaraiselvan:
### Description

Currently, /healthz endpoint does not check for db connectivity. It seems vital to check for db connectivity in the health check.

Also, /healthz throws 500 when inconsistent objects are detected. This seems not really correct because hasura starts and functions even with inconsistent objects. We should throw a WARN here instead.

### Affected components

- [x] Server
- [x] Docs

### Related issues

Not exactly sure what this issue means but I think we should close this with this PR
https://github.com/hasura/graphql-engine/issues/2645

Should we add some check here which checks whether graphql API is working?


server/src-lib/Hasura/Server/App.hs
```diff
@@ -530,9 +531,12 @@ httpApp corsCfg serverCtx enableConsole consoleAssetsDir enableTelemetry = do
     -- Health check endpoint
     Spock.get "healthz" $ do
       sc <- liftIO $ getSCFromRef $ scCacheRef serverCtx
-      if null $ scInconsistentObjs sc
-        then Spock.setStatus HTTP.status200 >> Spock.lazyBytes "OK"
-        else Spock.setStatus HTTP.status500 >> Spock.lazyBytes "ERROR"
+      dbOk <- checkDbConnection
+      if dbOk
```
tirumaraiselvan: There was no content-type set when using `lazyBytes`, also no point using lazyBytes for such small strings.

PR #3458, opened 2019-12-02
fix updating a column with multiple operators causing postgres query error (fix #3432)

### LINKED ISSUE #3432: updating same column with different operators causing postgres query error
If we try to update a column through an update mutation using multiple (> 1) operators, then it results in Postgres query error.

For a simple `article` table:-
Query:
```graphql
mutation {
  update_article(
    _set: {id: 1}
    _inc: {id: 2}
    where: {author_id: {_eq: 1}}
  ){
    affected_rows
  }
}
```

Response:
```json
{
  "errors": [
    {
      "extensions": {
        "internal": {
          "statement": "WITH \"public_article__mutation_result_alias\" AS (UPDATE \"public\".\"article\" SET \"id\" = $1,\"id\" = (\"id\"+('2')::integer)  WHERE (('true') AND (((((\"public\".\"article\".\"author_id\") = ($2)) AND ('true')) AND ('true')) AND ('true'))) RETURNING * ) SELECT  json_build_object('affected_rows', (SELECT  COUNT(*)  FROM \"public_article__mutation_result_alias\"      ) )        ",
          "prepared": true,
          "error": {
            "exec_status": "FatalError",
            "hint": null,
            "message": "multiple assignments to same column \"id\"",
            "status_code": "42601",
            "description": null
          },
          "arguments": [
            "(Oid 23,Just (\"\\NUL\\NUL\\NUL\\SOH\",Binary))",
            "(Oid 23,Just (\"\\NUL\\NUL\\NUL\\SOH\",Binary))"
          ]
        },
        "path": "$",
        "code": "unexpected"
      },
    }
  ]
}
```

Ideally, the server should make a validation check if a column is being updated using multiple operators.

rakeshkky:
### Description

Using a column in multiple operators in an update mutation generates invalid SQL statement and causing Postgres query error. This PR makes a validation check and throws an appropriate error.

Query:-
```graphql
mutation {
  update_article(
    _set: {id: 1}
    _inc: {id: 2}
    where: {author_id: {_eq: 1}}
  ){
    affected_rows
  }
}
```

Error:-
```json
{
  "errors": [
    {
      "extensions": {
        "internal": {
          "statement": "WITH \"public_article__mutation_result_alias\" AS (UPDATE \"public\".\"article\" SET \"id\" = $1,\"id\" = (\"id\"+('2')::integer)  WHERE (('true') AND (((((\"public\".\"article\".\"author_id\") = ($2)) AND ('true')) AND ('true')) AND ('true'))) RETURNING * ) SELECT  json_build_object('affected_rows', (SELECT  COUNT(*)  FROM \"public_article__mutation_result_alias\"      ) )        ",
          "prepared": true,
          "error": {
            "exec_status": "FatalError",
            "hint": null,
            "message": "multiple assignments to same column \"id\"",
            "status_code": "42601",
            "description": null
          },
          "arguments": [
            "(Oid 23,Just (\"\\NUL\\NUL\\NUL\\SOH\",Binary))",
            "(Oid 23,Just (\"\\NUL\\NUL\\NUL\\SOH\",Binary))"
          ]
        },
        "path": "$",
        "code": "unexpected"
      },
    }
  ]
}
```

### Affected components

- [x] Server
- [x] Tests

### Related Issues

Fix #3432 

### Solution and Design

Check for same columns present in different operators and if any such column found, raise a validation exception by plumbing in operator names in error message.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### Steps to test and verify

Reproduce the issue #3432, instead of unexpected, the server should throw validation error.

### DIFF
--- a/server/src-lib/Data/Sequence/NonEmpty.hs
+++ b/server/src-lib/Data/Sequence/NonEmpty.hs
@@ -5,13 +5,11 @@ module Data.Sequence.NonEmpty
   , init
   , head
   , toSeq
-  , toList
-  , length
   ) where
 
 import qualified Data.Foldable as F
 import qualified Data.Sequence as Seq
-import           Prelude       (Eq, Int, Show, fst, (.))
+import           Prelude       (Eq, Show, fst, (.))
 
 infixr 5 <|
 infixl 5 |>
@@ -19,6 +17,9 @@ infixl 5 |>
 newtype NESeq a
   = NESeq { unNESeq :: (a, Seq.Seq a)}
   deriving (Show, Eq)
+
+instance F.Foldable NESeq where
+  foldr f v = F.foldr f v . toSeq
 
 init :: a -> NESeq a
 init a = NESeq (a, Seq.empty)
@@ -34,9 +35,3 @@ v <| (NESeq (h, l)) = NESeq (v, h Seq.<| l)
 
 toSeq :: NESeq a -> Seq.Seq a
 toSeq (NESeq (v, l)) = v Seq.<| l
-
-toList :: NESeq a -> [a]
-toList = F.toList . toSeq
-
-length :: NESeq a -> Int
-length = Seq.length . toSeq
--- a/server/src-lib/Data/Sequence/NonEmpty.hs
+++ b/server/src-lib/Data/Sequence/NonEmpty.hs
@@ -5,10 +5,13 @@ module Data.Sequence.NonEmpty
   , init
   , head
   , toSeq
+  , toList
+  , length
   ) where
 
+import qualified Data.Foldable as F
 import qualified Data.Sequence as Seq
-import           Prelude       (Eq, Show, fst, (.))
+import           Prelude       (Eq, Int, Show, fst, (.))
 
 infixr 5 <|
 infixl 5 |>
@@ -31,3 +34,9 @@ v <| (NESeq (h, l)) = NESeq (v, h Seq.<| l)
 
 toSeq :: NESeq a -> Seq.Seq a
 toSeq (NESeq (v, l)) = v Seq.<| l
+
+toList :: NESeq a -> [a]
+toList = F.toList . toSeq
+
+length :: NESeq a -> Int
+length = Seq.length . toSeq
--- a/server/src-lib/Hasura/GraphQL/Resolve/Mutation.hs
+++ b/server/src-lib/Hasura/GraphQL/Resolve/Mutation.hs
@@ -8,23 +8,26 @@ module Hasura.GraphQL.Resolve.Mutation
 import           Data.Has
 import           Hasura.Prelude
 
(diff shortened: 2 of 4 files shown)


server/src-lib/Data/Sequence/NonEmpty.hs
```diff
@@ -31,3 +34,9 @@ v <| (NESeq (h, l)) = NESeq (v, h Seq.<| l)
 
 toSeq :: NESeq a -> Seq.Seq a
 toSeq (NESeq (v, l)) = v Seq.<| l
+
+toList :: NESeq a -> [a]
+toList = F.toList . toSeq
+
+length :: NESeq a -> Int
+length = Seq.length . toSeq
```
lexi-lambda: Might it be better to implement `Foldable` for `NESeq`? The implementation would be pretty easy, since you can get away with just implementing `foldr`, which you could do as just `foldr f v = F.foldr f v . toSeq`. Then you’d get `toList` and `length` for free, since those work on any `Foldable`. 😄

server/src-lib/Hasura/GraphQL/Resolve/Mutation.hs
```diff
@@ -115,53 +118,67 @@ convertUpdateP1
...
+            columnsWithMultiOps = map (second (map fst . NESeq.toList)) $
+                                  filter ((>1) . NESeq.length . snd) itemTuples
+        -- A column shouldn't be present in more than one operator.
+        -- If present, then generated UPDATE statement throws unexpected query error
+        if null columnsWithMultiOps then
+          pure $ resolvedPreSetItems <> map (second (snd . NESeq.head)) itemTuples
+        else throwVE $ "column found in multiple operators; "
+                       <> T.intercalate ". "
+                          (map (\(col, ops) -> col <<> " in " <> showNames ops)
+                               columnsWithMultiOps)
```
lexi-lambda: This generally looks good to me, but one suggestion I have to make this a little more data-/type-driven would be to take advantage of the result of `OMap.groupTuples` more aggressively. That operation produces a map with a shape like `InsOrdHashMap k (NESeq v)`, and from there you can very naturally write a function of type

```haskell
(MonadError QErr m) => k -> NESeq v -> m v
```

which raises an error if there are multiple elements and returns a single element otherwise. Then you can hand that function to `OMap.traverseWithKey` to get an `OMap k v` out, which is, in effect, the “parsed” representation of that input. 🙂

rakeshkky: Yes, I can do `(MonadError QErr m) => k -> NESeq v -> m v`, but it'd stop at the occurrence of first multiple elements. My point is to collect all those multiple elements (`columnsWithMultipleOps`) and throw an error including all of them in the error message.

PR #3465, opened 2019-12-03
add read_only flag to run_sql select queries from console

rikinsk:
### Description

Add `read_only` flag for all `run_sql` select queries run by console

**Blocker:**

Currently the largest and most common queries run by the console cannot use the flag as they are run in bulk along with one normal `select` query 

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/console/src/components/Services/Data/DataActions.js
+++ b/console/src/components/Services/Data/DataActions.js
@@ -68,8 +68,8 @@ const useCompositeFnsNewCheck =
 const compositeFnCheck = useCompositeFnsNewCheck
   ? 'c'
   : {
-      $ilike: '%composite%',
-    };
+    $ilike: '%composite%',
+  };
 
 const initQueries = {
   schemaList: {
@@ -285,6 +285,7 @@ const loadSchema = configOptions => {
       }
     }
 
+    // TODO: make read_only sub-queries. blocked as query 2 is v1/query select
     const body = {
       type: 'bulk',
       args: [
@@ -563,12 +564,14 @@ const getBulkColumnInfoFetchQuery = schema => {
     type: 'run_sql',
     args: {
       sql: fetchColumnTypesQuery,
+      read_only: true,
     },
   };
   const fetchTypeDefaultValues = {
     type: 'run_sql',
     args: {
       sql: fetchColumnDefaultFunctions(schema),
+      read_only: true,
     },
   };
 
@@ -576,6 +579,7 @@ const getBulkColumnInfoFetchQuery = schema => {
     type: 'run_sql',
     args: {
       sql: fetchColumnCastsQuery,
+      read_only: true,
     },
   };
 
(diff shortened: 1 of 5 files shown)


tirumaraiselvan:
@rikinsk I thought event-triggers had more than one bulk query which was read_only. But this PR only changes one query in event-trigger/utils.js

console/src/components/Services/EventTrigger/utils.js
```diff
@@ -140,6 +140,7 @@ FROM
     type: 'run_sql',
     args: {
       sql: runSql,
+      read_only: true,
```
tirumaraiselvan: you missed checkFeatureSupport here?

console/src/components/Services/Data/DataActions.js
```diff
@@ -563,19 +563,22 @@ const getBulkColumnInfoFetchQuery = schema => {
     type: 'run_sql',
     args: {
       sql: fetchColumnTypesQuery,
+      read_only: true,
```
tirumaraiselvan: don't you need `checkFeatureSupport` here?

console/src/components/Services/Data/utils.js
```diff
@@ -375,6 +377,7 @@ FROM
     type: 'run_sql',
     args: {
       sql: runSql,
+      read_only: checkFeatureSupport(READ_ONLY_RUN_SQL_QUERIES) ? true : false,
```
tirumaraiselvan: if instead of `false` we set it to `undefined` will this key be dropped at the client side itself?

console/src/components/Services/EventTrigger/utils.js
rikinsk: This is a single query. so nno need for version check

console/src/components/Services/Data/utils.js
rikinsk: No real need to drop it as the readOnly flag is already supported. it's just the select along with with run_sql part that needs to be behind version check

console/src/components/Services/Data/DataActions.js
rikinsk: it's just the select along with with run_sql part that needs to be behind version check

PR #3467, opened 2019-12-03
bulk query should not care about access mode of select or count queries

tirumaraiselvan:
### Description

Fixes the TODO in #3191 and fixes the commented code in #3465 

As RQSelect and RQCount statements could be combined with run_sql queries which mutate data, we had set the tx access mode for them to be READ WRITE to support backwards compatibility in the console.

But the console is doing lots of RQSelect alongside READ ONLY run_sql, which forces the console to continue using run_sql in READ WRITE. This would negate the performance benefit of READ ONLY run_sql in the first place.

This PR fixes that by making RQSelect and RQCount work with both types of access modes.

### Affected components

- [x] Server

### DIFF
--- a/server/src-lib/Hasura/Server/Query.hs
+++ b/server/src-lib/Hasura/Server/Query.hs
@@ -301,31 +301,34 @@ queryNeedsReload (RQV2 qi) = case qi of
   RQV2SetTableCustomFields _ -> True
   RQV2TrackFunction _        -> True
 
--- TODO: RQSelect query should also be run in READ ONLY mode.
--- But this could be part of console's bulk statement and hence should be added after console changes
 getQueryAccessMode :: (MonadError QErr m) => RQLQuery -> m Q.TxAccess
-getQueryAccessMode (RQV1 q) =
-  case q of
-    RQRunSql RunSQL{rTxAccessMode} -> pure rTxAccessMode
-    RQBulk qs -> assertAllTxAccess (zip [0::Integer ..] qs)
-    _                              -> pure Q.ReadWrite
+getQueryAccessMode q = (fromMaybe Q.ReadOnly) <$> getQueryAccessMode' q
   where
-    assertAllTxAccess = \case
-      [] -> throw400 BadRequest "expected atleast one query in bulk"
-      (_i, q1):[] -> getQueryAccessMode q1
-      q1:q2:qs -> assertSameTxAccess q1 q2 >> assertAllTxAccess (q2:qs)
+    getQueryAccessMode' :: (MonadError QErr m) => RQLQuery -> m (Maybe Q.TxAccess)
+    getQueryAccessMode' (RQV1 q') =
+      case q' of
+        RQSelect _ -> pure Nothing -- RQSelect is don't care i.e. works with both types of TxAccess
+        RQCount _ -> pure Nothing  -- RQCount is don't care i.e. works with both types of TxAccess
+        RQRunSql RunSQL{rTxAccessMode} -> pure $ Just rTxAccessMode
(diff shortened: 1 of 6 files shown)


server/src-lib/Hasura/Server/Query.hs
```diff
@@ -301,31 +301,34 @@ queryNeedsReload (RQV2 qi) = case qi of
...
+        RQCount _ -> pure Nothing  -- RQCount is don't care i.e. works with both types of TxAccess
+        RQRunSql RunSQL{rTxAccessMode} -> pure $ Just rTxAccessMode
+        RQBulk qs -> assertAllAccessMode Nothing (zip [0 ..] qs)
+        _                              -> pure $ Just Q.ReadWrite
+      where
+        assertAllAccessMode :: (MonadError QErr m) => Maybe Q.TxAccess -> [(Integer, RQLQuery)] -> m (Maybe Q.TxAccess)
+        assertAllAccessMode expectedAccessMode = \case
+          [] -> throw400 BadRequest "expected atleast one query in bulk"
+          q1:[] -> assertAccessMode expectedAccessMode q1
+          q1:qs -> assertAccessMode expectedAccessMode q1 >>= (flip assertAllAccessMode qs)
```
tirumaraiselvan: would foldM be cuter here?

server/tests-py/queries/v1/bulk/mixed_access_mode.yaml
```diff
@@ -3,15 +3,11 @@ url: /v1/query
 status: 400
 response:
   path: $
-  error: "incompatible access mode requirements in bulk query: $.args[1] requires READ WRITE, $.args[2] requires READ ONLY"
+  error: "incompatible access mode requirements in bulk query, expected access mode: Just READ WRITE but $.args[1] forces Just READ ONLY"
```
lexi-lambda: Including `Just` in the error message doesn’t seem right to me—was that intentional?

server/src-lib/Hasura/Server/Query.hs
lexi-lambda: Yes, I think the way I would expect to see this written is to have a function that checks compatibility between two access modes, returning their least upper bound and raising an error if they’re incompatible. Something like this:

```haskell
reconcileAccessModes :: Maybe Q.TxAccess -> Maybe Q.TxAccess -> m (Maybe Q.TxAccess)
reconcileAccessModes Nothing mode = pure mode
reconcileAccessModes mode Nothing = pure mode
reconcileAccessModes (Just mode1) (Just mode2)
  | mode1 == mode2 -> pure mode1
  | otherwise -> {- raise error -}
```

Then you could pass that function to `foldM` with `Nothing` as the initial element to fold over the list of required access modes.

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -304,32 +304,38 @@ queryNeedsReload (RQV2 qi) = case qi of
...
-            else
-            throw400 BadRequest $ "incompatible access mode requirements in bulk query, "
-            <> "expected access mode: " <> (T.pack $ show expectedAccessMode) <> " but "
-            <> "$.args[" <> (T.pack $ show index') <> "] forces " <> (T.pack $ show accessModeQ)
+        reconcileAccessModeWith expectedMode (i, query) = do
+          queryMode <- getQueryAccessMode' query
+          onLeft (reconcileAccessModes expectedMode queryMode) $ \errMode ->
+            throw400 BadRequest $
+            "incompatible access mode requirements in bulk query, " <>
+            "expected access mode: " <>
```
tirumaraiselvan: hindent auto formatting

PR #3472, opened 2019-12-04
Update server/CONTRIBUTING.md and include manual build/test steps

lexi-lambda:
### Description

This fixes some problems with `CONTRIBUTING.md` for the server code, and it adds detailed manual build steps to document how to run the tests without using `scripts/dev.sh`, should anyone find that helpful.

### Affected components
- [x] Server

### DIFF
--- a/server/.gitignore
+++ b/server/.gitignore
@@ -10,6 +10,7 @@ Pipfile.lock
 *.dyn_o
 *.dyn_hi
 .virtualenv
+.python-venv
 .hpc
 *.tix
 .hsenv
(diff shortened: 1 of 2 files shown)


server/.gitignore
```diff
@@ -10,6 +10,7 @@ Pipfile.lock
 *.dyn_o
 *.dyn_hi
 .virtualenv
+.python-venv
```
jberryman: It doesn't seem super important but I was wondering if we should standardize on a single name for python venvs (`.python-venv` seems good)

lexi-lambda: I don’t have a strong opinion about the particular name, so anything works for me, really. I just put the name I was using in the `.gitignore` file. We could use `.virtualenv` instead, or something else, so let me know if you’d prefer something different!

PR #3481, opened 2019-12-04
cli(migrations-docker): randomize server port (fix #3479)

### LINKED ISSUE #3479: cli-migrations docker-entrypoint.sh should start hasura in a random port during migrations
This is because if you are having `readiness` or `liveness` probes then the readiness probe might succeed but liveness probe might fail later. This is because the docker image starts a metadata only graphql-engine first for migrations, kills that, and finally starts a regular graphql-engine. 

The default to `--server-port` is not ideal for this reason. If the metadata process starts in a random port, then the readinessProbe will not hit the random port and we will be safe.

tirumaraiselvan:
### Description

Randomize the port of the temporary migrations server

### Affected components

- [x] Build System

### Related Issues

#3479


tirumaraiselvan:
Source for random number generator: https://dilbert.com/strip/2001-10-25

PR #3486, opened 2019-12-04
fix insert permission views are not unique for long role names (fix #3444)

### LINKED ISSUE #3444: Permission metadata name cropped after 63 characters
Hi,

I'm currently writing custom migrations, and am generating permissions combinatorially to populate Hasura as manually doing them would take too long. 

A problem I hit is this error:
```
FATA[0006] apply failed: [unexpected] postgres query error ($.args[184].args)
File: '1575004447175_generate_permissions/up.yaml'
{
    "permission": {
        "check": {},
        "columns": "*"
    },
    "role": "supermanadmin.researcheradmin.user.useradmin",
    "table": {
        "name": "QuestionnaireQuestionOnQuestionnaireSubscale",
        "schema": "public"
    }
}
[42P07] FatalError: relation "supermanadmin.researcheradmin.user.useradmin__insert__public__Q" already exists 
```

This appears to fail because it sees that another permission exists (although it's a different table - I have many tables starting with Q, it just happens to crop off at the same position). 

I have a work-around which is, don't make the names of the tables and/or permissions that long, but it would be nice not to hit this in the future.

Cheers,
Patrick

rakeshkky:
### Description

All PostgreSQL identifiers are limited by 63 character length and longer ones are truncated. See [Postgres docs](https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS). 

The server makes use of views to define an insert trigger with check permission. The auto-generated view name is of format `{role}__insert__{table-schema}__{table-name}`. For the really long role names, the generated view names may exceed 63 characters and truncated names may not be unique.

This PR fixes the above issue by using hash string instead of formatted text.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

Fix #3444 

### Solution and Design

Use Blake2b_224 algorithm, which generates 56 character length hash for insert permission view names.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### Steps to test and verify

Define insert permissions with long role names which differ in the tail characters (> 63 characters).

### DIFF
--- a/server/src-lib/Hasura/RQL/DDL/Permission.hs
+++ b/server/src-lib/Hasura/RQL/DDL/Permission.hs
@@ -87,8 +87,11 @@ type CreateInsPerm = CreatePerm InsPerm
 buildViewName :: QualifiedTable -> RoleName -> PermType -> QualifiedTable
 buildViewName qt rn pt = QualifiedObject hdbViewsSchema tableName
   where
+    -- Generate a unique hash for view name from role name, permission type and qualified table.
+    -- See Note [Postgres identifier length limitations].
+    -- Black2b_224 generates 56 character hash. See Note [Blake2b faster than SHA-256].
+    -- Refer https://github.com/hasura/graphql-engine/issues/3444.
     tableName = TableName $ T.pack $ show hash
-    -- Blake2b_224 generates 56 character length hash
     hash :: CH.Digest CH.Blake2b_224 =
       CH.hash $ txtToBs $ roleNameToTxt rn <> "__" <> T.pack (show pt) <> "__" <> qualObjectToText qt
 
--- a/server/src-lib/Hasura/RQL/DDL/Permission.hs
+++ b/server/src-lib/Hasura/RQL/DDL/Permission.hs
@@ -66,6 +66,7 @@ import           Data.Aeson.Casing
 import           Data.Aeson.TH
 import           Language.Haskell.TH.Syntax         (Lift)
 
+import qualified Crypto.Hash                        as CH
 import qualified Data.HashMap.Strict                as HM
 import qualified Data.HashSet                       as HS
 import qualified Data.Text                          as T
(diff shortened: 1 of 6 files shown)


server/src-lib/Hasura/RQL/DDL/Permission.hs
```diff
@@ -84,12 +85,12 @@ type InsPermDef = PermDef InsPerm
...
-buildViewName (QualifiedObject sn tn) rn pt =
-  QualifiedObject hdbViewsSchema $ TableName
-  (roleNameToTxt rn <> "__" <> T.pack (show pt) <> "__" <> snTxt <> "__" <> tnTxt)
+buildViewName qt rn pt = QualifiedObject hdbViewsSchema tableName
   where
-    snTxt = getSchemaTxt sn
-    tnTxt = getTableTxt tn
+    tableName = TableName $ T.pack $ show hash
+    -- Blake2b_224 generates 56 character length hash
+    hash :: CH.Digest CH.Blake2b_224 =
```
lexi-lambda: A thought: it isn’t immediately clear to someone reading this code *why* we’re doing this hashing. This might be a good situation for [something like GHC’s “notes” system](https://www.stackbuilders.com/news/the-notes-of-ghc). The idea is to create a long-form comment somewhere, like this:

```haskell
{- Note [Postgres identifier length limitations]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Postgres truncates identifiers to a maximum of 63 characters by default (see
https://www.postgresql.org/docs/12/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS). This causes
problems for A, B, and C because of X and Y, so we work around it by doing Z. -}
```

Except obviously `A`, `B`, `C`, etc. are replaced with real explanations. :) Then somewhere else in the code you can write

```haskell
-- see Note [Postgres identifier length limitations]
hash :: CH.Digest CH.Blake2b_224 =
```

wherever the note is relevant. That way, explanations for tricky concepts can be centralized in one place, and they can be referenced from other places in the code. I think there is already at least one or two places in `graphql-engine` where we do some work to work around Postgres’s identifier length limitation, so it would be great to put a note somewhere and add comments in all those places referencing it!

PR #3494, opened 2019-12-05
remove beta version checks from console

rikinsk:
### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


console/src/components/Main/Main.js
```diff
@@ -105,12 +91,6 @@ class Main extends React.Component {
     }
   }
```
wawhal: I think we need to `fetchServerConfig` for JWT decoder.

console/src/components/Services/EventTrigger/EventActions.js
```diff
@@ -107,21 +106,14 @@ const loadPendingEvents = () => (dispatch, getState) => {
             '*',
             { name: 'logs', columns: ['*'], order_by: ['-created_at'] },
           ],
-          where: { delivered: false, error: false, tries: 0 },
+          where: { delivered: false, error: false, tries: 0, archived: false },
```
wawhal: What is this?

console/src/components/Main/Main.js
rikinsk: Yep. I moved it to line 69. Removed unnecessary fn

console/src/components/Services/EventTrigger/EventActions.js
rikinsk: This was some event trigger server refactor related thing. Event data can be archived so we need to add thid flag to ensure we fetch only non archived data

rikinsk: it was initially added after version check below

PR #3504, opened 2019-12-09
avoid using CTE for querying custom SQL functions (close #3349)

### LINKED ISSUE #3349: Improve performance when querying Functions
Given this graphql query:

```graphql
{
  my_function(
    args: {
      p_customer_id: 3050
    }
    limit: 50
    where: { code: { _eq: "10000056724" } }
  ) {
    code
  }
}
```

Hasura console Analyse feature reveals the following generated SQL query:

```sql
WITH "public_my_function__result" AS (
  SELECT
    *
  FROM
    "public"."my_function"(
      ('3050') :: integer
    )
)
SELECT
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_1_e"
          FROM
            (
              SELECT
                "_0_root.base"."code" AS "code"
            ) AS "_1_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public_my_function__result"
        WHERE
          (
            (
              "public_my_function__result"."code"
            ) = (('10000056724') :: text)
          )
      ) AS "_0_root.base"
    LIMIT
      50
  ) AS "_2_root"
```

When my_function() returns many records, it comes to be very heavy for Postgresql because the select inside WITH clause will return every possible records and only after they will be filtered by the where condition, causing a time consuming CTE Scan.

It would be more efficient if the where conditions - and any other possible join/limit/order clauses/... - would be applyed inside the WITH clause, something like that:

```sql
WITH "public_my_function__result" AS (
  SELECT
    *
  FROM
    "public"."my_function"(
      ('3050') :: integer
    )
  WHERE code = '10000056724'
  LIMIT 50
)
...
...
```

rakeshkky:
### Description

Do not use [Common Table Expressions (CTE)](https://www.postgresql.org/docs/current/queries-with.html) for querying custom SQL functions.

Consider the following custom SQL query:-

```graphql
query {
  search_tracks(
    args: {search: "Night"}
    where: {id: {_gt: 100}}
  ){
    id
    name
  }
}
```

Generated SQL statements:-

With CTE
```sql
WITH "public_search_tracks__result" AS (
  SELECT
    *
  FROM
    "public"."search_tracks"(('Night') :: text)
)
SELECT
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_1_e"
          FROM
            (
              SELECT
                "_0_root.base"."id" AS "id",
                "_0_root.base"."name" AS "name"
            ) AS "_1_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public_search_tracks__result"
        WHERE
          (
            ("public_search_tracks__result"."id") > (('100') :: integer)
          )
      ) AS "_0_root.base"
  ) AS "_2_root"
```
Without CTE (in this PR)
```sql
SELECT
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_2_e"
          FROM
            (
              SELECT
                "_1_root.base"."id" AS "id",
                "_1_root.base"."name" AS "name"
            ) AS "_2_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public"."search_tracks"(('Night') :: text) AS "_0_search_tracks"
        WHERE
          (("_0_search_tracks"."id") > (('100') :: integer))
      ) AS "_1_root.base"
  ) AS "_3_root"
```

### Affected components

- [x] Server

### Related Issues

Close #3349 

### Solution and Design

- Remove types related `function select` and `function aggregate select`. 
- SQL function queries are treated as normal `select` which differs only in `FROM` expression. 

#### Catalog upgrade

- [x] No
- [ ] Yes
 
#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### Steps to test and verify

The custom SQL function queries shouldn't have Common Table Expressions (CTE). Use console `Analyze` to view the generated SQL statement.

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Resolve/Select.hs
+++ b/server/src-lib/Hasura/GraphQL/Resolve/Select.hs
@@ -108,7 +108,7 @@ fromSelSet fldTy flds =
                 colMapping = riMapping relInfo
                 rn = riName relInfo
             if isAgg then do
-              aggSel <- fromAggField relTN colGNameMap tableFilter tableLimit fld
+              aggSel <- fromAggField (RS.FromTable relTN) colGNameMap tableFilter tableLimit fld
               return $ RS.FArr $ RS.ASAgg $ RS.AnnRelG rn colMapping aggSel
             else do
               annSel <- fromField (RS.FromTable relTN) colGNameMap tableFilter tableLimit fld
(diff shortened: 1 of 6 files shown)


paf31:
If you use the same function with the same arguments twice, will it now evaluate the result twice? Or will the optimizer be smart enough to figure that out?

lexi-lambda:
@paf31 IIUC, this change shouldn’t affect the number of times the function is called—it should be called exactly once in both cases, since the subquery containing the function call is at the top level. My understanding is that this rearranging just helps the Postgres query optimizer see that the `WHERE` condition is used to filter the result of the function, so it can push the condition “inside” the function’s definition (assuming it’s written in `LANGUAGE SQL`).

paf31:
Right yes, sorry, I misread the example, and thought the function was being used in the `where` clause. I think the question does apply, but not in this case. Never mind :)

PR #3519, opened 2019-12-11
Fix tests on postgis 3, fixes #3415

### LINKED ISSUE #3415: support postgis 3.0
When running the current test framework against PostGIS 3.0 (using the docker image `hasura/postgres:12.1-1`), there seem to be errors along these lines:
```sql
ERROR:  LWGEOM_dwithin: Operation on mixed SRID geometries (Point, 0) != (Polygon, 4326)
ERROR:  crosses: Operation on mixed SRID geometries (Point, 0) != (LineString, 4326)
```
Also from PostGIS 3.0, `raster` support has to be explicitly enabled with `CREATE EXTENSION postgis_raster`.

paf31:
# Description

Runs the test suite against postgis 3.0.0 when testing with postgres 12.

We could create separate tasks for postgres 12 with postgis 2.5.3 if that's desirable?

The approach is simple:

- Install `postgis_raster` (which is now a separate extension) during the test suite setup only if postgis 3.0.0 is installed. 
- Specify SRIDs for geometry and geography types wherever needed. I think the recommendation should be to include `crs` data in any GeoJSON inputs from now on, because the default SRID when parsing JSON will vary between postgis versions.
- Also, modify some numeric inputs to work around [a floating point issue we found in postgis 3.0.0](https://trac.osgeo.org/postgis/ticket/4604#ticket)

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

#3415

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


paf31:
This is currently passing on Postgres v12, but I'm lost as to why I'm getting errors on earlier versions of Postgres for the two tests

- `user_can_query_geometry_values_filter`
- `user_can_query_geometry_values_filter_session_vars`

Issue #3203, opened 2019-10-22
add computed fields support on console


rikinsk:
Server support added in #2917

mnlbox:
Any short term plan for this in the next release?

PR #3522, opened 2019-12-12
add computed fields support on console

rikinsk:
### Description

Add support for adding computed fields via the console

**TODO**
- verify migrations

### Affected components

- [x] Server
- [x] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3203 #3565

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


santicomp2014:
@rikinsk Hey nice job with this PR, I was wondering if this affected the computed field and roles.
Currently, I'm using a computed field, but when I query from a role it does not even appear as a field in the console neither from my frontend.

Should I create a new Issue?.
Thanks.

darknoon:
Hey, this is really awesome and would fix one of my main gripes at the moment, which is creating views on my tables so I can add a computed field. My use case is signing URLs for S3 access, eg profile photo etc.

In the design, I think it would be clearer if the "function definition" area didn't show up until a matching function was found. I tried to start typing there, and took a sec to realize I needed to type in the exact name of the function.

What about a dropdown of existing functions?

Screen Shot 2019-12-18 at 12 21 25 PM

Anyway, looking forward to using this!

lexi-lambda:
> * SERVER: The metadata has been updated to track the `hdb_computed_field_function` view. Catalog migrations need to be figured out.

Interesting issue—we basically need an “empty” migration, just to get the metadata migrations to fire. That shouldn’t be difficult to do, and in fact it’s probably simplest to genuinely just create an empty SQL file for the migration, which I think should work? I’ll figure it out and push a change to this branch.

rikinsk:
@lexi-lambda Thanks for fixing the catalog migrations. Though it seems like there might be a need for some schema changes as well. 

@rakeshkky Currently I am querying the  `hdb_computed_field_function` view to fetch the computed-field configuration. But it seems like it currently doesnt expose the computed_field `comment` and `table_row_argument` information. Can these be added to the view as well or am I approaching this in the wrong way.

rikinsk:
@santicomp2014 I see you have raised a separate issue about permissions, so I guess we can track it there.

rikinsk:
@darknoon Thanks for your feedback! We'll definitely attempt to make this experience more intuitive.

lexi-lambda:
> Currently I am querying the `hdb_computed_field_function` view to fetch the computed-field configuration. But it seems like it currently doesnt expose the computed_field `comment` and `table_row_argument` information. Can these be added to the view as well or am I approaching this in the wrong way.

Would it be better to just query `hdb_catalog.hdb_computed_field` directly? Or is there some reason you’d rather use the `hdb_computed_field_function` view?

rikinsk:
> > Currently I am querying the `hdb_computed_field_function` view to fetch the computed-field configuration. But it seems like it currently doesnt expose the computed_field `comment` and `table_row_argument` information. Can these be added to the view as well or am I approaching this in the wrong way.
> 
> Would it be better to just query `hdb_catalog.hdb_computed_field` directly? Or is there some reason you’d rather use the `hdb_computed_field_function` view?

There is no real reason other than maintaining consistency. It seemed to me that it was a pattern being followed that we query a view that aggregates all relevant information about an object from the console. I could totally be wrong here though.

rikinsk:
I will check if all the information I need is present in the table directly and query it instead. There doesnt seem to be a need for querying the view currently as far as I can see. 

@rakeshkky Feel free to add any insights if any

docs/graphql/manual/api-reference/schema-metadata-api/computed-field.rst
```diff
@@ -22,8 +22,8 @@ add_computed_field
 There cannot be an existing column or relationship or computed field with
 the same name.
 
-Create a ``computed field`` called ``get_articles`` on an ``author`` *table*, using
-an SQL function called ``fetch_articles``:
+Create a ``computed field`` called ``full_name`` on an ``author`` *table*, using
```
marionschleifer: I was wondering if it would be good to also have the actual SQL function definition here, like in the schema part. It might be easier to visualize. What do you think?

rikinsk: Uptill now we have been following the practice of not going into details of the examples we use in the API reference section. The API reference currently just shows API signatures but not the morivation behind the feature or the examples.

But it's a fair point Marion. Let's discuss this outside the domain of this PR as it actually impacts the entire API reference section.

marionschleifer: Thank you for the explanation, Rikin. Let's discuss it in our next call 🙂

rikinsk:
@wawhal The table row argument should be ignorable ideally. what was the error you saw? 

Will add a tooltip next to it to explain what it is

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -164,6 +165,138 @@ const resetPrimaryKeys = () => ({
...
+  computedField,
+  table,
+  originalComputedField,
+  successCb
+) => (dispatch, getState) => {
+  const migrationUp = [];
+  const migrationDown = [];
+
+  if (originalComputedField) {
+    migrationUp.push({
```
wawhal: Move all the queries to `v1QueryUtils.js`

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -164,6 +165,138 @@ const resetPrimaryKeys = () => ({
   type: RESET_PRIMARY_KEY,
 });
 
+export const saveComputedField = (
+  computedField,
```
wawhal: **Question**:

All these arguments are available in redux. Probably the only thing you need is their index. So not sure if we should pass all these arguments.

I was thinking that these ServerIO functions should have as less arguments as possible. Otherwise, as you iterate, things get pretty messy. Some of the functions in our code base have 8 arguments and all.

Tell me if you agree.

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -164,6 +165,138 @@ const resetPrimaryKeys = () => ({
...
+
+export const deleteComputedField = (computedField, table) => (
+  dispatch,
+  getState
+) => {
+  const migrationUp = [];
+  const migrationDown = [];
+
+  migrationUp.push({
+    type: 'drop_computed_field',
```
wawhal: Move queries to `v1QueryUtils.js`.

console/src/helpers/versionUtils.js
```diff
@@ -6,7 +6,7 @@ export const READ_ONLY_RUN_SQL_QUERIES = 'readOnlyRunSqlQueries';
 
 // list of feature launch versions
 const featureLaunchVersions = {
-  // feature: 'v1.0.0'
+  // feature: 'v1.0.1'
```
wawhal: CMIAW. This feature need not have semver checks since it'll go out in RC?

console/src/components/Services/Data/TableModify/ComputedFieldsEditor.js
```diff
@@ -0,0 +1,409 @@
...
+              })}
+            </select>
+          </div>
+          <div className={`${styles.add_mar_top}`}>
+            <div className={`${styles.add_mar_bottom_mid}`}>
+              <b>Function name: </b>
+              <RawSqlButton
+                dataTestId={'create-function'}
+                customStyles={`${styles.display_inline} ${styles.add_mar_left}`}
+                sql={''}
```
wawhal: Can we pass a template function here?
For people who are trying out and don't know the exact syntax of SQL functions (peeps like me). Also for good demos.

console/src/components/Services/Data/DataActions.js
```diff
@@ -383,10 +383,10 @@ const fetchFunctionInit = () => (dispatch, getState) => {
   };
 
   // set schema in queries
-  const currentSchema = getState().tables.currentSchema;
-  body.args[0].args.where.function_schema = currentSchema;
-  body.args[1].args.where.function_schema = currentSchema;
-  body.args[2].args.where.function_schema = currentSchema;
+  const fnSchema = schema || getState().tables.currentSchema;
+  body.args[0].args.where.function_schema = fnSchema;
```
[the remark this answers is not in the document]
rikinsk: I had thought of this but then had reverted. We cannot make the assumption of the structure of the args. Imagine a new arg got added which has the `where` under a `not` operator. This entire code piece is a bad pattern. Ideally the schema should have been passed as a param to the sql generator and that function would be complete in itself. lets fix this in a refactor later

console/src/components/Services/Data/TableModify/ModifyActions.js
rikinsk: I want to go in the opposite direction of this. I would want to make all ServerIO functions as pure functions which accept params explicitly and do not depend on the redux state. No of params is not a big issue if done logically. Doing this enures reusability and ability to move such code around easily. 

The more I think I about it, I believe redux state should only be used for holding common state and data and nothing else. Local state seems always enough for most rendering purposes.

console/src/helpers/versionUtils.js
rikinsk: Yep. No version checks needed

console/src/components/Services/Data/TableModify/ComputedFieldsEditor.js
rikinsk: I has though about this but there is no generic template I could come up with that can be used without making assumptions about the function. The create function structure is defined [here](https://www.postgresql.org/docs/current/sql-createfunction.html) and you can see it cant really be templated without making  assumptions about return types and language

console/src/components/Services/Data/TableModify/ComputedFieldsEditor.js
```diff
@@ -0,0 +1,409 @@
...
+          <br />
+          <span key={'comment'} className={styles.text_gray}>
+            {origComputedFieldComment}
+          </span>
+        </div>
+      );
+    };
+
+    // expand button text "View"
+    // eslint-disable-next-line no-nested-ternary
```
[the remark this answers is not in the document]
rikinsk: Done

console/src/components/Services/Data/TableModify/ModifyActions.js
rikinsk: Done

console/src/components/Services/Data/TableModify/ModifyActions.js
rikinsk: Done

ManUtopiK:
Just tested the last deploy: https://hge-ci-pull-3522.herokuapp.com/
I think it missing computed fields permission management. It should be something like "Computed fields insert/select/update permissions" after column permissions.

ManUtopiK:
Also, f there are two columns; first_name with public permission and last_name with private permission, could computed field full_name be public?

rikinsk:
This PR is finally feature complete and ready for merge

@marionschleifer @wawhal @lexi-lambda Please add your reviews .

@lexi-lambda The empty server migration file is now replaced by one added in another PR.

PR #3529, opened 2019-12-13
Allow request body to be logged if required even with successful queries

nizar-m:
### Description
`LogHTTPSuccess` now has the json parsed request body (if present) as an extra argument.
This is needed if we also want to log the request body with successful queries.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


nizar-m:
> This seems like it could result in pretty noisy logs, right? I don’t really have a fundamental problem with this change, but does it make sense to put this behind a flag or something? Most servers don’t log every single request payload!

This maybe useful in some cases. For example you may want to always log the request body of metadata queries.

PR #3534, opened 2019-12-13
Understand bounded cache performance. Closes #3530

### LINKED ISSUE #3530: understanding the performance characteristics of bounded cache
@0x777 said :

What's the difference in metrics like req/s and latency? Does it warrant extra work to improve the data structures used?

jberryman:
### Description
Some initial benchmarks to ask/answer some questions about plan cache performance, and the performance of the underlying data structures.

Some analysis will go in the comments of #3530

### Affected components

- [x] Server  (minor changes to cache implementations)
- [x] Build System
- [x] Tests

### Related Issues

https://github.com/hasura/graphql-engine/issues/3530

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### Steps to test and verify

To build and run benchmarks:

    stack build --bench

...but that will take forever once it reaches the "realistic" benchmarks, so at that point do e.g.

    ./.stack-work/dist/x86_64-linux/Cabal-2.4.0.1/build/cache/cache -n1

to just get timings for a single run.

As mentioned above and in the last commit, some of the benchmarks are just too slow, but I'm also not inclined to rewrite them at this point

### DIFF
--- a/server/src-bench-cache/Main.hs
+++ b/server/src-bench-cache/Main.hs
@@ -1,0 +1,281 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+module Main where
+
+import           Control.Concurrent            (getNumCapabilities)
+import           Control.Concurrent.Async      (forConcurrently)
+import           Control.DeepSeq
+import           Control.Monad                 (foldM)
+import           Criterion.Main
+import           Data.Bits
+import           Data.IORef
+import           Data.List
+import           Data.List.Split               (chunksOf)
+import           Data.Ord
+import           Data.Traversable
+import qualified Data.Vector                   as V
+import           Data.Word
+import           GHC.Clock
+import qualified Hasura.Cache.Bounded          as B
+import qualified Hasura.Cache.Unbounded        as U
+import           Prelude
+import           System.Random.MWC             as Rand
+import           System.Random.MWC.Probability as Rand
+-- higher level interface to above, combined:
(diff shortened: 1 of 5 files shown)


server/src-bench-cache/Main.hs
```diff
@@ -0,0 +1,281 @@
...
+import Control.Concurrent (getNumCapabilities)
+import Control.DeepSeq
+import Control.Monad (foldM)
+import GHC.Clock
+import System.Random.MWC as Rand
+import System.Random.MWC.Probability as Rand
+import qualified Hasura.Cache.Bounded   as B
+import qualified Hasura.Cache.Unbounded as U
+-- higher level interface to above, combined:
+import qualified Hasura.Cache as Cache
```
lexi-lambda: Minor point, but: could you run `stylish-haskell` on this file?

jberryman: Just made those changes, thanks!

PR #3537, opened 2019-12-14
add note about abac

tirumaraiselvan:
add note on abac

### Description

Add note about ABAC.

### Affected components

- [x] Docs

### DIFF
--- a/docs/graphql/manual/auth/authorization/roles-variables.rst
+++ b/docs/graphql/manual/auth/authorization/roles-variables.rst
@@ -85,7 +85,7 @@ Examples:
 
 .. admonition:: ABAC
 
-  Session variables are analogous to *attributes* in a typical attribute-based access control (ABAC) system.
+  Session variables are analogous to *attributes* in a typical `attribute-based access control <https://en.wikipedia.org/wiki/Attribute-based_access_control>`_ (ABAC) system.
 
 
 Modelling Roles in Hasura
--- a/docs/graphql/manual/auth/authorization/roles-variables.rst
+++ b/docs/graphql/manual/auth/authorization/roles-variables.rst
@@ -34,8 +34,7 @@ Examples:
 Dynamic session variables
 -------------------------
 
-When you create a permission, or an access control rule, the permission rule itself needs access to some variables
-that are derived from the request itself. Let's refer to these as *session variables*.
+Permission rules can also refer to *session variables*. Session variables are key-value pairs returned from the authentication service for each request.
 
 For example: If a user makes a request, the session token maps to a ``user-id``. This ``user-id`` can be used in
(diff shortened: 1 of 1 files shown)


docs/graphql/manual/auth/authorization/roles-variables.rst
```diff
@@ -34,8 +34,12 @@ Examples:
...
 -------------------------
 
-When you create a permission, or an access control rule, the permission rule itself needs access to some variables
-that are derived from the request itself. Let's refer to these as *session variables*.
+When you create a permission or an access control rule, the permission rule itself can use variables
+that come from the authentication service. Let's refer to these as *session variables*. 
+
+.. admonition:: ABAC
+
+  Session variables are analogous to *attributes* in a typical ABAC system.
```
lexi-lambda: I think it would be good to expand the initialism here:

> Session variables are analogous to *attributes* in a typical attribute-based access control (ABAC) system.

lexi-lambda: Also, this isn’t all that relevant to this particular change, but the previous sentence reads a bit strangely to me: “Let’s refer to these as session variables” sounds weirdly noncommittal! Why not write:

> Permission rules can also refer to *session variables*, which contain information returned from the authentication service.

tirumaraiselvan: I don't mind what is present current but in case we are re-reading then, I would like to be more explicit: 

> Permission rules can also refer to *session variables*.  Session variables are key-value pairs  returned from the authentication service.

rikinsk: I would move the note to the end of the section. It seems kind of odd in its current location _(bang in the middle of the definition and an example explaining it)_. Also I would hope to see a link to something speaking about abac attributes otherwise the entire section seems kind of unqualified

Agree with @lexi-lambda 's suggestion about the initial statement as well.

tirumaraiselvan: Should it not be where it's defined?

rikinsk: If its part of the deriniton, it doesnt need to be a note. Make it as a statement itself

tirumaraiselvan: changed. I think it's fine to not link to any abac resource, dunno any authoritative resource anyway.

lexi-lambda: I think a link to the Wikipedia article would probably be good enough, honestly.

tirumaraiselvan: done !

Issue #3280, opened 2019-10-30
[RFC] Switch from `stack` to `cabal-install` for building Haskell code


lexi-lambda:
## What

I propose we switch away from using `stack` as our Haskell build tool to `cabal-install` (aka the `cabal` command-line executable).

## Why

Why bother switching build tools? Generally, the reasons fall into two categories: reasons `stack` is insufficient and reasons `cabal-install` has gotten better.

### Problems with `stack`

- First and foremost, the number one reason to switch away from `stack` is the way it handles profiling. I **do not know** how to properly profile Haskell code using `stack`, as it forcibly compiles all dependencies with `-fprof-auto` (commercialhaskell/stack#2853), which is simply untenable. To me, this is a dealbreaker: when I have needed to profile `graphql-engine`, I have needed to rebuild it with `cabal-install` *anyway*.

- My experience with `stack` 2.x has been poor. It uses a new caching system called `pantry`, which is similar in spirit to `cabal-install`’s Nix-style build caching. However, `pantry`’s caching model is worse—it does not properly account for different sets of build flags in dependencies (which is related to the above point on why `-fprof-auto` cannot be disabled)—and it relies on a global SQLite database that provides zero options from recovering from an invalid caching state.

- Also related to the above, there is no way to build library dependencies with different optimization levels, so it is not possible to reliably link against libraries compiled with `-O2`.

- Getting access to new versions of GHC and libraries with `stack` is slow. The use of Stackage LTSes means reliable builds, but it also means you’re locked into those packages unless you want to do a lot of constraint solving yourself. `stack` has also recently **removed** the `stack solver` escape hatch, leaving it with no constraint solver capabilities whatsoever.

- Various other minor infelicities, some of which I wrote about [here](https://twitter.com/lexi_lambda/status/1180370814882996225).

### Improvements in `cabal-install`

`cabal-install`’s support for Nix-style local builds (aka `cabal new-build`) combined with support for per-project configuration (in the form of `cabal.project` files) is a total game-changer. `cabal-install` now supports essentially all the flexibility provided by `stack.yaml` files and much more:

- `cabal.project` files support multi-project builds, as well as using packages directly from git repositories, just like `stack.yaml` files do.

- `cabal.project` files allow fine-grained control over the build options for every dependency in your project, and it actually caches them properly. If you modify the optimization or profiling settings for a dependency, `cabal new-build` will rebuild the minimal set of things necessary, caching the build results globally.

[...]

PR #3558, opened 2019-12-17
Switch from `stack` to `cabal-install` for building server code

jberryman:
See #3280 

The idea here is to start the process of moving away from stack
(potentially), but making this move official if we decie to, will
require more coordination (moving CI, dev.sh and docs all at once).

I've added a "fast" variation of cabal.project.local mostly as a
demonstration. I haven't found a "prof" variant that really works for
me. All this is pretty open to change, but it would be nice to have some
local cabal.project files that encode good/useful stuff.

The freeze file should be version-exact to our stackage LTS snapshot.


server/cabal.project.local.fast
```diff
@@ -0,0 +1,17 @@
...
+
+-- Compile quickly without optimizations
+--
+-- TO ENABLE THIS YOU CAN DO: 
+--   $ ln -s cabal.project.local.prof cabal.project.local
+package *
+  documentation: true
+
+package graphql-engine
+  optimization: 0
```
lexi-lambda: FWIW, this doesn’t actually work due to haskell/cabal#5353. The issue is that `optimization: 0` doesn’t override `-O2` in the `.cabal` file. We should probably remove `-O2` from the `.cabal` file and only put `optimization: 2` in the `cabal.project` file.

jberryman: Ya came here to post after just noticing this too. Trying to figure out wth build tools are doing is about my least favorite thing...

jberryman: But I guess this speaks to the benefit of encoding this knowledge in config files or comments at least

lexi-lambda: A simple workaround, which I’ve been using, is to use `ghc-options: -O0` instead. But I think moving `-O2` out of the `.cabal` file is a better solution.

jberryman: (sorry for the delay). I updated the `cabal.project.local.fast` to mention the issue and work around it.

jberryman:
@lexi-lambda (bump) can we merge or close this?

lexi-lambda:
@jberryman I’d like to merge the `cabal` support all in one go, so to speak, so we don’t have some things using `cabal` and other things using `stack`. That said, I’d really like to get this merged, too, so maybe I’ll try to make the rest of the changes today or tomorrow.

jberryman:
Cool, just didn't want this to get lost.

jberryman:
@lexi-lambda awesome! Let me know if there's anything more I can help with here

lexi-lambda:
@jberryman I think this PR is now good to go. Take a look and let me know if everything looks good to you?

tirumaraiselvan:
What about this file: https://github.com/hasura/graphql-engine/blob/master/server/packaging/Dockerfile

From initial glance, seems like we don't need it anymore.

lexi-lambda:
Thanks, good catch. I’ve deleted it.

0x777:
@jberryman Can you have a look at Alexis's changes? Let's get this merged as soon as possible.

jberryman:
LGTM except for porting `dev.sh` . I think I'll have time to try this branch out this afternoon and make those changes, but not 100% sure.

tirumaraiselvan:
I also want this merged. I am struggling with lts not having few packages. And adding them to extra-deps is causing some cascading failures :innocent:

tirumaraiselvan:
For those curious, I needed http://hackage.haskell.org/package/time-1.9.3 because of `NominalDiffTime` support.

server/cabal.project.freeze
```diff
@@ -0,0 +1,349 @@
...
+             any.text-latin1 ==0.3.1,
+             any.text-printer ==0.5,
+             any.text-short ==0.1.2,
+             text-short -asserts,
+             any.tf-random ==0.5,
+             any.th-abstraction ==0.2.11.0,
+             any.th-lift ==0.7.11,
+             any.th-lift-instances ==0.1.12,
+             any.these ==0.7.6,
+             any.time ==1.8.0.2,
```
tirumaraiselvan: Probably off-topic, say I need time = 1.9.3 but that has other deps , then how will i figure that out? `cabal-install` will resolve the plan and suggest changes?

lexi-lambda: The easiest way is just to delete the `caba.project.freeze` file an re-run `cabal new-freeze`. The main point of the freeze file is to ensure reproducible builds, so other people won’t get spurious build failures because the state of the package index changed and for some reason the new build plan is somehow broken. It’s to avoid things from changing out from underneath you. There’s nothing wrong with just blowing it away and regenerating it, since once you check it in, you’ve verified that the build plan is, in fact, good.

Stackage is really the same way. Technically it’s true that bumping the LTS version only bumps minor versions, so theoretically it’s “safer” if people are versioning their packages really carefully, but the truth is that Stackage does no auditing of packages to ensure that’s actually true; all that matters about an LTS is that all the packages build together. When Stackage was first created, this was a bigger deal than it is now, since a lot of packages *didn’t* build together, so `cabal-install` often failed to generate build plans. But the funny thing about Stackage is that its benefits are actually useful to `cabal-install` users almost as much as to `stack` users, since the ecosystem-wide CI means packages are kept more compatible, and `cabal-install` is able to generate good build plans.

In that light, I feel like there are really relatively few benefits to using Stackage directly, in a “tragedy of the commons” sort of way. `cabal-install` is more flexible, since you aren’t tethered to a particular snapshot, but you still get working build plans without much effort.

lexi-lambda:
Well, I did say I would merge this today if I didn’t receive any objections, and I’ve received none, so I’m going to merge this as soon as CI passes. @jberryman, feel free to open a followup PR to further improve `dev.sh` if you deem it necessary.

jberryman:
> Well, I did say I would merge this today if I didn’t receive any objections, and I’ve received none, so I’m going to merge this as soon as CI passes. @jberryman, feel free to open a followup PR to further improve `dev.sh` if you deem it necessary.

Meant to give the thumbs up there. Thanks for doing that! I'll have some updates to the script this week.

server/cabal.project.freeze
hvr: > so other people won’t get spurious build failures because the state of the package index changed

Just for the record (please just ignore me if this is already known); if the main worry is the package-index changing resulting in unsound build-plans, "freezing" merely the index-state can be accomplished by `--index-state=<timestamp>` instead of freezing the actual build-plan.

But of course, doing a full `cabal freeze` via version pinning (and you'd still want to combine that with index-freezing) gives more guarantees  re reproducibility.

jberryman: @hvr 

>  (and you'd still want to combine that with index-freezing) 

Why freeze the index as well? In case a hackage "revision" changes the constraints for a dependency version retroactively?

@lexi-lambda I'd still like a little clarity on how we want to manage package version upgrades and new dependencies. I think it might make sense to pin the `index-state` in `cabal.project` (similar to a stackage version) if it allows us to do two things separately:

- add new dependencies without upgrading a bunch of others
- upgrade some or all dependencies to newer versions (and bumping index version)

...but I'm not sure that's how that would work.

lexi-lambda: Maybe I’m in the minority on this, but I’ve always been of the opinion that the hand-wringing about Hackage revisions is largely overblown. I understand why a project like Stackage might find them frustrating, but Stackage has pretty special needs. From my point of view, Hackage revisions are an extremely good thing, as they prevent incorrectly-specified constraints from indefinitely polluting the solver.

My guess is that we won’t ever encounter a situation where a revision causes a build that previously passed to fail. Revisions are almost exclusively added for one of two reasons:

  1. A new version of a dependency was released, so someone makes a revision to weaken a conservative upper bound once it’s clear the new version is still compatible.

  2. Someone discovered that a selected build plan does not actually successfully compile, so they fix the incorrect bound via revision.

We don’t care about (1), since we’ve pinned all the versions in the freeze file anyway. Weakened bounds can’t affect our build plans. And while it’s theoretically possible that we could get bitten by (2), it’s highly unlikely, as if our build passes CI, then our build plan fundamentally must be a good one. If a package maintainer adds a revision that breaks our previously-working build plan, either they screwed up (and we should tell them about it), or they had a really good reason to do so that we probably want to know about!

Given the above, I don’t think pinning the index state is super useful, and it seems more likely to create busywork than anything else. If it ever becomes a problem, I’m more than happy to revisit this decision, but I don’t expect it to be.

> I'd still like a little clarity on how we want to manage package version upgrades and new dependencies as a team.

From my point of view, the right workflow here is really simple: delete the freeze file and re-run `new-freeze`. Ideally, this should always be possible, as if it isn’t, that means the constraints in our `.cabal` file are too loose! The purpose of the freeze file in my eyes is simply build reproducibility: if it turns out the constraints in our `.cabal` file *are* too loose, we don’t want to find out because CI jobs just start suddenly failing. But if we find out because we deleted the freeze file, that’s great, as there’s a human in the loop at that point to fix the `.cabal` file.

In other words, I think manually worrying about the freeze file is an antipattern. We should never rely on the constraints in the freeze file to save us, so we should only be concerned with the `.cabal` file. That said, if you *really* want to bump just one dependency, you can either:

  1. Update the constraint in the freeze file by hand.

  2. Delete the constraint for that particular package from the freeze file and run `new-freeze` again.

     This should work, as I believe `cabal-install` will take the existing freeze file into account when creating the new build plan. When it regenerates the freeze file, it will include all the existing constraints, plus whatever new ones are needed for the updated dependency.

But really, I think that’s overly pessimistic. If we generate a build plan that (a) compiles and (b) passes CI, then we can be pretty confident it’s a good one. And if anything, we *want* our dependencies to be as up to date as possible, since we want to integrate bugfix patches as much as we can!

Let me know if that makes sense to you or if you disagree with any of the above. I can also put some of these thoughts about the workflow in the CONTRIBUTING.md if you think that would help.

jberryman: > Hackage revisions are an extremely good thing,

I agree, I'm not really worried about revisions.

> That said, if you really want to bump just one dependency, you can either: ...

I have pretty high confidence in the quality of libraries on hackage generally, but if we have a situation where e.g. many dependencies are changing arbitrarily (maybe even being downgraded?) without much oversight on each PR that sounds a little scary for debugging issues (esp performance issues). 

I'm okay trying to proceed this way. But I think I'd be more comfortable with guidance like: "try to upgrade dependencies only as needed by editing the freeze file ... upgrades of many dependencies can happen, but should be in a separate commit so it can be easily reverted if necessary". Does that sound reasonable?

lexi-lambda: I don’t expect there to be a ton of churn in the freeze file even if people delete it and recreate it, but I’m willing to try and be conservative if possible if you think that’s a good idea.

From my point of view, there is a fundamental tension here: we want to incorporate bugfix patches as much as possible (perhaps there’s a performance bug or security vulnerability), and we just aren’t going to regularly audit all of our dependencies to see which ones need upgrading. For that reason, my preference is to give package authors the benefit of the doubt and to assume that newer versions are better. Sure, we could get bitten by upgrading if that assumption turns out to be wrong, but we could also get bitten if a package fixes an issue and we haven’t upgraded.

Still, I can also understand wanting to avoid profiling noise that comes from changes in dependencies rather than changes in our code. So I’m willing to be a little more deliberate about this, I just don’t want to be more conservative than we have to be.

PR #3570, opened 2019-12-19
Add external URL endpoint options for the console (close #2824)

### LINKED ISSUE #2824: Running hasura console via CLI in docker container
Hello!

I'm trying to run `hasura console` in docker container no#1 and then access the console on port `9695` on my host machine.

The JavaScript of the hasura console application needs access to both the graphql engine (running in docker container no#2) and the migration service spun up by running `hasura console` in docker container no#1.

The URLs that the JavaScript tries to talk to are based on two things:
- The `endpoint` configuration of `config.yaml` in docker container no#1
- The `--address` flag used to start `hasura console` with in docker container no#1, which defaults to localhost when not set.

Because the hasura console application is only available inside docker container no#1 when started with the default `--address` of "localhost", but is not available on the host machine of docker container no#1, i need to start it with `--address 0.0.0.0` to have it bind to all available interfaces so it's not only reachable inside of container no#1, but also on the host machine.

Because the graphql engine is running in docker container no#2, which happens to be set up in a `docker-compose.yml` with docker container no#1 and a third run running postgres, i need to configure the `endpoint` in `config.yaml` to the docker compose service name of the second container, which is `endpoint: http://graphql-engine:8080`.

This is the only configuration where the console actually starts, because it can access docker container no#2 from docker container no#1 and where the started console is then actually reachable by my docker host, because it is bound to all available interfaces.

**The problem**:

The console web application now tries to talk to both `http://graphql-engine:8080` and `http://0.0.0.0` when accessed from the browser of my host machine - both of which are not available.

You can work around the first of those requests by having both container no#1 and container no#2 on the same network, so the graphql engine can be reached with `endpoint: http://localhost:8080` - which also works on the docker host when port `8080` is exposed to the host.

I did not find any workarounds for the `--address` and the resulting requests to `0.0.0.0` though. Is there anything that i'm missing?

Would it be possible to add another configuration option to differentiate between the interfaces the graphql engine is actually served on and the URL that the clientside javascript uses to access it?

ashishra0:
### Description
This PR adds two external URL endpoint options for the Hasura console.
Currently, There are no options configured.
Also, all requests going through CLI must be protected by the access key.

### Affected components
- [x] Console
- [x] CLI

### Related Issues
#2824 

### Solution and Design
Add a flag each for external server endpoint and external cli endpoint.
Add a check for access key in the request body.


cli/commands/console.go
```diff
@@ -281,3 +290,14 @@ func serveConsole(assetsVersion, staticDir string, opts gin.H) (*gin.Engine, err
 
 	return r, nil
 }
+
+func verifyAdminSecret() gin.HandlerFunc {
+	return func(c *gin.Context) {
+		if ec.ServerConfig.AdminSecret != "" {
+			if c.GetHeader(XHasuraAdminSecret) != ec.ServerConfig.AdminSecret {
+				//reject
+				c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"Message": "Unauthorized"})
```
shahidhk:
```suggestion
				c.AbortWithStatusJSON(http.StatusUnauthorized, gin.H{"message": "unauthorized"})
```

shahidhk:
@rikinsk @beerose Please fast-track review on this PR

shahidhk:
@arvi3411301 Can you fix conflicts here?

rikinsk:
There are some issues while running console from the CLI when admin secret is set. Will figure out whats happening and push fixes.

cli/commands/console.go
```diff
@@ -60,6 +60,8 @@ func NewConsoleCmd(ec *cli.ExecutionContext) *cobra.Command {
 	f.StringVar(&opts.Address, "address", "localhost", "address to serve console and migration API from")
 	f.BoolVar(&opts.DontOpenBrowser, "no-browser", false, "do not automatically open console in browser")
 	f.StringVar(&opts.StaticDir, "static-dir", "", "directory where static assets mentioned in the console html template can be served from")
+	f.StringVar(&opts.ServerExternalEndpoint, "server-external-endpoint", "", "endpoint using which console can access graphql engine")
+	f.StringVar(&opts.CliExternalEndpoint, "cli-external-endpoint", "", "endpoint using which console can access graphql engine")
```
mubaidr: Both descriptions look same

cli/commands/console.go
```diff
@@ -113,6 +122,26 @@ func (o *ConsoleOptions) Run() error {
...
+	}
+	if o.ServerExternalEndpoint == "" {
+		o.ServerExternalEndpoint = o.EC.Config.ServerConfig.Endpoint
+	}
+
+	// FIXME: My Router struct
+	r := &cRouter{
+		g,
+		t,
+	}
```
kolharsam, line 137: @scriptonist I'm marking 4 `FIXME` comments that I need help with. Basically wanted to know if there are alternate structures or methods that need to be used in order to get this PR up to speed.

cli/commands/console.go
```diff
@@ -171,3 +200,126 @@ func (o *ConsoleOptions) Run() error {
...
+}
+
+func serveConsole(assetsVersion, staticDir string, opts gin.H) (*gin.Engine, error) {
+	// An Engine instance with the Logger and Recovery middleware already attached.
+	r := gin.New()
+
+	// FIXME: DoAssetExist
+	if !util.DoAssetExist("assets/" + assetsVersion + "/console.html") {
+		assetsVersion = "latest"
+	}
```
kolharsam, line 306: Second one.

cli/commands/console.go
```diff
@@ -171,3 +200,126 @@ func (o *ConsoleOptions) Run() error {
...
+
+	// Template console.html
+	// FIXME: LoadTemplates
+	templateRender, err := util.LoadTemplates("assets/"+assetsVersion+"/", "console.html")
```
kolharsam, line 310: Third one

cli/commands/console.go
```diff
@@ -171,3 +200,126 @@ func (o *ConsoleOptions) Run() error {
...
+		c.Next()
+	}
+}
+
+func setFilePath(dir string) gin.HandlerFunc {
+	return func(c *gin.Context) {
+		// FIXME
+		host := getFilePath(dir)
+		c.Set("filedir", host)
+		c.Next()
```
kolharsam, line 261: The last one

thinknirmal:
Are there any progresses with this one? This is an important feature, especially in a multi-developer setup where schema changes must be tracked centrally.

joanrodriguez:
Will this ever get merged? I need it too...

sevensidedmarble:
What's the status on this? A lot of people would really appreciate this merged in

Mexicool (approved):

dvasdekis:
Hi all, could you please approve this?? We are waiting for this feature in Userland

LeulAria:
what is the solution here is it merge or ?

PR #3584, opened 2019-12-24
add query execution time and response size to ws-server logs in websocket transport

hgiasac:
### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

Add more two fields `query_execution_time` and `response_size` into `ws-server` event detail. 
They are computed on `message_sent` data event only.

```json
{
  "type":"ws-server",
  "timestamp":"2019-12-17T11:46:29.190+0700",
  "level":"debug",
  "detail": {
    "metadata": {
        "response_size":21,
        "query_execution_time": 8.54369e-4
    },
    "event": {
      "type":"message_sent",
      "detail": "{
        "type":"data",
        "id":"1",
        "payload": {
            ...
          }
        }
      }"
    },
    "websocket_id":"4c9dffd6-aa14-4539-b935-f40dd5647b2b"
  }  
}
```

Live query `TQueue` content are changed from `BL.ByteString` to `WSQueueResponse` ADT:
```haskell
data WSQueueResponse
  = WSQueueResponse
  { _wsqrMessage   :: !BL.ByteString
  , _wsqrEventInfo :: !(Maybe WSEventInfo)
  }
```

Add `LiveQueryMetadata` ADT as additional parameter in live query `OnChange` callback:

```haskell
data LiveQueryMetadata
  = LiveQueryMetadata
  { _lqmExecutionTime :: !(Maybe Double)
  }

type OnChange = GQResponse -> LiveQueryMetadata -> IO ()
```

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

Add more two fields `query_execution_time` and `response_size` into `ws-server` event detail. However, normal user won't see any difference because those logs are only sent on `debug` level

### Steps to test and verify

Run graphql engine with `--log-level=debug`
Request a subcription query 
View stdout logs


server/src-lib/Hasura/GraphQL/Transport/WebSocket/Protocol.hs
```diff
@@ -56,18 +56,19 @@ instance J.FromJSON ClientMsg where
...
 -- server to client messages
 
 data DataMsg
   = DataMsg
-  { _dmId      :: !OperationId
-  , _dmPayload :: !GraphqlResponse
+  { _dmId                 :: !OperationId
+  , _dmPayload            :: !GraphqlResponse
+  , _dmQueryExecutionTime :: !(Maybe Double)
   }
```
ecthiender: This is the ADT which represents the data frame sent to the client as response. Are we sending the query execution time in the response as well?

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Protocol.hs
```diff
@@ -128,11 +129,14 @@ encodeServerMsg msg =
...
-    , ("payload", encodeGraphqlResponse payload)
-    ]
+  SMData (DataMsg opId payload execTime) ->
+    let encPayload = encodeGraphqlResponse payload
+    in [ encTy SMT_GQL_DATA
+       , ("id", encJFromJValue opId)
+       , ("response_size", encJFromJValue $ BL.length $ encJToLBS encPayload)
+       , ("query_execution_time", encJFromJValue execTime)
+       , ("payload", encPayload)
+       ]
```
ecthiender: Just for my understanding, looks like we are sending the query execution time, response size etc. to the client in response as well?

hgiasac: Yeah. Forgot this. However, the internal payload is sent in bytestring already. We have to decode JSON again. One solution I think that we don't need to include `response_size` on HGE. Lux will parse and caculate response size

About `query_execution_time`, I will move it outside

ecthiender: Those details should only be logged right? Nothing in the responses should change?

ecthiender: > he internal payload is sent in bytestring already. We have to decode JSON again to get response size.

We can just measure the length of the bytestring? That's how it is done in the HTTP layer currently.

hgiasac: yes, just for logs only. 
> We can just measure the length of the bytestring? That's how it is done in the HTTP layer currently.

The bytestring response include more data `type` and `id` beside payload. If we accept them in `response_size`, it will be ok.

ecthiender: I think that's okay (to include the entire apollo-protocol frame to count the response size). Because the `response_size` metric indicates how many bytes the server is sending out for each response. Thoughts @shahidhk ?

shahidhk: Yeah, it can be the entire frame size.

hgiasac:
New log json structure is:

```json
{
  "type":"ws-server",
  "timestamp":"2019-12-17T11:46:29.190+0700",
  "level":"debug",
  "detail": {
    "metadata": {
        "response_size":21,
        "query_execution_time": 8.54369e-4
    },
    "event": {
       ...
    },
    "websocket_id":"4c9dffd6-aa14-4539-b935-f40dd5647b2b"
  }  
}
```

``

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -64,7 +65,12 @@ data Subscriber
   , _sOnChangeCallback :: !OnChange
   }
 
-type OnChange = GQResponse -> IO ()
+data OnChangeMeta
+  = OnChangeMeta
+  { _ocmExecutionTime :: !(Maybe Double)
+  }
+
+type OnChange = GQResponse -> OnChangeMeta -> IO ()
```
ecthiender: What does the name `OnChangeMeta` mean? AFAICT, it just has the execution time. Maybe its a good idea to add some comments on this ADT?

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -84,7 +103,7 @@ data WSConn a
   { _wcConnId    :: !WSId
   , _wcLogger    :: !(L.Logger L.Hasura)
   , _wcConnRaw   :: !WS.Connection
-  , _wcSendQ     :: !(STM.TQueue BL.ByteString)
+  , _wcSendQ     :: !(STM.TQueue (BL.ByteString, Maybe WSEventInfo))
```
ecthiender: Instead of a tuple maybe we can create a separate ADT, maybe called `Response` or something like that? Which has the two fields.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
ecthiender: Oh so I think this means metadata? Do you think it makes sense to call it `LiveQueryMetadata`?

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -85,9 +86,20 @@ data WSConnData
...
 
 type WSConn = WS.WSConn WSConnData
+
 sendMsg :: (MonadIO m) => WSConn -> ServerMsg -> m ()
-sendMsg wsConn =
-  liftIO . WS.sendMsg wsConn . encodeServerMsg
+sendMsg wsConn msg =
+  liftIO $ WS.sendMsg wsConn (encodeServerMsg msg) Nothing
+
+sendMsg' :: (MonadIO m) => WSConn -> ServerMsg -> LQ.OnChangeMeta -> m ()
```
ecthiender: Maybe use more explicit name? Something like `sendMsgWithMetadata`?

ecthiender:
Would be great if @lexi-lambda can also take a brief look at it (to check if it makes sense to calculate the exec time and put it in the queue).

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -64,7 +65,13 @@ data Subscriber
   , _sOnChangeCallback :: !OnChange
   }
 
-type OnChange = GQResponse -> IO ()
+-- live query onChange metadata, used for adding more extra analytics data
+data LiveQueryMetadata
+  = LiveQueryMetadata
+  { _lqmExecutionTime :: !(Maybe Double)
```
lexi-lambda: It would be nice to use a more precise type here until we have to convert it to a string—consider making this a `DiffTime`?

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -64,7 +65,13 @@ data Subscriber
...
+  }
+
+type OnChange = GQResponse -> LiveQueryMetadata -> IO ()
```
lexi-lambda: A `GQResponse` can be an error, and it seems unlikely that the `LiveQueryMetadata` information is very useful in that case. Might it make sense to instead use something like

```haskell
data LiveQueryResponse
  = LiveQueryResponse
  { _lqrPayload :: !BL.ByteString
  , _lqrExecutionTime :: !DiffTime
  }

type OnChange = GQResult LiveQueryResponse -> IO ()
```

instead?

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -310,8 +322,11 @@ onStart serverEnv wsConn (StartMsg opId q) = catchAndIgnore $ do
       logOpEv ODStarted (Just reqId)
       -- log the generated SQL and the graphql query
       L.unLogger logger $ QueryLog query genSql reqId
+      t1 <- liftIO TC.getCurrentTime -- for measuring response time purposes
       resp <- liftIO $ runExceptT action
-      either (postExecErr reqId) sendSuccResp resp
+      t2 <- liftIO TC.getCurrentTime -- for measuring response time purposes
```
lexi-lambda: It might be worth extracting this pattern into a helper function with a type like this:

```haskell
withElapsedTime :: MonadIO m => m a -> m (DiffTime, a)
```

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
hgiasac: @lexi-lambda currently LiveQuery module are too coupling on `GQResponse` because it is content of `GraphqlResponse`. If we change the structure, it  will take many step too convert back and fore between `GQResult LiveQueryResponse` to `GQResponse`. Do you think we need to decouple live query `GraphqlResponse` too?

```haskell
data DataMsg
  = DataMsg
  { _dmId      :: !OperationId
  , _dmPayload :: !GraphqlResponse
  }

data GraphqlResponse
  = GRHasura !GQResponse
  | GRRemote !RemoteGqlResp
```

lexi-lambda: I see—I misunderstood the way these types were being used. It seems like maybe `GQResponse` should just be updated to be `GQResult LiveQueryResponse`, then, since as far as I can tell, `GQResponse` is only used for websockets? But in general these types seem a little unnecessarily twisty to me; it might be better to do a little untangling to find a better representation.

I’m not immediately sure what that representation ought to be, but I think the end goal of avoiding including `LiveQueryMetadata` with errors seems reasonable. If you disagree, then by all means, I could be convinced to leave the type alone: I don’t feel super strongly about it. But if the current approach is just a side-effect of the types currently being structured the way they are rather than an intentional design decision, I think it would be nice to do something better.

hgiasac: IMO, we can define a type class `EncoderJSON` to generalize `GQResult` and `encodeGQResp` function

```haskell
class EncoderJSON a where
  encJSON :: a -> EncJSON

-- encodeGQResp :: GQResponse -> EncJSON
-- change to
-- encodeGQResp :: EncoderJSON a => GQResult a -> EncJSON
```

We won't care about `GQSuccess` content anymore. Moreover, we can manage what fields are returned instead of encoding entire data

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -68,23 +70,49 @@ $(J.deriveToJSON
...
 instance L.ToEngineLog WSLog L.Hasura where
   toEngineLog wsLog =
     (L.LevelDebug, L.ELTInternal L.ILTWsServer, J.toJSON wsLog)
 
+data WSQueueResponse
+  = WSQueueResponse
+  { _wsqrMessage   :: !BL.ByteString
+  , _wsqrEventInfo :: !(Maybe WSEventInfo)
+  -- ^ extra metadata that we use for other actions, such as print log
+  -- we don't want to inlcude them into websocket message payload
```
lexi-lambda: Please use [Haddock comment syntax](https://haskell-haddock.readthedocs.io/en/latest/markup.html#documenting-parts-of-a-declaration) for comments like there where it makes sense. (I’ve made the change here myself, since it’s a small thing, but just so you know in the future.)

server/src-lib/Hasura/Server/Utils.hs
```diff
@@ -214,3 +214,10 @@ makeReasonMessage errors showError =
     [singleError] -> "because " <> showError singleError
     _ -> "for the following reasons:\n" <> T.unlines
          (map (("  • " <>) . showError) errors)
+
+withElapsedTime :: MonadIO m => m a -> m (NominalDiffTime, a)
```
jberryman: FWIW I wrote a similar utility function in https://github.com/hasura/graphql-engine/pull/3622/files#diff-b4adc70671eae2d8ba4ad21bd16ede86R104  I considered `NominalDiffTime` but ultimately went a different route since having a duration type tagged with a unit is less error-prone when using literals and when serializing. But there are tradeoffs for both approaches...

I also use the monotonic clock, which is a more strictly correct way to get timing, but again this probably doesn't matter (unless some other part of the code blows up or loops on unexpected negative duration etc...)

jberryman: @lexi-lambda @hgiasac actually I think I need an opinion here. I'm rebasing my PR again on this and now there's a block of code we time twice ) in two different ways which doesn't make much sense. 

I'm okay refactoring to use `NominalDiffTime` (or maybe even better, a `newtype` called `Seconds`) as a standard duration type

lexi-lambda: @jberryman I wrote a library for doing type-safe unit conversion stuff in [`Data.Time.Clock.Units`](https://github.com/hasura/graphql-engine/blob/bda02f81923bb79c96ea3bd7120bb29d548f1f93/server/src-lib/Data/Time/Clock/Units.hs). It’s all based on `DiffTime`, though, not `NominalDiffTime`, since as you mention it’s the Right Thing here.

That said, I’ve never actually been 100% clear on how you’re supposed to actually get a `DiffTime` with the usual `Data.Time.Clock` API. As far as I can tell, it doesn’t expose a monotonic clock. There’s `Data.Time.Clock.TAI`, but there’s no guaranteed way to access a clock that produces an `AbsoluteTime`.

My preference would be to wrap `GHC.Clock` so that it produces a `DiffTime` and use that, along with the `Data.Time.Clock.Units` module mentioned above. Does that seem reasonable to you?

jberryman: Oh cool, I looked but totally missed that module. It looks good!

Types help a _lot_ but time is crazy; my understanding, which is probably not entirely correct:

- `AbsoluteTime` was defined to be an epoch time, to support the _possibility_ of a correct difference operation (`DiffTime`); no ecosystem really developed out of these (or maybe there are some good libraries out there! I've only used `timezone-olsen` which uses completely separate types IIRC)
- `NominalDiffTime` (_nominal_ in the sense of "it doesn't really work"?) was defined separately with much self-loathing to support the useful but formally nonsensical diffing of `UTCTime` (which as they note is only correct (or legally binding?) if the `UTCTime`s are earlier than a horizon of 6 months from when you perform the calculation...)
- I think there fundamentally is not a way to get `AbsoluteTime` as they define it from your machine without a table of leap seconds; `POSIXTime` is what we usually think of when we think of an epoch time but is, I think, approximate wrt leap seconds (...very unclear about this)
- the monotonic clock is orthogonal to all of this; it's just an API (in ghc ) that gives you a ticking counter that only increases (the regular system clock can jump forward or backwards). afaiu a _reading_ of the monotonic clock has no meaning; it's only good for measuring durations.

Sooooo ... (what were we talking about?) 

> My preference would be to wrap GHC.Clock so that it produces a DiffTime and use that, along with the Data.Time.Clock.Units module mentioned above. Does that seem reasonable to you?

I think either `DiffTime` or `NominalDiffTime` are equally valid to use as a duration type in that module, but `NominalDiffTime` is the one you can actually use in calculations with time moments (for better or worse). So we might want to consider using `NominalDiffTime` in your module, but I have no real opinion there and yes that sounds good!

jberryman: @lexi-lambda also re. an algebra for time etc., I couldn't remember the name of this structure and had to email a colleague: https://ro-che.info/articles/2013-01-08-torsors

jberryman: Oh I see you can get an `AbsoluteTime` with http://hackage.haskell.org/package/time-1.9.3/docs/Data-Time-Clock-System.html#v:systemToTAITime , but now I'm a more lost...

Issue #3533, opened 2019-12-13
[RFC] run hasura without hdb_catalog tables


tirumaraiselvan:
This is a big architectural refactor which will obviate the need for hdb_* schemas and tables. Instead, hasura can start from just the metadata.yaml file which can be given as a server-flag.

Why do we need hdb_* schemas in the first place?

~1. Insert permissions are based on postgres triggers on hdb_views : How can we move the logic to the app layer?~ Solved via https://github.com/hasura/graphql-engine/pull/3598 
2. We store lot of user generated events for event-triggers and actions in hdb_catalog : This can probably be solved by asking for user-defined tables. This also makes sense as these are user-events and not hasura configuration.
3. Easier dev workflow: Currently, having hdb_* is easier because you don't need to continuously reload the metadata file or even generate one (e.g. how will people try-out on heroku?).

If we do this, we can solve the following use-cases quite trivially:

1. Hasura(s) on multi-tenant postgres
2. Handle lot of existing postgres users, who don't have admin access on postgres.
3. Co-existing versions / Blue Green deployments of Hasura
4. https://github.com/hasura/graphql-engine/issues/3496

But this would create new problems like:

1. What if users modify metadata on prod? How will they persist new metadata? How will they sync across instances?
2. Event logs are in public namespaces now. What if we need to do migrations on event log tables, etc?

shahidhk:
- CLI makes use of `hdb_catalog.schema_migrations` to store the migrations state
- Console uses `hdb_catalog.hdb_version.console_state` to store some state
- CLI/Console/Server all uses `hdb_catalog.hdb_version.hasura_uuid` for telemetry

davidecavestro:
IMHO the ability to alter metadata on prod without redeploy is a feature... indeed a great one!

As of now a custom UI can define a new table simply demanding some calls to _run_sql_, _track_table_ and _create_xyz_permission_.
It can be easily achieved through a companion service with admin access.

When data structure is quite dynamic, automated redeploys wouldn't provide the same QoS as live changes: i.e. clients should probably open new websockets for live subscriptions and so on.

soneymathew:
@tirumaraiselvan Do we even need metadata in prod if persisted query can be specified as a constraint?

Inspired by the spirit of disappearing frameworks that does all the work at compile time, The key lessons that I can take away from a few are
- Not everything needs to be done at runtime for every request. (Eg: sveltejs)
[...]

PR #3598, opened 2019-12-27
remove hdb_views for inserts

paf31:
### Description

This is a precursor to #3533. It is an early WIP, but I'd like to put up this PR for review. In particular, I still need to:

- [x] investigate whether I can avoid passing around as many aliases for CTEs
- [x] add more documentation
- [x] fix up the tests and write new tests
- [x] test the CLI changes
- [x] (if we decide it's the right approach), add the `err(text)` function to `hdb_views` or `hdb_catalog` and create a migration for it

### Affected components

- [x] Server

### Related Issues
#3710

#### Catalog upgrade

- [ ] No
- [x] Yes
  - [x] Updated docs with SQL for downgrading the catalog 

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


paf31:
Tests are now passing 🎉

0x777:
@paf31 I looked through the changes, from what I understand, the generated insert statement is of this form:

```sql
WITH 
  "<table>__mutation_result_alias" AS ( insert statement returning * ),
  "<table>__inserted" AS ( SELECT .. ),
  "<table>__checks" AS ( SELECT bool_and( .. WHEN check_expression ..) ) 
SELECT
  "<table>__inserted".*
FROM
  "<table>__inserted",
  "<table>__checks"
```

I experimented with a `check` condition which involves a condition on another table through a relationship, in such cases the following insert statement has a better plan (check expression is computed in returning):
```sql
WITH 
  "<table>__mutation_result_alias" AS 
    ( insert statement returning *, check_expression AS _check_expression_value ),
  "<table>__inserted" AS ( SELECT .. ),
  "<table>__checks" AS ( SELECT bool_and( .. WHEN _check_expression_value ..) ) 
SELECT
  "<table>__inserted".*
FROM
  "<table>__inserted",
  "<table>__checks"
```

The two statements that I used are as follows:
<details><summary>Statement 1</summary>

```sql
explain analyze with "address_mutation" as (
  update
    address a
  set
    location = 'blr'
  where
    id = 1 returning a.*
),
"address_check" as (
  select
    bool_and(
      case
        when (
          exists (
            select
              1
            from
              "user" u
            where
              u.id = a.user_id
          )
        ) THEN NULL
        ELSE "hdb_catalog"."check_violation"()
      end
    )
  from
    "address_mutation" a
),
"address_return" as (
  select
    json_agg(a.*) as returning
  from
    "address_mutation" a
)
select
  a.*
from
  "address_return" a,
  "address_check";
```
</details>


<details><summary>Statement 2</summary>

```sql
explain analyze with "address_mutation" as (
  update
    address a
  set
    location = 'blr'
  where
    id = 1 returning a.*,
    (
      exists (
        select
          1
        from
          "user" u
        where
          u.id = a.user_id
      )
    ) as _check_condition
),
"address_check" as (
  select
    bool_and(
      case
        when (a._check_condition) THEN NULL
        else "hdb_catalog"."check_violation"()
      end
    )
  from
    "address_mutation" a
),
"address_return" as (
  select
    json_agg(a.*) as returning
  from
    "address_mutation" a
)
select
  a.*
from
  "address_return" a,
  "address_check";
```
</details>

Regarding the `check_violation` function defined in `hdb_catalog`, it would be best not to rely on functions/views defined in `hdb_catalog` in the light of https://github.com/hasura/graphql-engine/issues/3533. The function body can be inlined in the statement itself?

During this refactor, I was also hoping that we would end up with simplified logic and fix some issues: 
### Better error messages
It would be good to have the object on which the constraint failed in the returned error message instead of the current "check constraint failed" (maybe something like "the following object '{"location":"blr"}' does not satisfy the check constraint"). This is particularly useful when you are inserting more than one objects or when inserting related data (through relationships). 

### Different code paths based on relationships in selection set
Assume the following schema:
```
user:
- id
- name
- addresses (array relationship to address table)
address:
- id
- user_id 
- location 
- user (object relationship to user table)
```
Let's say we issued a mutation such as this:
```graphql
mutation insert_address {
  insert_address(objects: {location: "blr"}) {
    affected_rows
    returning {
      user {
        addresses {
          id
          location
      }
    }
  }
}
```
Now, ideally you would expect to generate SQL such as this:
```
with "insert_result" AS ( INSERT INTO .. )
select .. from "insert_result"
```
However, this doesn't work as expected (the address inserted as a part of this mutation will not be returned in the `addresses` relationship) because of how CTEs [behave](https://www.postgresql.org/docs/current/queries-with.html) :
> All the statements are executed with the same snapshot (see Chapter 13), so they cannot “see” one another's effects on the target tables

So, whenever there are relationships in `returning`'s selection set, what we do is something different:
```sql
BEGIN
inserted rows <- INSERT INTO ...
WITH (inserted rows) AS "insert_result" SELECT .. from "insert_result"
COMMIT
```
This also applies in case of deletes and updates. I was wondering if having the same code path would reduce some of the issues that we have been seeing with inserts (https://github.com/hasura/graphql-engine/issues/3148, https://github.com/hasura/graphql-engine/pull/3375, https://github.com/hasura/graphql-engine/issues/3609). 

I have to admit, I'm not 100% familiar with the current insert mutations implementation so I do not know the implication of the changes that I'm proposing. @rakeshkky thoughts?

paf31:
@0x777 I'm not sure how I would inline the function body into the call site since it contains a `RAISE` statement.

I'm all for consolidating the different code paths. To be honest, I found the current way quite confusing, and I was trying to preserve the existing behavior.

server/src-lib/Hasura/SQL/DML.hs
```diff
@@ -283,7 +283,8 @@ data SQLExp
   | SELit !T.Text
   | SEUnsafe !T.Text
   | SESelect !Select
-  | SEStar
+  | SEStar !(Maybe Qual)
```
paf31: This isn't needed any more, but it seems like a useful addition anyway. Keep or revert?

lexi-lambda: Seems fine to keep it to me.

docs/graphql/manual/deployment/downgrading.rst
```diff
@@ -103,6 +103,12 @@ You can downgrade the catalogue from a particular version to its previous versio
   :backlinks: none
   :depth: 1
   :local:
+  
+From 29 to 28
```
shahidhk: Catalog version as per this PR is 30

Shouldn't we have 30 to 29 downgrade instructions?

paf31: Yes, thanks. I forgot to change this when I merged master (which bumped from 29 to 30).

PR #3601, opened 2019-12-30
cli: add browser flag to console command (close #3333)

### LINKED ISSUE #3333: CLI option to choose the browser when opening the console
Hi,

Is there any way to choose which browser will be used when opening the console with the CLI? A bit like the `BROWSER` environment variable in create-react-app.

Thanks,
Thibaut

ashishra0:
### Description
Add a browser flag to the `hasura console` command. This flag allows the user to run the hasura console on their choice of browser rather than the default one.

```
$ hasura console --browser firefox
INFO console running at: http://localhost:9695/ on firefox
```
The above command  with the browser flag will override the default browser for  eg `chrome` and will open the console only on `firefox`.

### Affected components

- [x] CLI

### Related Issues
#3333 

### Solution and Design
The fix to the issue was to add a new flag to the console command that when invoked with name of browser will open the console only on the specified browser.


tibotiber:
Tested here and works as expected 👍.

cli/commands/console.go
```diff
@@ -166,17 +168,31 @@ func (o *consoleOptions) run() error {
...
 	if !o.DontOpenBrowser {
-		o.EC.Spin(color.CyanString("Opening console using default browser..."))
-		defer o.EC.Spinner.Stop()
+		if o.Browser != "" {
+			o.EC.Spin((color.CyanString("Opening console on: %s", o.Browser)))
+			defer o.EC.Spinner.Stop()
+			err = open.RunWith(consoleURL, o.Browser)
+			if err != nil {
+				o.EC.Logger.WithError(err).Warn("Error opening browser, try to open the url with another browser")
+				return err
```
shahidhk:
```suggestion
```

cli/commands/console.go
```diff
@@ -166,17 +168,31 @@ func (o *consoleOptions) run() error {
...
 		}
 	}
 
 	o.EC.Spinner.Stop()
-	log.Infof("console running at: %s", consoleURL)
+	if o.Browser != "" {
+		log.Infof("console running at: %s on %s", consoleURL, o.Browser)
+	} else {
+		log.Infof("console running at: %s", consoleURL)
+	}
```
shahidhk: We don't need to do this here since console is running at this point irrespective of the browser flag.

Change it to just
```go
log.Infof("console running at: %s", consoleURL)
```

cli/commands/console.go
```diff
@@ -60,6 +60,7 @@ func NewConsoleCmd(ec *cli.ExecutionContext) *cobra.Command {
 	f.StringVar(&opts.Address, "address", "localhost", "address to serve console and migration API from")
 	f.BoolVar(&opts.DontOpenBrowser, "no-browser", false, "do not automatically open console in browser")
 	f.StringVar(&opts.StaticDir, "static-dir", "", "directory where static assets mentioned in the console html template can be served from")
+	f.StringVar(&opts.Browser, "browser", "", "open user specified browser")
```
shahidhk:
```suggestion
	f.StringVar(&opts.Browser, "browser", "", "open console in a specific browser")
```

shahidhk:
Should give credits to @tibotiber too when merging.

PR #3605, opened 2019-12-30
Report errors in parallel when batching

### LINKED ISSUE #1812: Feature Request: Batching queries
As it currently stands, the graphql-engine is unable to process batched queries ([batched query example](https://blog.apollographql.com/batching-client-graphql-queries-a685f5bcd41b)). The following response is returned when attempting to batch queries through Apollo: 
`{"errors":[{"extensions":{"path":"$","code":"parse-failed"},"message":"When parsing the constructor GraphQLRequest of type Hasura.GraphQL.Transport.HTTP.Protocol.GraphQLRequest expected Object but got Array."}]}`

paf31:
Fixes #1812

### Description

Reports errors in parallel, treating each operation independently, when receiving batched requests.

WIP, because I would still like to figure out a good way to integrate this into the test suite, but I'm putting this PR up now so that I can get a test environment to share.

### Affected components

- [x] Server
- [x] Tests

### Related Issues

#1812

### Steps to test and verify

Here is a simple test case on the terminal:

```
$ cat | curl -X POST 'http://localhost:8080/v1/graphql' -d @- | jq -S
[
  {
    "query": "query first { user { email id name } }",
    "variables": null
  },
  {
    "query": "query second { role { id name } }",
    "variables": null
  }
]
```

```json
[
  {
    "errors": [
      {
        "extensions": {
          "code": "validation-failed",
          "path": "$.selectionSet.user"
        },
        "message": "field \"user\" not found in type: 'query_root'"
      }
    ]
  },
  {
    "errors": [
      {
        "extensions": {
          "code": "validation-failed",
          "path": "$.selectionSet.role"
        },
        "message": "field \"role\" not found in type: 'query_root'"
      }
    ]
  }
]
```

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


paf31:
@thefliik If you are interested in trying out this change, there should be a test environment available soon once CI finishes. Thanks!

server/tests-py/queries/graphql_query/basic/select_query_batching_with_mutation.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - id: '2'
+        name: Temp User
+  - data:
+      update_user:
+        affected_rows: 1
+        returning:
+        - id: '2'
+query:
+  - query: |
+      mutation {
```
paf31: I'm not sure if this is poor form (updating the test data set from a test like this), so I try to put back the original data at the end so that it doesn't affect any other tests.

lexi-lambda: I think the test setup/teardown actions run around each individual test, not around the whole test group, so I *think* this is unnecessary.

paf31:
This seems to work well with `apollo-client`, verified using this little test case:

```js
const ApolloClient = require('apollo-client').default;
const BatchHttpLink = require('apollo-link-batch-http').BatchHttpLink;
const InMemoryCache = require('apollo-cache-inmemory').InMemoryCache;

const fetch = require('node-fetch');
const gql = require('graphql-tag').default;

const link = new BatchHttpLink({ 
  uri: 'http://localhost:8080/v1/graphql', 
  fetch 
});

const client = new ApolloClient({ link, cache: new InMemoryCache() });

client.query({
  query: gql`query first { user { id } }`,
}).then(user => console.log(user))
  .catch(err => console.error(err.message));
  
client.query({
  query: gql`query second { role { id } }`,
}).then(role => console.log(role))
  .catch(err => console.error(err.message));
```

Using netcat, I can see the requests get batched, and then errors get handled independently on the client.

thefliik:
> @thefliik If you are interested in trying out this change, there should be a test environment available soon once CI finishes. Thanks!
> ...
> Review app https://hge-ci-pull-3605.herokuapp.com is deleted

@paf31 looks like I was too slow on the draw. I'm not actually using batching at the moment though, so I probably couldn't have provided any new feedback anyway.

👍

Issue #3271, opened 2019-10-29
Incorrect ordering in returned JSON from nested mutation


jberryman:
We get failures from these two queries

	   server/tests-py/queries/graphql_mutation/insert/nested/articles_with_author.yaml
	   server/tests-py/queries/graphql_mutation/insert/nested/author_with_articles.yaml

after https://github.com/hasura/graphql-engine/pull/2831

Marked `xfail` in that branch

0x777:
Fixed with https://github.com/hasura/graphql-engine/commit/53b9cabd88d61f7ffbf8720b201410ea2d0bace2

PR #3618, opened 2020-01-02
fix few bugs related to nested insert returning (fix #3609, #3642, #3271)

### LINKED ISSUE #3609: Insert mutation with returning computed field and nested insert gives error
I am facing an issue with insert mutation that has a relationship based nested insert together with a returning computed field. A sample insert mutation query below:
 
`insert account ( objects: {
     name: nameValue,
     nestedAddress: { data : { street : streetValue }}}){
         returning {
            accountId
            addressId
           computedField
        }
   }`

The error I get is "cannot cast type record to ... "

If the computedField is removed from the returning field, or if the computedField is retained but the relationship based nested insert is removed, the mutation works. 

I am using Hasura v1.0.0-rc.1, on Fargate, connected to Aurora Postgres 10.7

rakeshkky:
### Description

This PR fixes:- 
#3609 - The nested insert with returning computed fields is failing with Postgres query error
#3642 - JSON/JSONB columns in nested insert returning are stringified.
#3271 - The order of keys in nested insert returning.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

fix #3609, #3642 and #3271 

### Solution and Design

- (3609) While building the SQL query for `returning` field, generate the `VALUES` expression in CTE in the order of columns ordinal position. Thus, the outcome rows of CTE are castable to table type in input arguments of the computed field function.
- (3642) Use `TxtEncodedPGVal` type to return the column values after inserting. Parse the text value as Postgres literals based on the column type.
- (3272) Use [Data.Aeson.Ordered](https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Data/Aeson/Ordered.hs) module's `Value` type to construct the returning of nested inserts. Thus the order of the keys is persisted from Postgres.

### Steps to test and verify

Follow the respective issue's description for reproduction guide. 

#### Catalog upgrade

- [ ] No
- [x] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-rsr/migrations/28_to_29.sql
```diff
@@ -0,0 +1,89 @@
...
+      c.table_name,
+      c.table_schema,
+      json_agg(
+        json_build_object(
+          'name', name,
+          'type', type,
+          'is_nullable', is_nullable :: boolean,
+          'references', primary_key_references,
+          'description', description,
+          'ordinal_position', ordinal_position
```
rakeshkky: Plumbing in `ordinal_position` of a column.

server/src-lib/Hasura/RQL/DML/Mutation.hs
```diff
@@ -111,9 +111,11 @@ mkSelCTEFromColVals parseFn qt allCols colVals =
         }
   where
     tableAls = S.Alias $ Iden $ snakeCaseQualObject qt
-    colNames = map pgiColumn allCols
+    sortedCols = flip sortBy allCols $ \lCol rCol ->
+                 compare (pgiOrdinalPosition lCol) (pgiOrdinalPosition rCol)
+    colNames = map pgiColumn sortedCols
     mkTupsFromColVal colVal =
-      fmap S.TupleExp $ forM allCols $ \ci -> do
+      fmap S.TupleExp $ forM sortedCols $ \ci -> do
```
lexi-lambda: Could you please add a comment here that explains why it’s important that the columns be sorted here? Ideally with an example SQL query that illustrates what goes wrong if they are not sorted.

lexi-lambda:
Ah, actually, I think I remembered what my other question was! I was trying to understand why we need the CTE in the first place, and I was reading through the code to try and figure that out, but I didn’t get far enough to have a very confident answer either way.

If we use the approach mentioned in https://github.com/hasura/graphql-engine/pull/3618#discussion_r364496920, could we get away with just using a literal `ROW(...)` expression instead of the CTE? And would that be any simpler? Maybe not, but I was mostly just trying to understand this bit of functionality better.

rakeshkky:
@lexi-lambda I choose the `CTE` because there's an AST implemented and it is easy to write the code for :smile:. So there are a lot of such `CTE`s in the code for SQL functions, mutations etc.

I'll look at whether I could get rid of using `CTE` and inline the values after `FROM`.

lexi-lambda:
> I choose the `CTE` because there's an AST implemented and it is easy to write the code for 😄. So there are a lot of such `CTE`s in the code for SQL functions, mutations etc.

I think that’s a reasonable reason. 🙂 I’d mostly just like to understand what this code is doing better (so I can write some comments!). I will try to take another look at this today.

rakeshkky:
@lexi-lambda We need `CTE` for mutations.

Generally the SQL expression generated for mutations look like following
```
WITH "some_alias" as (<mutation SQL statement (insert/update/delete)>)
SELECT json_build_object(<SQL statement which returns the graphql query output>)
```

If "returning" has nested fields then two statements are executed to perform a mutation
1
```
WITH "some_alias" as (<mutation SQL statement (insert/update/delete)>)
SELECT json_agg(<fetch the mutated column values as text>)
```
2
```
WITH "some_alias" as (<using 'VALUES' to generate virtual table rows with column values returned in step 1>)
SELECT json_build_object(<SQL statement which returns the graphql query output>)
```

Technically we can get rid of `CTE` in the step 2 of the later. But, the code which builds the SQL expression for "returning" accepts the CTE (See https://github.com/rakeshkky/graphql-engine/blob/3fa3685f54c2566d88fbf2f92daebb5f9e57724d/server/src-lib/Hasura/RQL/DML/Returning.hs#L98). We can avoid that by having two separate functions for building the mutation output SQL expression which I feel bit more confusing and having a `CTE` in the generated SQL makes easy to read and debug. :slightly_smiling_face:

Issue #3547, opened 2019-12-16
implement nginx's $request_time logging parameter


0x777:
Currently, we only log the execution time for a given request. It would be nice to measure the time from 'the arrival of first byte of the request' to the 'last byte sent as response'. This can help people debug network issues easily. 

From [nginx's docs](https://nginx.org/en/docs/http/ngx_http_log_module.html#var_request_time):
> $request_time
    request processing time in seconds with a milliseconds resolution; time elapsed between the first bytes were read from the client and the log write after the last bytes were sent to the client

jberryman:
This makes sense to me at a high level. Could you summarize the network issue and how you ended up debugging it?

0x777:
The network issue was that the clients connected to graphql-engine (deployed on k8s with istio) were noticing that the requests take a long time to finish(1-2 seconds). However in the logs of graphql-engine, the execution times were in the order of few milliseconds. The network connection between graphql-engine and the client was the issue. If `$request_time` was also present in the log, it would've been easier to pin point the issue to the network layer.

0x777:
> how you ended up debugging it?
By adding logging on the clients and tracing those requests in graphql-engine (through a session variable x-hasura-trace-id).

lexi-lambda:
@0x777 Why is this reopened?

0x777:
@lexi-lambda Because @jberryman added the time taken to read the request body, but `request_time` as described in the issue is not implemented. Looks like this may not be possible with spock currently.

lexi-lambda:
Got it, thank you! Maybe we need to add it at the WAI level, then?

Naveenaidu:
IIUC `request_time` is defined as follows:

```
request_time = time taken to read the request body +
                         time taken to execute that request and send the response back to client
```

While I was working on [Server Metrics](https://github.com/hasura/graphql-engine-mono/pull/222) issue - I noticed that in our `mkSpockAction` function, we have `ioWaitTime` as well as `serviceTime` which can be used for calculating the `request_time` mentioned in this issue.

Referencing from the code:

* `ioWaitTime` is retrieved at [this line](https://github.com/hasura/graphql-engine-mono/blob/58b0033be2ac9c0c757319a1a5f735e53f08307b/server/src-lib/Hasura/Server/App.hs#L281)
```haskell
(ioWaitTime, reqBody) <- withElapsedTime $ liftIO $ Wai.strictRequestBody req
```

[...]

PR #3622, opened 2020-01-03
Add request timings and count histograms to telemetry. Closes #3552

### LINKED ISSUE #3552: Gather usage telemetry data for cache
See action items in https://github.com/hasura/graphql-engine/issues/3530

jberryman:
We upload a set of accumulating timers and counters to track service
time for different types of operations, across several dimensions (e.g.
did we hit the plan cache, was a remote involved, etc.)

### Affected components

- [x] Server
- [x] Tests

### Related Issues
#3552 although metrics should hopefully give us general visibility into performance

### Steps to test and verify
run tests

- marked an unsafe forked thread as `TODO immortal`, but we should fix all of these together (see #3768 )
- I didn't document changes in `telemetry.rst`; that doc is already stale. It would be nice if we could generate it from tests

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Control/Concurrent/Extended.hs
+++ b/server/src-lib/Control/Concurrent/Extended.hs
@@ -1,5 +1,7 @@
 module Control.Concurrent.Extended
   ( module Control.Concurrent
+  , sleep
+  -- * Deprecated
   , threadDelay
   ) where
 
@@ -8,9 +10,15 @@ import           Prelude
 import qualified Control.Concurrent    as Base
 
 import           Control.Concurrent    hiding (threadDelay)
-import           Data.Time.Clock       (DiffTime)
-import           Data.Time.Clock.Units (Microseconds (..))
+import           Data.Time.Clock.Units (Microseconds (..), DiffTime)
 
--- | Like 'Base.threadDelay', but takes a 'DiffTime' instead of an 'Int'.
-threadDelay :: DiffTime -> IO ()
-threadDelay = Base.threadDelay . round . Microseconds
+-- | Like 'Base.threadDelay', but takes a 'DiffTime' instead of an 'Int' microseconds.
+--
+-- NOTE: you cannot simply replace e.g. @threadDelay 1000@ with @sleep 1000@ since those literals
+-- have different meanings!
+sleep :: DiffTime -> IO ()
+sleep = Base.threadDelay . round . Microseconds
+
+{-# DEPRECATED threadDelay "Please use `sleep` instead (and read the docs!)" #-}
+threadDelay :: Int -> IO ()
+threadDelay = Base.threadDelay
--- a/server/src-lib/Data/Time/Clock/Units.hs
+++ b/server/src-lib/Data/Time/Clock/Units.hs
@@ -25,55 +25,91 @@ You can also go the other way using the constructors rather than the selectors:
 0.5
 @
 
-Generally, it doesn’t make sense to pass these wrappers around or put them inside data structures,
-since any function that needs a duration should just accept a 'DiffTime', but they’re useful for
-literals and conversions to/from other types. -}
+NOTE: the 'Real' and 'Fractional' instances just essentially add or strip the unit label (as 
+above), so you can't use 'realToFrac' to convert between the units types here. Instead try 
+'fromUnits' which is less of a foot-gun.
+
+The 'Read' instances for these types mirror the behavior of the 'RealFrac' instance wrt numeric
+literals for convenient serialization (e.g. when working with env vars):
+
+@
+>>> read "1.2" :: Milliseconds 
+Milliseconds {milliseconds = 0.0012s}
+@
+
+Generally, if you need to pass around a duration between functions you should use 'DiffTime'
+directly. However if storing a duration in a type that will be serialized, e.g. one having
+a 'ToJSON' instance, it is better to use one of these explicit wrapper types so that it's 
+obvious what units will be used. -}
 module Data.Time.Clock.Units
   ( Days(..)
   , Hours(..)
   , Minutes(..)
-  , Seconds
-  , seconds
+  , Seconds(..)
   , Milliseconds(..)
   , Microseconds(..)
   , Nanoseconds(..)
+  -- * Converting between units
+  , Duration(..)
+  , fromUnits
+  -- * Reexports
+  -- | We use 'DiffTime' as the standard type for unit-agnostic duration in our
+  -- code. You'll need to convert to a 'NominalDiffTime'  (with 'fromUnits') in
+  -- order to do anything useful with 'UTCTime' with these durations.
+  --
+  -- NOTE: some care must be taken especially when 'NominalDiffTime' interacts
+  -- with 'UTCTime':
+  --
+  --  - a 'DiffTime' or 'NominalDiffTime' my be negative
+  --  - 'addUTCTime' and 'diffUTCTime' do not attempt to handle leap seconds
+  , DiffTime
   ) where
 
 import           Prelude
 
+import           Control.Arrow   (first)
+import           Data.Aeson
+import           Data.Hashable
 import           Data.Proxy
 import           Data.Time.Clock
 import           GHC.TypeLits
+import           Numeric         (readFloat)
 
-type Seconds = DiffTime
 
-seconds :: DiffTime -> DiffTime
-seconds = id
+newtype Seconds = Seconds { seconds :: DiffTime }
+  -- NOTE: we want Show to give a pastable data structure string, even
+  -- though Read is custom.
+  deriving (Duration, Show, Eq, Ord, ToJSON, FromJSON)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit (SecondsP 1))
 
+-- TODO if needed: deriving (ToJSON, FromJSON) via (TimeUnit ..) making sure
+--      to copy Aeson instances (with withBoundedScientific), and e.g.
+--         toJSON (5 :: Minutes) == Number 5
 newtype Days = Days { days :: DiffTime }
-  deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit (SecondsP 86400))
+  deriving (Duration, Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit (SecondsP 86400))
 
 newtype Hours = Hours { hours :: DiffTime }
-  deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit (SecondsP 3600))
+  deriving (Duration, Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit (SecondsP 3600))
 
 newtype Minutes = Minutes { minutes :: DiffTime }
-  deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit (SecondsP 60))
+  deriving (Duration, Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit (SecondsP 60))
 
 newtype Milliseconds = Milliseconds { milliseconds :: DiffTime }
-  deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit 1000000000)
+  deriving (Duration, Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit 1000000000)
 
 newtype Microseconds = Microseconds { microseconds :: DiffTime }
-  deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit 1000000)
+  deriving (Duration, Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit 1000000)
 
 newtype Nanoseconds = Nanoseconds { nanoseconds :: DiffTime }
-  deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit 1000)
+  deriving (Duration, Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, Hashable, RealFrac) via (TimeUnit 1000)
 
+-- Internal for deriving via
 newtype TimeUnit (picosPerUnit :: Nat) = TimeUnit DiffTime
   deriving (Show, Eq, Ord)
 
@@ -92,6 +128,9 @@ instance (KnownNat picosPerUnit) => Num (TimeUnit picosPerUnit) where
   signum (TimeUnit a) = TimeUnit $ signum a
   fromInteger a = TimeUnit . picosecondsToDiffTime $ a * natNum @picosPerUnit
 
+instance (KnownNat picosPerUnit) => Read (TimeUnit picosPerUnit) where
+  readsPrec _ = map (first fromRational) . readFloat
+
 instance (KnownNat picosPerUnit) => Fractional (TimeUnit picosPerUnit) where
   TimeUnit a / TimeUnit b = TimeUnit . picosecondsToDiffTime $
     diffTimeToPicoseconds a * natNum @picosPerUnit `div` diffTimeToPicoseconds b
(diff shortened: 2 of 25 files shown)


jberryman:
@0x777 
I decided to base the resolution to #3547 on the work mentioned in the body of this PR to use some of the timing utilities. Can you please review 29b2963 ?

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -192,7 +191,10 @@ data OperationLog
   { olRequestId          :: !RequestId
   , olUserVars           :: !(Maybe UserVars)
   , olResponseSize       :: !(Maybe Int64)
+  , olQueryIoTime        :: !(Maybe Double)
```
0x777: What is `query_io_time`? From what I followed, is it the time taken for the request body to be read?

jberryman: Yep, can the comment be improved?

I'm towards the end of a big refactor and rebase though, so some details might change. A question: do you think it's important to also include the IO wait time for the send of the response to the client? For whatever reason I didn't think to do this, but since there are often different bandwidth limits on uploads and downloads it might be a good idea. 

What do you think?

jberryman: I couldn't see an obvious way to time sending the response to the client so haven't done that below. I can dig further into spock if that seems important.

jberryman:
@lexi-lambda @0x777 this is ready for review again

server/src-lib/Control/Concurrent/Extended.hs
```diff
@@ -8,9 +10,15 @@ import           Prelude
...
+-- | Like 'Base.threadDelay', but takes a 'NominalDiffTime' instead of an 'Int' microseconds.
+--
+-- NOTE: you cannot simply replace e.g. @threadDelay 1000@ with @sleep 1000@ since those literals
+-- have different meanings!
+sleep :: NominalDiffTime -> IO ()
+sleep = Base.threadDelay . round . Microseconds
+
+{-# DEPRECATED threadDelay "Please use `sleep` instead (and read the docs!)" #-}
+threadDelay :: Int -> IO ()
+threadDelay = Base.threadDelay
```
lexi-lambda: I understand the intent here (of guiding people who might have used `threadDelay` towards the new thing), but it seems a little curious to have a `DEPRECATED` binding in our own application code. Maybe it’s fine, though, I don’t have enormously strong feelings about it.

server/src-lib/Data/Time/Clock/Units.hs
```diff
@@ -25,56 +25,72 @@ You can also go the other way using the constructors rather than the selectors:
...
   , Hours(..)
   , Minutes(..)
-  , Seconds
-  , seconds
+  , Seconds(..)
   , Milliseconds(..)
   , Microseconds(..)
   , Nanoseconds(..)
+  -- * Reexports
+  -- | We use 'NominalDiffTime' as the standard type for duration in our code.
```
lexi-lambda: Could you add a comment about why we’ve chosen to use `NominalDiffTime` rather than `DiffTime` here, since the latter is technically more “correct”?

server/src-lib/Data/Time/Clock/Units.hs
```diff
@@ -25,56 +25,72 @@ You can also go the other way using the constructors rather than the selectors:
...
-seconds = id
+newtype Seconds = Seconds { seconds :: NominalDiffTime }
+  deriving (Show, Eq, Ord)
+  deriving (Read, Num, Fractional, Real, RealFrac) via (TimeUnit (SecondsP 1))
 
-newtype Days = Days { days :: DiffTime }
+newtype Days = Days { days :: NominalDiffTime }
   deriving (Show, Eq, Ord)
-  deriving (Num, Fractional, Real, RealFrac) via (TimeUnit (SecondsP 86400))
+  deriving (Read, Num, Fractional, Real, RealFrac) via (TimeUnit (SecondsP 86400))
```
lexi-lambda: Do we really need the `Read` instances here? If so, why?

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -28,14 +31,23 @@ runGQ
...
-  execPlan <- E.getResolvedExecPlan pgExecCtx planCache
-              userInfo sqlGenCtx enableAL sc scVer req
-  case execPlan of
-    E.GExPHasura resolvedOp ->
-      flip HttpResponse Nothing <$> runHasuraGQ reqId req userInfo resolvedOp
-    E.GExPRemote rsi opDef  ->
-      E.execRemoteGQ reqId userInfo reqHdrs req rsi opDef
+  -- The response and misc telemetry data:
+  let telemTransport = Telem.HTTP
+  (telemTimeTot, (telemCacheHit, telemLocality, ((telemTimeIO, telemQueryType), !resp))) <- withElapsedTime $ do
```
lexi-lambda: Wouldn’t it be nice if Haskell had structurally typed records? 😭

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -72,10 +84,12 @@ runHasuraGQ
   -> GQLReqUnparsed
   -> UserInfo
   -> E.ExecOp
-  -> m EncJSON
+  -> m ((NominalDiffTime, Telem.QueryType), EncJSON)
```
lexi-lambda: It seems strange to me to use a nested pair here instead of a 3-tuple. Is there a particular reason to do it this way?

server/src-lib/Control/Concurrent/Extended.hs
0x777: @jberryman Might as well get rid of `threadDelay` in the codebase?

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -192,7 +191,10 @@ data OperationLog
   { olRequestId          :: !RequestId
   , olUserVars           :: !(Maybe UserVars)
   , olResponseSize       :: !(Maybe Int64)
-  , olQueryExecutionTime :: !(Maybe Double)
+  , olQueryIoTime        :: !(Maybe NominalDiffTime)
+  -- ^ Request IO wait time, i.e. time spent reading the full request from the socket.
```
0x777: Can the field name can be slightly more explicit (I feel that `query_io_time` suggests that this is the total time spent in io), maybe something like `request_read_time`?

0x777: How hard would it be implement the entire request time (https://github.com/hasura/graphql-engine/issues/3547)?

server/src-lib/Control/Concurrent/Extended.hs
jberryman: In the last call I tried to outline the pattern. When (if) eventually this `Extended` module can shadow the standard module I think we should really strive to keep the changes and tweaks as obvious and discoverable as possible. I anticipate more `threadDelay` being written before the mixin shadowing thing happens, then all of these will become visible with a warning. I think there's no downside to keeping this in since we're fairly disciplined about keeping the haskell code warning free.

If I can do the mixin shadowing pattern successfully (now that we've moved to cabal-install) I'll write up a blog about the pattern and articulate thoughts a little more clearly and get feedback.

server/src-lib/Data/Time/Clock/Units.hs
jberryman: I'll add a comment, but I'm not totally sure what to say... it's not clear to me that one is more "correct" than the other as a measure of duration. I suppose `NominalDiffTime` indicates "may have been derived from sketchy math with `UTCTime`s" but that's not what we do

server/src-lib/Data/Time/Clock/Units.hs
jberryman: It was motivated from an application in this refactoring of a function like `readEnv :: Read a=> String -> IO a` (I can find the actually code if you need), but it seems generally useful for this kind of thing

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
jberryman: yeah it wasn't very satisfying to write this stuff...

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
jberryman: > Is there a particular reason to do it this way?

Laziness!

server/src-lib/Hasura/Server/Logging.hs
jberryman: I agree that name is clearer

> How hard would it be implement the entire request time 

It seemed more useful to me to measure and report the network IO time directly, and you can add the two times together to get total (except for network IO time for sending the response) time. 

But if you are asking about how we can measure the network IO time for sending the response: this looked really difficult/invasive to do, since at the highest level we're passing spock continuations like `... -> Handler m (HttpResponse EncJSON)`. So it looked to me like we'd have to get into the internals of spock to do this, and it would be quite inconvenient. But I haven't looked into it deeply.

My thought was for the purposes of the underlying issue in #3547  (slow network issues being discoverable) measuring in one direction would be sufficient. Let me know what you think!

0x777: > But if you are asking about how we can measure the network IO time for sending the response: this looked really difficult/invasive to do, since at the highest level we're passing spock continuations like ... -> Handler m (HttpResponse EncJSON). So it looked to me like we'd have to get into the internals of spock to do this, and it would be quite inconvenient. But I haven't looked into it deeply

Fair enough, `request_read_time` should do for now.

jberryman:
@lexi-lambda I think this should be ready now.

server/src-lib/Data/Time/Clock/Units.hs
lexi-lambda: `NominalDiffTime` is the correct type for *calendar* duration. It’s the appropriate type when you want to say “this event repeats every day,” since you don’t want it to drift earlier and earlier every time a leap second is introduced! It is not, however, an absolute measure of time.

`DiffTime`, on the other hand, *is* an absolute measure of time. This doesn’t make a significant difference if you’re measuring whole days, or even whole minutes, but if you’re measuring something that takes a second or two, the difference can be meaningful!

Of course, the datatypes don’t innately have any difference; they’re really just numbers. But they’re *supposed* to mean different things, which is why I think it’s surprising to use `NominalDiffTime` to measure wall clock time. I am not strongly opposed to using it this way, but I do think we should have a good justification for doing so.

Could you elaborate slightly on why you prefer using `NominalDiffTime`? It doesn’t seem like it would be enormously difficult to use `DiffTime`, but I assume you have a reason! I just want that reason to end up in the comment.

server/src-lib/Data/Time/Clock/Units.hs
lexi-lambda: I don’t really care for `Read` as a typeclass, but I think this instance is especially sketchy, since it breaks the informal “law” that `Show` and `Read` instances are intended to roundtrip, if both are defined. This is not the case here, since we are deriving the stock `Show` instance but the underlying `Read` instance.

I’d prefer some slightly more explicit parsing logic. What about deriving [`ParseTime`](https://hackage.haskell.org/package/time-1.9.3/docs/Data-Time-Format.html#t:ParseTime), instead?

server/src-lib/Data/Time/Clock/Units.hs
jberryman: So the way you've described the distinction between `DiffTime` and `NominalDiffTime` is one way to squint at the situation but I don't think it's really reflected in the docs (they are at best ambiguous). In particular I don't think it's very meaningful to declare that `10 :: NominalDiffTime` is not an absolute measure of time. I think they are both _primarily_ durations, and the distinction between the two is only relevant as they relate to `UTCTime` (in the same way that newtype wrappers alter the semantics of a type wrt a particular operation), but again this depends on how you squint.

The comment I have is here https://github.com/hasura/graphql-engine/pull/3622/commits/179af277c5b011faf79cc626902c35db2d94deeb#diff-2ee44e54da5eae501804c04d09d0a270R48

I can't really think of something more interesting to say about this in a comment (but am open to suggestions!). Using `DiffTime` instead just means we have to convert from `DiffTime` to `NominalDiffTime` in various places where we use `addUTCTime` and `diffUTCTime` with no real benefit that I can see.

server/src-lib/Data/Time/Clock/Units.hs
jberryman: I'm fine removing the `Read` instance or implementing a `Show` that works in the same way

server/src-lib/Data/Time/Clock/Units.hs
lexi-lambda: What do you think is ambiguous about the documentation? It’s possible that there’s a reading of the documentation that I’m not seeing, but it seems fairly explicit:

> [`NominalDiffTime`] ignores leap-seconds, so it's not necessarily a fixed amount of clock time. For instance, 23:00 UTC + 2 hours of `NominalDiffTime` = 01:00 UTC (+ 1 day), regardless of whether a leap-second intervened.

1 hour in `NominalDiffTime` is one hour of calendar time, not 3,600 SI seconds (though admittedly the two are usually the same). For *measuring* elapsed time, that isn’t usually the type you want.

server/src-lib/Data/Time/Clock/Units.hs
lexi-lambda: I’d rather leave the `Read` instance off if there isn’t any code using it right now.

jberryman: (as I said there is some code using it right now that motivated the instances, but that can be rewritten)

lexi-lambda: Ah, I misunderstood if you were saying it was currently using it or if it was previously using it. If it’s currently using it, I won’t die on this hill; I’ll leave it up to you whether you want to change it or not.

The reason I don’t like `Read` is that I never have any idea what it does for any given type. Does it try to parse a Haskell expression that would produce the value? Does it use some custom serialization format? In the case of `readFloat` and other `Read` operations on numeric types, it goes and actually tries to parse a Haskell number literal, which I think is rarely what people intend!

Admittedly, in this particular instance, it probably isn’t so bad, so maybe it’s okay. I’m just always wary of relying on `Read`, since different instances work in wildly different ways.

server/src-lib/Data/Time/Clock/Units.hs
jberryman: I think the docs are clumsily referring to the behavior of NominalDiffTime vis a vis UTCTime (which was the motivation for defining a new type as I understand it, it's a sin bin for sketchy time operations). 

In the context of an API in which the only way of constructing a `NominalDiffTime` was `diffUTCTime` (and no comparison operations were supported, and it couldn't be serialized, etc) I think you could say something abstract like "it's not necessarily a fixed amount of clock time" and really mean it.

lexi-lambda: > I think the docs are clumsily referring to the behavior of NominalDiffTime vis a vis UTCTime (which was the motivation for defining a new type as I understand it, it's a sin bin for sketchy time operations).

I agree that this is technically true. There isn’t actually any representational difference between a `DiffTime` and a `NominalDiffTime`, so really the only difference is where the value comes from! It just so happens that you get a `NominalDiffTime` from `diffUTCTime`, which makes it calendar time. I’m in full agreement with you on all of that!

Still, I think the distinction is useful, even if it doesn’t necessarily add much in the way of actual type safety. Really, it’s much the same as writing something like `newtype UserName = UserName Text`. Sure, you *can* store arbitrary `Text` values in a `UserName` value, whether or not they’re actually usernames. But you shouldn’t!

So I think it does make sense to use `DiffTime` here. We can update `withElapsedTime` and `startTimer` to return `DiffTime` values without any additional difficulty, and we’ll get a small bit of type safety in return. If we ever find ourselves needing to deal with intervals of calendar time, we can revisit this—it wouldn’t be hard to generalize these types to be parameterized over the representation type so they work with both `DiffTime` and `NominalDiffTime`.

server/src-lib/Hasura/Server/Telemetry/Counters.hs
```diff
@@ -0,0 +1,183 @@
...
+    -- ^ Total service time for request (including 'telemTimeIO')
+    }
+
+-- | Sum
+instance Semigroup RequestTimings where
+  RequestTimings a b <> RequestTimings x y = RequestTimings (a+x) (b+y)
+
+-- | 'RequestTimings' along with the count
+data RequestTimingsCount  =
+  RequestTimingsCount {
```
lorenzo: The current model does not seem to enable the ability to calculate percentiles, which are extremely useful for performance tracking. Would you guys be open to explore an alternative design that would enable calculations beyond totals and averages?

jberryman: For now for the purposes of telemetry we're just collecting counts and timings for a few hand chosen buckets, which should be fine for our purposes, rather than using hdrhistogram or something similar. But I agree with you that percentiles are useful!

jberryman:
@0x777 or @lexi-lambda 

This should address Alexis's changes (use DiffTime instead of NominalDiffTime). History got somewhat worse as I didn't want to spend the time to split up all the changes and squash them individually. I also added a few comments, added a `fromUnits` for conversions between units (including difftime and nominaldifftime), as I discovered `realToFrac` was a footgun.

PR #3639, opened 2020-01-06
translations(mandarin): add readme.md (close #3170)

### LINKED ISSUE #3170: Translation of README.md into Mandarin
I was hoping that translating the README file to the most spoken language in the world, Mandarin would help with the outreach of the repository. In addition, I would like to work on this as a part of Hacktoberfest.

shahidhk:
Mandarin translations by @johnbanq and @jagreetdg


translations/README.chinese.md
```diff
@@ -49,7 +49,7 @@ Hasura可帮助您构建基于Postgres的GraphQL应用程序，或将使用Postg
 - [客户端工具](#客户端工具)
 - [添加业务逻辑](#添加业务逻辑)
     - [远端Schema](#远端Schema)
-    - [通过数据库事件调用Webhook](#通过数据库事件触发Webhooks)
+    - [在数据库事件上触发Webhooks](#在数据库事件上触发Webhooks)
```
johnbanq: Might be a bad move, "在...上" in Chinese don't have the meaning of “sth. happens because of sth.”. It is usually used to mean something is better than another or physically above than another, which is different to the english preposition "on".

shahidhk: @johnbanq, so `通过数据库事件调用Webhook` is the right usage?

johnbanq: NOT RIGHT since it didn't get the EXACT meaning acrossed, but it still serves as a title to tell people you can add business logic to the process by writing a webhook that is triggered by database event.

shahidhk: :smile: What do you suggest? I will stick to that.

johnbanq: I think we can just stay with "[通过数据库事件触发Webhooks](#通过数据库事件触发Webhooks)" for now and leave this to a better translater in the future. btw, thank you for fixing the encoding issues :), so they are using UTF-8?

shahidhk: Yes, UTF-8.

I will make the suggested change.

shahidhk: Done! Thanks @johnbanq!

PR #3668, opened 2020-01-09
[RFC] Eliminate unnecessary recompilation caused by Hasura.Server.Version

lexi-lambda:
This PR is an experiment. I am opening it as a proof-of-concept, but I am not actually certain that it’s the right solution.

## Context

We compile the current version of `graphql-engine` directly into the executable using Template Haskell. This is very convenient, but because the calculation that computes the version depends on complicated things like the current state of the git repository, GHC doesn’t know when to recompile it. To fix this, #3414 added `{-# OPTIONS_GHC -fforce-recomp #-}` to the top of the `Hasura.Server.Version` module, but the unintended side-effect was that *way too much* ended up being regularly recompiled: currently, a whole lot of modules are recompiled on *any* change to *any* module!

## The solution

This PR reworks the way the version is accessed to avoid (transitive) dependencies on the module containing the Template Haskell as much as possible, thereby avoiding most of the recompilation. Instead of accessing the version directly, the new code threads it around via an implicit parameter.

This works alright. It does recompile a lot less, but it does still have to recompile `Hasura.Server.Version.Computed` on every change, and even though the resulting object file is exactly the same, `stack` still goes through the motions of *relinking* every target, which takes a nontrivial amount of time. In practice, this probably doesn’t matter very much if you are working on the library code, since that would cause `stack` to relink everything anyway. If you enable parallel builds using `--ghc-options=-j`, this should be pretty cheap.

## Possible alternatives

There are two alternatives to this change I’ve thought of: one simple and one complicated.

  1. The simple solution is easy: just give up on `-fforce-recomp`. This isn’t really so bad from my point of view, but I think @ecthiender mentioned it causes some trouble for Pro.

     I think that’s actually okay: since Pro is a separate library with a separate dependency, relying on the fact that `stack` will recompile the dependency when `-fforce-recomp` is specified if it’s not being installed from Hackage is a pretty major hack. I think the correct approach is to parameterize all the library code over the version and pass it in separately in Pro, and this change enables that even without `-fforce-recomp`.

  2. A more complicated alternative would be to compile the version into the executable using a GHC plugin rather than using Template Haskell. Plugins have the ability to provide custom hooks to the recompilation checker, so we could do something smarter than what we can do with Template Haskell.

     This solution would probably be pretty effective, but GHC plugins are weird and unwieldy, and I’d rather not have to deal with them. So my preference is to avoid doing this if possible.


lexi-lambda:
Alright, now that #3558 has been merged, I’ve tried this change out using `cabal-install` rather than `stack`. There’s good news and bad news:

  - The good news is that `cabal-install` is more aggressive about caching the results of previous builds, so it doesn’t even *invoke* `ghc` if nothing in a particular component changed. This means that changing some code in the tests doesn’t cause anything in the library to get rebuilt.

  - The bad news is that `cabal-install` is more aggressive about caching the results of previous builds, so `-fforce-recomp` isn’t enough to get it to always rebuild the `Hasura.Server.Version.Computed` module, which was the whole point of specifying `-fforce-recomp` in the first place!

Personally, I think this is just another indicator that `-fforce-recomp` isn’t worth it. My understanding is that @ecthiender was mostly worried about it in the context of Pro, since they needed it to ensure the version was correct in their resulting executable, but I’ve always felt like relying on `-fforce-recomp` for that is a hack that was in no way guaranteed to keep working. As it happens, though, the `HasVersion` constraint implemented by this PR is precisely the solution to *that* problem, too, since it allows Pro to specify the version independently, in its own executable entrypoint.

Therefore, I plan to repurpose this PR: I’m going to remove the `-fforce-recomp` flag completely, but I’m going to keep the `HasVersion` plumbing, just to handle that use case more nicely.

server/src-lib/Hasura/Server/Version.hs
```diff
@@ -1,61 +1,97 @@
...
+currentVersion = ?version
+
+withVersion :: Version -> (HasVersion => r) -> r
+withVersion version x = let ?version = version in x
+
+consoleVersion :: HasVersion => Text
+consoleVersion = case currentVersion of
+  VersionDev txt -> txt
+  VersionRelease v -> case getReleaseChannel v of
+    Nothing -> vMajMin
```
ecthiender: I am not entirely sure, but looks like in the previous implementation, when `getReleaseChannel v` returns `Nothing` it would return `rawVersion` which was defined as `"versioned/" <> version`.
Also, when the version was a dev version, `consoleVersion` returned `rawVersion`.

So shouldn't we prefix `txt` and `vMajMin` here with `"versioned/"` as well? Or am I missing something? To be honest, I don't really know this scheme. Looks like the console version here, reflects where the assets are stored. I think @shahidhk has better info on this and he can clarify if required.

ecthiender: Yeah, can confirm this. Opening the PR app, does not load the console assets. The `"versioned"` string from the full path is missing.

ecthiender:
Yeah I agree. `--fforce-recomp` was pretty much a hack, which probably never worked the way we anticipated. I think this way, of explicitly reading the environment variable in `Main` and passing it down to the actual library, makes way more sense.

Thanks for the PR, it looks great! There is only a minor issue I think in the way the `consoleVersion` is formed, I have left a comment.

PR #3669, opened 2020-01-09
Add bigint to frequently used column (close #3524)

### LINKED ISSUE #3524: Change the frequently used column `id` to bigint
The current `id` suggestion for frequently used columns is `int`, but I propose this to be changed to `bigint`. On all my projects I have to change this manually.

Other frameworks like Rails use bigint by default, and I think that's the best choice for most projects.

ashishra0:
### Description
Add a new option `bigint` for `id` to choose in frequently used column in the console.

### Affected components

- [x] Console

### Related Issues
#3524 

### Solution and Design
Instead of completely removing `integer` as one of the options for `id`, The solution was fixed by adding bigint as another option to choose from the list of frequently used columns.


ashishra0:
@rikinsk @shahidhk The build errors out because of `semver.NewConstraint("<= v1.0.0-rc.1")`

shahidhk:
@ashishra0 Your branch is not up-to-date with master. Hence the CLI change. Merge master into your branch.

Issue #3654, opened 2020-01-07
Auto migrations takes a very long time (20 mins)


flybayer:
I'm deploying this Dockerfile to Render.com which hosts both the app and the DB.

```Dockerfile
FROM hasura/graphql-engine:v1.0.0.cli-migrations

COPY ./migrations /hasura-migrations
```

### Problem

Migrations are taking a **very** long time. For example, the below commit took 20 minutes for the migrations to run.

```
| 3     0       hasura/migrations/1578006637841_create_table_public_x_cache_shopify_products/down.yaml
| 10    0       hasura/migrations/1578006637841_create_table_public_x_cache_shopify_products/up.yaml
| 3     0       hasura/migrations/1578006855498_create_table_public_x_cache_shopify_productVariants/down.yaml
| 14    0       hasura/migrations/1578006855498_create_table_public_x_cache_shopify_productVariants/up.yaml
| 11    0       hasura/migrations/1578006867682_alter_table_public_x_cache_shopify_products_alter_column_options/down.yaml
| 11    0       hasura/migrations/1578006867682_alter_table_public_x_cache_shopify_products_alter_column_options/up.yaml
| 11    0       hasura/migrations/1578008617640_alter_table_public_x_cache_shopify_products_alter_column_options/down.yaml
| 11    0       hasura/migrations/1578008617640_alter_table_public_x_cache_shopify_products_alter_column_options/up.yaml
| 12    0       hasura/migrations/1578008636900_alter_table_public_x_cache_shopify_productVariants_alter_column_images/down.yaml
| 12    0       hasura/migrations/1578008636900_alter_table_public_x_cache_shopify_productVariants_alter_column_images/up.yaml
| 12    0       hasura/migrations/1578008646868_alter_table_public_x_cache_shopify_productVariants_alter_column_selectedOptions/down.yaml
| 12    0       hasura/migrations/1578008646868_alter_table_public_x_cache_shopify_productVariants_alter_column_selectedOptions/up.yaml
| 6     0       hasura/migrations/1578008668655_add_relationship__table_public_x_cache_shopify_products/down.yaml
| 12    0       hasura/migrations/1578008668655_add_relationship__table_public_x_cache_shopify_products/up.yaml
| 6     0       hasura/migrations/1578008671585_add_relationship__table_public_x_cache_shopify_products/down.yaml
| 8     0       hasura/migrations/1578008671585_add_relationship__table_public_x_cache_shopify_products/up.yaml
[...]

PR #3686, opened 2020-01-13
Add fast path for server internal metadata migrations

lexi-lambda:
### Description

The way we perform internal metadata migrations within `graphql-engine` has a long and storied history.

Originally, we just reused the same infrastructure we use for CLI migrations, but this caused #2826. In #2379, we resolved the issue by dropping and recreating all system-defined metadata whenever we migrated the SQL schema, but that in turn caused #3354. The recent incremental schema cache construction changes in #3394 mostly fixed that problem, but as discovered in #3654, the process can still be very slow on underpowered Postgres instances.

This PR adds a fast path for the internal metadata migrations to improve the performance of internal server migrations. It does not affect the performance of user-defined CLI migrations, so it’s not a comprehensive fix, but it’s much simpler to make our own migrations faster (mostly because we don’t have to worry so much about the interface), so this takes advantage of that low-hanging fruit.

### Affected components
- [x] Server

### Related Issues
#3654 

### Solution and Design

This PR eliminates the use of the CLI migrations code path for performing internal metadata migrations, instead doing it all directly in code in a single batch of queries against the database. This is a little unsatisfying, since it means our internal migrations are “blessed” compared to user-defined CLI migrations, but it’s so easy to do that I decided to give in and make the change.

#### Catalog upgrade
- [x] No
#### Metadata
- [x] No
#### GraphQL
#### Breaking changes
- [x] No Breaking changes


0x777:
Currently we have `replace_metadata` which is backed by a `ReplaceMetadata` type (should've been called something like `GraphqlEngineMetadata`) which sets graphql-engine's metadata to the given metadata. 

I'm wondering if it makes sense to have a query like `add_metadata` which instead of replacing, 'merges' the given metadata into the existing metadata. Is it possible to have `add_metadata` to perform better than a `bulk` query of adding tables, relationships etc? If so, we can just specify our internal catalog as a single query instead of using a `bulk` query like we do currently.

(The implementation in this PR looks like an interface to `add_metadata` which can only be used internally)

lexi-lambda:
> Currently we have `replace_metadata` which is backed by a `ReplaceMetadata` type (should've been called something like `GraphqlEngineMetadata`) which sets graphql-engine's metadata to the given metadata.
> 
> I'm wondering if it makes sense to have a query like `add_metadata` which instead of replacing, 'merges' the given metadata into the existing metadata. Is it possible to have `add_metadata` to perform better than a `bulk` query of adding tables, relationships etc? If so, we can just specify our internal catalog as a single query instead of using a `bulk` query like we do currently.

I considered this, too, and I think it’s a totally viable alternative. I wasn’t sure whether it was better to split out some `ReplaceMetadata` functionality to do what we want or just do it in code.

Eventually, I decided to just do it in the code because that seemed very easy, and it’s very simple to change later. I figured that what we *really* want is a more general solution to this that helps CLI migrations, too, and maybe at that point we could go back to sharing the same code path, but I couldn’t figure out exactly what I thought that should look like, so I opted to do the simplest thing that seemed least likely to introduce tech debt.

That said, I’m really not attached to this, and in fact I’d actually love for someone else to make this decision for me. 😄 If you think I should tweak this to keep the `.yaml` file but go through a `ReplaceMetadata`-like code path, I’m happy to do that.

lexi-lambda:
By the way, I should add why I didn’t want to implement the `AddMetadata` API you’re suggesting: I could not figure out how to have good error reporting for it without doing a lot of additional engineering to associate all that information properly. So I wouldn’t want to expose it as a user-facing API without thinking a lot harder about what the semantics ought to be, and at that point it seemed less clearly valuable.

We could still have something that works like that internally that we use for this code path but not actually expose it as a valid RQL query, and we could adapt it into a full API at some point in the future. But I don’t think it makes sense to add that API right now without more thought and discussion.

0x777:
> I could not figure out how to have good error reporting for it without doing a lot of additional engineering to associate all that information properly

PR #3699, opened 2020-01-14
Add readable terms for the operators

beerose:
### Description
In Browse rows, there were listed postgres operators. This PR adds readable english terms and the operator in brackets so that anybody can understand it quickly.

Screenshot 2020-01-14 at 11 38 03

### Affected components

- [x] Console

#### Breaking changes

- [x] No Breaking changes


coco98:
@beerose @rikinsk Should we change `ilike` to `like (ignore case)` or `like (case insensitive)`?

beerose:
Yeah, it makes sense since we want to make it much more readable. I think I'd rather go with `(case insensitive)` since it's more common. 

But take into account that it would quite widen the dropdown:

Screenshot 2020-01-14 at 12 19 43

beerose:
I've changed dropdown options to be of the format `[<graphql equivalent operator>] logical name` and added `(case insensitive)` to `ilike` and `nilike`.

I also added the default value `%%` when `like`, `nlike`, `ilike` or `nilike` is chosen. Small demo:
likes

console/src/components/Services/Data/constants.js
```diff
@@ -1,18 +1,22 @@
...
+  { name: 'not in', value: '$nin', graphqlOp: '_nin' },
+  { name: '>', value: '$gt', graphqlOp: '_gt' },
+  { name: '<', value: '$lt', graphqlOp: '_lt' },
+  { name: '>=', value: '$gte', graphqlOp: '_gte' },
+  { name: '<=', value: '$lte', graphqlOp: '_lte' },
+  { name: 'like', value: '$like', graphqlOp: '_like' },
+  { name: 'not like', value: '$nlike', graphqlOp: '_nlike' },
+  { name: 'like (case-insensitive)', value: '$ilike', graphqlOp: '_ilike' },
+  {
+    name: 'not ilike (case-insensitive)',
```
rikinsk: `not ilike` -> `not like`

console/src/components/Services/Data/constants.js
```diff
@@ -1,18 +1,22 @@
...
+    value: '$nilike',
+    graphqlOp: '_nilike',
+  },
+  { name: 'similar', value: '$similar', graphqlOp: '_similar' },
+  { name: 'not similar', value: '$nsimilar', graphqlOp: '_similar' },
```
rikinsk: `_similar` -> `_nsimilar`

beerose:
I handled the `$in` and `$nin` case. Also, in the future, we could think about adding validation -- validate if, for expected integer, there's a number, for expected array there's a valid array, etc.

Issue #3603, opened 2019-12-30
Using permissions with Postgres enum field doesn't show comparing options


garlou:
I'm using typeorm to manage my database migrations and when I use typescript enums, typeorm generates a Postgres table like this:
<img width="601" alt="Screenshot 2019-12-30 11 42 23" src="https://user-images.githubusercontent.com/1465244/71578862-2a69f880-2afa-11ea-97a4-cf197008d1d2.png">

When I try to setup a permission, the dropdown select is empty:
<img width="386" alt="Screenshot 2019-12-30 11 43 01" src="https://user-images.githubusercontent.com/1465244/71579018-dc092980-2afa-11ea-9472-fe88e20fed36.png">

Workaround:
Save permission as it is, export hasura metadata and then manually edit the metadata.yaml:
<img width="232" alt="Screenshot 2019-12-30 11 44 32" src="https://user-images.githubusercontent.com/1465244/71579122-723d4f80-2afb-11ea-80fe-24b180378db8.png">
to
<img width="239" alt="Screenshot 2019-12-30 11 44 50" src="https://user-images.githubusercontent.com/1465244/71579130-7a958a80-2afb-11ea-966f-3c9018487714.png">

Then apply the fixed metadata file on server.

nick4u:
Any chance for having this in 1.1?

rikinsk:
@nick4u Unfortunately the feature set for v1.1 is already frozen. We'll aim to get this fixed in v1.2

PR #3712, opened 2020-01-15
handle unmatched types in permission builder

rikinsk:
### Description

In permission builder, for columns with types not present in the mainitained list, a default set of operators will always get shown

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3603

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


console/src/components/Services/Data/TablePermissions/PermissionBuilder/utils.js
```diff
@@ -44,40 +45,103 @@ const boolOperatorsInfo = {
...
   _eq: {
     type: 'object',
-    PGTypes: ['boolean', 'character', 'dateTime', 'numeric', 'uuid'],
+    PGTypes: [
+      'boolean',
+      'character',
+      'dateTime',
+      'numeric',
+      'uuid',
+      'user_defined',
```
beerose: Isn't this list the same for every operator? Would it make sense to extract it to the variable, so that another changes would be done just in the one place?

rikinsk: It does seem like there can be a better way to represent this info. Will take a stab at it and request a review

rikinsk: @beerose Made changes to avoid repetition in the future as well. This was a good catch. Thanks.

Issue #3710, opened 2020-01-15
Missing hdb_view in 1.1.0-beta.1?


nirvdrum:
I tried out 1.1.0-beta.1 today and I've been unable to run my app with it. Using a Docker Compose environment that upgraded from 1.0.0 to 1.1.0-beta.1, I'm getting an error when trying to insert data:

```
ERROR:  relation "hdb_views.eca2225ad9f81f9075915f94e6c65a4f91c855f7e0bdb795d09fee54" does not exist at character 47
```

Thinking that my DB somehow got corrupted, I tore down my Docker Compose environment and removed the persistent volume. With a clean install, I'm unable to run my migrations:

```
[42883] FatalError: function hdb_views.dbb780ae7437f2a5faf3926515a7a206f5bc9b227ba22a748b9f4587() does not exist
```

I wish I had more details about what's going on there. Unfortunately, that's a Hasura managed table that I know very little about. If I go back to 1.0.0, my migrations apply fine and I'm able to insert data successfully.

lexi-lambda:
Urk—this is my fault. This issue was already reported before we made the beta release in https://github.com/hasura/graphql-engine/issues/3354#issuecomment-569736691, but at the time I didn’t worry much about it because I knew we were getting rid of those views anyway in #3598. However, I’ve now realized that change didn’t make it into this release, so the issue is a real issue, after all!

This will definitely be fixed by #3598. I think it’s unlikely to be worth trying to patch independently of that change, since it should be merged very soon anyway. In the meantime, I’d recommend just staying on 1.0.0 if you’ve bumped into it.

(An aside: we should really not be notifying everyone to upgrade to `1.0.0-beta.1`, since it’s a *beta*, and we have stable releases now. That’s a holdover from pre-1.0.0 that we forgot about, and we’ll fix that soon.)

MarcelloTheArcane:
@lexi-lambda I'm still getting this error when I downgrade back to 1.0.0. Is there anything I can do to sort this out?

lukaspili:
Are there specific instructions to downgrade to 1.0.0? 
Setting the version of `hdb_catalog.hdb_version` to 28 does not seem to be enough. Hasura is crashing on startup with:

```
failed to build schema-cache because of inconsistent metadata
```

nirvdrum:
In my case, I was only trying it out on a local installation, so I just blew everything away and started from scratch. I lost some test data, but nothing critical.

nyamba:
[...]

PR #3718, opened 2020-01-16
run default tests in test_server_upgrade

tirumaraiselvan:
### Description

Currently in the server upgrade test path, we only run a basic query to see if the upgrade test was successful. 

This PR is a change to run all default tests.

### Affected components

- [x] Build System
- [x] Tests


tirumaraiselvan:
@shahidhk @lexi-lambda I have opened this PR so that we can have a more powerful after upgrade path testing suite ( in light of #3710 )

server/tests-py/test_upgrade/test_basics.py
```diff
@@ -0,0 +1,76 @@
+import ruamel.yaml as yaml
+import pytest
+from validate import check_query_f
+
+
+@pytest.mark.parametrize("transport", ['http', 'websocket'])
+class TestGraphQLQueryBasic():
```
lexi-lambda: Could we avoid duplicating the definitions of these tests? It should be possible to specify particular tests to run given a single invocation of `pytest`, which is what we do in `test_server.sh` already.

.circleci/config.yml
```diff
@@ -456,7 +456,7 @@ jobs:
   # test server upgrade from last version to current build
   test_server_upgrade:
     docker:
-    - image: hasura/graphql-engine-upgrade-tester:v0.4
+    - image: tirumarai/graphql-engine-upgrade-tester
```
lexi-lambda: This image should be pushed to the `hasura` org before merging.

server/tests-py/test_upgrade/test_basics.py
tirumaraiselvan: Another, although smaller, reason was that in our test suite, we have different `setup` and `teardown` for `TestGraphQLQueryBasic` and `TestGraphQLInsertBasic` and so on. Again, since we are avoiding setup and teardown all-together we would need to have a common migration which works for all these test classes.

lexi-lambda: I see, that makes sense. (This is another good reason to move these tests to Haskell—it would likely be much easier to add the necessary abstraction in Haskell than in Python!)

One thing you could probably do is add a common superclass, like this:

```python
class GraphQLQueryBasicMixin():
  def test_select_query_author(self, hge_ctx, transport):
    ...
  ...

class TestGraphQLQueryBasic(DefaultTestSelectQueries, GraphQLQueryBasicMixin): pass
```

Then you could reuse the common superclass in the migration test. But to be honest, I don’t really like this, as it still requires the migrations to be manually kept in sync with the setup/teardown actions.

A better solution would be to abstract over things further so that you could reuse the definition of the setup action as the migration itself. You could do this, too—the tests are just code, like everything else!—but it would require a little more plumbing. I don’t think it would take too much effort to get sharing between the two paths, though.

.circleci/server-upgrade/run.sh
```diff
@@ -9,114 +14,233 @@ set -euo pipefail
...
-hasura --project $HASURA_PROJECT_DIR migrate apply --endpoint $HGE_ENDPOINT
+log "Run pytests with server upgrade"
 
-# make a test query
-log "executing the test query"
-node $ROOT/make_test_query.js $HGE_ENDPOINT/v1alpha1/graphql
-log
+WORKTREE_DIR="$(mktemp -d)"
+rm_worktree(){
+	rm -rf $WORKTREE_DIR
```
jberryman: rm  -rf "$WORKTREE_DIR"  # with quotes

server/tests-py/conftest.py
```diff
@@ -193,21 +258,124 @@ def evts_webhook(request):
...
+    yield from db_state_context(request, hge_ctx)
+
+@pytest.fixture(scope='function')
+def per_method_tests_db_state(request, hge_ctx):
     """
-    setup_ctrl = { "setupDone" : False }
-    yield setup_ctrl
-    hge_ctx.may_skip_test_teardown = False
-    request.cls().do_teardown(setup_ctrl, hge_ctx)
+    This fixture sets up the database state for metadata operations
```
jberryman: I appreciate the docs here!

.circleci/server-upgrade/run.sh
nizar-m: Good catch. I have  pushed the commit with quotes.

PR #3735, opened 2020-01-17
add multi select to data table and bulk delete (close #1739)

### LINKED ISSUE #1739: multi-select in console browse rows
Add the ability to multi-select the rows in console browse rows table (similar to adminer) and allow bulk delete for the selected rows.

beerose:
Solves https://github.com/hasura/graphql-engine/issues/1739

### Description
Added a checkbox to each row on Browse Rows view that allows selecting one or more rows from the table and bulk delete them.

Screenshot 2020-03-10 at 11 30 35

### Affected components

- [x] Console

### To be done

- [x] Fix tests
- [x] Bring back delete button
- [x] Section for selected items

### Solution and Design
I added a column with select action to the data table. Users can select particular rows or select all at once. If any rows are selected delete button is visible and allows to delete all selected data.

multi-select

### To be discussed

* Do we want to get rid of the delete button in the actions column? Or do we want to leave it for now?

* Bulk-delete button design. We can go with icon button as on the demo above:

Screenshot 2020-01-17 at 13 53 12

Or have action button with `Delete rows` text:

Screenshot 2020-01-17 at 13 58 24


rikinsk:
@beerose  Couple of points:

- Can we leave the individual row delete btn as it was. I wouldnt want to make 2 clicks to delete a single row.
- The Delete btn outside the table looks a little awkward. Maybe we can have a section for selected items which we can use for other things in the future. Something like this (with better alignment etc. of course):
Screenshot from 2020-02-05 17-41-22

console/src/components/Services/Data/TableBrowseRows/ViewRows.js
```diff
@@ -184,6 +195,25 @@ const ViewRows = ({
       width: 152,
     });
 
+    _gridHeadings.push({
+      Header: (
+        <div className={styles.tableCenterContent}>
+          <input
+            className={styles.inputCheckbox}
+            style={{ marginBottom: '2px' }}
```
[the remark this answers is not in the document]
wawhal: Change it for wherever you have used the `style` prop.

console/src/components/Services/Data/TableBrowseRows/ViewRows.js
```diff
@@ -871,6 +953,22 @@ const ViewRows = ({
...
       <div className={`row ${styles.add_mar_top}`}>
+        {selectedRows.length > 0 && (
+          <div
+            className={`${styles.display_flex}`}
+            style={{ paddingLeft: '18px' }}
+          >
+            <b className={styles.padd_small_right}>Selected:</b>
+            {selectedRows.length}
+            <button
+              className={`add_mar_right_small___3slZA btn btn-xs btn-default ${styles.bulkDeleteButton}`}
```
wawhal: You have probably copied this classname from inspect element. It won't work because these names are autogenerated on the fly while building.

You can replace `add_mar_right_small__3slZA` with `${styles.add_mar_right_small}`. It is the same classname.

beerose: Yeah, I have no idea why it's there 🙄

Issue #384, opened 2018-09-09
Add custom permission check to update operation on the changed data


techninja1008:
Currently, on the permissions check on an update operation, the check is only run on the row *before* it is modified. It would also make sense to add a check to make sure that the modification is valid.

*Example use case*:

1. You have a table of users and a table of (for example) cars.
2. Each car has a reference (with a foreign key constraint) to a car.
3. You want users to be able to update their car entries using criteria including the user_id, so you allow them access to it on the update permission.
4. Currently: You have a problem, as they can now change the user_id column to someone else, when the application was designed for other users to be completely invisible.
    
    With suggestion: You add a check so that the user_id field *after modification* must still equal X-HASURA-USER-ID, and everything works fine.

coco98:
@techninja1008 If I understand correctly, I have a feeling this can be solved with the current permissions. The update check permission will only allow update operations only on columns where the user_id is x-hasura-user-id and then you'll also have to remove the user_id column from being updatable?
<img width="1038" alt="screen shot 2018-09-09 at 3 10 58 pm" src="https://user-images.githubusercontent.com/131160/45263209-b5753a00-b442-11e8-9308-cc8945e68e80.png">

I've setup a heroku app without any access-key that you can do something on and share with me to illustrate better if I haven't understood well:
https://permissions-test.herokuapp.com/console

On another note, @0x777 I think there's an overlap with this https://github.com/hasura/graphql-engine/issues/216.

techninja1008:
@coco98 Yes, this particular example (Not that great) can be solved with that, but lets say you also have a mileage counter on the cars table and you want to make sure that it can only be positive. A slightly more advanced example would be that you can only increase it.

coco98:
@techninja1008 Hm...gotcha. We can think about this!

A current workaround would be to use postgres check constraints https://www.postgresql.org/docs/current/static/ddl-constraints.html.

However, the problem is that those constraints are applied for all hasura permission roles.

spiffytech:
The suggestion of making `user_id` non-updatable works, but makes my code more complicated since I have to strip the id out of my `variables` objects in all of my GraphQL mutations. I'd much rather be able to pass the original object through to the GraphQL Engine and have it just reject if the modified row violates the permissions.

pnappa:
Another solution is to provide two degrees of permission checks for updates. 

One being a check on the current state of the row being updated (this is what the permission check currently does).

[...]

Issue #3536, opened 2019-12-14
Add Permission Rules For Checking the New (or Incoming, or Next) Value of a Field


yasinarik:
### Problem:
Currently, it is possible to add row permission rules for checking the existing value of a column field.

However, per my research, **it's not possible** to add control logics into row permission rules for checking the **incoming (or new, or next)** value of a field. In other words, this will ultimately result in a custom check constraint that is applied _BEFORE_ a mutation.

_This could be related to "actions" feature that is still under progress._ If so, when will the **actions** become ready to use?

### **To illustrate:**

_This is a **one way** relationship table. I took my time and tried to represent the exact same situation that we are going to use in production._  

Let the column fields be: (Simplified)

```
id_Doer   |   id_Target   |   followState
```

`followState` is a String field and may have 3 values => `"false"`, `"requested"`, and `"true"`.

Let `UserTarget` has a private profile. So any other user can only `request` following UserTarget or if they already requested, they can `cancel request`.

So here is an example situation: (Simplified)
 

- The below code blocks are connected with an `_or` operator.

- Let `UserA = UserDoer` and `UserB = UserTarget` . 

- This row represents the relationship of `UserA over UserB`.

**Code Block 1:** UserA is the logged in user who is trying to perform mutations on the same row.

```
_and: [
 id_Doer = X-Hasura-Id, (UserDoer is the logged in user. I can do this, no problem.),
 UserTarget = private,  (I can also check this since there are array & object relations. No problem)
    _or: [
      _and: [
       existing_value_of_followState = "false",  (UserDoer not yet requested)
       incoming_value_of_followState = "requested",    (UserDoer can request following)
     ],

      _and: [
       existing_value_of_followState = "requested",  (If UserDoer requested following)
       incoming_value_of_followState = "false",    (UserTarget can cancel the request)
     ], 

      _and: [
       existing_value_of_followState = "true",  (If UserDoer is following UserTarget)
       incoming_value_of_followState = "false",    (UserDoer can unfollow UserTarget)
     ],  
 ],

```

**Code Block 2:** UserB is the logged in user who is trying to perform mutations on the same row.
```
_and: [
 id_Target = X-Hasura-Id, (UserTarget is the logged in user. I can do this, no problem.),
 UserTarget = private,  (I can also check this since there are array & object relations. No problem)

    _or: [ 

      _and: [
      existing_value_of_followState = "requested",  (Request already made by UserDoer before)
[...]

PR #3750, opened 2020-01-21
rfc: check condition in update permissions

0x777:
Can be previewed [here](https://github.com/0x777/graphql-engine/blob/rfc-update-permission-check-condition/rfcs/update-permission-check-condition.md). Issue: https://github.com/hasura/graphql-engine/issues/384


rfcs/update-permission-check-condition.md
```diff
@@ -0,0 +1,152 @@
...
+
+> A row is only updated if the row is allowed to be updated with `filter` and the
+updated row holds the condition specified with `check`.
+
+#### Other options considered:
+
+- Why introduce a `check` field in the update permission? Why not just apply the
+  insert permission's `check` condition on updates?
+
+  1. It may not make sense to allow inserts, but a check condition on update needs
```
paf31: Presumably this could be handled by just turning off inserts for that role, if we moved the check constraint out to the table level.

rfcs/update-permission-check-condition.md
```diff
@@ -0,0 +1,152 @@
...
+  to be specified.
+  2. The check conditions maybe different for both insert and update permisisons.
```
paf31: I'm curious if we have a concrete use case in mind for this.

0x777: Such a use case is documented in https://github.com/hasura/graphql-engine/issues/3536. I've specified the update permissions on that thread. The insert permission's `check` would be something like this:
```json
{
  "id_Doer" : "x-hasura-id",
  "followState": "requested"
}
```
Note that the update permission's `filter`, `check` and insert permission's `check` are all different.

0x777:
@paf31 The `check` condition in our insert permissions (and in Postgres's RLS) is not an invariant like Postgres's [table level constraints](https://www.postgresql.org/docs/current/ddl-constraints.html), it will only be checked during the operation. 

> and be the same across updates and inserts

You can have different conditions to check on `insert` and `update` (a use case is in #3536). This is also allowed in Postgres's RLS and as long as the implementation is straight forward, we should try to remain as close to RLS as possible.

> The idea of check constraints as invariants doesn't quite work when we allow exists in conditions, because we can break the invariant by updating the joined table. Not sure if we want to consider that here

The check constraint will only be checked during the operation.

> It's also possible for one user to make it impossible for another user in a different role to update a row, in a new way which was not possible before

I'm sorry I don't quite follow this. The admin role could also modify some column involved in the `filter` condition because of which `filter` condition would never satisfy. 

@lexi-lambda 

> It’s possible that sometimes people want these things to be different, but it seems likely that, in the vast majority of cases, they want them all to be the same

The console already helps with this situation, it lets you copy a condition from other permissions. So you would only write these rules once if they are all the same (but yes you would still have to say 'same as ..' multiple times which I don't think can be avoided). 

> Do you think we can come up with a design that matches that logical model a little bit better? I like Phil’s idea about making these a property of the table rather than something that belongs to a particular operation.

I agree, our permissions API on the server is a bit low level (the console improves the DX by providing options like copying conditions etc), we can definitely add higher level abstractions at the server itself, maybe like Phil suggested, we can allow users to define conditions on a table and allow referring to these conditions by names when specifying permissions. However I feel that we can open a new RFC for this  as
1. The scope of these improvements would affect all permissions, not just insert and update.
2. Console can help our users in cases where they have to specify the same constraint across insert and update permissions.

rikinsk:
I just wanted to add my 2 cents around how I think about check contraints and permissions. This might not be completely related but seems relevant. 

- Any property of an entity which should always hold true should be part of check constraints / put behind biz logic during mutations
- Any property of an entity that needs properties of the user making the request to be decided, go into permission rules.

We often think of using permissions as a way to enforce general data consistency but that should be an anti-pattern as I can always then cause data inconsistencies by making a request as an admin.
Permission rules should always be seen from a lens of user level protection and not a way to define properties of your entities. The same follows for role-level columns presets vs PG defaults as well.

paf31:
Sorry if this is a dumb question, but is there a reason why we're not reusing Postgres' row level security directly then? I know this would require mapping our roles to Postgres roles, or something like that, but it seems simpler than implementing a mirror of its permission system. Is it missing something we want?

0x777:
>  is there a reason why we're not reusing Postgres' row level security

Couple of reasons:
1. We want to make as few changes as possible to the underlying Postgres database - creating new roles, defining permissions and enabling RLS on the tables are fairly intrusive changes and requires privileges that most of users may not be comfortable giving them to graphql-engine. In fact, some of the managed Postgres vendors (Heroku) do not allow creating new roles. 
2. We cannot do the kind of optimizations that we do with our [subscriptions implementation](https://github.com/hasura/graphql-engine/blob/master/architecture/live-queries.md) if we use RLS.

paf31:
Any thoughts on what the UI should look like for permissions? Right now it refers to "checks" even though those are currently really _filters_, but now we will have checks too :)

paf31:
If an insert becomes an update due via `ON CONFLICT`, do you think we should use the insert check or the update check?

rikinsk:
> If an insert becomes an update via ON CONFLICT, do you think we should use the insert check or the update check?

I believe the update check should kick in as technically the insert is now reduced to an update

> Any thoughts on what the UI should look like for permissions? Right now it refers to "checks" even though those are currently really filters, but now we will have checks too :)

Seems like the row permissions section will need to be further split into 2 sections now, a "before check" and an "after check". I currently can't seem figure out what the messaging should be for the 2 sections to get away with minimal changes. Will give it some more thought and get back. Ideas are welcome of course :)

0x777:
> If an insert becomes an update via ON CONFLICT, do you think we should use the insert check or the update check?

@paf31 This is discussed in the '[Implementation](https://github.com/0x777/graphql-engine/blob/rfc-update-permission-check-condition/rfcs/update-permission-check-condition.md#implementation)' section. 

> ...The tricky part would be the behaviour when on_conflict is used:
> 
> 1. When there is no conflict, the insert permission's check condition has to hold true on the inserted row.
> 2. When there is a conflict, the update should only happen if the row can be updated, i.e, the update permission's filter condition holds true and after the row is updated, the update permission's check condition has to hold true.
>

paf31:
> Seems like the row permissions section will need to be further split into 2 sections now, a "before check" and an "after check". I currently can't seem figure out what the messaging should be for the 2 sections to get away with minimal changes. Will give it some more thought and get back. Ideas are welcome of course :)

In that case, I will just carry on without implementing the console changes for now. It should be fine to merge this without UI support, since there are no breaking metadata changes. Perhaps if anyone needs this feature immediately, we can guide them on using the console to set these permissions? @0x777 @rikinsk 

In the meantime, I will work on updates to the `ON CONFLICT` logic and the documentation.

yasinarik:
> Perhaps if anyone needs this feature immediately, we can guide them on using the console to set these permissions? @0x777 @rikinsk

@paf31 I really want to test it with your help. I can provide extensive feedback about the feature. As you may already know, I have suggested and gave a real-world example for it on #3536 

The lack of this feature is a blocker at the moment, unfortunately. I really appreciate all of your kind efforts.

paf31:
@yasinarik If you're able to build my branch then you should be able to use the API or the console to set update permissions on your table. There is a small example in the test suite in my branch: https://github.com/hasura/graphql-engine/blob/5f0fb0cfd706fc4db708d2164b0115325b78fb3d/server/tests-py/queries/graphql_mutation/update/permissions/schema_setup.yaml#L109

yasinarik:
> If you're able to build my branch then you should be able to use the API or the console to set update permissions on your table. There is a small example in the test suite in my branch:

@paf31 Really sorry for bothering you too much Dear Freeman; however, I think need a guide on this. I have never used Hasura without the web console (the one with the UI). May I request a bit more help?

I've checked the example. Btw, how long will it take do you guess the `update check` feature to be released officially under the master branch?

paf31:
@yasinarik One option is to export metadata from the settings page in the console, edit the JSON to add the `check` condition in the update permission object, and reimport the result.

The simpler way might be to use the `create_update_permission` API, which I've updated the documentation for [in my PR](https://github.com/hasura/graphql-engine/pull/3883) (docs [here](https://deploy-preview-3883--hasura-docs.netlify.com/graphql/manual/api-reference/schema-metadata-api/permission.html#id10)). You can now specify a `check` value in the permissions object, which has the same JSON format as the `filter` property, namely either a boolean expression or the name of a session variable.

Please let me know if you need more help.

yasinarik:
@paf31 Dear Phil, I couldn't manage to build your branch that includes the required changes. Actually I couldn't find the right resources to do that. Also, I can say that I don't know how to build your branch and deploy that version on, let's say, Heroku. 

--------

I understood how to export metadata and make changes then import it. It's OK.

--------
### Is This Feasible?

Apart from that, per my understanding, there seems to be **no way to combine** `filter` - `check` pairs in single update permission. With the new branch, there is only one possibility at the moment. All the **before** conditions are collected inside the `filter` property and all the **after** conditions are collected inside the `check` property.

I don't know if it makes sense to you guys; however, my idea is to **_combine_** the separately written **before and after**  conditions. The current improvement is great, I have to say that I can't express how I am glad about it. ☺️
 
I just want to tell you that there are a lot of real-world scenarios that it is **not appropriate** to define all the **before conditions** in a single `filter` property and to define all the **after conditions** in a single `check` property. It can cause exceptions to occur.

I imagine defining the `filter&check` pairs separately so it will be **truly possible** to exactly point a logically complex situation out without any exceptions.

Let me illustrate:

```
"_or": [
  {
    "filter": "some condition",
    "check": "some condition",
  },
  {
    "filter": "some condition",
    "check": "some condition",
  },
  {
    "filter": "some condition",
    "check": "some condition",
  },
]
```

--------------
### Extended Example:
```
{
  "_or": {
    [
      /// CASE 1:
      {
        "filter": {
          "_and": {
            [
              {
                "id_Doer": "x-hasura-id",
              },
              {
                "followState": "false",
              }
            ],
          }
        },
        "check": {
          "followState": "requested",
        },
      },

      /// END OF CASE 1 ----
      /// -----------------------

      /// CASE 2:
      {
        "filter": {
          "_and": {
            [
              {
                "id_Doer": "x-hasura-id",
              },
              {
                "followState": "requested",
              }
            ],
          }
        },
        "check": {
          "followState": "false",
        },
      },

      /// END OF CASE 2
      /// -----------------------

      /// CASE 3:
      {
        "filter": {
          "_and": {
            [
              {
                "id_Doer": "x-hasura-id",
              },
              {
                "followState": "true",
              }
            ],
          }
        },
        "check": {
          "followState": "false",
        },
      },

      /// END OF CASE 3 ----
      /// -----------------------

      /// CASE 4:
      {
        "filter": {
          "_and": {
            [
              {
                "id_Target": "x-hasura-id",
              },
              {
                "followState": "true",
              }
            ],
          }
        },
        "check": {
          "followState": "false",
        },
      },

      /// END OF CASE 4 ----
      /// -----------------------

      /// CASE 5:
      {
        "filter": {
          "_and": {
            [
              {
                "id_Target": "x-hasura-id",
              },
              {
                "followState": "req",
              }
            ],
          }
        },
        "check": {
          "_or": {
            [
              {
                "followState": "true",
              },
              {
                "followState": "false",
              }
            ],
          },
        },
      },
      /// END OF CASE 5 ----
      /// -----------------------
    ],
  },
};
```

```
/// CASE 1:
/// id_Doer is the logged in user.
/// If followState is "false",
/// it can only become "requested".
/// MEANING: userDoer sends a follow request to userTarget.
///
/// CASE 2:
/// id_Doer is the logged in user.
/// If followState is "requested",
/// it can only become "false".
/// MEANING: userDoer cancels the awaiting follow request sent to userTarget.
///
/// CASE 3:
/// id_Doer is the logged in user.
/// If followState is "true",
/// it can only become "false".
/// MEANING: userDoer unfollows userTarget.
///
/// CASE 4:
/// id_Target is the logged in user.
/// If followState is "true",
/// it can only become "false".
/// MEANING: userTarget can remove userDoer from its followers.
///
/// CASE 5:
/// id_Target is the logged in user.
/// If followState is "req",
/// it can become "true" or "false".
/// MEANING: userTarget can accept or decline the follow request of userDoer.
```

@0x777 @rikinsk @lexi-lambda

0x777:
@yasinarik 

> I just want to tell you that there are a lot of real-world scenarios that it is not appropriate to define all the before conditions in a single filter property and to define all the after conditions in a single check property. It can cause exceptions to occur.

I definitely agree with this. 

> I imagine defining the filter&check pairs separately so it will be truly possible to exactly point a logically complex situation out without any exceptions.

I'm not sure what the exact interface should be, but I think the biggest limitation currently in the check expression is that you cannot use the values of columns from before the update (old value of a column). 

Adding the ability to use 'old' values of a column in check expression is not easy to implement, we'll need to 1. come up with a meaningful syntax and 2. map to appropriate postgres sql statements (this is particularly hard because of upserts)

About your use case here, wouldn't the recommendation in https://github.com/hasura/graphql-engine/issues/3536#issuecomment-576675947 work?

tirumaraiselvan:
One idea could be to use a list of `(filter, check)` expressions per role (or use multiple roles with single `(filter,check)`). This would be related to the write version of this [RFC](https://github.com/hasura/graphql-engine/pull/3868). 

But, that would be outside the scope of this discussion.

PR #3760, opened 2020-01-23
Add downgrade command (close #1156)

### LINKED ISSUE #1156: allow catalog downgrade through CLI

paf31:
### Description

Adds the `graphql-engine downgrade` command to revert the schema to a previous version.

As the help text indicates, the usage is as follows:

```
Usage: graphql-engine downgrade --to <VERSION> [--dryRun]
  Downgrade the GraphQL Engine schema to the specified version
```

`--dryRun` will simply show the SQL which _would be run_, without actually running it.

### Affected components

- [x] Other (new command)

### Related Issues

#1156

### Solution and Design

I copied the downgrade scripts from the docs.hasura.io website into the migrations folder, and then removed the ones which don't seem to work against the latest schema any more. I can downgrade as far as schema version 12 but no further.

The code picks out the correct range of scripts to be run, and runs them in order.

### Steps to test and verify

Tested by migrating back to various older versions, and then trying to upgrade back by restarting the latest server. As noted above, this works for all versions back to version 12.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -1061,3 +1069,52 @@ serveOptionsParser =
...
+          help "The target schema version (e.g. v1.0.0-beta.1)"
+        )
+    <*> switch
+        ( long "dryRun" <>
+          help "Don't run any migrations, just print out the SQL."
+        )
+  where
+    -- This implements the mapping between application versions
+    -- and catalog schema versions.
+    readVersion "v1.0.0-alpha0"  = Just "1"
```
paf31: I actually think having the mapping here is strictly worse than keeping it in documentation, because it's easier to forget to update it, and the user won't have any way to downgrade in that case. At least using catalog versions, we can tell them the exact command to use. I'm happy to hear alternatives if anyone has ideas.

lexi-lambda: Why do you think it’s easier to forget to update this mapping than one in the documentation? I don’t see why one would be more obvious than the other.

In any case, I think being able to treat catalog versions as an implementation detail our users don’t have to worry about is a big enough win that I lean toward this approach, anyway. However, it does make me want two other features:

  1. I’d like to be able to print the list of versions `graphql-engine` knows about somehow. Otherwise, if I enter something like `1.1.0-beta.1` (leaving off the `v`) I just get a cryptic “cannot parse value” error, which is very unhelpful.

     I think the error message for supplying an invalid version should be improved to either list the known versions or tell people how to list them, and there should be something in the `--help` output that says how to list them. (We probably don’t want to list them all in the `--help` text all the time, since it would be a lot of output. But maybe we could show an example instead, and then we could probably get away with only printing out all the known versions if someone supplies an invalid version.)

  2. I’d like to be able to specify a catalog version by specifying a `--catalog-version` flag or something like that. This is useful both for developers working on the server and users trying out PR builds. In the former case, we know the catalog versions, and in the latter, we’re already interacting with users directly, so we can just tell them what to run. Since this would be so easy to add, it seems worthwhile!

paf31: @lexi-lambda How's this?

0x777:
This looks good @paf31, couple of things:
1. I would second having the `--catalog-version` option, it'll be particularly useful in a `git bisect` style debugging workflow to figure out the commit that introduced a bug.
2. > because it's easier to forget to update it, and the user won't have any way to downgrade in that case

   We can have a test which gets the list of available options for the downgrade command and checks them with the tags in the git repo (ignoring the ones that we know will not exist)

paf31:
`--to-catalog-version` is implemented already, and I'll work on adding the test today.

server/src-test/Hasura/Server/MigrateSpec.hs
```diff
@@ -64,20 +67,37 @@ spec
...
+            swap <$> migrateCatalog time
+      time <- getCurrentTime
+      transact (dropAndInit time) `shouldReturn` MRInitialized
+      downgradeResult <- (transact . lift) (downgradeTo "12" time)
+      downgradeResult `shouldSatisfy` \case
+        MRMigrated{} -> True
+        _ -> False
+      transact (upgradeToLatest time) `shouldReturn` MRMigrated "12"
+      
+    it "supports downgrades for every Git tag" $ singleTransaction do
```
paf31: @0x777 The test is now implemented, but I'm concerned that this will cause test failures when we tag a new release. Perhaps we should omit the latest tag? We wouldn't ever need to downgrade to the latest tag anyway, except for in development.

docs/graphql/manual/deployment/downgrading.rst
```diff
@@ -35,1276 +35,27 @@ During upgrades the server automatically migrates the catalogue to the latest ve
...
-+-----------------------------------+---------------------+
-
 Downgrading across catalogue versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can downgrade the catalogue from a particular version to its previous version by executing the appropriate SQL statements provided below.
+You can downgrade the catalogue from a particular version to a previous version by executing the ``graphql-engine`` executable on the command line,
+with the ``downgrade`` command, specifying the desired catalog version using one of the ``--to-`` flags.
 
 .. note::
```
marionschleifer: This seems like a rather long note. How about just having the third point as a note (as it's the unexpected case) and have the rest part of the normal tutorial?

server/src-test/Hasura/Server/MigrateSpec.hs
0x777: Oh right, nice catch. We can omit the latest tag, it makes sense, when we are releasing `v1.2.0`, we only need to ensure that downgrade option exists to versions < `v1.2.0`. Now say we are releasing `v1.3.0` and say we forgot to add the downgrade option to `v1.2.0`, this test will fail when we are actually making the release, so we'll have to push a new commit, delete the old tag, add it at the new commit - it is a little inconvenient but I guess it is better than not having it. So immediately after making a release, someone opens a PR with the new mapping? Is that the suggested workflow?

@lexi-lambda thoughts?

docs/graphql/manual/deployment/downgrading.rst
```diff
@@ -35,1276 +35,27 @@ During upgrades the server automatically migrates the catalogue to the latest ve
...
-+-----------------------------------+---------------------+
-| v1.0.0-alpha0 - v1.0.0-alpha15    | 1                   |
-+-----------------------------------+---------------------+
-
 Downgrading across catalogue versions
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-You can downgrade the catalogue from a particular version to its previous version by executing the appropriate SQL statements provided below.
+You can downgrade the catalogue from a particular version to a previous version by executing the ``graphql-engine`` executable on the command line,
+with the ``downgrade`` command, specifying the desired catalog version using one of the ``--to-`` flags.
```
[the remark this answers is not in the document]
paf31: In order to get autocomplete in the usage message, the simplest approach was to turn each of these into flags. I was hoping the example would clarify the usage.

docs/graphql/manual/deployment/downgrading.rst
paf31: Done, thanks.

server/src-test/Hasura/Server/MigrateSpec.hs
paf31: I think perhaps the best option is to check if the current version is itself a tag. If so, check for all but the latest tag (all but the current release), and if not, check all tags. This way, the build will not fail for the tag itself, but for any commits after it. Ideally the first commit after the tag can add the downgrade option.

PR #3767, opened 2020-01-23
Test and dev script improvements

jberryman:
You can do `scripts/dev.sh test` to verify coverage.


server/cabal.project.dev-sh
```diff
@@ -0,0 +1 @@
+cabal.project
```
lexi-lambda: What is the use case behind not just using the `cabal.project` file directly? I can’t imagine a situation where I would want `dev.sh` to use a different `cabal.project` file than the one that’s checked in.

server/src-test/Main.hs
```diff
@@ -31,7 +31,10 @@ import qualified Hasura.Server.MigrateSpec    as MigrateSpec
 
 data TestSuites
   = AllSuites !RawConnInfo
-  | SingleSuite !TestSuite
+  -- ^ Run all test suites. It probably doesn't make sense to be able to specify additional 
+  -- hspec args here.
```
lexi-lambda: I’m not totally convinced by this comment! I think it makes as much sense to specify additional hspec arguments  when running all the test suites as it does when running just one.

Originally I considered making these truly different test suites, in the Cabal sense. We could have `test-suite graphql-engine-unit-tests` and `test-suite graphql-engine-postgres-tests`. However, working with multiple test suites can be a little awkward, especially with `stack`, so I decided to only have one test suite that can be configured via options. `cabal new-run` makes this a lot simpler, though, so maybe that would be a simpler solution now. I don’t feel strongly either way.

server/cabal.project.dev-sh.local
```diff
@@ -0,0 +1,23 @@
...
+  -- NOTE: this seems to work so long as there is no 'ghc-options: -O2' in the cabal file,
+  -- but new-build will report 'Build profile: -O1' for some reason. 
+  -- See:https://github.com/haskell/cabal/issues/6221 
+  optimization: 0
+  documentation: false
+  flags: +developer
+  -- NOTE: 'cabal new-build --enable-coverage' seems to rebuild all deps with coverage
+  -- which is not what we originally wanted. But building the test modules
+  -- themselves with coverage is actually nice for validation.
+  coverage: true
```
lexi-lambda: I’d be happy to just add this to `cabal.project.dev`, personally. It seems reasonable, and that would prevent us from having to maintain two copies of this file.

server/cabal.project.dev-sh
jberryman: We symlink `cabal.project` to `cabal.project.dev-sh` since the only way to specify a new `cabal.project.local`  is to do so indirectly via specifying the cabal.project file name

server/src-test/Main.hs
jberryman: Does `--match` fail when nothing matches? If not I suppose match and skip would both be fine to pass to all.

server/cabal.project.dev-sh.local
jberryman: We still need two copies I think. The dev script needs control over this script (e.g. if a dev alters their `cabal.project.local` to remove `coverage` the script will fail). Other things will probably be added to the file as well.

I initially was forced into this because `--enable-coverage` seemed to be building all dependencies with coverage, but I think it's actually good for maintenance of the script.

server/cabal.project.dev-sh
lexi-lambda: Ah, I see—the trick is that if you specify `--project-file=foobar`, then `cabal-install` looks for `foobar.local` instead of `cabal.project.local`. That makes sense.

I don’t like this proliferation of symlinks and duplicate configuration files—I think it’s confusing—but I don’t have a better solution right now. One option would be for `dev.sh` to manage its own temporary directory (via `mktemp -d`), where it could copy files or create symlinks and muck about with them as desired. I don’t know if that would be more complicated than you’d like, however.

IMO, the real right thing to do here is upstream some changes to `cabal-install`. I’d love to improve some of these things—it’s not clear to me that the behavior with things like `--enable-coverage` isn’t just a bug (IIUC flags are only *supposed* to affect local packages). But I recognize that in the meantime we need a workflow that works!

jberryman: > --it’s not clear to me that the behavior with things like --enable-coverage

ya I agree, I think I referenced a somewhat relevant issue in comments

> the real right thing to do here is upstream some changes to cabal-install

Right, but I think there's no way to avoid hacks and workarounds unless we're going to ship a cabal-install fork with graphql-engine. My hope is this script means just one person has to figure out these hacks and corner cases. It would be great to devote some resources to cabal-install though

lexi-lambda: Yes, agreed. I’m alright with this trick for now, and maybe we can fix `cabal-install` in the future.

jberryman:
@lexi-lambda Can you clarify where review stands here?

server/src-test/Main.hs
jberryman: I'm fine if the behavior changes going forward if someone decides it makes sense (hopefully the comment reflects that)

lexi-lambda: To your original question about `--match`: I don’t know what it does if nothing matches. I’d expect it to pass (and run nothing), but I haven’t checked.

I think that if we want to expose `hspec` options on the command line (which is reasonable), we should not treat `TestSuites` like subcommands, but more like other flags like `--match` and `--skip`. So if you wanted, you could pass `--suite=unit`, and it would select those tests.

jberryman: I'm fine with any of those changes you're suggesting, but don't have much of an opinion. Would you agree that can happen later? I'd love to get these fixes merged if possible.

PR #3770, opened 2020-01-24
improve docs loading time, reduce bundle size

praveenweb:
### Description

This PR addresses the following:
- Remove duplicate fontawesome and update to version 4.7.0
- Lazy load intercom in setTimeout
- Remove underscore.js (not used anywhere)
- Remove bootstrap pagination script inside vendor.js (not used anywhere)
- Update algolia search to version 3.9.4
- Removes lightbox and its own jQuery (1.11.0) and css files
- Defer vendor script

TODO:
- Combine all CSS files into one / preferably inline smaller ones

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


praveenweb:
@rikinsk - Seems like the contents of `pygments.css` which added highlighting was not applied in the earlier releases. Removed them now to maintain the original highlighting colours.

rikinsk:
@praveenweb What was the process / guidelines followed while making these changes? Was it based on suggetions form some tool or something else. I ask majorly to understand what needle we are trying to move with this and how do I verify if that has been achieved. 

From my local testing (comparing across 2 preview docs builds with network throttling) there doesnt seem to be any significant change to load times, rather I am seeing a slight increase in DOMContentLoaded times.

rikinsk:
Also the image click-to-zoom is now broken as we seem to have removed lightBox code. 

I am still in the process of verifying that this has no impact on UX so might find more of such issues

praveenweb:
@rikinsk - The optimisations were done based on Google Search Console Speed index which reported many links to have speed index averaging between slow and moderate.

A quick way to analyse this would be via [PageSpeed Insights](https://developers.google.com/speed/pagespeed/insights/) or using Chrome's Lighthouse Audits. For example, i took [this page on Prod](https://docs.hasura.io/1.0/graphql/manual/api-reference/postgresql-types.html) and tested.

[PageSpeed Insights for Prod](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdocs.hasura.io%2F1.0%2Fgraphql%2Fmanual%2Fapi-reference%2Fpostgresql-types.html&tab=mobile&hl=en)

[PagSpeed Insights for this PR](https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fdeploy-preview-3770--hasura-docs.netlify.com%2Fgraphql%2Fmanual%2Fapi-reference%2Fpostgresql-types.html&hl=en)

Note the optimisations / opportunities suggested for the existing Prod link and test it with current PR's results to see if changes in this PR are helping. Many of the changes came from those suggestions and from Chrome's Lighthouse Audits, apart from making general improvements like dead code removal etc. 

The lightbox code was intentionally removed. If image click to zoom feature is necessary/important from the UX side, we can bring it via a different implementation. The current implementation uses jQuery 1.11 (which is very old and reported to have vulnerabilities) and hence there needs to exist two different versions of jQuery (v3 and this) in the bundle.

I assume that Netlify PR apps are not cached/optimised to the level of the Prod link and hence prod should give better response time once these changes are incorporated especially for the Desktop tests.

ashishra0:
@praveenweb not a valid heroku command (deploy, delete)

ashishra0:
@ashishra0 not a valid heroku command (deploy, delete)

ashishra0:
@ashishra0 not a valid heroku command (deploy, delete)

ashishra0:
@ashishra0 not a valid heroku command (deploy, delete)

marionschleifer:
@praveenweb @rikinsk I've looked at this PR in more detail now and here is my feedback:

- I tried re-ordering the CSS files in `default.css` and could not make the tabs work. Either there's a combination that I've missed or the problem comes from somewhere else.

- The search doesn't work anymore. I can type in the search field normally:

image

But then when I submit there's a `404` for every search that I've tried:

image

- Loading times of `<script>` only improve with `defer` when the `src` attribute is present, otherwise it's ignored ([source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script)). Also `async` should only be used together with the `src` attribute ([source](https://www.w3schools.com/tags/att_script_async.asp)). So in order to make these keywords work, we'd need to add `src` to the scripts.

- Why did we downgrade Algolia? @praveenweb I think you told me this once but I don't remember.

praveenweb:
@marionschleifer

Search result page doesn't work with the Netlify apps that are generated in PRs. (i think the URLs formed in search result is the cause which needs a redirect rule). But nothing has been changed in this regard and should continue to work in Prod. Algolia was actually upgraded from v3.22.1 to v3.9.4 and no breaking changes.

And you are right, <script defer> works with src. We do have one optimization that can be done. There are multiple <script> tags with $(document.ready) code which can be merged into one, just from a code consistency / parsing angle.

Right now the blocker is that tabs are broken. Let me recheck and verify that.

Stefmore02:
I'm closing this PR due to recent updates and improvements in the Docs realm.

PR #3804, opened 2020-01-30
Add check expresion to update permissions (close #384)

### LINKED ISSUE #384: Add custom permission check to update operation on the changed data

paf31:
### Description

- Add a `check` constraint property to the update permission data structure
- Validate the check constraint after updates

The console UI has not yet been updated and I was hoping someone could pair with me on that. It is still possible to test this using direct changes to `hdb_permission` right now, or presumably using the CLI.

Left for a later PR:

- Console UI changes 
- Update documentation for permissions in the console

### Affected components

- [x] Server
- [x] Tests
- [ ] Console

### Solution and Design

See https://github.com/0x777/graphql-engine/blob/rfc-update-permission-check-condition/rfcs/update-permission-check-condition.md

#### Catalog upgrade

- [x] No

#### Metadata

- [x] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


paf31:
@0x777 Current status: I'm trying to figure out why `xmax` works in some cases but not in every case (see the test failures for example).

0x777:
@paf31 
> Edit: the problem is that xmax doesn't work for views, so in particular, if a user is using automatically updatable views for inserts, xmax won't work.

Ah! That makes sense! So if we unconditionally include the `CASE xmax = 0 ..` in `returning`, inserting into a view will result in errors. However, if we only use the `xmax` column based `CASE` expression in `returning` when `on_conflict` is used, we should be able to avoid these errors? Because `on_conflict` doesn't make sense on views anyway (in fact, we shouldn't even generate `on_conflict` arguments in the generated graphql schema for a view). What do you think?

paf31:
@0x777 One more question is what to do if there is an insert check, but no update check, and the `ON CONFLICT` behavior is used. In the old code, we would use the insert check, but now we will not. This is technically breaking, so do you think we should keep the old behavior?

0x777:
> ON CONFLICT does seem to work on automatically-updatable views for me

Interesting! I did a little bit of digging:

```sql
db> create table genre (id serial primary key, name text not null unique);
CREATE TABLE
db> create view genre_view as select * from genre;
CREATE VIEW
db> insert into genre (id, name) values (DEFAULT, 'drama');
INSERT 0 1
db> insert into genre_view (id, name) values (DEFAULT, 'crime');
INSERT 0 1
db> insert into genre (id, name) values (DEFAULT, 'drama') on conflict on constraint genre_name_key DO NOTHING;
INSERT 0 0
db> insert into genre_view (id, name) values (DEFAULT, 'drama') on conflict on constraint genre_name_key DO NOTHING;
ERROR:  constraint "genre_name_key" for table "genre_view" does not exist
```

What's interesting is that this works:
```sql
db> insert into genre_view (id, name) values (DEFAULT, 'drama') on conflict (name) DO NOTHING;
INSERT 0 0
```

`on conflict` with unique index inference seems to be supported on views. So our current `on_conflict`
API (which relies on unique constraint names) is limited when it comes to views which seems 
to work in our favour. Just to summarize, in case if we ever support unique index inference in our  `on_conflict` API, we won't be able to enforce permissions on views if the insert and update permission's check constraints differ, right?

> In the old code, we would use the insert check, but now we will not. This is technically breaking, so do you think we should keep the old behavior?

Oh yes, I guess it makes sense to use insert `check` constraint if the update check constraint is missing.

paf31:
> Just to summarize, in case if we ever support unique index inference in our on_conflict API, we won't be able to enforce permissions on views if the insert and update permission's check constraints differ, right? 

Correct.

paf31:
@0x777 This is ready for another review when you have time, please. I'm not planning to implement the console changes or documentation changes yet (since the docs only seem to describe the console UI anyway), but I think this could be merged without those while we decide on the right approach to the UI.

server/tests-py/queries/graphql_mutation/update/permissions/user_cannot_unpublish.yaml
```diff
@@ -0,0 +1,26 @@
+description: Try to change the published flag on a published article
```
0x777: This test doesn't seem to be testing the `check` constraint defined above. We should testing for both a success and an error.

0x777:
@paf31 LGTM except for the test case.

server/tests-py/queries/graphql_mutation/update/permissions/user_cannot_unpublish.yaml
paf31: Oops, thanks. I forgot to finish this and got confused because the tests were passing and I had uncommitted work locally.

server/tests-py/queries/graphql_mutation/update/permissions/user_cannot_publish.yaml
```diff
@@ -0,0 +1,27 @@
+description: Try to change the published flag on an unpublished article
+url: /v1/graphql
+response:
+  errors:
+  - extensions:
+      path: $
+      code: permission-error
+    message: Check constraint violation. insert check constraint failed
```
0x777: I think we'll need to change the message appropriately for update checks, `.. update check constraint failed`?

Issue #3808, opened 2020-01-31
unable to insert data with foreign_key to enum table


polRk:
I tried to set the value for the enum of the linked column

```log
main.js.gz:1 Uncaught (in promise) TypeError: Cannot use 'in' operator to search for 'info' in Internal Server Error
    at n (main.js.gz:1)
    at n (main.js.gz:1)
    at main.js.gz:1
```

Https query 
```json
{"type":"update","args":{"table":{"schema":"public","name":"_action"},"$set":{"type":"viber_bot","description":null},"$default":[],"where":{"id":"d54869a7-77b5-4fa4-b8cb-8707bca172fa"}}}
```

It helped me to make the table normal from enum first, and then make enum again

rikinsk:
@polRk Can you share some more information on what exactly you were attempting to do. A little more info about your schema and the action you were attempting will help

polRk:
My table 

<img width="409" alt="Снимок экрана 2020-01-31 в 4 17 34 PM" src="https://user-images.githubusercontent.com/39828645/73542185-345a8080-4445-11ea-8f64-61e268c327ce.png">

My additional table options
<img width="552" alt="Снимок экрана 2020-01-31 в 4 18 00 PM" src="https://user-images.githubusercontent.com/39828645/73542253-55bb6c80-4445-11ea-99f9-44be5ca9815d.png">

My enum table _action_type
<img width="366" alt="Снимок экрана 2020-01-31 в 4 19 37 PM" src="https://user-images.githubusercontent.com/39828645/73542309-7be10c80-4445-11ea-84b0-3df8bd81f151.png">

If I am not mistaken when I created the connection, the action for updating Foreign Key - On Update Violation: restrict. After which I updated and deleted some values ​​in my enum table. And when adding a new record to the data table, there was an error. I recreated the enum table with toggle button and Make On Update Violation: cascade. And it worked

rikinsk:
Thanks for the extra information. I unfortunately still did not understand the actual issue. This is what I have understood till now:

- You have an enum table `_action_type`
- You have a table `_action` that has a foreign key to the enum table.
- You already had some data in the `_action` table and then you added/edited some data in the `_action_type` table. 
- Now while adding data to the `_action` table you faced an error. Was the value set for `type` column was not present in the `_action_type` table (causing a foreign-key violation)? Or was the error due to something else?
[...]

Issue #3789, opened 2020-01-28
Console Permissions - Cannot read property 'length' of undefined


jcmoore0:
**Problem**:  **Error:  Cannot read property 'length' of undefined**
Additional Info:
  1.  Occurs when editing 'SELECT' permission.
  2.  Is intermittent and usually occurs on 1 of my 2 roles for a table.
  3.  Bulk sql/ loadSchema data from server is missing 'permissions.select.computed_fields' empty array
  4.  I don't have any computed_columns

**v1.1.0-beta.2**

**line 89 -  console/src/components/Services/Data/PermissionsSummary/utils.js**
permission['computed_fields'] is undefined
```
 Object.keys(tableFields).forEach(fieldType => {
      noFields = noFields && !permission[fieldType].length;

      allFields =
        allFields &&
        (permission[fieldType] === '*' || 
          permission[fieldType].length === tableFields[fieldType].length);   <-- there is not a computed_fields property on permission
    });
```

Thanks,
John

jcmoore0:
Thanks @rikinsk for the quick response.
As a short term work around,  I can use the previous console version to edit permissions.

rikinsk:
@jcmoore0 I was unfortunately not able to reproduce this. I assume you must be seeing a 500 page shown when the error occurs. Could you share the error stack trace displayed.

jcmoore0:
Hi @rikinsk  - I am getting all 200 responses.   It happens on several tables on one of my 2 roles.

data returns with 200:
```
  return dispatch(requestAction(url, options)).then(
      data => {
        const mergedData = mergeLoadSchemaData(
          JSON.parse(data[0].result[1]),
          data[1],
          JSON.parse(data[2].result[1]),
          JSON.parse(data[3].result[1])
        );

**Missing property computed_fields:[]**
data[1][5].permissions[0].permissions.select {
  filter: {attachment: {…}}
  columns: (5) 
  allow_aggregations: false
}
```

```
Client error:
TypeError: Cannot read property 'length' of undefined
    at http://127.0.0.1:3001/rstatic/dist/main-885638bc67bbc108f8ba.js:408806:171
    at Array.forEach (<anonymous>)
    at getPermissionColumnAccessSummary (http://127.0.0.1:3001/rstatic/dist/main-885638bc67bbc108f8ba.js:408801:30)
    at getColumnSection (http://127.0.0.1:3001/rstatic/dist/main-885638bc67bbc108f8ba.js:524287:160)
[...]

Issue #3777, opened 2020-01-25
support _is_null operator for jsonb fields in permission builder


piroszhog:
### Version
1.0.0

### What I try to do
Put `null` in permission check of `jsonb` field named `message`

![image](https://user-images.githubusercontent.com/17676101/73127346-81fb6700-3fc7-11ea-903a-633aef192013.png)

### What I get
App crashes with an error below

```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at p (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:388881)
    at n (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:388554)
    at https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:389022
    at Array.forEach (<anonymous>)
    at p (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:388889)
    at n (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:388554)
    at https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:389022
    at Array.forEach (<anonymous>)
    at p (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/main.js.gz:1:388889)
```

### What I excpect

1. No app crash
2. Ability to add filter by `null` value of field, not only by json object

tirumaraiselvan:
@rikinsk Can you pass `null` in the permission builder as jsonb operand?

rikinsk:
@piroszhog Can you explain your use case a little more. When would  `jsonb_col contains null` evaluate to `true`?

piroszhog:
@rikinsk I expected that it will work like non-null check, the same behaviour as from `_is_null: false` operator. Actually, it would be great to add `_is_null` operator for `jsonb` fields.

rikinsk:
Got it. This seems to be a regression. Changing the scope of this issue to allow `_is_null` operator for json fields

tirumaraiselvan:
@rikinsk Wouldn't there be a difference between Postgres NULL (value not given to a column) and JSONB `null` (value of a json/jsonb column is set explicitly to null)?

PR #3813, opened 2020-01-31
misc console fixes

rikinsk:
### Description

- handle string errors (#3808)
- handle empty columns / computed fields in permissions (#3789)
- allow is_null for all operators (#3777)

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3789, #3777, #3808


console/src/components/Services/Common/Notification.js
```diff
@@ -31,48 +31,51 @@ const getNotificationDetails = (detailsJson, children = null) => {
...
+      } else if ('info' in error) {
+        notificationMessage = error.info;
+      } else if ('message' in error) {
+        if (error.code) {
+          if (error.message.error) {
+            notificationMessage = error.message.error.message;
+          } else {
+            notificationMessage = error.message;
+          }
+        } else if (error && error.message && isString(error.message)) {
```
beerose: No need to check for `error` because it is already done in line 35.

console/src/components/Services/Data/TablePermissions/Permissions.js
```diff
@@ -886,87 +885,58 @@ class Permissions extends Component {
...
+
+              if (permittedFields === '*') {
                 checked = true;
               } else {
-                checked = permissionsState[query].columns.includes(column);
+                checked = permittedFields.includes(fieldName);
               }
             } else {
               checked = false;
             }
```
beerose: What about doing `let checked = false` so the `else` statement can be removed?
```js
            let checked = false;
            if (permissionsState[query]) {
              const permittedFields = permissionsState[query][fieldType] || [];

              if (permittedFields === '*') {
                checked = true;
              } else {
                checked = permittedFields.includes(fieldName);
              }
            } 
```

Issue #1734, opened 2019-03-12
allow copy-pasting permissions json in console


praveenweb:
Currently i'm forced to use the permissions builder UI on the console to create permissions. I want to be able to copy paste already created permission json.

rikinsk:
needs to think about how to handle invalid permission rule pasted

rikinsk:
Closed via #3823

PR #3823, opened 2020-02-03
console: permission builder fixes

rikinsk:
### Description

- allow manually editing permissions (#1734)
- separate operators for json and jsonb column types
- allow `null` as valid JSON type

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### DIFF
--- a/console/src/components/Services/Data/TablePermissions/Permissions.js
+++ b/console/src/components/Services/Data/TablePermissions/Permissions.js
@@ -91,16 +91,14 @@ class Permissions extends Component {
     dispatch(fetchFunctionInit());
   }
 
-  componentWillReceiveProps(prevProps) {
+  componentWillReceiveProps(nextProps) {
     const currPermissionsState = this.props.permissionsState;
-    const prevPermissionsState = prevProps.permissionsState;
+    const nextPermissionsState = nextProps.permissionsState;
 
     // clear local filterString
     if (
-      (currPermissionsState.role &&
-        currPermissionsState.role !== prevPermissionsState.role) ||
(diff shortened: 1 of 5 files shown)


console/src/components/Services/Data/TablePermissions/Permissions.js
```diff
@@ -89,6 +91,19 @@ class Permissions extends Component {
     dispatch(fetchFunctionInit());
   }
 
+  componentWillReceiveProps(nextProps) {
```
beerose: `componentWillReceiveProps` will be deprecated and in React 16.3 there was `getDerivedStateFromProps` introduced to replace it. So can we use this instead?
```js
static getDerivedStateFromProps(nextProps, prevState) {
    const currPermissionsState = prevState.permissionsState;
    const nextPermissionsState = nextProps.permissionsState;

    if (
      currPermissionsState.role !== nextPermissionsState.role ||
      currPermissionsState.query !== nextPermissionsState.query
    ) {
      return { filterString: '' };
    }
    return null;
}
```

rikinsk: I didnt go down that route as I would then have to store the previousProps in state which I wanted to avoid. Is there any other way I am not aware of?

In any case, we should be converting all these to hooks anyway when we refactor so it might not be worth adding a lot of code to achieve this

rikinsk: Figured out an easy workaround

PR #3851, opened 2020-02-07
cli: load assets from server if cdn is disabled (close #3382)

### LINKED ISSUE #3382: cli console should use local assets if the server is configured to do so
currently the console run via the CLI uses cdn assets even if the server is configured to use local assets

scriptonist:
### Description
Currently, the console run via the CLI uses CDN assets even if the server is configured to use local assets.  This PR changes this behaviour and will load assets from server if the server is configured to do so, ie if `HASURA_GRAPHQL_CONSOLE_ASSETS_DIR` set on the server.

### Affected components

- [ ] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues
#3382

### Steps to test and verify
- Start server with `HASURA_GRAPHQL_CONSOLE_ASSETS_DIR` set to a valid directory
- Open console with `hasura console`, check if assets are loaded from server
#### Breaking changes

- [x] No Breaking changes


rikinsk:
Quick suggestion, can we not introduce a new flag and instead just use assets from server if the HASURA_GRAPHQL_CONSOLE_ASSETS_DIR env var is set. I as a user would expect the ENV var to apply to both server and cli consoles without having to specify a flag again

scriptonist:
@rikinsk I think that makes sense, updated the PR to address that.

shahidhk:
@arvi3411301 Can you review this?

cli/commands/console.go
```diff
@@ -127,6 +130,34 @@ func (o *ConsoleOptions) Run() error {
 	consoleTemplateVersion := o.EC.Version.GetConsoleTemplateVersion()
 	consoleAssetsVersion := o.EC.Version.GetConsoleAssetsVersion()
 
+	// Determine from where assets should be served
+	getConfig := func(url string, target interface{}) error {
```
arvi3411301: can we have this a method in `cli.ServerConfig` and call this method on `ec.Validate`? https://github.com/hasura/graphql-engine/blob/master/cli/cli.go#L63

Anyway, we are contacting server `v1/version` to get the server version. And `v1alpha1/config` also returns version.

shahidhk: Okay, let's use the config api to get all the details. I'd still recommend putting this behind a feature flag -- just to be on the safe side.

arvi3411301: @shahidhk feature flags can be used only when we have the server version in the CLI.

cli/cli.go
```diff
@@ -95,6 +97,32 @@ type Config struct {
 	MigrationsDirectory string `yaml:"migrations_directory,omitempty"`
 	// ActionConfig defines the config required to create or generate codegen for an action.
 	ActionConfig types.ActionExecutionConfig `yaml:"actions"`
+
+	HasuraServerConfigAPI HasuraServerConfigAPI
```
arvi3411301: Can we move this to type `ServerConfig`?

arvi3411301: add `yaml:"-"`

cli/cli.go
```diff
@@ -95,6 +97,32 @@ type Config struct {
...
+	}
+	defer r.Body.Close()
+
+	return json.NewDecoder(r.Body).Decode(&c.HasuraServerConfigAPI)
+}
+
+// HasuraServerConfig is the type returned by the v1alpha1/config API
+// TODO: Move this type to a client implementation for hasura
+type HasuraServerConfigAPI struct {
+	Version          string `json:"version,omitempty"`
```
arvi3411301: do we need `omitempty` since this is not going to be exported?

cli/cli.go
```diff
@@ -95,6 +97,32 @@ type Config struct {
 	MigrationsDirectory string `yaml:"migrations_directory,omitempty"`
 	// ActionConfig defines the config required to create or generate codegen for an action.
 	ActionConfig types.ActionExecutionConfig `yaml:"actions"`
+
+	HasuraServerConfigAPI HasuraServerConfigAPI
+}
+
+func (c *Config) InitHasuraServerConfigAPI() error {
```
arvi3411301:
```suggestion
func (s *ServerConfig) GetHasuraServerConfigAPI() error {
```

cli/cli.go
```diff
@@ -360,11 +388,12 @@ func (ec *ExecutionContext) Validate() error {
 }
 
 func (ec *ExecutionContext) checkServerVersion() error {
-	v, err := version.FetchServerVersion(ec.Config.ServerConfig.Endpoint)
+	err := ec.Config.InitHasuraServerConfigAPI()
```
arvi3411301: can we move this to `func (ec *ExecutionContext) readConfig()`

cli/commands/console.go
```diff
@@ -126,6 +127,9 @@ func (o *ConsoleOptions) Run() error {
 
 	consoleTemplateVersion := o.EC.Version.GetConsoleTemplateVersion()
 	consoleAssetsVersion := o.EC.Version.GetConsoleAssetsVersion()
+	if ec.Config.HasuraServerConfigAPI.ConsoleAssetsDir != "" {
```
arvi3411301:
```suggestion
	if o.EC.Config.HasuraServerConfigAPI.ConsoleAssetsDir != "" {
```

cli/cli.go
```diff
@@ -437,6 +461,11 @@ func (ec *ExecutionContext) readConfig() error {
 			},
 		},
 	}
+	// this populates the ec.Config.ServerConfig.HasuraServerInternalConfig
+	err = ec.Config.ServerConfig.GetHasuraInternalServerConfig()
+	if err != nil {
+		return errors.Wrap(err, "cannot get config information from server")
+	}
 	return ec.Config.ServerConfig.ParseEndpoint()
```
arvi3411301: parse the endpoint first

cli/cli.go
```diff
@@ -68,7 +70,32 @@ type ServerConfig struct {
...
-	ParsedEndpoint *url.URL `yaml:"-"`
+	ParsedEndpoint             *url.URL                   `yaml:"-"`
+	HasuraServerInternalConfig HasuraServerInternalConfig `yaml:"-"`
+}
+
+func (c *ServerConfig) GetHasuraInternalServerConfig() error {
+	// Determine from where assets should be served
+	url := c.Endpoint + "/v1alpha1/config"
+	var client = &http.Client{Timeout: 10 * time.Second}
+	r, err := client.Get(url)
```
arvi3411301: If admin_secret is set, we need to set header

arvi3411301: Also, handle error message in case `admin_secret` is incorrect

cli/cli.go
```diff
@@ -68,7 +72,53 @@ type ServerConfig struct {
...
+		req.Header.Set("x-hasura-admin-secret", c.AdminSecret)
+	}
+	if err != nil {
+		return errors.Wrap(err, "error fetching config from server")
+	}
+	r, err := client.Do(req)
+	if err != nil {
+		return err
+	}
+	type apiError struct {
```
arvi3411301: use `HasuraError` from hasuradb package

arvi3411301:
@scriptonist Since we are going to force use `adminSecret` header, we can remove `HasAccessKey` check [here](https://github.com/hasura/graphql-engine/blob/8f6b2302a4b37c15d76ea0fdbfb4ddeef2bfc94d/cli/version/feature_flags.go#L11)

cli/cli.go
```diff
@@ -68,7 +73,48 @@ type ServerConfig struct {
...
+	if c.AdminSecret != "" {
+		req.Header.Set("x-hasura-admin-secret", c.AdminSecret)
+	}
+	if err != nil {
+		return errors.Wrap(err, "error fetching config from server")
+	}
+	r, err := client.Do(req)
+	if err != nil {
+		return err
+	}
```
arvi3411301: move `defer r.Body.Close()` here

cli/cli.go
```diff
@@ -68,7 +73,48 @@ type ServerConfig struct {
...
+	HasuraServerInternalConfig HasuraServerInternalConfig `yaml:"-"`
+}
+
+func (c *ServerConfig) GetHasuraInternalServerConfig() error {
+	// Determine from where assets should be served
+	url := c.Endpoint + "/v1alpha1/config"
+	client := http.Client{Timeout: 30 * time.Second}
+	req, err := http.NewRequestWithContext(context.Background(), "GET", url, nil)
+	if c.AdminSecret != "" {
+		req.Header.Set("x-hasura-admin-secret", c.AdminSecret)
```
arvi3411301: use `XHasuraAdminSecret` constant from [here](https://github.com/hasura/graphql-engine/blob/8f6b2302a4b37c15d76ea0fdbfb4ddeef2bfc94d/cli/commands/migrate.go#L146)

cli/cli.go
```diff
@@ -68,7 +73,48 @@ type ServerConfig struct {
...
+	}
+	if err != nil {
```
arvi3411301: shouldn't the error check happen before setting `adminSecret` header?

scriptonist: 🤦‍♂ Yes!

cli/cli.go
scriptonist: I think this will cause an import cycle 🤔

arvi3411301: yes, we can move the const to `hasuradb` package or in util?

cli/cli.go
```diff
@@ -149,7 +152,51 @@ type ServerConfig struct {
...
 	APIPaths *ServerAPIPaths `yaml:"api_paths,omitempty"`
 
-	ParsedEndpoint *url.URL `yaml:"-"`
+	ParsedEndpoint             *url.URL                   `yaml:"-"`
+	HasuraServerInternalConfig HasuraServerInternalConfig `yaml:"-"`
+}
+
+func (c *ServerConfig) GetHasuraInternalServerConfig() error {
+	// Determine from where assets should be served
+	url := c.Endpoint + "/v1alpha1/config"
```
arvi3411301, line 161: Can we get the config API path from `ServerConfig.Paths.Config`

shahidhk:
@scriptonist Will CLI fallback to `/v1/version` if config api is disabled?

scriptonist:
Yes @shahidhk, the following commits should address that, let me know if you feel otherwise
- https://github.com/hasura/graphql-engine/pull/3851/commits/b4376bca73fc66ec9af6ec8d3369c748a6da8ba8 
- https://github.com/hasura/graphql-engine/pull/3851/commits/4f695b02a1b06ac81d543c2fdb7c00732a738fb6

datakurre:
@scriptonist Why did this pull removed support for `hasuraconsole --static-dir` (which is stil documented, but seems to do nothing due to this change)?

PR #3860, opened 2020-02-10
Forking audit

jberryman:
### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

A few existing issues are flagged with TODO, some reference new bug reports

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -546,6 +549,8 @@ onClose
 onClose logger lqMap wsConn = do
   logWSEvent logger wsConn EClosed
   operations <- liftIO $ STM.atomically $ ListT.toList $ STMMap.listT opMap
+  -- TODO We race several large STM transactions, with the possibility of livelock.
```
0x777: I agree, no point in running them concurrently but I don't see how a livelock can happen, given that each of these transactions only touch a single `TVar`.

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Protocol.hs
```diff
@@ -22,6 +22,8 @@ import           Hasura.EncJSON
 import           Hasura.GraphQL.Transport.HTTP.Protocol
 import           Hasura.Prelude
 
+-- TODO what is the provenance? It looks like this comes from the client.
+--      Can any client who can guess another client's OperationId do anything naughty?
```
0x777: Nope, an `OperationId` is unique per websocket.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -457,13 +458,15 @@ onMessage authMode serverEnv wsConn msgRaw =
                            wsConn authMode params
       CMStart startMsg  -> liftIO $ onStart serverEnv wsConn startMsg
       CMStop stopMsg    -> liftIO $ onStop serverEnv wsConn stopMsg
+      -- TODO do we need onStop/removeLiveQuery cleanup here? I guess this is handled by 'onClose'?
```
0x777: > I guess this is handled by 'onClose'

That's correct.

jberryman: Under what conditions might `onClose` not get run? I haven't traced this. From what I recall that depends somewhat on internals of the websockets lib we use

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
jberryman: Hm not clear to me what you mean (I'm looking at body of `removeLiveQuery` in which we have racing `lookupfrommap ... stmstuff... deletefrommap`), would be interesting to look into it further, but just making it serial seems more pragmatic :+1:

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Protocol.hs
jberryman: Thanks. Could you elaborate? Do you mean the operationId + someotherkey together form a unique key so we don't have to worry about operationId collisions?

0x777: I'm not sure I follow what you are asking, the type represents `id` as documented here: https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/State.hs
```diff
@@ -142,5 +147,6 @@ removeLiveQuery lqState (LiveQueryId handlerId cohortId sinkId) = do
       if handlerIsEmpty
         then do
           STMMap.delete handlerId lqMap
+          -- TODO Nothing here would seem to imply a bug
```
0x777: I'm assuming your question is why this isn't: `Just . _pThread <$> STM.takeTMVar ioState`? Honestly, I don't remember why `tryReadTMVar` is being used here, but looking at [this](https://github.com/hasura/graphql-engine/pull/3860/files#diff-5696cf6bf0d51987a7405e83f24d4a58R88) comment in `addLiveQuery`, this does look like a bug.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
0x777: > From what I recall that depends somewhat on internals of the websockets lib we use

Yes. The `onClose` handler will be run whenever a `WS.ConnectionException` is received, so if there are cases where this exception does not occur but the connection disappears, then `onClose` wouldn't be called.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -457,13 +458,15 @@ onMessage authMode serverEnv wsConn msgRaw =
...
       CMConnTerm        -> liftIO $ WS.closeConn wsConn "GQL_CONNECTION_TERMINATE received"
   where
     logger = _wseLogger serverEnv
 
 onStop :: WSServerEnv -> WSConn -> StopMsg -> IO ()
 onStop serverEnv wsConn (StopMsg opId) = do
-  -- probably wrap the whole thing in a single tx?
+  -- TODO probably wrap the whole thing in a single tx? IO in removeLiveQuery can be returned and
```
0x777: A transaction is not really needed because there cannot be any concurrent invocations of `onMessage` on each websocket connection. In addition, `onClose` is called after the thread which invokes `onMessage` is cancelled (in `Transport.WebSocket.Server` module).

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
0x777: Ignore my previous comment! I completely misunderstood what `removeLiveQuery` is doing. So the reason why this is concurrent is because of the assumption that each websocket connection would most likely have subscriptions that map to different pollers than the same, in such case, executing `removeLiveQuery` concurrently is better because `STMMap` can handle such cases really well. 

>  A transaction operating on the map will only ever have to retry if another transaction is operating on the same key at the same time.

I'm not particular about this at all, we can execute it serially too. I think the best solution would be to group the queries that need to be deleted based on the `PollerKey`, each such group can be processed concurrently (in each group we delete the queries sequentially). Regardless, we'll still have contention from other closing connections.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -52,7 +52,7 @@ import qualified Hasura.GraphQL.Transport.WebSocket.Server   as WS
 import qualified Hasura.Logging                              as L
 import qualified Hasura.Server.Telemetry.Counters            as Telem
 
-
+-- TODO what are the invariants here? How are they maintained?
```
0x777: `OperationMap` maintains the state of active subscriptions. When a connection is initialized it is empty, `onStart` and `onStop` can potentially add and remove operations and `onClose` cleans up the connection's operations.

jberryman:
@0x777 this is ready for final review

server/src-lib/Hasura/Server/App.hs
```diff
@@ -65,9 +66,11 @@ import qualified Hasura.Server.PGDump                   as PGD
 data SchemaCacheRef
   = SchemaCacheRef
   { _scrLock     :: MVar ()
+  -- ^ TODO consider removing in place of `_scrCache :: MVar ...` if it's okay
+  -- or in fact correct to block during schema update in e.g. _wseGCtxMap.
```
0x777: The idea behind explicit locking here is to 
1. Allow maximum throughput for serving requests (/v1/graphql) (as each request reads the current schemacache) 
2. We don't want to process more than one request at any point of time which would modify the schema cache as such queries are expensive.

0x777: > if it's okay or in fact correct to block during schema update in e.g. _wseGCtxMap

It is theoretically possible to have a situation (in between [building new schemacache](https://github.com/hasura/graphql-engine/pull/3860/files#diff-faa5431efd1e2b8a934e9a860b158b3dR131) and [before writing it to the IORef](https://github.com/hasura/graphql-engine/pull/3860/files#diff-faa5431efd1e2b8a934e9a860b158b3dR132)) where we serve a request with a stale schemacache but I guess it is an okay trade-off to pay for a higher throughput (I remember doing a bunch of benchmarks to test this hypothesis).

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/State.hs
```diff
@@ -107,16 +116,22 @@ addLiveQuery lqState plan onResultAction = do
...
     detM <- getQueryDet
     fmap join $ forM detM $ \(Poller cohorts ioState, cohort) ->
       cleanHandlerC cohorts ioState cohort
-  traverse_ A.cancel threadRef
+  case threadRefM of
+    Just threadRef -> Immortal.stop threadRef
+    -- This would seem to imply addLiveQuery broke, or a bug elsewhere:
+    Nothing -> L.unLogger logger $ L.UnstructuredLog L.LevelError $ fromString $
+      "In removeLiveQuery no worker thread installed. Please report this as a bug: "<>
+      show lqId
```
0x777: Might as well log a more structured message instead of using the show instance on `LiveQueryId`?

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/State.hs
```diff
@@ -107,16 +116,22 @@ addLiveQuery lqState plan onResultAction = do
...
+removeLiveQuery logger lqState lqId@(LiveQueryId handlerId cohortId sinkId) = do
+  threadRefM <- STM.atomically $ do
     detM <- getQueryDet
     fmap join $ forM detM $ \(Poller cohorts ioState, cohort) ->
       cleanHandlerC cohorts ioState cohort
-  traverse_ A.cancel threadRef
+  case threadRefM of
+    Just threadRef -> Immortal.stop threadRef
+    -- This would seem to imply addLiveQuery broke, or a bug elsewhere:
+    Nothing -> L.unLogger logger $ L.UnstructuredLog L.LevelError $ fromString $
```
0x777: `Nothing` could also mean several things:
1. There is no associated live query with the specified `lqId`.
2. The given live query is not the last live query of the handler (a handler can manage more than one live query). 
3. If `fmap _pThread <$> STM.tryReadTMVar ioState` returns a `Nothing`. This is the case where it is definitely a bug, but what I do not understand (or recall) is why is `tryReadTMVar` being used instead of `readTMVar`?

So if we get rid of the 3rd case, this log isn't needed as `Nothing` is a valid value?

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -457,19 +473,23 @@ onMessage authMode serverEnv wsConn msgRaw =
                            wsConn authMode params
       CMStart startMsg  -> liftIO $ onStart serverEnv wsConn startMsg
       CMStop stopMsg    -> liftIO $ onStop serverEnv wsConn stopMsg
+      -- The idea is cleanup will be handled by 'onClose', but...
+      --   TODO I don't think closeConn here is correct. Document what's going on here.
```
0x777: Do you have any ideas? This is what we have: we need to close the websocket connection when we receive the `CMConnTerm` message and calling `WS.closeConn` will definitely throw an exception.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/State.hs
jberryman: I know it's a bit ugly, but it didn't seem worth the effort for an unexpected/internal error. Let me know if you strongly object!

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
jberryman: My understanding looking at the docs was that `closeConn` is supposed to be a friendly way of initiating connection closed by server, so it seems weird to do this after getting a connection termination from client. But I don't know! I think best is to just add the info you commented and keep as TODO

server/src-lib/Hasura/Server/App.hs
jberryman: I'll add this additional information, thanks!

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/State.hs
jberryman: Good catch, thanks! There are some comments elsewhere that explains the awkwardness behind this var. I think it's a bit safer to log than to use `readTMVar` and risk blocking in the case of a bug we didn't anticipate. Would be great if you had another look at the new version when I push it

jberryman:
@0x777 hopefully addressed comments in latest, thanks!

jberryman:
Is thw down migration failure a known issue on master? I thought I heard chatter about it but can't find it now

jberryman:
@0x777 rebased on master with fixed tests. Should be ready to merge if it looks good to you.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/State.hs
0x777: This is incorrect: 
```haskell
            case threadRefM of
              Just threadRef -> Immortal.stop threadRef
              -- This would seem to imply addLiveQuery broke or a bug
              -- elsewhere. Be paranoid and log:
              Nothing -> L.unLogger logger $ L.UnstructuredLog L.LevelError $ fromString $
                "In removeLiveQuery no worker thread installed. Please report this as a bug: "<>
                show lqId
```
You can have `Nothing` in any of the 3 scenarios mentioned above but only in (3) is when we need to log this message (maybe even (1) too because we wouldn't be calling removeLiveQuery on a query that cannot exist) but (2) is a valid scenario where `Nothing` is expected.

jberryman: @0x777 hm were you looking at an old revision? I believe the revision I made after your first comment addresses the issue (we only log a warning in case (3))

0x777: @jberryman My bad!

jberryman:
@0x777 seems tests fail due to replaceMetadataToOrdJSON which I've noticed seems to consume a crazy amount of memory. I assumed that was just for me when I run it locally with `-O0`, but I might investigate and make a ticket.

Issue #877, opened 2018-10-27
Handling multiple roles on the graphql client (possibly in the same query)


revskill10:
- In an application, one user often have multi roles at the same time, at a screen.
Let's say, in a blog application, both editor and author could edit a post. So if a user is both an editor and an author, how to authenticate with hasura then ? 
In this case, hasura should support multi-roles in headers: `x-hasura-roles` instead of `x-hasura-role`.
Else how do you decide which role the user should use ? Because all roles are equal and valid.

- What i mean by multi-roles query is that.
```
query {
table1 {
 field1
}
table2 {
 field2
}
```
Suppose that user has roles role1 for table1, and role2 for table2 .
This is currently impossible to make above query.

- OK, what's if a websocket connection is made by above query. Now the user's role changed, what happen ?
The websocket connection should not be recreated for that. Instead the subscription data should reflect the change after that.

And this is the second hard part to solve this problem (when roles change for a subscription query)

- Suggested implementation (per @coco98 )
The query now should look like this
```
query {
  asUser {
    someTable {...}
  }
  asManager {
    someTable {...}
  }
}
```

## Note 2 : My proposed solution
In Console UI for permission setting, there must be ability to assign current permission to another roles in system (or creation new role).
Only then that, Hasura could validate one role or array of roles query to resolve correct data to return.

Maxpain177:
+1. Same problem

revskill10:
One more note: This feature allows you to put multi-roles queries in a HTTP graphql request.
For subscription to work, hasura needs one more thing:
- It requires a pool of websocket connections which could handle any roles that are sent to it.

go4cas:
@shahidhk ... my understanding of the current JWT role workflow is:

1. The token is decoded, and checked whether both `x-hasura-default-role` and `x-hasura-allowed-roles` are included
2. If both headers exist, it is checked whether the `x-hasura-default-role` is part of the roles defined in `x-hasura-allowed-roles`
3. If the `x-hasura-role` header is present, then this value is checked against the `x-hasura-allowed-roles` and it overrides `x-hasura-default-role` as the role to use
4. The role (either `x-hasura-default-role` or `x-hasura-role`) is then checked against the role defined on the permissions on the table(s) in the graphql query.

My suggestion (and it's really a question whether this is doable or not):
The `x-hasura-default-role` becomes optional, and the following is implemented:
- If both `x-hasura-default-role` **AND** `x-hasura-role` are present, use the workflow as described above
[...]

PR #3868, opened 2020-02-12
rfc: multiple roles for reads

tirumaraiselvan:
Can be previewed [here](https://github.com/tirumaraiselvan/graphql-engine/blob/multiple-roles-reads/rfcs/multiple-roles-reads.md). 

Related issue: https://github.com/hasura/graphql-engine/issues/877

### DIFF
--- a/rfcs/multiple-roles-reads.md
+++ b/rfcs/multiple-roles-reads.md
@@ -1,0 +1,178 @@
+## Multiple roles (for reads)
+
+Currently, each graphql operation is run with a single role that is provided in the `X-Hasura-Role` session variable. 
+This is a RFC to allow a list of roles to be provided (`X-Hasura-Roles`) for query/subscription operations. We will keep mutations out of scope for this RFC.
+
+Original issue: [#877](https://github.com/hasura/graphql-engine/issues/877)
+
+### Motivation
+
+Sometimes permission rules for an operation cannot be satisfactorily expressed with a single role because:
+
+  i) it is complex or hard to maintain,
+
+  ii) it is impossible to create rules because some data is not present in the db,
+
+  iii) you may want to fetch data with different permissions together.
+
+Examples of the above are as follows:
+
+i) **Complex permissions**:
+
+Suppose you are building a forum application like *Reddit*. You might have the following schema:
+
+- Tables:
+
+`posts (id, title, user_id, group_id, published`. A table for posts.
+
+`groups (id, name)`. A table for groups (like _subreddits_).
+
+`user_group_roles(user_id, group_id, role)`. A table where mapping of users with their group-level roles are defined. E.g. a row in `user_group_roles` like (1, 189, "moderator") would mean user with `user_id: 1` has "moderator" level access on group with `group_id: 189`.
+
+- Relationships:
+
+Each `post` has an object relationship with `groups`.
+
+Each `group` has an array relationship with `user_group_roles`.
+
+For this application, you might create a role **user** with select rules on table `posts` like so:
+
+```yaml
+_or:
+- published:                   # All posts that are published
+    _eq: true            
+- user_id:                     # All posts that belong to me
+    _eq: X-Hasura-User-Id
+- group:                       # All posts that belong to the group that I am a moderator of
+    user_group_roles:
+      _and:
+      - role: moderator
+      - user_id: X-Hasura-User-Id
+```
+
+This is a simple example but it is easy to see that for a more complex application defining all role accesses into one rule will be hard to manage. 
+
+A single role approach like above also make your permissions less *DRY*. For e.g. suppose you want to add a new role in your application called **editor** with select rules as follows:
+
+```yaml
+_or:
+- published:                   # All posts that are published
+    _eq: true            
+- user_id:                     # All posts that belong to me
+    _eq: X-Hasura-User-Id
+```
+Now if I have to change the access rule for my application for say all published posts, I would have to do this for both **user** and **editor** roles.
+
+ii) **Data not present in same postgres**:
+
+Suppose you have a table:
+
+`books (id, title, content, author_id, publisher_id)`
+
+but `publisher_id` above does not have any related data in the same postgres (say, it is present in some other source).
+
+You want a user to have read access to all rows for which they are the author or publisher. With single role, you might have a role **user** which uses two attributes `X-Hasura-Author-Id` and `X-Hasura-Publisher-Id` like this:
+
+```yaml
+_or:
+- author_id:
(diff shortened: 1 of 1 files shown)


pnappa:
Hi,

I've implemented a similar system for my Hasura instance, yet took the approach from the other direction. I defined the permissions for individual roles, then generated each valid combination of roles by unioning the columns and forming a disjunction cross the permission queries (i.e. OR[...] for each of the roles). See picture (I made a templating language to make the permission query definitions more DRY conformant (especially across different tables, many queries were shared).

spreadsheet definitions

Now, this is identical to the RFC, yet implemented outside of the Hasura engine (except mine extends to work for mutations too), and there is a limitation that isn't mentioned.

Consider a user of role A, who can access the columns X and Y in a table with permission query {}. Users of a role B can access the columns X, Y, Z with permission IS_USER (e.g. if this was a user table, a user can access additional columns for only _their_ row). Now, as the union of all columns is X,Y,Z between roles, but the disjunction of the permissions is OR[{}, IS_USER], i.e {} - this means that the user is able to access Z for rows that aren't theirs.

There is a workaround which is a bit dodgy, which requires "permission normalisation", so moving data into a foreign-keyed object. I believe there are more complex permissions that can't be expressed at the moment, due to limitations of _ceq expressions across tables.

mnesarco:
Hi Friends, 

I am not an expert in the topic, but I think that any real RBAC system should separate Permissions from Roles as different entities.

hp

Roles are a mere combination of permissions, and roles can be combined resulting again
in a list of permissions (derived role).

Hasura exposes a schema based on permissions, so the thing is that the schema should be
generated based on a list of roles that can be decomposed to permissions and merged together.

As Hasura only have Grant permissions (not Deny permissions), it is just additive.

```
ExposedSchema : (roles: [Role]) => buildSchema( MergePermissions( roles.flatMap(role -> role.permissions) ) )

MergePermissions : (perms : [Permisssion]) =>
    - Add allowed columns
    - Add aggretation permissions
    - Add column presets
    - Add (with OR) custom checks
```

Of course merges are type aware (Insert permissions are merged only with insert permissions... and so)

unfernandito:
Can i use this feature in a test environment?

What is the tag of this docker image or i should be clone the repo and compile in my machine?

Thanks.

c-spencer:
I think additive roles, as described, will be great and they seem like the last hurdle, at least for my current use cases, for allowing flexible-enough auth to cover a full multi-tenancy style application.

My one concern with this is the interplay with how schemas are generated based on roles, and whether that would still leave complexity in client logic to make sure they only query 'existing' (from their perspective) fields. As noted in the RFC, this introduces the concept of 'columns I can query but won't necessarily get a value for' (due to lacking the particular column permission on that row, but having it on others).

If you have complicated permissions, then you may end up with a lot of re-usable queries (efficient in both developer time, caches, subscriptions, etc) but have difficulty to know ahead of time which columns a given client might have access to. As a contrived example, you might generally want to show who last modified something, except for anonymous users who can never see. It would be clean to re-use the same queries across both anonymous and logged in users (of various roles), but doing so would cause the anonymous users to throw an error (as they cannot see the given field).

A proposal would be allowing columns to be marked as 'visible in schema, but not queryable' as a middle ground on permissions. I know this isn't directly associated with multiple roles, but I think is a knock-on effect of allowing better ergonomics around more complicated auth.

oliverpool:
As far as I understand, the current RFC proposes the role to be chosen dynamically by the backend depending on the query.

**As an alternative, you could consider the role to be chosen by the frontend (I chose `acting_as`, but there might be better syntax):**

The backend will always choose the first role of the header `X-Hasura-Roles: [“editor”, “publisher”]`, except when otherwise specified in the query:
```
query {
  author(acting_as:"moderator") {
    author_id
  }
  publisher(acting_as:"publisher") {
    publisher_id
  }
} 
```

Pros:
- it prevents the issue raised by @pnappa (incompatible row and columns permissions depending on the role)
- the frontend can get different result in different contexts
- there is no complex resolution in case of conflicting roles

Cons:
- if you need to see your posts and the public posts together, the frontend must merge the results of the queries
```
query {
  public_posts: posts(acting_as:"viewer") {
  }
  my_posts: posts(acting_as:"author") {
  }
} 
```

Bessonov:
Hey @tirumaraiselvan, thank you for your proposal. Is there any progress on it? I have two applications. ATM I introduce hasura/yugabyte to the smaller hobby application. This application has a very lightweight requirements for the roles: the admin, the owner and the anonymous role. That's fine. But hasura would fail at the bigger app at the work, because the current role model doesn't support more comprehensive permissions. In fact, there are no roles in the application, but permissions. The permissions are compounded to roles outside of the application through active directory.

Some use cases:

1. It have an area for loyalty management. Current permission model is simple and consist of three permissions:

- manage all campaigns
- read access to all campaigns
- take actions like redeem voucher

Although the permission model is very simple, most of the employees has multiple permissions in different areas. For example marketing staff has access to all campaigns, but also to marketing and sales channels. I don't bother if it called "roles" or "permissions" and I propose to handle only the lowest of them in this proposal. Later we can discuss about hierarchical structures,if they are needed at all.

2. There is a very flexible permission model for access to customer data. We call it "data partitions". I think current requirements would be handled with this proposal + rows- and column-level permissions described in the documentation. (The model supports cell-level permissions, where the access can be defined for every cell in the table on it's own, but I hope it would be never needed).

There are other models, but I think they can be covered by use cases above.

> - Columns are unions of columns for each role.

I hope I don't understand it correctly, but does it mean that column level permissions are concatenated independent of row-level permissions (versus coalesce the data while merging)? Indeed, we employ this for the second use case. It makes some things easier to **manage**. But well, I don't think that this model is "right" for most applications and would introduce an inflexibility to hasura. For example, imagine a table:

| user | phone | mail | contact_by |
| ------ | -------- | ----- | --------------- |
| Max | 0123 | a@b.c | phone |
| Tim | 4567  | e@f.g | mail |

Now you want to limit the access to the information depending on staff role in sales department, for example for "operators" and "writers". Now the operators can access to phone, but not the mail of Max and the writes can access to the mail, but not the phone of Tim. Now you want to have a staff member with both roles. BAM - the staff member get access to all data.

> - Duplicate rows are merged appropriately, with NULLs as values for non-permissive columns.

What does it means for the client? From my point of view, `null` means "there is no value" and `undefined`/missing means "I don't know". A `null` would be very misleading in case of missing permissions. But I'm not sure that error would be better in this case. May be omit the filed? But AFAIK it go against the graphgl spec.

tirumaraiselvan:
This has been released as Inherited Roles in `v2.0.0-alpha.4`

Issue #1700, opened 2019-03-06
UI for adding field-level validations to mutations


tirumaraiselvan:
Currently there are multiple ways to perform validations before mutations:

1) add check constraints on postgres explicitly
2) create postgres triggers
3) remote-schemas 

Many use-cases can be solved using 1 and 2 but its not easy to use them currently. 

We can solve the problem of 1) more efficiently: In the data tab, we should be able to "select" check constraints during table create/modify. We can keep this similar to permissions UI but underneath it will translate to a simple "ADD CONSTRAINT" query.

mnlbox:
Woooo, @tirumaraiselvan I love this idea man :wink: :heart:

yasinarik:
This can be extended per suggested in #3536

shahidhk:
closed with https://github.com/hasura/graphql-engine/commit/71240f310dff17e876bd1bdcf28cddf7871461c1

PR #3881, opened 2020-02-13
console: manage postgres check constraints

beerose:
### Description

There was added a component that allows adding check constraints while creating a new table in the same way as it can be done on the `Modify` view. 

Screenshot 2020-03-10 at 11 54 42

### Affected components

- [x] Console

### Related Issues
https://github.com/hasura/graphql-engine/issues/1700


beerose:
What should be the content of this tooltip?

rikinsk:
> What should be the content of this tooltip?
A check constraint allows you to specify if the value in a certain column must satisfy a specific condition

console/src/components/Services/Data/Add/CheckConstraints.js
```diff
@@ -0,0 +1,149 @@
...
+        <div>
+          <b>{existingConstraintName}</b>
+        </div>
+      );
+    };
+
+    const expandButtonText = isLast ? 'Add a check constraint' : 'Edit';
+
+    const expandedContent = () => {
+      return (
```
wawhal: We should pull this out into a reusable component. Because the same UI is replicated on the `Modify` page as well. What do you think?

console/src/components/Services/Data/Add/Tooltips.js
```diff
@@ -33,3 +33,10 @@ export const uniqueKeyDescription = (
 export const dataTypeDescription = description => (
   <Tooltip id="tooltip-datatype-description">{description}</Tooltip>
 );
+
+export const checkConstraintsDescription = (
```
wawhal: Same. This tooltip should be reused between `Add` and `Modify`.

beerose: I did a small refactor, because in `Modify` we are using component [`ToolTip`](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Common/Tooltip/Tooltip.js), which is more convenient, so I replaced bootstrap `OverlayTrigger` and `Tooltip` with our `ToolTip` component on `Add`.

tirumaraiselvan:
If I try to add a constraint which has spaces for e.g. `name is long` then the constraint is unquoted in the resulting SQL and it throws an error. The constraint name should be quoted.

shahidhk:
Changelog for this PR is added in #4079 feel free to copy the same:

### Manage Postgres Check Constraints from Console

Postgres Check constraints allows you to specify that the value in a certain column must satisfy a Boolean (truth-value) expression. They can be used to put in simple input validations for mutations and with this release, these constraints can now be added while creating a table or later from Modify tab on the console.

**Example**:
When a product is created, ensure that the price is greater than zero. The SQL would look like this:
```sql
CREATE TABLE products (
    product_id UUID DEFAULT gen_random_uuid(),
    name TEXT,
    price NUMERIC CONSTRAINT positive_price CHECK (price > 0)
);
```
To create this table with Hasura Console, on the 'Add a new table' screen, after adding all the columns, scroll down to 'Check constraints' section and 'Add a new check constraint' with the following properties:
- Constraint name: `positive_price`
- Check expression: `price > 0`

Read more about check constraints on [Postgres Docs](https://www.postgresql.org/docs/12/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS).

(close #1700) (#3881)

PR #3883, opened 2020-02-13
Update permission.rst for check constraint API

paf31:
### Description

I missed this in my PR for #384. I noticed the existing docs were incorrect, so I fixed a few mistakes too. Namely:

- There was a reference to `check` which didn't exist previously, presumably copied from the insert API docs.
- The `set` example didn't really make sense since you wouldn't want to set the article ID to a user ID value.

### Affected components

- [x] Docs

#### Catalog upgrade

- [x] No

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


docs/graphql/manual/api-reference/schema-metadata-api/permission.rst
```diff
@@ -349,20 +349,27 @@ An example:
...
+                   "updated_at" : "NOW()"
                }
            }
        }
    }
 
 This reads as follows - for the ``user`` role:
 
-* Allow updating only those rows where the ``check`` passes i.e. the value of the ``author_id`` column of a row matches the value of the session variable ``X-HASURA-USER-ID`` value.
+* Allow updating only those rows where the ``filter`` passes i.e. the value of the ``author_id`` column of a row matches the value of the session variable ``X-HASURA-USER-ID`` value.
```
marionschleifer:
```suggestion
* Allow updating only those rows where the ``filter`` passes i.e. the value of the ``author_id`` column of a row matches the value of the session variable ``X-HASURA-USER-ID``.
```

docs/graphql/manual/api-reference/schema-metadata-api/permission.rst
```diff
@@ -349,20 +349,27 @@ An example:
...
+
+* If the above ``filter`` passes for a given row, allow updating only the ``title``, ``content`` and ``category`` columns (*as specified in the* ``columns`` *key*).
 
-* If the above ``check`` passes for a given row, allow updating only the ``title``, ``content`` and ``category`` columns (*as specified in the* ``columns`` *key*).
+* After the update happens, verify that the ``check`` condition holds for the updated row, i.e. that the value in the ``content`` column is not empty.
```
marionschleifer:
```suggestion
* After the update happens, verify that the ``check`` condition holds for the updated row i.e. that the value in the ``content`` column is not empty.
```

docs/graphql/manual/api-reference/schema-metadata-api/permission.rst
```diff
@@ -349,20 +349,27 @@ An example:
...
 
-* When this update happens, the value of the column ``id`` will be automatically ``set`` to the value of the resolved session variable ``X-HASURA-USER-ID``.
+* When this update happens, the value of the column ``updated_at`` will be automatically be ``set`` to the current timestamp.
```
marionschleifer:
```suggestion
* When this update happens, the value of the column ``updated_at`` will be automatically ``set`` to the current timestamp.
```

PR #3888, opened 2020-02-14
show pre-release notifications on console with opt out option

rikinsk:
### Description

Show update notifications for pre-release versions with a disclaimer and option to opt out of pre-release notifications

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

If user opts out, persist choice in the DB via `console_state` field of `hdb_version`

### DIFF
--- a/console/src/components/Common/utils/localStorageUtils.js
+++ b/console/src/components/Common/utils/localStorageUtils.js
@@ -1,0 +1,12 @@
+export const getLocalStorageItem = key => {
+  return window.localStorage.getItem(key);
+};
+
+export const setLocalStorageItem = (key, value) => {
+  window.localStorage.setItem(key, value);
+};
+
+/** Local storage keys **/
+
+export const LS_VERSION_UPDATE_CHECK_LAST_CLOSED =
+  'versionUpdateCheck: lastClosed';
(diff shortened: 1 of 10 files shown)


console/src/components/Common/utils/localStorageUtils.js
```diff
@@ -0,0 +1,12 @@
+export const getLocalStorageItem = key => {
```
beerose: We also have utility functions for local storage here: https://github.com/hasura/graphql-engine/blob/master/console/src/utils/localstorage.js,
and they are not used anywhere. Maybe we can get rid of them, so there aren't multiple local storage utils files?

rikinsk: @beerose That is true. This one seems to be a better implemementation of the functions but seems to be unused anywhere in the codebase as of now. Can you approve this for now. I'll raise another PR right after this with a refactor of the localStorage utils usage across the code.

beerose: Ok. Cool.

PR #3892, opened 2020-02-15
allow special characters in json path's property name (close #3890)

### LINKED ISSUE #3890: When JSON object key start with "_", Hasura returns invalid error.
Hi, I try to search JSONB data.
It includes following data.
`{ 
   "hello": "world",
  "_hello": "_world" 
 }`

I executed query including the key starts with "_" on Hasura console.
`
query MyQuery {
  test1 {
    data(path: "$._hello")
  }
}
`

I received following response.
`
{
  "errors": [
    {
      "extensions": {
        "path": "$.selectionSet.test1",
        "code": "parse-failed"
      },
      "message": "parse json path error: the first character of property name must be a letter."
    }
  ]
}
`

So I confirmed JSON specification [RFC8259](https://tools.ietf.org/html/rfc8259).
The document says following.
I suppose the first character of JSON object key doesn't have any constraints without unescaped letter, but Hasura cannot use "_".

Is it a constraint Hasura or GraphQL?
Or is there any workaround?

      4. Objects
      object = begin-object [ member *( value-separator member ) ]
               end-object
      member = string name-separator value
      
      7. String
      string = quotation-mark *char quotation-mark
      char = unescaped /
          escape (
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
              %x75 4HEXDIG )  ; uXXXX                U+XXXX
      escape = %x5C              ; \
      quotation-mark = %x22      ; "
      unescaped = %x20-21 / %x23-5B / %x5D-10FFFF

Thanks.

hgiasac:
### Description

Current JSON path parser only allow alphabet as first character and restrict special characters such as space, #!@... 

Because Postgres JSON type doesn't restrict anything. Moreover, in real world, there are many developers use object as dictionary for translation, etc
  
### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3890

### Solution and Design

For dot notation `.`, we still restrict special characters, only allow `_` at beginning and `-` in remain positions

```javascript
abc.xyz 
abc._xyz 
abc.-xyz // bad
abc.0ab // bad
```
For special character name, we accept in squared bracket notation that was allowed for integer only. 

```javascript
[Hello world!]
address[$street]
dummy[0][!@#$$]
```

### Steps to test and verify

Test case included
 
#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


0x777:
@hgiasac You'll need to update changelog.

hgiasac:
@0x777 changelog updated

docs/graphql/manual/api-reference/graphql-api/query.rst
```diff
@@ -120,6 +120,7 @@ E.g.
       address(path: "city") # scalar JSON field -> property; '$.' prefix is optional
       contacts(path: "[0]") # scalar JSON field -> array_item
       contacts(path: "[0].phone") # scalar JSON field -> array_item_property
+      contacts(path: "[Hello world!]") # scalar JSON field -> property; used for special characters key
```
lexi-lambda: This syntax doesn’t seem quite right to me. Both the `a` and `[0]` syntax reflect the syntax used in JavaScript for accessing properties:

```js
> let obj = { a: 1, b: 2 }
> obj.a
1
> let arr = [1, 2]
> obj[0]
1
```

But `[Hello, world!]` isn’t valid JavaScript syntax. The correct syntax should be `["Hello, world!"]`:

```js
> let obj = { ["Hello, world!"]: 1 }
> obj["Hello, world!"]
1
```

Leaving out the quotes actually creates an ambiguity in the grammar, since it makes it impossible to access a numeric object field such as this one:

```json
{ "0": "some value" }
```

So we should use an extra layer of quoting here, just like JavaScript does.

hgiasac: @lexi-lambda I agree with your opinion. However, we can't use quote `"`. It will be harder to read and parse, because GraphQL argument use it already `(path: "abc")`. It will break GQL specs.

We can use single quote `'` for this case. JSON path also use this style.
https://github.com/json-path/JsonPath
`$['store']['book'][0]['author']`

lexi-lambda: [GraphQL string literals support escaping.](http://spec.graphql.org/June2018/#StringValue) Moreover, the parsing logic should be exactly the same amount of effort, since the `graphql-parser` library already does the work of parsing GraphQL string syntax, so escapes have already been handled at that phase. Once we get the string in `graphql-engine`, the backslashes will already be gone.

That said, your point about it being harder to read and write for humans is valid. Writing `"$['store']"` looks nicer than writing `"$[\"store\"]`. I picked `"` originally because JSON supports `"` but not `'`, so it seemed more appropriate, but I don’t see any reason not to support `'` as well. Let’s just support both—it’s trivial to do if we’ve already supported one!

hgiasac:
@lexi-lambda updated change requests. Please check, thanks

hgiasac:
@lexi-lambda LGTM

lexi-lambda:
@hgiasac Could you please resolve the merge conflicts and fix the failing `test_server_upgrade` job? It looks like the updated test needs a `@pytest.mark.skip_server_upgrade_test`.

@hasura/hge-console-owners Could one of you review this PR? It would be great to get it merged.

hgiasac:
@lexi-lambda updated

hgiasac:
There are several improvements:
1. Friendly error message 
2. Add more test cases
3. JSONPath Encoder fixes:
- Redundant dot with special characters `$.['!@#'].['abc']`
- First character only allows alphabet and `-_`. Numbers and special characters are disallowed

PR #3895, opened 2020-02-16
adds postgres password to docker-compose setup (fix #3894)

### LINKED ISSUE #3894: docker-compose file for use in the quickstart docs for docker-compose missing postgres password
from a comment i made on the discord, here's the error you get when you use the present version of the docker-compose.yaml from the docs here - https://docs.hasura.io/1.0/graphql/manual/getting-started/docker-simple.html

```
postgres_1        |        You may also use POSTGRES_HOST_AUTH_METHOD=trust to allow all connections
postgres_1        |        without a password. This is *not* recommended. See PostgreSQL
postgres_1        |        documentation about "trust":
postgres_1        |        https://www.postgresql.org/docs/current/auth-trust.html
postgres_1        | Error: Database is uninitialized and superuser password is not specified.
postgres_1        |        You must specify POSTGRES_PASSWORD for the superuser. Use
postgres_1        |        "-e POSTGRES_PASSWORD=password" to set it in "docker run".
```

fix is just to add the password to the env section.

silky:
# Description

This just adds the `POSTGRES_PASSWORD` env var to the docker-compose file linked in the docs [here](https://docs.hasura.io/1.0/graphql/manual/getting-started/docker-simple.html).

### Affected components
- [x] Docs

### Solution and Design

- Add the env var. 

### Steps to test and verify

- To verify the bug, see the #3894 (just copy the docker-compose file and run it)

I have the following docker and docker-compose versions:

```
> docker --version
Docker version 17.12.0-ce, build c97c6d6
> docker-compose --version
docker-compose version 1.25.4, build unknown
```

-- edit:

feel free to reject if this isn't necessary; i imagine this file worked for someone once; maybe the postgres image should be tied to a version, instead?


marionschleifer:
@shahidhk added the password to the other `docker-compose` files too and added the Postgres version. Could you please review again? 🙂

silky:
@marionschleifer @shahidhk since we're in the mood to set explicit versions, i added the explicit version for the mdillion/postgis image as well.

note that it's only at postgres 11, so maybe there's some quirkyness about that

maybe that image could be switched to a postgis that's based on postgres-12?

Issue #660, opened 2018-10-05
Cannot use serializable mode in a hot standby


rijnhardtkotze:
I am following the instructions from [the official documentation](https://docs.hasura.io/1.0/graphql/manual/deployment/heroku/using-existing-heroku-database.html).

I am trying to connect Hasura to a follower database, it serves the following error: "cannot use serializable mode in a hot standby" 

Here is the exception: 

    {  
      "internal":{  
          "statement":"BEGIN ISOLATION LEVEL SERIALIZABLE ",
          "prepared":true,
          "error":{  
             "exec_status":"FatalError",
             "hint":"You can use REPEATABLE READ instead.",
             "message":"cannot use serializable mode in a hot standby",
             "status_code":"0A000",
             "description":null
          },
          "arguments":[  

          ]
      },
      "path":"$",
      "error":"postgres tx error",
      "code":"postgres-error"
    }

Any ideas?

0x777:
@rijnhardtkotze We haven't tested graphql-engine with a Hot Standby Postgres instance. It might be possible to connect to a Hot Standby instance after initializing the state with the master instance. We'll get back to you on this.

rijnhardtkotze:
Thanks, @0x777 

Please let me know if I can be of any assistance.

paulmwatson:
Any progress on this please, having the same issue.

vinay0410:
Hi @0x777 ,
Is there a workaround to connect if this exception pops up.

Iris-Qian:
Any progress on this?

guibernardino:
+1

jocrau:
I tried setting `HASURA_GRAPHQL_TX_ISOLATION=repeatable-read` but that didn't help because the initialization always happens at `SERIALIZABLE` level.

This is currently a blocker for using Hasura at our company. We have a high ingest rate on a Postgres master and read only from replicas.

ttelfer:
Just ran into this problem as well. We have actually been using Hasura in essentially a read-only for generating dashboard frontend with no user inputed data. The problem we have is similar to ours is have an extremely high ingest rate and due to this we need to separate our reads. Has anyone looked at modifying the isolation level to repeatable read?

cheslip:
Seeing this issue in Azure when setting up Postgres read replicas for a backup hasura environment.

ttelfer:
[...]

PR #3901, opened 2020-02-18
Refactor initialisation and relax tx isolation levels where possible

tirumaraiselvan:
### Description

Currently, we are using `Serializable` tx isolation levels for all system processes. This is not required as most (one exception is catalog migration) of the system processes are either single statement read or write i.e (no repeated reads). Reducing the tx isolation level of such processes is generally good. 

### Affected components

- [x] Server

### Related Issues

https://github.com/hasura/graphql-engine/issues/660

### DIFF
--- a/server/src-lib/Hasura/App.hs
+++ b/server/src-lib/Hasura/App.hs
@@ -8,14 +8,14 @@ import           Control.Monad.Stateless
 import           Control.Monad.STM                    (atomically)
 import           Control.Monad.Trans.Control          (MonadBaseControl (..))
 import           Data.Aeson                           ((.=))
-import           Data.Time.Clock                      (UTCTime, getCurrentTime)
+import           Data.Time.Clock                      (UTCTime)
 import           GHC.AssertNF
 import           Options.Applicative
 import           System.Environment                   (getEnvironment, lookupEnv)
 import           System.Exit                          (exitFailure)
 
-import qualified Control.Concurrent.Extended          as C
 import qualified Control.Concurrent.Async.Lifted.Safe as LA
+import qualified Control.Concurrent.Extended          as C
 import qualified Data.Aeson                           as A
 import qualified Data.ByteString.Char8                as BC
 import qualified Data.ByteString.Lazy.Char8           as BLC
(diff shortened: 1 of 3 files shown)


server/src-lib/Hasura/App.hs
```diff
@@ -211,6 +213,9 @@ runHGEServer ServeOptions{..} InitCtx{..} initTime = do
 
   authMode <- either (printErrExit . T.unpack) return authModeRes
 
+  schemaCache <- onNothing _icRebuildableSchemaCache
+                 (printErrExit "expected schema cache to be initialised")
```
lexi-lambda, line 217: I would probably just pass the schema cache as a separate argument to this function rather than include it in the `InitCtx` wrapped in a `Maybe`. That way the type system can prevent this error from ever happening.

tirumaraiselvan: The schema cache is now part of `InitCtx{..}` and is of type `Maybe (RebuildableSchemaCache Run)`. It is is a `Maybe` because we get `InitCtx{..}` from `initialiseCtx` which is used in operations which do not require the schema cache to be initialised like `HCExport`, `HCClean`.

We can take schema cache outside of `InitCtx{..}` ofcourse but I felt it is better to be here.

Thoughts?

server/src-lib/Hasura/Server/App.hs
```diff
@@ -526,6 +514,30 @@ mkWaiApp isoLevel logger sqlGenCtx enableAL pool ci httpManager mode corsCfg ena
     getTimeMs :: IO Int64
     getTimeMs = (round . (* 1000)) `fmap` getPOSIXTime
 
+    initialiseSchemaCache :: m (E.PlanCache, SchemaCacheRef, Maybe UTCTime)
```
tirumaraiselvan, line 517: Can we rename this to reflect that this also migrates catalog (if required)? Or split the catalog migration into a different top-level function?

The reason being at some point (in near future) we might want to run a check if catalog migration is required (using a different txiso level) and then skip it (that will solve #660)

lexi-lambda: Renamed to `migrateAndInitialiseSchemaCache`.

PR #3902, opened 2020-02-18
TypeScript setup

beerose:
This PR adds TypeScript to the console 🎉

### Affected components

- [x] Console


console/package.json
```diff
@@ -101,7 +103,9 @@
     "redux-thunk": "^2.2.0",
     "sanitize-filename": "^1.6.1",
     "semver": "5.5.1",
+    "source-map-loader": "^0.2.4",
```
wawhal: Can't these go into dev dependencies? They won't be needed in the prod bundle right?
Just curious.

console/src/components/Common/Button/Button.tsx
```diff
@@ -9,7 +10,12 @@ import styles from '../Common.scss';
                and not the ones that change the appearance (color, font, size) of the button
 */
 
-const Button = props => {
+export interface ButtonProps extends React.ComponentProps<'button'> {
+  size: string;
+  color: 'yellow' | 'red' | 'green' | 'gray' | 'white' | 'black';
```
wawhal: Yay. Typed languages are so helpful for things like these :D

console/package.json
```diff
@@ -101,7 +103,9 @@
...
     "subscriptions-transport-ws": "^0.9.12",
+    "typescript": "^3.7.5",
```
wawhal: We'll need `tslint` as well? Or `eslint` does the job?

console/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "compilerOptions": {
+    "lib": ["es6", "dom", "es2017"],
+    "checkJs": true,
+    "allowJs": true,
+    "jsx": "react",
+    "allowSyntheticDefaultImports": true,
+    "strict": true,
+    "esModuleInterop": true,
+    "noEmitOnError": false
+  },
```
wawhal: Should we also do `noExplicitAny`?

console/package.json
beerose: Eslint does the job.

console/package.json
beerose: It can. Will update this.

console/tsconfig.json
beerose: Sometimes it's okay to use `any`. Example:
```js
declare const log: <T>(x: T) => void;
```
The function above is not generic, `T` is not used anywhere else than as the type of `x`, so it can as well be `any` (which is more accurate since we definitely don't care about this type).

console/package.json
wawhal: Hey! All the typescript related dependencies I meant. We don't need them in production right?

beerose: Updated!

rajanand02:
@beerose Shall I migrate the cypress setup and tests to TypeScript?

beerose:
Go ahead @rajanand02!

PR #3917, opened 2020-02-20
Don't send empty strings if field was empty (close #3332)

### LINKED ISSUE #3332: GUI forces to fill in a non-nullable field on 'Insert Row' tab even if there is BEFORE INSERT trigger
My BEFORE INSERT trigger will correctly fill in this field, but I can't insert a new row via GUI.
I don't think, that there is need any pre-validation before processing request to Postgre.

Version: v1.0.0-beta.9

beerose:
Solves #3332.

### Affected components

- [x] Console


rikinsk:
Closing this as we need a new solution for this and its not high priority. We can pick this up later.

PR #3927, opened 2020-02-24
Reserved keywords in column references break parser (fix #3597)

### LINKED ISSUE #3597: Field naming starting with reserved keywords breaks updates
When in a table there is a column that starts with a reserved keyword, for example nullSomeColumnName or trueSomeColumnName, if you try to create an update with an on_conflict and columns, the update will fail like:
```
extensions": {
        "path": "$.selectionSet.insert_some_table.args.on_conflict.update_columns[indexOfColumnThatStartsWithNull]",
        "code": "validation-failed"
      },
      "message": "null value found for non-nullable type: some_table_update_column!"

```
If the field starts with true or false the error is almost the same but the message is related to value must be an enum.

It seems like the engine is trying to parse the column names somehow.

sordina:
### Description

This pull-request is intended to fix issues related to #3597 "Field naming starting with reserved keywords breaks updates ".

The bug is located in the parser, therefore the fix is as follows:

* [x] Fix bug in parser
* [x] Add tests to engine
* [x] Update project.cabal to reference new version of the parser

```
source-repository-package
  type: git
  location: https://github.com/hasura/graphql-parser-hs.git
  tag: 1380495a7b3269b70a7ab3081d745a5f54171a9c
```

### Affected components

- [x] Server
- [x] Tests

### Related Issues

* #3597

### Solution and Design

The solution was to add more nuance to the parser as can be seen in the related parser pull request: https://github.com/hasura/graphql-parser-hs/pull/21

This is however only a temporary fix as the parser should probably be split into lexer/parser phases to much more effectively express the tokenisation of such identities as caused the problem here. Also this would make the code a lot more readable as it was quite difficult to narrow down the bug for a new contributor to the codebase (me!).

### Steps to test and verify

Run `pytest` against the new and old versions to see the fix addressed:

> pytest ... -vv test_graphql_mutations.py::TestGraphqlInsertNullPrefixedColumnOnConflict

You can also reproduce manually in the console if preferred.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/tests-py/test_graphql_mutations.py
+++ b/server/tests-py/test_graphql_mutations.py
@@ -188,6 +188,18 @@ class TestGraphqlInsertConstraints:
 
 
 @pytest.mark.parametrize("transport", ['http', 'websocket'])
+@usefixtures('per_class_tests_db_state')
+class TestGraphqlInsertNullPrefixedColumnOnConflict:
+
+    def test_address_not_null_constraint_err(self, hge_ctx, transport):
+        check_query_f(hge_ctx, self.dir() + "/null_prefixed_column_ok.yaml")
+
+    @classmethod
+    def dir(cls):
+        return "queries/graphql_mutation/insert/nullprefixcolumn"
+
+
+@pytest.mark.parametrize("transport", ['http', 'websocket'])
 @use_mutation_fixtures
 class TestGraphqlInsertGeoJson:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug fixes and improvements
 
 - docs: add One-Click Render deployment guide (close #3683) (#4209)
+- server: reserved keywords in column references break parser (fix #3597) #3927
 
 ## `v1.2.0-beta.3`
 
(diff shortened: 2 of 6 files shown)


sordina:
@lexi-lambda do you want to review, or should I assign to someone else?

lexi-lambda:
@0x777 Has been doing most of the server reviews temporarily while I’ve been working on this refactoring project, so he’s probably the one to review this.

server/tests-py/test_graphql_mutations.py
```diff
@@ -184,6 +184,18 @@ def dir(cls):
         return "queries/graphql_mutation/insert/constraints"
 
 
+@pytest.mark.parametrize("transport", ['http', 'websocket'])
+@usefixtures('per_class_tests_db_state')
+class TestGraphqlInsertNullPrefixedColumnOnConflict:
```
paf31: I think this could probably have been added to one of the existing insert classes, since we can setup multiple unrelated tables for different tests in one class setup file. I don't think we need to change it now though.

sordina:
Hi @lexi-lambda, is there anything you'd like me to do on this one before it can be reviewed for merge?

PR #3930, opened 2020-02-24
fix new release notification logic

rikinsk:
### Description

While displaying new release notifications on console, it was assumed that the prerelease version would be always >= the stable release version. It actually only returns the latest prerelease version. This fixes the invalid assumption

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### DIFF
--- a/console/src/components/Main/Main.js
+++ b/console/src/components/Main/Main.js
@@ -95,12 +95,11 @@ class Main extends React.Component {
       !console_opts || !console_opts.disablePreReleaseUpdateNotifications;
 
     let latestServerVersionToCheck;
-    if (allowPreReleaseNotifications) {
-      if (versionGT(latestPreReleaseServerVersion, latestStableServerVersion)) {
-        latestServerVersionToCheck = latestPreReleaseServerVersion;
-      } else {
-        latestServerVersionToCheck = latestStableServerVersion;
-      }
+    if (
+      allowPreReleaseNotifications &&
+      versionGT(latestPreReleaseServerVersion, latestStableServerVersion)
+    ) {
+      latestServerVersionToCheck = latestPreReleaseServerVersion;
     } else {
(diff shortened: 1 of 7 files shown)


console/src/components/Main/Main.js
```diff
@@ -86,17 +86,25 @@ class Main extends React.Component {
...
+    if (allowPreReleaseNotifications) {
+      latestServerVersionToCheck = versionGT(
+        latestPreReleaseServerVersion,
+        latestStableServerVersion
+      )
+        ? latestPreReleaseServerVersion
+        : latestStableServerVersion;
+    } else {
+      latestServerVersionToCheck = latestStableServerVersion;
+    }
```
beerose: One small suggestion. This ternary looks less readable since it's multiline. 

```js
    let latestServerVersionToCheck;
    const isVersionGT = versionGT(latestPreReleaseServerVersion, latestStableServerVersion);
    if (allowPreReleaseNotifications && isVersionGT) {
      latestServerVersionToCheck = latestPreReleaseServerVersion
    } else {
      latestServerVersionToCheck = latestStableServerVersion;
    }
```

rikinsk: Yeah. This was auto-formatted. I genuinely hate this auto-formatiing of code that happens. Would really like to disable it. Makes a lot of readable code much less readable.

Will make this an if-else

beerose: Sorry to be annoying but I still don't like that these parts are the same 😛
Screenshot 2020-02-24 at 12 41 15

rikinsk: Done

rikinsk: > Sorry to be annoying but I still don't like that these parts are the same

Fixed :)

PR #3933, opened 2020-02-24
Maintain downgrade commands in a text file, update when tagging

paf31:
This moves the catalog version and downgrade shortcuts into a file so that we can update the latter automatically when tagging a build.


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -1084,23 +1085,13 @@ serveOptionsParser =
...
-  , ("v1.1.0-beta.3", "31")
-  , ("v1.1.0", "31")
-  ]
+  $(do addDependentFile "catalog_versions"
+       s <- runIO (readFile "catalog_versions")
+       let parseVersions = mapMaybe (parseVersion . words) . lines
+    
+           parseVersion [tag, version] = Just (tag, version)
+           parseVersion _ = Nothing
+       TH.lift (parseVersions s))
```
lexi-lambda: I’d rather this raise an error if parsing fails than silently drop the line, since the latter seems like it could cause headaches.

server/src-lib/Hasura/Server/Migrate/Version.hs
```diff
@@ -9,10 +9,14 @@ module Hasura.Server.Migrate.Version
 
 import           Hasura.Prelude
 
-import qualified Data.Text      as T
+import qualified Data.Text                  as T
+import           Language.Haskell.TH.Syntax as TH
 
 latestCatalogVersion :: Integer
-latestCatalogVersion = 32
+latestCatalogVersion =
```
lexi-lambda: Could you add a documentation comment to this binding that explains why this is kept in a separate file? Also, could you update the instructions in `Hasura.Server.Migrate` to tell people to update the `catalog_version` file instead of this module?

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -1084,23 +1085,13 @@ serveOptionsParser =
...
-  , ("v1.0.0-beta.10", "27")
-  , ("v1.0.0-rc.1", "28")
-  , ("v1.0.0", "28")
-  , ("v1.1.0-beta.1", "29")
-  , ("v1.1.0-beta.2", "30")
-  , ("v1.1.0-beta.3", "31")
-  , ("v1.1.0", "31")
-  ]
+  $(do addDependentFile "catalog_versions"
+       s <- runIO (readFile "catalog_versions")
```
lexi-lambda: This is a minor point, but I think you can nest splices, so you could use `Data.FileEmbed.embedStringFile` here to avoid having to call `addDependentFile` manually (since we already depend on `file-embed` anyway).

scripts/tag-release.sh
```diff
@@ -37,6 +37,10 @@ find "$ROOT/install-manifests" \
      -type f -exec sed -i -E \
      's#(hasura/graphql-engine:)v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)(\-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(.*)*$#\1'"${TAG}"'\9#' {} \;
 
+# add the latest tag to the catalog_versions file
+[ -n "$(tail -c1 catalog_versions)" ] && echo >> catalog_versions 
+echo $TAG $(cat catalog_version) >> catalog_versions
```
lexi-lambda: I think you want to use `$ROOT` to build the paths here. (Also, the first line seems needlessly defensive to me, but I guess it’s fine if you want to keep it.)

server/src-lib/Hasura/Server/Init.hs
paf31: `embedStringFile` produces an `Exp`, but I need the string so that I can parse it.

scripts/tag-release.sh
paf31: If someone modifies the file in an editor, it's not unlikely that it would add an extra newline. If we're turning that into an error, I think we should be careful not to add an extra one.

server/src-lib/Hasura/Server/Init.hs
lexi-lambda: Right, but I think you can nest splices, so you could do:

```haskell
$(let s = $(embedStringFile "catalog_versions")
      parseVersions = ...
  in ...)
```

scripts/tag-release.sh
lexi-lambda: I don’t understand—this code *adds* an extra newline if one isn’t already there. But your comment talks about being careful not to add an extra one, so I’m confused.

server/src-lib/Hasura/Server/Migrate/Version.hs
paf31: Done

server/src-lib/Hasura/Server/Init.hs
paf31: Done

server/src-lib/Hasura/Server/Init.hs
paf31: I see, thanks.

scripts/tag-release.sh
paf31: I can either not add one at all, in which case I might end up missing a necessary newline character, or I can add one always, in which case I might end up with two.

lexi-lambda: Ah, I see, sure. (I think it’d be okay to assume the newline will always be there, but you’re right, being defensive seems harmless.)

PR #3952, opened 2020-02-25
docs: integration guide for auth guardian

marionschleifer:
### Description

Add a guide on integrating AuthGuardian with Hasura to the docs.

### Affected components

- Docs


rikinsk:
Subsumed in #3958

PR #3965, opened 2020-02-27
Remove any outstanding hdb_views (fixes #3940)

### LINKED ISSUE #3940: hdb_views.role__insert__schema__table remain in db after upgrading to v1.1.0
After upgrading to v1.1.0, I was trying to drop a column and came across this error:
```
PostgreSQL said: cannot drop table "MyTable" column "MyColumn" because other objects depend on it
Detail: view hdb_views."role__insert__public__ MyTable" depends on table "MyTable" column "MyColumn"
Hint: Use DROP ... CASCADE to drop the dependent objects too.
```

I saw that these views were removed here:  https://github.com/hasura/graphql-engine/pull/3598

Note: I am NOT using Hasura for db migrations.

paf31:
Deletes any remaining views and trigger functions in `hdb_views` on startup. Tested by downgrading to an earlier version, creating some insert permission views, and then upgrading again to see them deleted.

### Affected components

- [x] Server

#### Catalog upgrade

- [x] Yes

#### Metadata

- [x] No
#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/RQL/DDL/Schema/Cache.hs
```diff
@@ -406,9 +405,6 @@ buildSchemaCacheRule = proc (catalogMetadata, invalidationKeys) -> do
 -- if not, incorporates them into the schema cache.
 withMetadataCheck :: (MonadTx m, CacheRWM m, HasSQLGenCtx m) => Bool -> m a -> m a
 withMetadataCheck cascade action = do
-  -- Drop hdb_views so no interference is caused to the sql query
-  liftTx $ Q.catchE defaultTxErrorHandler clearHdbViews
```
tirumaraiselvan: We should still keep this because Event Trigger trigger functions are still created in `hdb_views`. The proposal to move them to `hdb_catalog` is pending discussion.

paf31: Oops, thanks. Updated ..

tirumaraiselvan:
@paf31 Could you see about the conflicts?

server/src-rsr/migrations/33_to_34.sql
```diff
@@ -0,0 +1,15 @@
...
+    FOR r IN (SELECT viewname FROM pg_views WHERE schemaname = 'hdb_views' ORDER BY viewname) LOOP
+      EXECUTE 'DROP VIEW IF EXISTS hdb_views.' || quote_ident(r.viewname) || ' CASCADE';
+    END LOOP;
+  END $$;
+
+DO $$ DECLARE
+  r RECORD;
+  BEGIN
+    FOR r IN (SELECT routine_name FROM information_schema.routines WHERE specific_schema = 'hdb_views' ORDER BY routine_name) LOOP
+      EXECUTE 'DROP FUNCTION hdb_views.' || quote_ident(r.routine_name) || '() CASCADE';
```
tirumaraiselvan: @paf31  Is this required? I am looking at the tests and the `test_server_upgrade` failed on Event Triggers. 

Note that ET functions are present in `hdb_views`. I tried to repro this and indeed it is the case that Event Triggers stop functioning after this upgrade. If I remove this particular routine, they work normally after upgrade. 

I am not certain why this is so. I would have expected that even after removing all the routines during catalog upgrade, the schema building process should have recreated them. We can either investigate this or just not touch the functions of `hdb_views` to get the tests to pass.

paf31: I've removed it, thanks.

PR #3968, opened 2020-02-27
cli(migrations-img): add env to skip update prompts (fix #3964)

### LINKED ISSUE #3964: Migrations not applied if update available
I'm using the cli-migrations (`hasura/graphql-engine:v1.1.0.cli-migrations`) docker image, and I've noticed twice now (once when 1.1.0 became available and once now that the 1.2.0 preview is available) that migrations aren't applied if there's an update available. Does the `--skip-update-check` flag need to be passed [here](https://github.com/hasura/graphql-engine/blob/master/scripts/cli-migrations/docker-entrypoint.sh#L76)?

arvi3411301:
### Description
This PR adds the fix to disable update notification in cli-migrations docker image.

### Affected components

- [ ] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [x] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues
#3964 

### Solution and Design
Instead of adding `show_update_notifcation: false` in config.yaml (which is not supported), add env `HASURA_GRAPHQL_SHOW_UPDATE_NOTIFCATION=false`


scripts/cli-migrations/docker-entrypoint.sh
```diff
@@ -72,7 +72,8 @@ if [ -d "$HASURA_GRAPHQL_MIGRATIONS_DIR" ]; then
     cp -a "$HASURA_GRAPHQL_MIGRATIONS_DIR/." "$TEMP_MIGRATIONS_DIR/migrations/"
     cd "$TEMP_MIGRATIONS_DIR"
     echo "endpoint: http://localhost:$HASURA_GRAPHQL_MIGRATIONS_SERVER_PORT" > config.yaml
-    echo "show_update_notification: false" >> config.yaml
```
shahidhk: @arvi3411301 Why do you think this didn't work on the CLI?

arvi3411301: @shahidhk we are checking for `show_update_notification` in global config not in project config

shahidhk: @arvi3411301 Okay!

PR #3975, opened 2020-02-27
cli, server: use prerelease tag as channel for console assets cdn

arvi3411301:
### Description
This PR adds the support to use the prerelease tag as a channel for the console assets.

### Affected components

- [x] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [x] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design
In the previous implementation, the console assets path were identified based on the prerelease tag with conditional [checks](https://github.com/hasura/graphql-engine/blob/master/cli/version/console.go#L47).

In the current implementation, we have changed the logic to identify console assets path by using the prerelease tag as a channel without the build number.

```
v1.0.0-alpha35 -> channel/alpha/v1.0
v1.1.0-beta.1  -> channel/beta/v1.1
```


server/graphql-engine.cabal
```diff
@@ -194,6 +194,9 @@ library
                      
                      , directory
 
+                     -- regex
+                     , regex-posix
```
lexi-lambda: We already depend on `regex-tdfa`, so can we use that instead of adding another dependency?

arvi3411301: @lexi-lambda I have updated the code to use `regex-tdfa`

cli/version/console.go
```diff
@@ -44,14 +44,14 @@ func (v *Version) GetConsoleAssetsVersion() (av string) {
...
-			}
-			if strings.HasPrefix(preRelease, "beta") {
-				channel = "beta"
-			}
-			if strings.HasPrefix(preRelease, "rc") {
-				channel = "rc"
+			if preRelease != "" {
+				// Get the correct channel from the prerelease tag
+				var re = regexp.MustCompile(`^[a-z]+`)
+				re.FindStringIndex(preRelease)
```
shahidhk: Is this line required?

PR #4006, opened 2020-03-03
fix syntax error in codeowners file

shahidhk:
This PR fixes the syntax error that was introduced in https://github.com/hasura/graphql-engine/pull/3946


shahidhk:
@rikinsk Do you think it should be `/CHANGELOG.md`?

PR #4007, opened 2020-03-03
update actions docs

rikinsk:
### Description

- add actions relationships docs
- fix action error response format
- add note about no setup needed for console actions workflow
- fix QualifiedTable link
- add note about downgrading catalogue from older HGE versions

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


rikinsk:
@marionschleifer 
> Changing the title ("Connecting actions with the rest of the graph") to "Action relationships" for just the side bar, not in the title

I wanted to do this as well but I reconsidered as it is a little misleading. Relationships are defined on object types and not the action itself. Its quite confusing actually hence I went for what would be clear and not misleading.  I am not a great fan of the current title as well but maybe we can revisit this later with feedback from others. Currently letting it be as it is.

PR #3553, opened 2019-12-17
Scheduled triggers (close #1914)

tirumaraiselvan:
### Description

The functionality is to create “scheduled eventing” which will invoke a webhook at configured times. This will solve use-cases such as:

1. Invoke a webhook at EoD everyday (say, for daily reports)
2. Send follow-up email 2 weeks after creation of user (using the API in business logic)
3. Build a tweet management system (like buffer.com)

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

closes #1914

### Solution and Design

Each hasura instance starts two threads:

Generator: Fetches the list of Scheduled Triggers and generates scheduled-events for the next 6 hour (configurable via HASURA_GRAPHQL_SCHEDULED_TRIGGER_GENERATE_RANGE env var)

Pusher: Fetches the scheduled-events which are  <=currentTime and not delivered (similar to event trigger delivery thread). For v1, the payload will be empty i.e. no action will be performed for the scheduled event.

#### Catalog upgrade

Does this PR change Hasura Catalog version?
- [ ] No
- [x] Yes
  - [ ] Updated docs with SQL for downgrading the catalog 

#### Metadata

Does this PR add a new Metadata feature?
- [ ] No
- [ ] Yes
  - Does `run_sql` auto manages the new metadata through schema diffing?
    - [ ] Yes
    - [ ] Not required
  - Does `run_sql` auto manages the definitions of metadata on renaming?
    - [ ] Yes
    - [ ] Not required
  - Does `export_metadata`/`replace_metadata` supports the new metadata added?
    - [ ] Yes
    - [ ] Not required

#### GraphQL
- [ ] No new GraphQL schema is generated
- [ ] New GraphQL schema is being generated:
   - [ ] New types and typenames are correlated
   
   

#### Breaking changes

- [ ] No Breaking changes
- [ ] There are breaking changes:

  1. Metadata API

     Existing `query` types:
     - [ ] Modify `args` payload which is not backward compatible
     - [ ] Behavioural change of the API
     - [ ] Change in response `JSON` schema
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     - [ ] Change in any `NamedType`
     - [ ] Change in table field names
     

     Schema Resolve:-
     
     - [ ] Change in treatment of `null` value for any input fields 
     

  3. Logging

[...]

PR #4013, opened 2020-03-03
server(events): utilize proper backpressure scheme (close #3839)

### LINKED ISSUE #3839: Utilize proper backpressure scheme in Hasura.Events.Lib
We slurp events as fast as possible, while bounding the number of processing threads. This can lead to a memory leak (in case we can't process fast enough), or possibly event data loss (if we crash with events in the TQueue), though maybe the latter is acceptable.

TODO
- bound work in `pushEvents` not `consumeEvents`
- log when or before we block

Possible relation to #3388

jberryman:
### Description

With loo low `HASURA_GRAPHQL_EVENTS_FETCH_INTERVAL` and/or slow webhooks
and/or too small `HASURA_GRAPHQL_EVENTS_HTTP_POOL_SIZE` we might
previously check out events from the DB faster than we can service them,
leading to space leaks, weirdness, etc.

Other changes:
- avoid fetch interval sleep latency when we previously did a non-empty
  fetch
- prefetch event batch while http pool is working
- warn when it appears we can't keep up with events being generated
- make some effort to process events in creation order so we don't
  starve older ones.

ALSO NOTE: HASURA_GRAPHQL_EVENTS_FETCH_INTERVAL changes semantics
slightly, since it only comes into play after an empty fetch. The old
semantics weren't documented in detail, so I think this is fine.

### Changelog

### Affected components

- [x] Server

- [x] Docs

- [x] Tests

### Related Issues

#3839 

#### Breaking changes

- [x] No Breaking changes


jberryman:
Only the second commit needs review here. First is being reviewed in  https://github.com/hasura/graphql-engine/pull/3860

jberryman:
I'd also like someone to take a look here, while we're on the subject:

https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Hasura/App.hs#L278

It looks like this means if we spin up a bunch of servers we're going to be processing the same first batch of events on all of them. I haven't really thought about what we should be doing here.

tirumaraiselvan:
> I'd also like someone to take a look here, while we're on the subject:
> 
> https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Hasura/App.hs#L278
> 
> It looks like this means if we spin up a bunch of servers we're going to be processing the same first batch of events on all of them. I haven't really thought about what we should be doing here.

Yeah, this bit has been causing few issues especially multiple deliveries. Will discuss.

jberryman:
(FYI trying to debug these intermittent failures on CI)

jberryman:
I think this is blocked on a fix Nizar implemented somewhere in a PR

docs/graphql/manual/event-triggers/index.rst
```diff
@@ -24,6 +24,11 @@ Events can be of the following types:
 - DELETE: When a row is deleted from a table
 - MANUAL: Using the console or API, an event can be triggered manually on a row
 
+.. note::
+
+  Event webhook notifications will be delivered at least once, and may arrive out of order with
```
tirumaraiselvan: Thanks, was just looking to add this in the documentation myself :)

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -166,49 +168,86 @@ retryAfterHeader = "Retry-After"
 
 initEventEngineCtx :: Int -> DiffTime -> STM EventEngineCtx
 initEventEngineCtx maxT fetchI = do
-  q <- TQ.newTQueue
-  c <- newTVar 0
-  return $ EventEngineCtx q c maxT fetchI
-
-forkEventQueueProcessors
+  c <- newTVar maxT
```
tirumaraiselvan: I know the naming here were screwed up to begin with but maybe we can few of them here like `c`

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -166,49 +168,86 @@ retryAfterHeader = "Retry-After"
...
+      -- worth the effort for something more fine-tuned
+      eventsNext <- withAsync popEventsBatch $ \eventsNextA -> do
+        -- process approximately in order, minding HASURA_GRAPHQL_EVENTS_HTTP_POOL_SIZE:
+        forM_ events $ \event -> 
+          mask_ $ do
+            atomically $ do  -- block until < HASURA_GRAPHQL_EVENTS_HTTP_POOL_SIZE threads:
+              capacity <- readTVar _eeCtxEventThreadsCapacity
+              check $ capacity > 0
+              writeTVar _eeCtxEventThreadsCapacity (capacity - 1) 
+            let restoreCapacity = liftIO $ atomically $
```
tirumaraiselvan: FYI, this has been changed [here](https://github.com/hasura/graphql-engine/pull/3553/files#diff-6bb6b345ca93d9f4bf123a91bc86056dR180) to use the `bracket_` pattern

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -166,49 +168,86 @@ retryAfterHeader = "Retry-After"
...
+                    processEvent event `finally` restoreCapacity
+            link t
+
+        -- return when next batch ready; some 'processEvent' threads may be running.
+        wait eventsNextA
+
+      let lenEvents = length events
+      if | lenEvents == fetchBatchSize -> do
+             -- When we choose to warn is fairly arbitrary:
+             let clearlyBehind = fullFetchCnt >= 3
```
tirumaraiselvan: a better name than `fullFetchCnt` would help since it doesn't correlate really well with `clearlyBehind`

server/tests-py/test_events.py
```diff
@@ -81,6 +84,34 @@ def test_create_operation_spec_not_provider_err(self, hge_ctx):
     def dir(cls):
         return 'queries/event_triggers/create-delete'
 
+# Smoke test for handling a backlog of events
```
tirumaraiselvan: Seems like we are testing if we process 2000 events. But how do we know that backpressure changes are working i.e. we are emitting warnings?

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -166,49 +168,86 @@ retryAfterHeader = "Retry-After"
...
+              check $ capacity > 0
+              writeTVar _eeCtxEventThreadsCapacity (capacity - 1) 
+            let restoreCapacity = liftIO $ atomically $ 
+                  modifyTVar' _eeCtxEventThreadsCapacity (+ 1)
+            t <- async $ flip runReaderT (logger, httpMgr) $ 
+                    processEvent event `finally` restoreCapacity
+            link t
+
+        -- return when next batch ready; some 'processEvent' threads may be running.
+        wait eventsNextA
```
tirumaraiselvan: IIUC, this call blocks till all events for a batch are finished executing? How would it effect with a slow webhook? Suppose 99 events finished and one event was stuck for 60 seconds (or till timeout) because webhook was very slow. Will the whole queue get blocked for 60 seconds?

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -166,49 +168,86 @@ retryAfterHeader = "Retry-After"
...
+    popEventsBatch = do
+      let run = runExceptT . Q.runTx pool (Q.RepeatableRead, Just Q.ReadWrite)
+      run (fetchEvents fetchBatchSize) >>= \case
+          Left err -> do 
+            L.unLogger logger $ EventInternalErr err
+            return []
+          Right events -> 
+            return events
+
+    go :: [Event] -> Int -> Bool -> IO void
```
tirumaraiselvan: @jberryman I am slighly confused how are we capping the memory/threads in flight. IIUC, we keep fetching events and get blocked on processing them because of `HASURA_GRAPHQL_EVENTS_HTTP_POOL_SIZE`

tirumaraiselvan: How is this more beneficial than using the in-memory queue that we had before `_eeCtxEventQueue`

jberryman: The main issue with the old version is we fetch from DB at intervals without regard for how fast we're processing. If webhooks are very slow we will check out more and more events to process, maybe leaking all memory.

Here we prefetch at most two batches, we don't recurse on `go` and fetch another batch until we've (concurrently) processed the last batched (or to be precise, when there is some extra capacity in our workers).

I agree this could be more clear! will improve comments , then would be great if you could take another look

server/src-lib/Hasura/Events/Lib.hs
jberryman: maybe I can resolve this with a better comment. The idea is this line should read "we are `clearlyBehind` in our processing when the last 3 fetches have been full (i.e. only limited by our `LIMIT` in the query)". Let me know if you have suggestions to improve aside from a comment like that

server/tests-py/test_events.py
jberryman: We don't, this is just a smoke test for code coverage and so I could convince myself it was working. Do you think it's worth the effort to test for logging?

server/src-lib/Hasura/Events/Lib.hs
tirumaraiselvan: I think it's allright. Maybe we can just do `fullFetchCount` but optional.

server/tests-py/test_events.py
tirumaraiselvan: Okay, just wanted to clarify. I think checking for log line might be too much infra. Is there any other way to detect if things are actually getting throttled?

server/src-lib/Hasura/Events/Lib.hs
tirumaraiselvan: > Here we prefetch at most two batches, we don't recurse on go and fetch another batch until we've (concurrently) processed the last batched (or to be precise, when there is some extra capacity in our workers).

Is it because it gets blocked here https://github.com/hasura/graphql-engine/pull/4013/files#diff-a5793e7f6a1e13c4a19d13cae27dae88R215?

jberryman: yep we fork a new worker only when capacity is freed, I'll beef out comments

server/tests-py/test_events.py
jberryman: I can't think of a slick way to do it... maybe better to punt on that?

jberryman:
@tirumaraiselvan thanks for the review! let me know if the additional docs make things clear

PR #4023, opened 2020-03-05
hide starterkit button from actions codegen if a framework has no starter kit

wawhal:
### Description

Some codegen frameworks have starter kits and some don't. This PR just hides the `Get starter kit` button if a framework has no codegen.

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


console/src/components/Services/Actions/Codegen/Codegen.js
```diff
@@ -101,6 +101,15 @@ const Codegen = ({ allActions, allTypes, currentAction }) => {
     };
 
     const getStarterKitButton = () => {
+      const selectedFrameworkMetadata = allFrameworks.find(
+        f => f.name === selectedFramework
+      );
+      if (
+        selectedFrameworkMetadata &&
```
beerose: What if `selectedFrameworkMetadata` was not found?

PR #4034, opened 2020-03-05
docs: add note on pg version for actions

marionschleifer:
### Description
Add note on actions docs for supported Postgres versions.

### Affected components
- Docs


docs/graphql/manual/actions/index.rst
```diff
@@ -20,8 +20,9 @@ enrichment from external sources and any other complex business logic.
 
 .. admonition:: Supported from
 
-   Actions are currently available for preview in the pre-release versions of
-   ``v1.2.0``
+   - Actions are currently available for preview in the pre-release versions of ``v1.2.0``
```
tirumaraiselvan: let's keep the nomenclature consistent by calling it `beta` ?

tirumaraiselvan: Also, is this the standard way of saying something is available from this version?

docs/graphql/manual/actions/index.rst
```diff
@@ -20,8 +20,9 @@ enrichment from external sources and any other complex business logic.
...
+
+   - Actions are only supported for ``Postgres versions 10 or higher``.
```
tirumaraiselvan: I wouldn't mind if this Postgres version thing was in a separate admonition. Because it is kind of a permanent limitation.

wawhal: I moved it up because it's hard to come across if we put it at the bottom. It's something people should know upfront when they are evaluating actions. And two admonitions one below the other looks sorta ugly.

tirumaraiselvan: let's keep it standard. @marionschleifer what's the standard way of mentioning availability?

marionschleifer: @tirumaraiselvan I put it on the bottom to be consistent how we inform about Postgres version on the getting started guide: https://hasura.io/docs/1.0/graphql/manual/getting-started/index.html.

docs/graphql/manual/actions/index.rst
rikinsk: We can use the terminology `.. currently available in beta in the ..`. Makes sense. 
And yeah. This is what we use to show available from currently

docs/graphql/manual/actions/index.rst
rikinsk: I think here it is fine calling it out upfront. High chance of missing it otherwise. I am in favour of using 2 separate admonitions. The 2 limitations are kind of unrelated. One is a hasura version thing vs a permanent limitation. Adding a separate note about the PG limitation like in gettting started makes sense.

docs/graphql/manual/actions/index.rst
```diff
@@ -20,11 +20,14 @@ enrichment from external sources and any other complex business logic.
 
 .. admonition:: Supported from
 
-   Actions are currently available for preview in the pre-release versions of
-   ``v1.2.0``
+   Actions are currently available for preview in the pre-release versions of ``v1.2.0``
```
tirumaraiselvan: pls change this to beta.

PR #4045, opened 2020-03-06
Pause/Resume event triggers, fix #3423

### LINKED ISSUE #3423: pause event triggers
Sometimes for seeding data or general dev functionalities, we may not want the event trigger to be invoked. Currently, the only way to stop events from creating/delivering is by dropping the event trigger. We can perhaps give a pause functionality to pause an event trigger without needing to drop it.

codingkarthik:
add a new metadata api to unpause an event trigger

 #3423 

### Changelog

### Affected components

- [ X ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ X ] Tests
- [ ] Other (list it)

### Solution and Design

Add new APIs `pause_event_trigger` and `unpause_event_trigger` which will pause and unpause a given event trigger respectively. 
A new column `paused` is added in the `event_triggers` and `event_log` tables. 
Whenever a ET is paused, the trigger will write new event_logs with `paused` as `true`. These events won't be run.

### Steps to test and verify

Pause an ET and mutate the table related to it. There should not be any webhook calls made while the ET is paused

#### Catalog upgrade

- [ ] No
- [ X ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ X ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


codingkarthik:
@marionschleifer You're right. I think `resume` would be a better choice here.

I'll make the required changes.

Thanks for the suggestion 🙂

docs/graphql/manual/api-reference/schema-metadata-api/event-triggers.rst
```diff
@@ -224,3 +224,77 @@ EventTriggerColumns
    :class: haskell-pre
 
    "*" | [:ref:`PGColumn`]
+
+.. _pause_event_trigger:
+
+pause_event_trigger
+--------------------
```
marionschleifer:
```suggestion
-------------------
```

marionschleifer: Underline should always be the same length as the title 🙂

marionschleifer:
Changelog approved.

docs/graphql/manual/api-reference/schema-metadata-api/event-triggers.rst
codingkarthik: Oh got it, thank you 😄

server/src-lib/Hasura/RQL/DDL/EventTrigger.hs
```diff
@@ -307,6 +309,24 @@ runInvokeEventTrigger (InvokeEventTriggerQuery name payload) = do
...
       _         -> throw400 NotSupported "manual mode is not enabled for event trigger"
 
+runPauseEventTrigger
+  :: (MonadTx m)
+  => PauseEventTriggerQuery -> m EncJSON
+runPauseEventTrigger (PauseEventTriggerQuery name) = do
+  affectedRows <- updatePauseOfEventTriggerInCatalog name True
+  if | affectedRows == 1 -> pure successMsg
+     | affectedRows == 0 -> throw400 NotFound "event trigger not found"
+     | otherwise -> throw500 "unexpected: more than one event triggers have been paused"
```
lexi-lambda, line 319: This is a weird quirk of the English language that doesn’t make any sense, but the phrase “more than one *x*” [is a grammatically *singular* subject](https://ell.stackexchange.com/a/109365/4507). That means the correct wording of this error message is actually:

> unexpected: more than one event trigger has been paused

Why? I have no idea.

That said, I’d actually reword this error message more significantly, since I think as-written it is confusing. If I got this error message, I would expect that more than one of my event triggers actually got paused, but in reality, the transaction will be rolled back, so *no* event triggers will have been paused. Therefore, I’d word this a different way:

```suggestion
     | otherwise -> throw500 "multiple event triggers exist with the given name"
```

server/src-lib/Hasura/RQL/DDL/EventTrigger.hs
```diff
@@ -307,6 +309,24 @@ runInvokeEventTrigger (InvokeEventTriggerQuery name payload) = do
...
+
+runResumeEventTrigger
+  :: (MonadTx m)
+  => ResumeEventTriggerQuery -> m EncJSON
+runResumeEventTrigger (ResumeEventTriggerQuery name) = do
+  affectedRows <- updatePauseOfEventTriggerInCatalog name False
+  if | affectedRows == 1 -> pure successMsg
+     | affectedRows == 0 -> throw400 NotFound "event trigger not found"
+     | otherwise -> throw500 "unexpected: more than one event triggers have been resumed"
```
lexi-lambda, line 328: This logic is duplicated between `runPauseEventTrigger` and `runResumeEventTrigger`, but I don’t think it needs to be—this is really an implementation detail of `updatePauseOfEventTriggerInCatalog`! So I think you could move this check into that function, and that way it wouldn’t need to be duplicated at both call sites.

server/src-lib/Hasura/RQL/DDL/EventTrigger.hs
```diff
@@ -354,3 +374,12 @@ updateEventTriggerInCatalog trigConf =
       configuration = $1
       WHERE name = $2
     |] (Q.AltJ $ toJSON trigConf, etcName trigConf) True
+
+updatePauseOfEventTriggerInCatalog :: (MonadTx m) => TriggerName -> Bool -> m Int
+updatePauseOfEventTriggerInCatalog triggerName pause = liftTx $ do
```
lexi-lambda, line 379: This function name is a little unwieldy; I’d rename it to something a little more direct:

```suggestion
setEventTriggerPausedInCatalog :: (MonadTx m) => TriggerName -> Bool -> m Int
setEventTriggerPausedInCatalog triggerName pause = liftTx $ do
```

server/src-lib/Hasura/RQL/Types/EventTrigger.hs
```diff
@@ -218,3 +220,17 @@ data InvokeEventTriggerQuery
...
+  } deriving (Show, Eq, Lift)
+
+$(deriveJSON (aesonDrop 4 snakeCase){omitNothingFields=True} ''PauseEventTriggerQuery)
+
+data ResumeEventTriggerQuery
+  = ResumeEventTriggerQuery
+  { retqName  :: !TriggerName
+  } deriving (Show, Eq, Lift)
+
+$(deriveJSON (aesonDrop 4 snakeCase){omitNothingFields=True} ''ResumeEventTriggerQuery)
```
lexi-lambda, line 236: I realize the other query types are defined in this module, so you’re just doing what the existing code does, but I don’t think there’s any reason we can’t define these alongside the functions that handle them, in `Hasura.RQL.DDL.EventTrigger`. Keeping the types close to the functions that need them is always better if possible in my opinion! So I’d move these definitions into that module, and while you’re at it, maybe you can move the other query types there, too.

server/src-rsr/initialise.sql
```diff
@@ -294,7 +295,8 @@ CREATE TABLE hdb_catalog.event_log
   created_at TIMESTAMP DEFAULT NOW(),
   locked BOOLEAN NOT NULL DEFAULT FALSE,
   next_retry_at TIMESTAMP,
-  archived BOOLEAN NOT NULL DEFAULT FALSE
+  archived BOOLEAN NOT NULL DEFAULT FALSE,
+  paused BOOLEAN NOT NULL DEFAULT FALSE
```
lexi-lambda, line 299: It seems quite curious to me that each `event_log` record has a `paused` field—you don’t pause an event, you pause a trigger! When a trigger is paused, it shouldn’t create any more events at all, not “paused events.”

I’d remove the `paused` field from `event_log` and change `insert_event_log` so that it just returns without doing anything if the event is paused. Then you don’t have to worry about filtering out paused events in `fetchEvents`.

tirumaraiselvan:
I haven't gone through this PR, but careful consideration has to be given if the `paused` attribute should also make it to exported metadata.

lexi-lambda:
> I haven't gone through this PR, but careful consideration has to be given if the `paused` attribute should also make it to exported metadata.

Good catch! I think you’re right, it probably needs to be reflected there, too.

server/src-rsr/initialise.sql
codingkarthik: @lexi-lambda I was creating the event log events with paused as one of the columns because I thought it would be a good idea to track the events that were tracked when the trigger was paused, maybe I over-thought it 😅 .

Stefmore02:
Closing this PR, this feature is put on hold until further notice.

PR #4089, opened 2020-03-11
relation selector component (close #2296)

### LINKED ISSUE #2296: [FEATURE][CONSOLE] Relation selector component rather than id for foreign keys
Hi guys,
For me one of the core feature for data entry is ability to add foreign key based on human friendly component rather than simple text box and enter messy UUID.
One of the reason that now many user like me forced to use extra tools like React-Admin is that manual data entry in Hasura console is a little hard specially when you have many relation and foreign key.
Later I suggested this as a proposal and wire-frame here: https://github.com/hasura/graphql-engine/issues/506#issuecomment-471521156
I open this issue because previous one is closed and I think this is a extra feature.

Maybe it's better to show dropdown as input type for **Insert Row** tab. For example in below image maybe it's better to show a dropdown that show `title` but pass `UUID id` for **parent_subject_id** field:

Screenshot from 2019-03-11 16-00-27

Maybe we can select which column is show as **display column** in foreign key creation page right to the `to` column here:
53887711-db257980-4048-11e9-863c-8c6dcf402ec9

For more idea now in react-admin you have [ReferenceInput](https://marmelab.com/react-admin/Inputs.html#referenceinput) and [ReferenceArrayInput](https://marmelab.com/react-admin/Inputs.html#referencearrayinput) component and able to enter relation with drop down or select2 based selector rather than id.

beerose:
Solves #2296

### Description

This PR adds a select component for columns with foreign key relation. For example, if we have table `users` with columns `id` and `name`, and table `articles` with columns `id`, `author_id` (referencing users.id), then when adding/editing foreign we can specify `display_column`:
captured (4)

Then in `insert row` and `edit row` view, we can choose a value from a dropdown where column specified as `display column` is used as a label. By default, only 20 rows are displayed in the dropdown. When typing data is filtered by the current input value.

Screenshot 2020-03-31 at 19 33 01

We're not limited to using the dropdown and filtering the data by `display column`. Providing own value (in this case user.id) also works. 

Screenshot 2020-03-31 at 19 29 30

### Affected components

- [x] Console

### Modify Table

**Insert/edit item**

1. When the user goes to insert/edit row, the user will now see a new dropdown called "Display Name" if the column is having a foreign 
2. Whenever the user choose the display name, display name preference in every table is stored in the local storage as a single object.
3. based on the display name, when the user types the input column value, UI shows a helper dropdown through which the user can select the foreign row by display name (example: shows name instead of confusing with user id)

### Limitations

Since I'm using `ilike` for filtering data, the user can only choose column of type `text`, `varchar` or `citext` for display column.

### Affected console components

1. Migrated SearchableSelect.js to TypeScript. Extended custom filter to full-text filtering. Added two optional props:
* `onSearchValueChange`
2. Migrated jsUtils.js to TypeScript.
3. Migrated v1QueryUtils.js to TypeScript.
4. Migrated TypedInput to TypeScript. Extended implementation. (Described above)
5. EditItem, InsertItem -- added handling selecting value from the dropdown. 
6. Migrated Data/utils.js to TypeScript. Changed implementation a bit in two places (same issue): https://github.com/hasura/graphql-engine/pull/4089/files#diff-089d4fb070627c2802e8853d6bb14f30R190
  It was done because we were taking advantage of JavaScript accepting the following code:
  ```js
  rcol = [foreignKeyConstraintOn.column]; // rcol = ["example"]
  lcol = [column_mapping[rco]]; // column_mapping[["example"]] <- valid JS
  ``` 
I changed it because TypeScript doesn't accept it: [link](http://www.typescriptlang.org/play#code/MYewdgzgLgBAHjAvDA3gKBpmByAhtgLhgEY0BfNNOAbT2wF0BuNAehZgDdcAbASwBMYvMDABSAZQA0QsFz6DhMACriq1Wvnr0gA).
  

### Ideas

* Make common abstraction on insert row and edit row.


console/src/components/Services/Data/DataActions.js
```diff
@@ -196,6 +238,127 @@ const initQueries = {
   },
 };
 
+const loadConsoleOpts = () => {
```
beerose, line 241: Get rid of this

console/src/components/Services/Data/DataActions.js
```diff
@@ -196,6 +238,127 @@ const initQueries = {
...
+      error => {
+        console.error(
+          'Failed to load console options: ' + JSON.stringify(error)
+        );
+      }
+    );
+  };
+};
+
+const getForeignKeyOptions = () => {
```
beerose, line 269: Get rid of this — move to LS

console/src/components/Services/Data/DataActions.js
```diff
@@ -196,6 +238,127 @@ const initQueries = {
...
+      error => {
+        console.error(
+          'Failed to get foreign key options: ' + JSON.stringify(error)
+        );
+      }
+    );
+  };
+};
+
+const filterFkOptions = (fkColOptions, searchValue) => {
```
beerose, line 317: Fetching data from the server. This we'll leaving.

console/src/components/Services/Data/DataActions.js
```diff
@@ -650,6 +834,86 @@ const fetchColumnTypeInfo = () => {
...
+ *
+ * @typedef DisplayConfig
+ * @param {string} tableName
+ * @param {string} schemaName
+ * @param {string} constraintName
+ * @param {Mapping[]} mappings
+ *
+ * @param {DisplayConfig} displayConfig
+ */
+const setConsoleFKOptions = displayConfig => (dispatch, getState) => {
```
beerose, line 851: Get rid of in favor of LS

beerose:
@soorajshankar could you fix the conflicts?

Issue #3388, opened 2019-11-20
Memory usage accumulating over time until crash / restart


woba-nwa:
Since installing Hasura beta 9 (along with other updates to our system), we experience a gradual increase in memory usage over time. We are hosted on Heroku, and may have been saved by the daily dyno restart prior to yesterday.
Yesterday, we had enough, heavy queries however to max out our memory during the day and a manual restart was required. Heavy queries does seem to be responsible for the behaviour.
![Skærmbillede 2019-11-20 kl  11 02 51](https://user-images.githubusercontent.com/46349196/69229237-60c06980-0b85-11ea-9ce9-55015ba13696.png)

Today, we are already at 3 times our normal memory consumption. 
Please advise what to do!

Seems related to issue [#2794 ]

steelerc69:
We are seeing the same behavior after upgrading from beta-6 to beta-10.  We are running docker on portainer.  The memory seems to grow until the containers crash.

0x777:
Can you check if this issue persists after setting the env variable `HASURA_GRAPHQL_QUERY_PLAN_CACHE_SIZE` to a value like `100`? You'll have to upgrade to `v1.0.0-rc.1`.

makkus:
I'm seeing the same issue at the moment, in a k3s cluster. It's a service we just set up, and which hasn't seen much traffic yet.  Up till now I ran a single replica, without a memory limit, because I didn't though I needed to worry.

Today we are seeing some traffic (2, 3 requests per second, mostly mutations) because it's the time of the year for this service, and it took about 3 hours for the pod to consume all of the nodes' memory.  Happened again after I restarted the whole thing. 

Now I've scaled up the service to 4 replicas, and set the memory limit to 1200mi, and the OOMKiller kills a pod roughly every 15 minutes. There's no service disruption, since Kubernetes handles this fine, but I'm wondering what the cause of this could be.

Here's what the memory consumption of the node in question looked like (2 times eating all the nodes memory, and the spikes at the end indicate the OOMKiller kicking in):

![hasura](https://user-images.githubusercontent.com/311257/72669294-434a3780-3a30-11ea-9237-9dfbe9000b40.png)


I tried the ``HASURA_GRAPHQL_QUERY_PLAN_CACHE_SIZE`` setting (to 100), as suggested above, but that didn't seem to have any impact.

jberryman:
@makkus thanks for the information. Can you confirm the version of graphql-engine you're using?

[...]

PR #4097, opened 2020-03-12
Use AssertNF for leak prophylaxis when doing mutable writes. Closes #3937

### LINKED ISSUE #3937: Audit usage of mutable vars for space leaks
These are a particularly potent source of leaks. Use https://hackage.haskell.org/package/ghc-heap-view-0.6.1/docs/GHC-AssertNF.html

jberryman:
Writing to a mutable var is a particularly potent source of leaks since
it mostly defeats GHC's analysis. Here we add assertions to all mutable
writes, and fix a couple spots where we wrote some thunks to a mutable
var (compiled with -O2).

Some of these thunks were probably benign, but others looked liked they
might be retaining big args. Didn't do much analysis, just fixed.

Actually pretty happy with how easy this was to use and as a diagnostic,
once I sorted out some issues. We should consider using it elsewhere,
and maybe extending so that we can use it with tests, enable when
`-fenable-assertsions` etc.

Relates #3388

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -148,8 +149,10 @@ updateJwkRef (Logger logger) manager url jwkRef = do
     logAndThrow err
 
   let parseErr e = JFEJwkParseError (T.pack e) $ "Error parsing JWK from url: " <> urlT
-  jwkset <- either (logAndThrow . parseErr) return $ J.eitherDecode respBody
-  liftIO $ writeIORef jwkRef jwkset
+  !jwkset <- either (logAndThrow . parseErr) return $ J.eitherDecode respBody
```
lexi-lambda: Might as well use `J.eitherDecode'` here if we want to be explicit about strictness, right?

jberryman:
> but could you add a Note [Force thunks before mutable writes] someplace that includes the rationale in your PR description?

Okay, I think I'll add a longer comment to the `disableAssertNF` comment, and also a couple little notes inline. This is something I would use `git blame / git log -p` for to get the full context though, so hopefully all those things together will be enough.

I also am going to talk about this briefly at the next call:

- cargo cult `Just $! ...`
- generally inform about the library, what output looks like, how we might use it going forward
- mention why mutable vars are a particularly easy place to introduce leaks

jberryman:
missed a few spots, will finish this up Mon

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -603,8 +612,9 @@ createWSServerApp
   -> WSServerEnv
   -> WS.PendingConnection -> m ()
   -- ^ aka generalized 'WS.ServerApp'
-createWSServerApp authMode serverEnv =
-  WS.createServerApp (_wseServer serverEnv) handlers
+{-# INLINABLE createWSServerApp #-}
```
lexi-lambda: This `INLINABLE` pragma sadly will not do anything because the `HasVersion` implicit parameter defeats specialization. Is there a reason we really care about this being specialized?

jberryman: Thanks, That was leftover from an experiment I abandoned, just removed it in latest push.

PR #4110, opened 2020-03-13
rfc: select permission improvements

0x777:
Rendered:
https://github.com/0x777/graphql-engine/blob/select-permission-improvements/rfcs/columns-in-boolean-expression.md
https://github.com/0x777/graphql-engine/blob/select-permission-improvements/rfcs/disable-query-and-subscription-root-fields.md

### DIFF
--- a/rfcs/columns-in-boolean-expression.md
+++ b/rfcs/columns-in-boolean-expression.md
@@ -1,0 +1,16 @@
+## Allow customizing a table's boolean expression.
+
+Currently, a table's boolean expression for a role is generated with the
+columns that are allowed to be read as defined in the select permission.
+Sometimes you may want to restrict this further, for example you may want to
+only allow columns on which indexes are defined.
+
+#### Proposed solution
+
+Introduce an optional `boolean_expression_columns` field in select permission
+which specifies the columns that are allowed in a boolean expression. To
+preserve backwards compatibility, when this field is absent in a select
+permission, `columns` is used for generating the boolean expression.
+
+Note: If a table's primary key column set is not a subset of
+of `boolean_expression_columns`, we shouldn't generate `<table>_by_pk` field.
--- a/rfcs/disable-query-and-subscription-root-fields.md
+++ b/rfcs/disable-query-and-subscription-root-fields.md
@@ -1,0 +1,66 @@
+## Allow disabling query root fields
+
+Currently when a select permission is defined for a role on a table, we
+automatically generate 3 fields for the table (`<table>`, `<table_by_pk>`,
+`<table_aggregate>`) in `query_root` and likewise in `subscription_root`.  This
+should be customisable to allow some of the patterns as discussed below.
+
+### Motivation
+
+#### 1. Allow selecting data only through relationships (issues: [207](https://github.com/hasura/graphql-engine/issues/207), [696](https://github.com/hasura/graphql-engine/issues/696), [3742](https://github.com/hasura/graphql-engine/issues/3742)).
+
+Let's say you have a slack like application with the following schema:
+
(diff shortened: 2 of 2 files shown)


aleksander-newstore:
Thanks for the RFC! I've read through them and the features proposed seem to go a long way to meet our needs.

As I understand it, we'll be able to disable subscriptions altogether, enable querying for some of the entities just via relations and enable/disable the use of certain fields in filter conditions. Thats great!

One scenario that seems to be not covered is multi-field filtering - you need to enable filtering on all the fields from the condition and that means users can filter only on the second or further column in the index.  Eg to use index on order_items_extended_attributes with columns (tenant, order_item_id, name) we would have to enable column name. And in that case someone could make a query filtering just by name which wouldn't use the index.

aleksandersumowski:
Thanks for the RFCs! I've read through them and the features proposed seem to go a long way to meet our needs.

As I understand it, we'll be able to disable subscriptions altogether, enable querying for some of the entities just via relations and enable/disable the use of certain fields in filter conditions. Thats great!

One scenario that seems to be not covered is multi-field filtering - you need to enable filtering on all the fields from the condition and that means users can filter only on the second or further column in the index.  Eg to use index on order_items_extended_attributes with columns (tenant, order_item_id, name) we would have to enable column name. And in that case someone could make a query filtering just by name which wouldn't use the index.

marbemac:
This would also make our permissioning setup so much cleaner (and likely more performant). As of now for our more nested data models, e.g. organization -> project -> things in project, we need to basically repeat all of the org and project permission checks on the tables for things inside of the project. This will be a fantastic change 👍. Any idea on rough timeline post RFC acceptance? Months, quarters?

jlowin:
We currently resort to schema stitching to achieve this; would love to see this feature

franciscolourenco:
Really need this. Any prospects on the implementation of this feature? Thanks!

MeisterTea:
Could the rfc add possibility to disable the where argument https://github.com/hasura/graphql-engine/issues/5751

Is anyone looking into the implementation?

mathieutu:
Hi, this would be a killer feature for us.

However why limit that to query and subscriptions ? I would add the ability to manage the same permissions for mutations too.

switz:
Any plans to implement blocking root level queries? This would be a huge permissioning win.

jberryman:
This sounds good, but may be a fragile or dangerous way to create permissions. For a complex DB schema it may be difficult to see where some data might leak. It should be possible to compute the implicit permissions to show the admin, or even ensure that those don't change if the schema changes.

Just some thoughts

marbemac:
> Any plans to implement blocking root level queries? This would be a huge permissioning win.

👌 this pretty much sums it up - a "simple" (obv I don't really know if it is, but I like to imagine it is :)) option to track a table so that it can be exposed over relations, etc, but not expose it at the root level, would *dramatically* simplify so much of the permissions we have to setup today.

Ambro17:
Is this RFC accepted? Is it on roadmap? I'm trying hasura and really liked it so far, but without this i will be forced to drop it

VNadygin:
Any news on it?

PhillipHoffmann:
This is very much needed.

marbemac:
From how it reads, it seems like release 2.8 might solve this (I have not tried the functionality yet, we're still on 2.1) - specifically https://github.com/hasura/graphql-engine/issues/696. @tirumaraiselvan is that correct?

tirumaraiselvan:
Merged via https://github.com/hasura/graphql-engine-mono/commit/f7f4e60250b4c7375a973d0e5770ac134217767a

PR #4120, opened 2020-03-16
rfc: admin-only/backend-only insert permissions

0x777:
Rendered: https://github.com/0x777/graphql-engine/blob/reuse-permissions-in-actions/rfcs/reuse-insert-permission-in-action.md

### DIFF
--- a/rfcs/reuse-insert-permission-in-action.md
+++ b/rfcs/reuse-insert-permission-in-action.md
@@ -1,0 +1,67 @@
+### Insert permissions and actions
+
+If an action is used for data validation, it is hard to re-use constraints
+specified through insert permissions.
+
+#### Motivation
+
+Let's consider a table with the following schema:
+
+```sql
+create table message (
+  id serial primary key,
+  content text not null,
+  channel_id integer not null references channel(id)
+)
+create table channel (..)
+create table channel_members (..)
+```
+
+Let's say you want to allow a `user` role to insert into
+`message` table if the following `check` constraint holds:
+```json
+{"channel": {"members": {"user_id": "x-hasura-user-id"}}}
+```
+i.e, a user can only post a message to a channel if they are a member of that
+particular channel. Now let's say you want to do some sort of validation on the
+content of the message (maybe through some external service), this is a problem
+that would broadly fall under the 'data validation' category (data cannot be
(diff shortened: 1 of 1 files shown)


rfcs/reuse-insert-permission-in-action.md
```diff
@@ -0,0 +1,67 @@
...
+#### Proposed solution
+
+Introduce a new field in insert permission called `admin_only`, this
+will remove the insert mutation field for the table from `mutation_root` for
+that role but the mutation can still be accessed when `admin-secret` is
+specified. The action's handler can now specify the `admin-secret`, set
+`x-hasura-role` as the user's role and execute the request so that the
+constraints defined on the role in insert permissions are enforced.
+
+Currently `admin-secret` + `x-hasura-role` is used in GraphiQL to mimic a
```
tirumaraiselvan: This would be a breaking change.

rfcs/reuse-insert-permission-in-action.md
```diff
@@ -0,0 +1,67 @@
...
+is defined, the `insert_message` mutation is generated which can be directly
+used by anyone with the `user` role and thus bypassing the validation of the
+`message` content.
+
+This problem is originally reported
+[on discord](https://discordapp.com/channels/407792526867693568/516590536875048990/685462912420282546).
+
+#### Proposed solution
+
+Introduce a new field in insert permission called `admin_only`, this
```
tirumaraiselvan: How is the solution affected by https://github.com/hasura/graphql-engine/blob/233ab645993ec08e981ca55d4bc8414a01bbd944/rfcs/disable-query-and-subscription-root-fields.md

Does it make sense to hide fields under `mutation` also? In that case, we do not need `admin_only` in the permission but we will need a header like `x-hasura-include-hidden-fields: true` or something?

tirumaraiselvan:
implemented in https://github.com/hasura/graphql-engine/pull/4224

coco98:
@0x777 Can you add a link to this on your RFC: https://docs.google.com/spreadsheets/d/1X3iKfmTxRq4Sfqqq23C2tp1Wp4sHFELQDEx7TSn2DhU/edit#gid=0

Let's close this PR?

PR #4122, opened 2020-03-16
include postgres version of the database while sending telemetry data

codingkarthik:
### Changelog

### Affected components

- [ X ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)
Telemetrics

#### Catalog upgrade

- [ X ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/App.hs
```diff
@@ -166,7 +167,11 @@ initialiseCtx hgeCmd rci = do
   eDbId <- liftIO $ runExceptT $ Q.runTx pool (Q.Serializable, Nothing) getDbId
   dbId <- either printErrJExit return eDbId
 
-  return (InitCtx httpManager instanceId dbId loggers connInfo pool, initTime)
+  -- get the pg version
```
0x777: You can use [`serverVersion`](https://github.com/hasura/pg-client-hs/blob/master/src/Database/PG/Query/Transaction.hs#L251) to get the underlying Postgres version.

There seems to be a lot of `runExceptT $ runTx` followed by `printErrJExit`, so might as well clean up this section by having a function as follows:
```haskell
-- | runs a transaction, on error, it is logged and the program is aborted
runTxIO :: Q.Pool -> Q.IsolationLevel -> Tx e a -> IO a
runTxIO = ...
```

server/src-lib/Hasura/App.hs
```diff
@@ -112,6 +112,7 @@ data InitCtx
   , _icLoggers     :: !Loggers
   , _icConnInfo    :: !Q.ConnInfo
   , _icPgPool      :: !Q.PGPool
+  , _icPgVersion   :: !Text
```
0x777: Use a `newtype` wrapper for `PGVersion`

codingkarthik:
I have made the changes

Issue #4117, opened 2020-03-16
reload_remote_schema does not work if remote schema is in inconsistent metadata


tirumaraiselvan:
### Description

`reload_remote_schema` does not work on remote schemas which are inconsistent (i.e. they were not available on startup). Considering prior discussions (see: #3792 ), this is actually a bug because even `reload_metadata` does not refresh remote schemas. There is no way to resolve an inconsistent remote schema without restarting Hasura.

### Repro:

1. Run hasura (any version > 1.0.0)
2. Add a remote schema. E.g. this: https://github.com/hasura/preview-remote-joins/tree/master/graphql-services/weather
3. Stop hasura and remote schema
4. Start hasura -> remote schema becomes inconsistent
5. curl -d'{"type": "reload_remote_schema", "args":-schema" }}' localhost:8080/v1/query -> throws an error: `{"path":"$.args","error":"remote schema with name \"my-remote-schema\" does not exist","code":"not-exists"}` 

**Note for versions prior to v1.1.0**: `reload_metadata` API would actually fetch inconsistent remote schemas and hence would resolve the inconsistency (although it wouldn't refresh consistent remote schemas).

tirumaraiselvan:
Fixed via https://github.com/hasura/graphql-engine/pull/4141

PR #4141, opened 2020-03-18
option to reload remote schemas in 'reload metadata' (fix #3792, #4117)

### LINKED ISSUE #3792: reload_metadata API should take optional argument to reload Remote Schemas
The `reload_metadata` API does not reload Remote Schemas as it is a costly step (and might be unnecessary). For reloading remote schemas, there is the explicit `reload_remote_schema` API.

But there might be a need to `reload_metadata` while also reloading Remote Schemas. Hence, an optional argument could be provided to `reload_metadata` like `reload_remote_schemas: true` to achieve this.

Side note: While writing down this issue, I noticed a small inconsistency in the behaviour (before v1.1):

1. Suppose you reload_metadata or reload_remote_schema in one Hasura instance, this changed the schema cache and triggered a postgres event to refresh other instances.
2. The other instances build schema cache from scratch and hence have up-to-date schema. Whereas the originator instance may not have up-to-date remote schema or up-to-date postgres schema (as it only did a partial reload).

@lexi-lambda  Is this issue moot after v1.1 because of incremental schema cache building?

rakeshkky:
### Description

Checkout the updated `reload_metadata` API reference [here](https://github.com/hasura/graphql-engine/blob/9dc6f9c53bcc0d202e95968972888a6219dca1d7/docs/graphql/manual/api-reference/schema-metadata-api/manage-metadata.rst#id5).

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

Fix #3792 & #4117 

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


console/src/components/Services/Settings/Actions.js
```diff
@@ -37,7 +37,9 @@ const getInconsistentObjectsQuery = {
 
 const reloadCacheQuery = {
   type: 'reload_metadata',
-  args: {},
+  args: {
+    reload_remote_schemas: true,
```
tirumaraiselvan: @wawhal Perhaps this should only be set if there are one ore more remote schemas which are inconsistent? Otherwise, this can get quite needlessly heavy operation.

rikinsk:
@wawhal It seems we have a reload metadata option on the inconsistenet metadata page
Screenshot from 2020-03-20 13-53-11

Lets just change the behaviour of this button in case there are inconsistent remote schemas

console/src/components/Services/Settings/Actions.js
wawhal: @tirumaraiselvan @rakeshkky  can the server not make this check? I will have to make an extra network request to get this information because we fetch remote-schemas/event-triggers/actions only in their respective routes.

I was thinking that server might have access to the information about number of remote schemas so it'd be better done there.

I can do it if you absolutely cannot. Let me know.

tirumaraiselvan: The server cannot by the API design itself. Without that arg, the API behaviour is to ignore remote schemas. And with that arg, the API behaviour is to not ignore remote schemas. There is no way to say "reload remote schemas if remote schemas are inconsistent".

I am not sure I understand the problem. The `get_inconsistent_metadata` API should tell you if there are inconsistent remote schemas right?

wawhal: I am saying, people might want to reload their metadata (including remote schemas) generally. So this must not be tied to the inconsistent objects.

I have done this now.

Screenshot from 2020-03-22 21-53-58

Similar for the inconsistent metadata page. Is this okay? They'll check the flag if they want to reload remote schemas as well. One more improvement we can do is, if there's an inconsistent remote schema, we can have the checkbox checked by default.

tirumaraiselvan: Yep, this is perfect.

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
```diff
@@ -85,9 +85,14 @@ runReloadRemoteSchema
...
-  void $ onNothing (Map.lookup name rmSchemas) $
-    throw400 NotExists $ "remote schema with name " <> name <<> " does not exist"
+  sc <- askSchemaCache
+  case Map.lookup name (scRemoteSchemas sc) of
+    Nothing -> if name `elem` getInconsistentRemoteSchemas (scInconsistentObjs sc) then
+                 throw400 NotSupported $
+                 "remote schema with name " <> name <<> " is in inconsistent state; cannot reload"
+               else throw400 NotExists $
+                    "remote schema with name " <> name <<> " does not exist"
+    Just _ -> pure ()
```
lexi-lambda: Is it necessary to fail if the remote schema is inconsistent? That seems unnecessary, and it’s certainly unhelpful—maybe someone is trying to reload the remote schema precisely *because* it’s inconsistent! So I’m wondering if we can go with a simpler check like this:

```suggestion
  remoteSchemas <- getAllRemoteSchemas <$> askSchemaCache
  unless (name `elem` remoteSchemas) $ throw400 NotExists $
    "remote schema with name " <> name <<> " does not exist"
```

Would that cause any problems?

rakeshkky: Console won't show up a remote schema in the UI if it is present in inconsistent metadata. 
> maybe someone is trying to reload the remote schema precisely because it’s inconsistent!

The Idea is the `reload_remote_schema` API should only refresh the consistent remote schema. Use `reload_metadata` with `reload_remote_schema: true` to refresh all remote schemas along with inconsistent ones. So, the explicit error message is thrown when the remote schema is inconsistent.

docs/graphql/manual/api-reference/schema-metadata-api/manage-metadata.rst
```diff
@@ -77,9 +77,28 @@ table using ``psql`` and this column should now be added to the GraphQL schema.
...
+   :header-rows: 1
+
+   * - Key
+     - Required
+     - Schema
+     - Description
+   * - reload_remote_schemas
+     - false
+     - Boolean
+     - If set to ``true``, all remote schemas' (including inconsistent ones) cached GraphQL schemas are refreshed (default: ``false``)
```
0x777: @rakeshkky Why do we not reload the remote schemas when we do `reload_metadata`? Any particular reason?

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
0x777: > Console won't show up a remote schema in the UI if it is present in inconsistent metadata.

We can improve this in the console. 

> That seems unnecessary, and it’s certainly unhelpful—maybe someone is trying to reload the remote schema precisely because it’s inconsistent

I completely agree.

rikinsk: > Console won't show up a remote schema in the UI if it is present in inconsistent metadata.

This is something console will be changing soon. It doesnt make sense to hide all inconsistent objects.

docs/graphql/manual/api-reference/schema-metadata-api/manage-metadata.rst
rakeshkky: @0x777 https://github.com/hasura/graphql-engine/issues/3792#issue-556619670

server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
rakeshkky: I'll commit the suggestion.

console/src/components/Services/Settings/MetadataStatus/MetadataStatus.js
```diff
@@ -161,25 +148,23 @@ const MetadataStatus = ({ dispatch, metadata }) => {
...
-            Reload metadata
-          </Button>
+        <div
+          className={`${metaDataStyles.display_flex} ${metaDataStyles.add_mar_top_small}`}
+        >
           <Button
             color="red"
             size="sm"
-            className={metaDataStyles.add_mar_top_small}
+            className={`${metaDataStyles.add_mar_right}`}
```
beerose: One day I'll create an eslint rule for that.

Issue #91, opened 2018-07-11
materialised views are not listed in console


0x777:
`information_schema.tables` from where we currently read untracked tables does not list materialized views. We should probably switch to something more Postgres specific

mterron:
pg_class probably have what you are looking for:
```sql
SELECT n.nspname as "Schema",
  c.relname as "Name",
  CASE c.relkind WHEN 'r' THEN 'table' WHEN 'v' THEN 'view' WHEN 'm' THEN 'materialized view' WHEN 'f' THEN 'foreign table' WHEN 'p' THEN 'table' END as "Type",
  pg_catalog.pg_get_userbyid(c.relowner) as "Owner"
FROM pg_catalog.pg_class c
     LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
WHERE c.relkind IN ('r','p','f','v','m')
      AND n.nspname <> 'pg_catalog'
      AND n.nspname <> 'information_schema'
      AND n.nspname !~ '^pg_toast'
  AND pg_catalog.pg_table_is_visible(c.oid)
ORDER BY 1,2;
```

eads:
👍 to this PR. Materialized views are a powerful tool for data analysis and it's a bummer I can't access them from Hasura currently.

0x777:
@eads as a workaround you can create a regular view as follows: 
```sql
create view somename as select * from materialized_view;
```
You can now access data from the view `somename` which in turn fetches data from the materialized view.

revskill10:
Does materialized view automatically get updated when we mutate the related tables ?

amesas:
Nope by default.

You have to call REFRESH VIEW
https://www.postgresql.org/docs/11/static/sql-refreshmaterializedview.html

We use triggers with  REFRESH CONCURRENTLY in some scenarios and a "CRON" in others.

revskill10:
@amesas Could hasura automatically handle Refreshing view for us ?

sfcgeorge:
It's not just showing untracked that doesn't work, even if you create the materialized view within Hasura SQL tab and click "Track" it doesn't show up. So I'll use the view wrapper trick.

karibertils:
+1 Would love to see this implemented

@0x777 hack is useful until then :smile:

nagautier:
:+1:  Hack works fine, but it would be cool to have materialized views discovered :).

pksunkara:
[...]

PR #4155, opened 2020-03-20
server: add tests for track_table of a materialized view

nicuveo:
### Description
In the context of #91, we discovered that materialized views were already "automagically" supported; to ensure we don't regress on this accidental but welcome change, this patch adds simple tests.

This is basically just a copy of `track_untrack_table`, but for materialized views.

### Affected components
- [ ] Tests


nicuveo:
Ah, gotcha; I'll pay attention to it in future PRs!

PR #4182, opened 2020-03-24
Improve performance of replace_metadata on tracking tables. Closes #3802

### LINKED ISSUE #3802: Tracking a large number of tables in a single metadata load query is slow / doesn't work
Replicate:

1. Running hasura/graphql-engine:v1.1.0-beta.2
2. Load this sample schema in postgres (sample.sql.txt).
3. Import the attached metadata json
 (big-metadata.json.txt)

[big-metadata.json.txt](https://github.com/hasura/graphql-engine/files/4133275/big-metadata.json.txt)
[sample.sql.txt](https://github.com/hasura/graphql-engine/files/4133276/sample.sql.txt)

Note: This doesn't affect GraphQL performance for a large number of tables, just the initial metadata load for tracking new tables.

jberryman:
On the 1000 table track case this went from >20min to 8 seconds, the
bottleneck being all the former calls (for each table) to
buildSchemaCache.

I don't have high confidence this is all correct, so this could definitely use a close look. Trying to figure out a larger but better refactoring would have taken me quite a bit longer (detective work, documenting a bunch of things after asking the folks who know, etc)

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


jberryman:
Oh should also ask: is 8 seconds acceptable for the 1000 table track example? I'm sure it could be made better with some unknown amount of engineering

server/src-lib/Hasura/RQL/DDL/Metadata.hs
```diff
@@ -142,11 +146,16 @@ applyQP2 (ReplaceMetadata _ tables functionsMeta
...
+    
+    -- NOTE: this is essentially a more efficient bulk 'Schema.trackExistingTableOrViewP2':
+    withBuildSchemaCache $ do
+      typeMap <- GC._gTypes . scDefaultRemoteGCtx <$> askSchemaCache
+      indexedForM tables $ \TableMeta{..} -> do
+        -- It should be okay to do this check against old cache as we work through new tables; we've 
+        -- already checked for duplicates in the track table payload:
+        GS.checkConflictingNode typeMap $ GS.qualObjectToName _tmTable
+        saveTableToCatalog _tmTable _tmIsEnum _tmConfiguration
+        return (MOTable _tmTable)
```
rakeshkky: I think we don't need to build schema cache here and check for inconsistent tables/views. Ideally in replace metadata, we should be _only saving the metadata in the catalog_ and do `buildSchemaCacheStrict` at the end. 

Define a new function with name like `saveMetadata` with definition as follows
```haskell
saveMetadata :: (MonadTx m, HasSystemDefined m) => ReplaceMetadata -> m ()
```
It should only insert the metadata items in the catalog. Move the code from `withPathK "tables"` to `forM_ actions .. ` snippet to this function. Also, we should be only using `addRemoteSchemaToCatalog` instead of `addRemoteSchemaP2` for remote schemas. Eventually, `applyQP2` function will look like

```haskell
applyQP2 replaceMetadata = do
  liftTx clearMetadata
  saveMetadata replaceMetadata
  buildSchemaCacheStrict
  pure successMsg
```

jberryman: Thanks, Rakesh. Do you think it makes more sense for me to reassign this issue to you? Or maybe to open a new issue that is about a bigger refactoring of replace_metadata?

I'm happy to attempt what you're describing, but with the state of documentation and my unfamiliarity with this part of the codebase I wonder if that makes sense. 

Let me know what you think.

lexi-lambda: I agree with @rakeshkky—I don’t think we should worry about building the schema cache at all until after we’ve dropped/recreated all the catalog information. I think what you have is mostly fine, you can just drop `withBuildSchemaCache`. We rebuild the schema cache again once we’re done, anyway.

rakeshkky: @jberryman No worries, I'll push a commit for the changes I requested. Please make sure I have write permissions to your fork. :)

jberryman: Cool! I invited you as a "collaborator" @rakeshkky . Hopefully that did it.

Appreciate the attention here; with all that code being in `CacheRWM` it took a lot of staring to decide that the small change I made was probably okay. It seemed as though there might be many places where we do some validation against what would be a stale schema cache, but like I said I'm still getting acquainted with stuff here.

lexi-lambda: Yeah, there’s a lot going on. I’m sure `replace_metadata` could be cleaned up, but I don’t think it absolutely has to be done right now.

One thing that stands out to me is that we call `buildSchemaCacheStrict` after dropping the old metadata but before creating the new metadata. This has the effect of purging the metadata cache, so even if the new metadata is mostly the same as the old one, it will reprocess everything.

I’d hesitate to just delete the call without ensuring some of this code isn’t incorrectly relying on it, but it’s something we could improve if we decide to clean things up.

rakeshkky: @jberryman I've pushed a couple of commits. Please test this against the table set of ~1000 and confirm the replace metadata is taking minimal time.

rakeshkky: @lexi-lambda
If we do `buildSchemaCacheStrict` only at the end of dropping metadata and creating the new metadata, the process of building schema cache will be totally **incremental**! And that is how the "v2 migrations workflow" meant to be which uses the `replace_metadata` API. Now the API will be even more faster. 

Using CLI v2-migrations flow:
I have 1000 tables before and I'll edit my metadata file to track one more table. I'll apply the metadata, then server only builds the cache for the extra table since it is new and older tables' metadata ain't changed. 
I would like to know your thoughts. :slightly_smiling_face:

lexi-lambda: I very much agree! The current approach of calling `buildSchemaCacheStrict` after dropping the old metadata is quite a bad deoptimization, and doing things incrementally is much better.

Sadly, we’ll still pay a nontrivial cost on every `replace_metadata` call, since we still have to *load* the whole schema into memory from Postgres. That query will take time, reading and parsing it into Haskell will take time, and checking what has/has not changed will take time. Still, I expect it to be an improvement! I just don’t know how high the remaining overhead will be.

jberryman:
Thanks @rakeshkky ! LGTM I think

0x777:
Sorry I messed up the changelog, fixing it shortly!

Issue #3867, opened 2020-02-12
support token expiry in webhook auth mode


0x777:
In webhook auth mode, if a websocket connection is established with a token and if the token expires at a later time, the connection will still continue to work as graphql-engine doesn't know that the token has expired. 

We already handle token expiry in JWT mode (by forcefully closing the websocket connection after the time specified with `exp`), so we just need to provide a means to specify expiry time in webhook mode. The `Expiry`/`Cache-Control` HTTP headers seem to be a good fit, we already handle expiry of JWKs with these headers, so should be straightforward to use it for webhook auth too.

itsarawutsiripak34:
Google Chrome

0x777:
closed with #4196

PR #4196, opened 2020-03-26
server: add support for webhook connection expiration

nicuveo:
### Description
This implements #3867: a webhook can now specify how long a connection should remain active, either by using a Cache-Control field containing a relative "max-age" duration, or an Expires field that contains an absolute date. Cache-Control takes precedence if both are present and successfully parsed. If neither is present, the connection never expires, as it used to.

If parsing a field fails, a warning is logged, and that field is ignored. If no expiration time information is found, the code also assumes the connection should not expire.

### Changelog
While this is user-facing, this seems a bit too minor to be part of the Changelog. What do you think?

### Affected components
- Server
- Tests

### Solution and Design
The only notable point is that I moved some of the code around. I moved webhook-specific code to a dedicated file, and moved header parsing function to a common utility file. I am not entirely satisfied with the result, even if it fulfills the goal of avoiding duplication of the parsing, and I'm open to suggestions.

### Steps to test and verify
The python tests, if run with the webhook server, will test a variety of configurations (no expiry headers, either of them, both, and also test that we fallback to no timeout if we fail to parse a header).

### Breaking changes
This might be a breaking change, for users that already had those headers set. That seems unlikely?

### Limitations
The documentation should be updated to reflect the change. I don't know if I should do it in that same PR though?


server/src-lib/Hasura/Server/Auth/Utils.hs
```diff
@@ -0,0 +1,52 @@
...
+
+data ExpiryTime = AbsoluteTime UTCTime
+                | RelativeTime NominalDiffTime
+
+
+-- | Extracts a relative expiry time from the CacheControl header.
+timeFromCacheControlHeader
+  :: (Monad m)
+  => Maybe T.Text         -- ^ the header
+  -> (T.Text -> m ())     -- ^ what should be done if parsing fails
```
lexi-lambda: This seems like a curious way to model this to me! Why not give these functions the following simpler type signature?

```haskell
timeFromCacheControlHeader, timeFromExpiresHeader :: MonadError Text m => Text -> m ExpiryTime
```

Then the caller can figure out what to do in the case of a `Maybe`, since neither of these functions do anything interesting when `Nothing` is passed, anyway, and the caller can instantiate `m` to `Either Text` if they want to do something custom on errors.

server/src-lib/Hasura/Server/Auth/Utils.hs
```diff
@@ -0,0 +1,52 @@
...
+  where timeFmt = "%a, %d %b %Y %T GMT"
+
+
+toUTCTime :: (MonadIO m) => ExpiryTime -> m UTCTime
+toUTCTime (AbsoluteTime t) = return t
+toUTCTime (RelativeTime d) = addUTCTime d <$> liftIO getCurrentTime
+
+toNominalDiffTime :: (MonadIO m) => ExpiryTime -> m NominalDiffTime
+toNominalDiffTime (RelativeTime d) = return d
+toNominalDiffTime (AbsoluteTime t) = diffUTCTime t <$> liftIO getCurrentTime
```
lexi-lambda: I would inline both of these functions at their use sites, since they’re each only used once, and I think it’s useful to see the `getCurrentTime` call at the use sites to make it clear what’s going on.

server/src-lib/Hasura/Server/Auth/Utils.hs
```diff
@@ -0,0 +1,52 @@
...
+import           Data.Time.Format         (defaultTimeLocale, parseTimeM)
+
+import qualified Data.String.Conversions  as CS
+import qualified Data.Text                as T
+
+import           Hasura.Prelude
+
+
+data ExpiryTime = AbsoluteTime UTCTime
+                | RelativeTime NominalDiffTime
```
lexi-lambda: I don’t think this sum type is useful:

  1. `timeFromCacheControlHeader` has no reason not to return `NominalDiffTime`, since it can never return an `AbsoluteTime`. Returning an `ExpiryTime` makes its type less precise for relatively little gain.

  2. By the same reasoning, `timeFromExpiresHeader` can return a `UTCTime`.

  3. The primary advantage of the sum type seems to be homogenization at call-sites. But both call-sites immediately convert the `ExpiryTime` to either a `UTCTime` or a `NominalDiffTime`, so the values get wrapped only to be immediately unwrapped!

So I think we can drop this sum type and handle each case explicitly at each use site.

server/src-lib/Hasura/Server/Auth/Utils.hs
```diff
@@ -0,0 +1,52 @@
...
+-- | Extracts a relative expiry time from the CacheControl header.
+timeFromCacheControlHeader
+  :: (Monad m)
+  => Maybe T.Text         -- ^ the header
+  -> (T.Text -> m ())     -- ^ what should be done if parsing fails
+  -> m (Maybe ExpiryTime)
+timeFromCacheControlHeader header onError =
+  onMaybe header $ \h -> case parseMaxAge h of
+    Left e       -> onError (T.pack e) >> return Nothing
+    Right maxAge -> return $ Just $ RelativeTime $ fromInteger maxAge
```
lexi-lambda: This converts the result of `parseMaxAge` to a `NominalDiffTime`, which is good! But arguably `parseMaxAge` should just return a `NominalDiffTime` in the first place, so I think you can eliminate this function completely and change the type of `parseMaxAge` directly.

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+    logWebHookResp logLevel mResp message =
+      logger $ WebHookLog logLevel (Just statusCode)
+        url method Nothing (bsToTxt . BL.toStrict <$> mResp) message
+
+    getExpiryTime ccHeader eHeader = do
+      cct <- timeFromCacheControlHeader ccHeader $ logWarn . Just
+      expTime <- case cct of
+        Just t  -> return $ Just t
+        Nothing -> timeFromExpiresHeader eHeader $ logWarn . Just
+      onMaybe expTime $ fmap Just . toUTCTime
```
lexi-lambda: You might be able to get more bang for your buck by using `MaybeT`’s `Alternative` instance here. You could do something like this:

```haskell
mkUserInfoFromResp =
    ...
    expiration <- runMaybeT (timeFromCacheControl headers <|> timeFromExpires headers)
    ...
  where
    timeFromCacheControl headers = do
      header <- afold $ Map.lookup "Cache-Control" headers
      duration <- parseMaxAge header `onLeft` \err -> logWarn err *> empty
      addUTCTime duration <$> liftIO getCurrentTime

    timeFromExpires headers = do
      header <- afold $ Map.lookup "Expires" headers
      timeFromExpiresHeader header `onLeft` \err -> logWarn err *> empty

    -- I think this function should go in Hasura.Prelude; I’ve wanted it on several occasions.
    afold :: (Foldable t, Alternative f) => t a -> f a
    afold = getAlt . foldMap pure
```

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -155,35 +152,22 @@ updateJwkRef (Logger logger) manager url jwkRef = do
...
 
   -- first check for Cache-Control header to get max-age, if not found, look for Expires header
-  let cacheHeader   = resp ^? Wreq.responseHeader "Cache-Control"
-      expiresHeader = resp ^? Wreq.responseHeader "Expires"
-  case cacheHeader of
-    Just header -> getTimeFromCacheControlHeader header
-    Nothing     -> mapM getTimeFromExpiresHeader expiresHeader
+  let cacheHeader   = bsToTxt <$> resp ^? Wreq.responseHeader "Cache-Control"
+      expiresHeader = bsToTxt <$> resp ^? Wreq.responseHeader "Expires"
+  cct <- timeFromCacheControlHeader cacheHeader $ logAndThrowInfo . parseCacheControlErr
```
lexi-lambda: Please don’t use variable names like `cct` if you can help it! Our [coding style](https://github.com/hasura/graphql-engine/blob/950d3842e4662c2ff5d8d5b20ac64bf9f9387b80/server/CONTRIBUTING.md#naming) is to avoid abbreviations. A name like `cacheDuration` isn’t too long, I think, and it’s much easier to read!

server/src-lib/Hasura/Prelude.hs
```diff
@@ -80,6 +81,9 @@ onNothing m act = maybe act return m
 onJust :: (Monad m) => Maybe a -> (a -> m ()) -> m ()
 onJust m action = maybe (return ()) action m
 
+onMaybe :: Monad m => Maybe a -> (a -> m (Maybe b)) -> m (Maybe b)
+onMaybe m action = maybe (return Nothing) action m
```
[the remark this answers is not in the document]
nicuveo: I originally thought about `MaybeT`, but I wasn't sure how to make it readable: my main problem was the lack of `afold`: I ended up using `maybe` quite a lot within `MaybeT`, which wasn't really much better... Now fixed!

server/src-lib/Hasura/Server/Auth/Utils.hs
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/Utils.hs
nicuveo: Done! I remove `Hasura.Server.Auth.Utils`, and created instead `Data.Parser.Expires`.

server/src-lib/Hasura/Server/Auth/Utils.hs
nicuveo: Done! Yup, it definitely was an attempt at homogenizing call sites, which proved a bad idea in the end.

server/src-lib/Hasura/Server/Auth/Utils.hs
nicuveo: I didn't touch parseMaxAge in the end, but simply `fmap fromInteger` at the call site, which I think is acceptable?

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/JWT.hs
nicuveo: My bad!

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+  bool (ifGet u) (ifPost u) $ isPost t
+
+hookMethod :: AuthHook -> N.StdMethod
+hookMethod = bool N.GET N.POST . isPost . ahType
+
+
+-- | Extracts 'UserInfo' from the webhook answer.
+--   This uses the provided manager to make the call, extracts
+--   information such as response body and status, and delegates to a
+--   local 'mkUserInfoFromResp'.
```
lexi-lambda: I don’t think this comment is very helpful—it mostly just describes what the implementation does. It’s a documentation comment, so it probably ought to be useful to *callers* of `userInfoFromAuthHook`, but the fact that it delegates to a local function is just an implementation detail.

If I were to write a comment at all, I’d leave it at something as simple as this:

> Makes an authentication request to the given `AuthHook` and returns `UserInfo` parsed from the response, plus an expiration time if one was returned.

Mentioning that the second result is an expiration time is the most useful bit, since the purpose of the other arguments/results can be inferred from their types.

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+    postHdrsPayload = J.toJSON $ Map.fromList $ hdrsToText reqHeaders
+
+    logAndThrow err = do
+      unLogger logger $
+        WebHookLog LevelError Nothing (ahUrl hook) (hookMethod hook)
+        (Just $ HttpException err) Nothing Nothing
+      throw500 "Internal Server Error"
+
+    filteredHeaders = flip filter reqHeaders $ \(n, _) ->
+      n `notElem` commonClientHeadersIgnored
```
lexi-lambda: This is a style comment, so I will preface it with the disclaimer that this is subjective.

There is a Haskell style that seems to enjoy aggressively pulling out lots of expressions into `where` blocks (sometimes seemingly at random). Perhaps this is a little more aesthetically pleasing, but it harms readability: reading the function requires constantly bouncing around between the body and the `where` clause.

Moving nontrivial helper functions that represent meaningfully-distinct chunks of functionality into `where` clauses is great, and I wholeheartedly support doing that! But otherwise, I much prefer keeping related bits of logic close together in the code. Consider this rearranging of the logic in this function:

```haskell
userInfoFromAuthHook logger manager hook reqHeaders = do
  response <- (`onLeft` logAndThrow) =<< liftIO (try performHttpRequest)
  let status   = response ^. Wreq.responseStatus
      respBody = response ^. Wreq.responseBody
  mkUserInfoFromResponse logger (ahUrl hook) (hookMethod hook) status respBody
  where
    performHttpRequest = do
      let url = T.unpack $ ahUrl hook
      case ahType hook of
        AHTGet -> do
          let isCommonHeader header = (`elem` commonClientHeadersIgnored)
              filteredHeaders = filter (not . isCommonHeader . fst) reqHeaders
          Wreq.getWith (wreqOptions manager filteredHeaders) url
        AHTPost -> do
          let contentType = ("Content-Type", "application/json")
              headersPayload = Map.fromList $ hdrsToText reqHeaders
          Wreq.postWith (wreqOptions manager [contentType]) url $
            object ["headers" J..= headersPayload]
```

I’ve made the following changes:

  1. I moved most of the functionality into `performHttpRequest`, which means the function body is basically just wiring code that glues various pieces together.

  2. Since `performHttpRequest` is now a meaningful chunk of functionality, it can be given an equally-meaningful name.

  3. I’ve moved the auxiliary bindings into `let` blocks, defined as close to where they’re needed as possible. This makes it possible to read the code by just following the control flow.

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+      object ["headers" J..= postHdrsPayload]
+    mkOptions = wreqOptions manager
+    contentType = ("Content-Type", "application/json")
+    postHdrsPayload = J.toJSON $ Map.fromList $ hdrsToText reqHeaders
+
+    logAndThrow err = do
+      unLogger logger $
+        WebHookLog LevelError Nothing (ahUrl hook) (hookMethod hook)
+        (Just $ HttpException err) Nothing Nothing
+      throw500 "Internal Server Error"
```
lexi-lambda: This isn’t a very useful error message! `throw500` will already raise an internal server error, so it would be nice to be a little more descriptive here. Maybe something like “authentication request to webhook failed”?

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+mkUserInfoFromResp (Logger logger) url method statusCode respBody
+  | statusCode == N.status200 =
+    case eitherDecode respBody of
+      Left e -> do
+        logError
+        throw500 $ "Invalid response from authorization hook: " <> T.pack e
+      Right rawHeaders -> getUserInfoFromHdrs rawHeaders
+
+  | statusCode == N.status401 = do
+    logError
```
lexi-lambda: Do we need to log an error here? If the authentication hook responds with 401, that seems like it’s just operating as intended, not an error (the error is on the user’s part, not the server). (This is a genuine question, I’m not 100% sure I understand what this functionality does.)

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+        (Just $ HttpException err) Nothing Nothing
+      throw500 "Internal Server Error"
+
+    filteredHeaders = flip filter reqHeaders $ \(n, _) ->
+      n `notElem` commonClientHeadersIgnored
+
+
+-- | Processes the result of the wehook call, and tries to construct a
+--   'UserInfo'. The expiration time is retrieved from either the
+--   CacheControl or Expires header.
```
lexi-lambda: I’d delete this comment, personally—I think it mostly just describes what the function does, and I think the name `mkUserInfoFromResponse` captures most of that.

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+  | AHTPost
+  deriving (Eq)
+
+instance Show AuthHookType where
+  show AHTGet  = "GET"
+  show AHTPost = "POST"
+
+isPost :: AuthHookType -> Bool
+isPost AHTGet  = False
+isPost AHTPost = True
```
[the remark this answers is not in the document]
nicuveo: ACK, fair!

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: SGTM! I had kept the code pretty close to its original style, partly because, well, don't fix what ain't broke: code often is as it is for a good reason, and I didn't want to change too much of it just for style reasons. ^^

lexi-lambda: Ah, you are right—that bit was part of the code that was moved, not something new! I was confused by the diff. It sounds like my argument was misplaced, so mea culpa!

nicuveo: No, no, that's valid criticism! I had already made some small changes, I'll apply your suggestions too. :)

server/src-lib/Hasura/Server/Auth/WebHook.hs
```diff
@@ -0,0 +1,150 @@
...
+  = AuthHookG
+  { ahUrl  :: !a
+  , ahType :: !b
+  } deriving (Show, Eq)
+
+type AuthHook = AuthHookG T.Text AuthHookType
+
+withAuthHook :: AuthHook -> (T.Text -> a) -> (T.Text -> a) -> a
+withAuthHook (AuthHookG u t) ifGet ifPost =
+  bool (ifGet u) (ifPost u) $ isPost t
```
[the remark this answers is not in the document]
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: Done!

server/src-lib/Hasura/Server/Auth/WebHook.hs
nicuveo: That's a good point... but there again, this is existing code, that I didn't want to touch. Thinking about it: we are going to receive a 401 every time authentication fails... perhaps a warning is enough?

PR #4204, opened 2020-03-26
console: support materialized views

nicuveo:
### Description
This is the frontend part of #91. **WARNING**: it is an ugly workaround!

The problem is as follows: this query used `information_schema` to retrieve information about the underlying postgres database. Unfortunately, neither `information_schema.tables` nor `information_schema.columns` contain data about materialized views. What this PR does is to change the query to instead rely on postgres internals, such as `pg_class` and `pg_attribute`. It has a few upsides; notably, retrieving comments is much easier now, since we have access to the oid of each element, and we can just use `col_description` and `obj_description`.

However, the previous version of that query returned *every single column* of the `columns` view, unmodified. To emulate that behaviour, I had to recreate the entire `columns` view in the query, while fixing it so that it includes materialized views.

I obtained the definition of `columns` by running `psql -E` and running `\dS+ information_schema.columns`. The only change is as follows:

before: `c.relkind = ANY (ARRAY['r'::"char", 'v'::"char", 'f'::"char", 'p'::"char"])`
after: `c.relkind = ANY (ARRAY['r'::"char", 'v'::"char", 'm'::"char"])`

### Affected components

- [x] Console

There are several things that are missing from this PR, in its current state:

- currently, the `view_info` part of the query is computed using the `information_schema.views` view, which... suffers from the exact same problem: not listing materialized views. I am not sure whether we actually want to treat them as views as far as the console is concerned, but that's something to check,
- I have **NOT** checked that the console properly supports this new `MATERIALIZED VIEW` table type everywhere; from my limited testing, it treats is as a view (the name is italicized for instance),
- I have **NOT** done regression testing beyond some very simple examples on my test database,
- I have **NOT** written new tests, nor done extensive tests in the console; I have simply tested, on one materialized view:
  - tracking it
  - untracking it
  - running a query against it in GraphiQL
  - checking that the data is correct (both rows and columns) in the Data tab

Ideally, this would be rewritten to call a backend function, and none of the postgres internals would be in the frontend. But in recent discussions we have agreed that it was preferable to do this workaround before doing the necessary server work.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -483,6 +483,8 @@ jobs:
     - run:
         name: test console
         command: .circleci/test-console.sh
+    - store_artifacts:
+        path: /build/_console_output/server.log
 
   # test server upgrade from last version to current build
   test_server_upgrade:
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,22 @@
+help:
+	@echo "available commands:"
+	@echo "  build:   builds console and server"
+	@echo "  console: builds console"
+	@echo "  server:  builds server"
+
+build: console server
+
+server:
+	cd server && cabal new-install && cabal new-build
+
+console:
+	cd console && npm ci && npm run server-build
+
+
+hs-tests:
+	cd server && cabal new-run -- 'test:graphql-engine-tests' --database-url='postgres://localhost:4200/hasura-tests'
+
+py-tests:
+	./run-py-tests.sh
+
+.PHONY: help build server console hs-tests py-tests
--- a/console/src/components/Services/Data/utils.js
+++ b/console/src/components/Services/Data/utils.js
@@ -417,113 +417,48 @@ FROM (
     ON pga.attrelid = pgc.oid
   LEFT OUTER JOIN (
     SELECT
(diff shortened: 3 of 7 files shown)


nicuveo:
An update on this:

I had used the definition of `information_schema.columns` from postgres 12, which is not backwards compatible. That's why the tests were failing on CircleCI: our tests use postgres 10.

I pushed a new commit that rewrites the query, based on how it was defined in [posgres 9.5](https://github.com/postgres/postgres/blob/REL9_5_STABLE/src/backend/catalog/information_schema.sql#L635), which is the oldest version we support. I stripped some of the fields that either had no information (as in: they were declared as `cast NULL as ...`) or that relied on internal functions that no longer exist in version 12.

Importantly, this is still a **WORK IN PROGRESS**. I am currently running all cypress tests on versions 12 and 10, and will perform some additional manual ones / will had cypress tests for materialized views.

nicuveo:
Sorry for the noise on this PR: I was blocked by what turned out to be #4250. I'll close this PR and open a new one when I have something a bit cleaner.

PR #4213, opened 2020-03-27
unlock event if event trigger info is not found in cache

tirumaraiselvan:
### Description

In the unexpected scenario in which table or event trigger info is not found in schema cache for a event, we were logging an error. We also need to unlock the event so that it can be reprocessed.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### DIFF
--- a/server/src-lib/Hasura/Events/Lib.hs
+++ b/server/src-lib/Hasura/Events/Lib.hs
@@ -259,7 +259,15 @@ processEventQueue logger logenv httpMgr pool getSchemaCache EventEngineCtx{..} =
       let meti = getEventTriggerInfoFromEvent cache e
       case meti of
         Nothing -> do
+          --  This rare error can happen in the following known cases:
+          --  i) schema cache is not up-to-date (due to some bug, say during schema syncing across multiple instances)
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -260,6 +260,8 @@ processEventQueue logger logenv httpMgr pool getSchemaCache EventEngineCtx{..} =
       case meti of
         Nothing -> do
           logQErr $ err500 Unexpected "table or event-trigger not found in schema cache"
+          liftIO . runExceptT $ Q.runTx pool (Q.RepeatableRead, Just Q.ReadWrite) $ unlockEvent e
```
tirumaraiselvan: To prevent starvation of other events due to corrupt schema cache, we can set this to retry after `5 mins` otherwise it will be picked immediately.

When we have exponential back-off retry settings, we can use that as well.

0x777: @tirumaraiselvan Can you document known scenarios where this could happen?

PR #4214, opened 2020-03-27
unlock the locked events while scale-down (close #3548)

### LINKED ISSUE #3548: manage locked events during scale-down
If hasura instances are scaled down, then it is possible that events that were being processed by aborting instances stay locked and stuck in pending.

codingkarthik:
- what:
  - While the HGE is shut down, it's possible that some of the events
    were fetched by the HGE and it will always remain in the locked state

### Changelog

### Affected components

- [ X ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3548 

### Solution and Design

Whenever events are fetched from the database, the fetched event IDs will be stored in a cache and after the event has been processed, the event ID will be removed from the cache. 

When a graceful shutdown is initiated, all the events that have been locked will be stored in the cache and then those events will be unlocked.

### Steps to test and verify

Create an event trigger and do mutations(preferably a lot, because we will require events to be fetched by the HGE while we initiate a graceful shutdown) and initiate a graceful shutdown.

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ X ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ X ] Yes
    - [ ] Not required
    - [ X ] Yes
    - [ ] Not required

#### GraphQL
- [ X ] No new GraphQL schema is generated
   
   

#### Breaking changes

- [ X ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


tirumaraiselvan:
> Earlier, the problem was solved by unlocking all events during the startup

I think for keeping the behavior predictable, we should still unlock all events during startup.

codingkarthik:
This issue is meant to solve that problem, right?

tirumaraiselvan:
This issue, more precisely, is to unlock events which are in process in an instance which is scaled down i.e. exited gracefully. 

Now, suppose a container did not crash gracefully, there will be events which are still unlocked. Hence, they still need to unlocked.

codingkarthik:
Yes, I understand that. But what about the multiple hasura instances running on the same db? If one hasura instance has locked some events and it's in the midst of processing those events and another hasura instance on startup will unlock those events and will start processing them(very likely that the events fetched by both the instances will be more or less the same), so it's possible that an event will be delivered twice

tirumaraiselvan:
That is fine. Our system is atleast-once delivery. We can tolerate duplicate deliveries of this kind.

codingkarthik:
Ah, yes I forgot about the atleast-once delivery thing 🤦‍♂ . I'll unlock the events on startup.

tirumaraiselvan:
Having a locked (and unprocessed) event is much worse. We have made this trade-off in the design.

codingkarthik:
Now, in light of the atleast-once delivery thing, I was thinking that why not process the events in a transaction and if the server is crashed then the transaction will rollback. The only problem with this approach is that some of the events which have been processed will also be rolled back, so those events will be processed again.

jberryman:
> Having a locked (and unprocessed) event is much worse. We have made this trade-off in the design.

It should be possible to do better than this (make some attempt to not redeliver during normal functioning), but if you think this is acceptable for now certainly happy to defer to you.

tirumaraiselvan:
@jberryman We can definitely improve on this (if it's not trivial then maybe in a separate PR). What are you thinking?

server/src-lib/Hasura/App.hs
```diff
@@ -269,27 +265,47 @@ runHGEServer ServeOptions{..} InitCtx{..} initTime = do
     asyncActionsProcessor (_scrCache cacheRef) _icPgPool _icHttpManager
 
   -- start a background thread to check for updates
-  _updateThread <- C.forkImmortal "checkForUpdates" logger $ liftIO $ 
+  _updateThread <- C.forkImmortal "checkForUpdates" logger $ liftIO $
```
jberryman: Aside: These trailing whitespace changes are a bummer, sorry. I've tried a plugin in vim that removed them but it ended up being really annoying for various reasons. I should try again. But It's not clear to me why the problem is trailing whitespace and not editors and tools that make a big deal of trailing whitespace...

server/src-lib/Hasura/App.hs
```diff
@@ -269,27 +265,47 @@ runHGEServer ServeOptions{..} InitCtx{..} initTime = do
...
 
   where
     prepareEvents pool (Logger logger) = do
       liftIO $ logger $ mkGenericStrLog LevelInfo "event_triggers" "preparing data"
-      res <- runTx pool unlockAllEvents
+      res <- liftIO $ runTx pool (Q.Serializable, Nothing) unlockAllEvents
       either printErrJExit return res
 
+    -- | shutdownEvents will be triggered when a graceful shutdown has been inititiated, it will
+    -- get the locked events from the event engine context and then it will unlock all those events.
```
jberryman: Could we add a little more on the implications here: "this means some pending events may get processed twice, etc"

server/src-lib/Hasura/Events/Lib.hs
```diff
@@ -169,20 +177,20 @@ retryAfterHeader = "Retry-After"
...
   return $ EventEngineCtx{..}
 
 -- | Service events from our in-DB queue.
 --
 -- There are a few competing concerns and constraints here; we want to...
 --   - fetch events in batches for lower DB pressure
 --   - don't fetch more than N at a time (since that can mean: space leak, less
---     effective scale out, possible double sends for events we've checked out 
---     on exit (TODO clean shutdown procedure))
+--     effective scale out
```
jberryman: Hm I guess this TODO isn't actually done here if we continue to unlock all events everytime an HGE comes online...

jberryman:
Sorry, so after going through this: I don't see the point in this change if hge is going to unlock all events on startup. Isn't that strictly equivalent?

tirumaraiselvan:
> Sorry, so after going through this: I don't see the point in this change if hge is going to unlock all events on startup. Isn't that strictly equivalent?

I had a discussion above with @codingkarthik about this earlier in this PR. Unlocking events at startup is still required because instances can crash un-gracefully. Then what's the point of this PR? This is mostly to solve the issue specifically with managed scale down i.e. when you know that events in-memory need to be unlocked. Suppose you brought down num of hasura instances from 10 to 1, then the events in-memory will stay locked till a new hasura instances is spawned.

jberryman:
Cool, it would be good also to have a comment to that effect (that although we unlock on startup we need graceful shutdown in order to handle the scale down use case)

And I guess point out that this doesn't work for a crash or impatient scaledown...

@codingkarthik Tiru suggests opening a new ticket that will tackle how we can manage redelivery in a smarter way than simply unlocking all events on startup . One idea is simply to mark items as fresh after some timeout.

server/src-lib/Hasura/Events/Lib.hs
codingkarthik: Ah yes, I removed it when I was not unlocking all the events at the startup, I'll add it again

server/src-lib/Hasura/App.hs
codingkarthik: Yes, I'll add that

codingkarthik:
@jberryman I'll add a comment about why we are unlocking all the locked events at startup. 

I'll create a new issue for the crash shutdown locked events and raise a PR.

CHANGELOG.md
```diff
@@ -132,3 +132,4 @@ Read more about it in the [docs](https://hasura.io/docs/1.0/graphql/manual/auth/
 - server: check expression in update permissions (close #384) (rfc #3750) (#3804)
 - console: show pre-release update notifications with opt out option (#3888)
 - console: handle invalid keys in permission builder (close #3848) (#3863)
+- server: unlock events the locked events fetched by the HGE when the HGE is scaled down (close #3548)
```
marionschleifer, line 135: I'm not sure I understand what this means? How is the user affected by this?

tirumaraiselvan: I would say something like:  "manage processing events when HGE instance is gracefully shutdown"

tirumaraiselvan: @codingkarthik Also, this has been added to `v1.2.0-beta.1` changelog entry. Pls put this at the top `Next release`

server/src-lib/Hasura/App.hs
```diff
@@ -275,14 +271,48 @@ runHGEServer ServeOptions{..} InitCtx{..} initTime = do
...
+    -- It may happen that an event may be processed more than one time, an event that has been already
+    -- processed but not been marked as delivered in the db will be unlocked by `shutdownEvents`
+    -- and will be processed when the events are proccessed next time.
+    shutdownEvents :: Q.PGPool -> Logger Hasura -> EventEngineCtx -> IO ()
+    shutdownEvents pool (Logger logger) EventEngineCtx {..} = do
+      liftIO $ logger $ mkGenericStrLog LevelInfo "event_triggers" "unlocking events that are locked by the HGE"
```
tirumaraiselvan, line 305: should this be `unstructured`? Is this a new log-type: `event_triggers`?

tirumaraiselvan: Okay nvm. I think the backpressure changes has an `unstructured` kind that should probably also use `event_triggers`: https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Hasura/Events/Lib.hs#L236

PR #4221, opened 2020-03-28
Fix JSON(B) rendering in data browser (close #4201)

### LINKED ISSUE #4201: JSONB fields aren't presented adequately
JSON of Schrodinger:

Screen Shot 2020-03-26 at 11 05 37

One is a string (by mistake) and the other one is an object. Which one is the correct one?

Screen Shot 2020-03-26 at 11 05 46

Screen Shot 2020-03-26 at 11 05 55

You figure when you edit the row.

What you guys think about showing in the listing as it is shown in the edit?

I can make a PR if you agree.

sobrinho:
### Description

Hasura was rendering strings and objects as the same in browse rows, we couldn't differentiate one from another.

### Changelog

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#4201

### Solution and Design

Checking if the column is a JSON or JSONB type instead of only if the value is an object fix the issue.

Before (You couldn't tell that one of the rows is a string by mistake):

Screen Shot 2020-03-28 at 18 41 24

After (You can see the difference clearly):

Screen Shot 2020-03-28 at 18 37 46

### Steps to test and verify

Create one json and one jsonb columns, create two rows, one with both values as `{"object":true}` and another one with both values as `"{\"string\": \"true\"}"`.

You can't tell the difference in `Browse Rows` before and you should after this fix.

None.

### Server checklist

None.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/console/src/components/Services/Data/TableBrowseRows/ViewRows.js
+++ b/console/src/components/Services/Data/TableBrowseRows/ViewRows.js
@@ -581,7 +581,11 @@ const ViewRows = ({
           } else if (rowColumnValue === undefined) {
             cellValue = 'NULL';
             cellTitle = cellValue;
-          } else if (typeof rowColumnValue === 'object') {
+          } else if (
+            col.data_type === 'json' ||
+            col.data_type === 'jsonb' ||
+            rowColumnValue === 'object'
+          ) {
             cellValue = JSON.stringify(rowColumnValue, null, 4);
             cellTitle = cellValue;
           } else {
@@ -759,9 +763,7 @@ const ViewRows = ({
           <b className={styles.padd_small_right}>Selected:</b>
           {selectedRows.length}
           <button
-            className={`${styles.add_mar_right_small} btn btn-xs btn-default ${
-              styles.bulkDeleteButton
-            }`}
+            className={`${styles.add_mar_right_small} btn btn-xs btn-default ${styles.bulkDeleteButton}`}
             title="Delete selected rows"
             onClick={handleDeleteItems}
           >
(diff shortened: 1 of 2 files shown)


console/src/components/Services/Data/TableBrowseRows/ViewRows.js
```diff
@@ -759,9 +763,7 @@ const ViewRows = ({
           <b className={styles.padd_small_right}>Selected:</b>
           {selectedRows.length}
           <button
-            className={`${styles.add_mar_right_small} btn btn-xs btn-default ${
-              styles.bulkDeleteButton
-            }`}
+            className={`${styles.add_mar_right_small} btn btn-xs btn-default ${styles.bulkDeleteButton}`}
```
sobrinho: This was done by the linter, not me.

beerose: Yup, this is proper formatting. It was wrongly formatted on the master.

PR #4238, opened 2020-03-30
add permissions for functions

tirumaraiselvan:
### Description

Since UDFs are commonly used with Hasura, adding docs to grant sample permissions for their access.

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


tirumaraiselvan:
I am not sure of the utility of this because turns out Postgres grants `execute` to `PUBLIC` (i.e. all roles) by default. And `execute` is the only privilege for a function: https://www.postgresql.org/docs/9.1/sql-createfunction.html

But, I guess no harm :man_shrugging:

PR #4242, opened 2020-03-30
skip reloading the schema cache if the "skip_cache_reload" is set to true

### LINKED ISSUE #4200: run_sql should take a flag to skip schema cache building
Currently, run_sql takes few flags: https://hasura.io/docs/1.0/graphql/manual/api-reference/schema-metadata-api/run-sql.html#run-sql-syntax

Only when `read_only` is `true` does it skip schema cache building but this only works for `SELECT` queries : https://github.com/hasura/graphql-engine/blob/b84db36ebb51acd5b51e1254c103f3097a7c2358/server/src-lib/Hasura/Server/Query.hs#L262

We should have a flag like `skip_cache_reload` that ignores building the cache for any kind of statements. This can be used for optimizing pure SQL statements and is especially important for migration v2 workflow where pure SQL migrations are applied followed by a `replace_metadata`

codingkarthik:
closes #4200 

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#4200 
#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ Read more about this command in the [docs](https://hasura.io/docs/1.0/graphql/ma
 
 (close #1156) (#3760)
 
-- Check expression in update permissions (close #384) (rfc #3750) (#3804) 
+- Check expression in update permissions (close #384) (rfc #3750) (#3804)
 
 - console: add multi select in browse rows to allow bulk delete (close #1739) (#3735)
 
@@ -73,4 +73,4 @@ Read more about this command in the [docs](https://hasura.io/docs/1.0/graphql/ma
 
   A new CLI migrations image is introduced to account for the new CLI workflow. If you're have a project with `version: 2` in `config.yaml`, you should use the new image: `hasura/graphql-engine:v1.2.0-cli-migrations-v2`. Mount the migrations at `/hasura-migrations` and metadata at `/hasura-metadata`.
 
-- server: Skip the schema cache building in `run_sql` request, if `skip_cache_reload` is set to `true`. (close #4200)
+- server: server: The `run_sql` API supports a new skip_cache_reload option to suppress reloading metadata from Postgres after executing the query. (close #4200)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,4 +72,5 @@ Read more about this command in the [docs](https://hasura.io/docs/1.0/graphql/ma
 - cli(migrations-docker): add support for v2 config (close #3969)
 
   A new CLI migrations image is introduced to account for the new CLI workflow. If you're have a project with `version: 2` in `config.yaml`, you should use the new image: `hasura/graphql-engine:v1.2.0-cli-migrations-v2`. Mount the migrations at `/hasura-migrations` and metadata at `/hasura-metadata`.
-  
+
+- server: Skip the schema cache building in `run_sql` request, if `skip_cache_reload` is set to `true`. (close #4200)
--- a/docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
+++ b/docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
@@ -215,6 +215,10 @@ Args syntax
      - false
      - Boolean
      - When set to ``true``, the query will be run in ``READ ONLY`` transaction access mode which means only ``select`` queries will be successful. This flag ensures that the GraphQL schema is not modified and is hence highly performant.
+   * - skip_cache_reload
+     - false
+     - Boolean
+     - When set to ``true``, the sql is executed and the GraphQL schema is not modified. This field should be set to ``true`` in ``READ ONLY`` queries or while migrating data, because after the migrations, :ref:`replace_metadata` will be called.
 
 Response
 ^^^^^^^^
--- a/server/src-lib/Hasura/RQL/DDL/Schema.hs
+++ b/server/src-lib/Hasura/RQL/DDL/Schema.hs
@@ -62,6 +62,7 @@ data RunSQL
   , rCascade                  :: !Bool
   , rCheckMetadataConsistency :: !(Maybe Bool)
   , rTxAccessMode             :: !Q.TxAccess
+  , rSkipCacheReload          :: !Bool
   } deriving (Show, Eq, Lift)
 
 instance FromJSON RunSQL where
@@ -70,6 +71,7 @@ instance FromJSON RunSQL where
     rCascade <- o .:? "cascade" .!= False
     rCheckMetadataConsistency <- o .:? "check_metadata_consistency"
     isReadOnly <- o .:? "read_only" .!= False
+    rSkipCacheReload <- o .:? "skip_cache_reload" .!= True
     let rTxAccessMode = if isReadOnly then Q.ReadOnly else Q.ReadWrite
     pure RunSQL{..}
 
@@ -83,6 +85,7 @@ instance ToJSON RunSQL where
         case rTxAccessMode of
           Q.ReadOnly  -> True
           Q.ReadWrite -> False
+      , "skip_cache_reload" .= rSkipCacheReload
       ]
 
 runRunSQL :: (MonadTx m, CacheRWM m, HasSQLGenCtx m) => RunSQL -> m EncJSON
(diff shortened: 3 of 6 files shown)


docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
```diff
@@ -215,6 +215,10 @@ Args syntax
      - false
      - Boolean
      - When set to ``true``, the query will be run in ``READ ONLY`` transaction access mode which means only ``select`` queries will be successful. This flag ensures that the GraphQL schema is not modified and is hence highly performant.
+   * - skip_cache_reload
+     - false
+     - Boolean
+     - When set to ``true``, the sql is executed and the GraphQL schema is not modified. This field should be set to ``true`` in ``READ ONLY`` queries or while migrating data, because after the migrations, :ref:`replace_metadata` will be called.
```
lexi-lambda: If it should always be set to `true` for `READ ONLY` queries, why not default it to `true` when the `read_only: true` option is provided? Also, it isn’t clear to me from reading the documentation how this relates to `check_metadata_consistency`—they seem similar (both involve suppressing metadata checking in some way), but not exactly the same. Is there any situation where I would want to use one but not the other?

More broadly, I think the recommendation to set this to `true` during migrations doesn’t say *quite* enough for an average user to connect the dots about why it makes sense. I’d say something like this, instead:

> When set to `true`, metadata is not reloaded after the SQL query is executed. This can improve performance, but it introduces the possibility for a SQL query that modifies the schema to cause `graphql-engine`’s metadata cache to enter an inconsistent state. For example, if a query adds a column to a tracked table, the new column won’t be immediately reflected in the GraphQL schema.
>
> `skip_cache_reload` is most useful when used as part of a larger `bulk` operation that will subsequently reload or alter the metadata anyway. This is especially common in migrations, which use `run_sql` to modify the database schema followed by `replace_metadata` to update the metadata to match.

CHANGELOG.md
```diff
@@ -72,4 +72,5 @@ Read more about this command in the [docs](https://hasura.io/docs/1.0/graphql/ma
 - cli(migrations-docker): add support for v2 config (close #3969)
 
   A new CLI migrations image is introduced to account for the new CLI workflow. If you're have a project with `version: 2` in `config.yaml`, you should use the new image: `hasura/graphql-engine:v1.2.0-cli-migrations-v2`. Mount the migrations at `/hasura-migrations` and metadata at `/hasura-metadata`.
-  
+
+- server: Skip the schema cache building in `run_sql` request, if `skip_cache_reload` is set to `true`. (close #4200)
```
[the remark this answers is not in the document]
codingkarthik: Oh yes, definitely makes more sense. I'll make the change

docs/graphql/manual/api-reference/schema-metadata-api/run-sql.rst
codingkarthik: I'll set `skip_reload_cache` to `true` when `read_only` is set to true.

@lexi-lambda , I read the docs of the `run_sql` API's args syntax (https://hasura.io/docs/1.0/graphql/manual/api-reference/schema-metadata-api/run-sql.html#args-syntax) and after reading it, I thought that after setting the `check_metadata_consistency` flag, the API checked the consistency without re-building the schema cache. So, after going through its implementation, I realised that both (`skip_cache_reload` and `check_metadata_consistency`) will be re-building the schema cache.

So, I think `skip_cache_reload` is just a more explicit flag that the schema cache is going to be built?

lexi-lambda: @tirumaraiselvan What do you think? Does `check_metadata_consistency` already address #4200? Should these be aliases for the same thing? It seems confusing to have two options with such a similar effect.

tirumaraiselvan: It's definitely a bit confusing. @codingkarthik Can you verify if this is correct understanding:

1. `check_metadata_consistency: true` will fail if there are inconsistencies (unless `cascade` option is also provided which will drop inconsistencies).
2. `check_metadata_consistency: any` will build schema cache (if it didn't fail above).
3. `skip_cache_reload: true` will not fail if there are inconsistencies.
4. `skip_cache_reload: true` will not build schema cache.

So in the migration workflow, I might think to use `check_metadata_consistency: false` to avoid unnecessary computation. But this will still trigger a rebuild of cache and can possibly mark some metadata inconsistent (say, I dropped a table then all its permissions are inconsistent).

This means the only option I have is to `skip_cache_reload: true` . But this is an unsafe API because it can cause inconsistencies which are not even detected. If we can combine this with `check_metadata_consistency` then we get what we want, consistent SQL operations without rebuilding cache. 

Although, the CLI migrations should still only do `skip_cache_reload: true` because the final metadata is in the `replace_metadata` call.

Let me know if this makes sense or if I dug a deeper hole.

codingkarthik: @tirumaraiselvan 
1. It is correct
2. If the `check_metadata_consistency` is omitted, then the schema cache will be built if it's one of `alter|drop|replace|create function|comment on` and on `false` it won't build the schema cache.
3. Correct
4. Correct (but the schema cache may be built, see `#2`).

> So in the migration workflow, I might think to use check_metadata_consistency: `false` to avoid unnecessary computation. But this will still trigger a rebuild of cache and can possibly mark some metadata inconsistent (say, I dropped a table then all its permissions are inconsistent).

I didn't understand how will the cache rebuild be triggered if `check_metadata_consistency` is set to `false`

tirumaraiselvan: Because the only thing that is checking if cache needs to be rebuilt is the `rTxAccessMode` : https://github.com/hasura/graphql-engine/blob/b84db36ebb51acd5b51e1254c103f3097a7c2358/server/src-lib/Hasura/Server/Query.hs#L262

codingkarthik: Ugh, I was only thinking about `check_metadata_consistency`, smh. My bad

tirumaraiselvan: Okay, my bad and apologies. I didn't notice the code had changed a bit since I last looked at it and `queryModifiesSchemaCache` does NOT retrigger a schema cache build but only notifies other instances for a rebuild.

So, I guess both are **equivalent**. And I guess we can close this PR atleast for the purposes of its main intent (migrations v2).

tirumaraiselvan: Although, we may have to add the check for `check_metadata_inconsistency` here as well: https://github.com/hasura/graphql-engine/blob/b84db36ebb51acd5b51e1254c103f3097a7c2358/server/src-lib/Hasura/Server/Query.hs#L262 ?

codingkarthik: https://github.com/hasura/graphql-engine/blob/b84db36ebb51acd5b51e1254c103f3097a7c2358/server/src-lib/Hasura/Server/Query.hs#L262

The problem with the current implementation is that it's not optimal, other hasura instances will be notified whenever `read_only` is set to `false`. 

This can be improved by notifying the other instances only when there's a DDL query request

0x777: @tirumaraiselvan Why can't the new cli migrations workflow use `run_sql` with `check_metadata_consistency` to `false`?

tirumaraiselvan: @codingkarthik I would recommend doing that optimization as a separate issue. 

Let's just fix the `check_metadata_consistency` issue?

PR #4266, opened 2020-04-01
fix unnecessary schema update event in run_sql API

codingkarthik:
- when consistent_metadata_check is set to false, then a new schema update
  event should not be created

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

The `run_sql` API creates a new schema update event(`occurred_at` col of the `hdb_catalog.hdb_schema_update_event` is updated with `now()`) when the `read_only` flag is set to `true`. 
This is a bug because when `check_metadata_consistency` is set to `false`, the schema cache is not updated in the current hasura instance but there's a new update event created which triggers a schema cache build in the other hasura instances.

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/Server/Query.hs
+++ b/server/src-lib/Hasura/Server/Query.hs
@@ -259,10 +259,10 @@ queryModifiesSchemaCache (RQV1 qi) = case qi of
   RQAddCollectionToAllowlist _    -> True
   RQDropCollectionFromAllowlist _ -> True
 
-  RQRunSql RunSQL{rTxAccessMode}  ->
+  RQRunSql RunSQL{rTxAccessMode, rCheckMetadataConsistency}  ->
     case rTxAccessMode of
       Q.ReadOnly  -> False
-      Q.ReadWrite -> True
+      Q.ReadWrite -> maybe True id rCheckMetadataConsistency
 
   RQReplaceMetadata _             -> True
   RQExportMetadata _              -> False
--- a/server/src-lib/Hasura/Server/Query.hs
+++ b/server/src-lib/Hasura/Server/Query.hs
@@ -262,7 +262,7 @@ queryModifiesSchemaCache (RQV1 qi) = case qi of
   RQRunSql RunSQL{rTxAccessMode, rCheckMetadataConsistency}  ->
     case rTxAccessMode of
       Q.ReadOnly  -> False
-      Q.ReadWrite -> maybe True id rCheckMetadataConsistency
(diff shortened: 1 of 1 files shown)


server/src-lib/Hasura/Server/Query.hs
```diff
@@ -259,10 +259,10 @@ queryModifiesSchemaCache (RQV1 qi) = case qi of
   RQAddCollectionToAllowlist _    -> True
   RQDropCollectionFromAllowlist _ -> True
 
-  RQRunSql RunSQL{rTxAccessMode}  ->
+  RQRunSql RunSQL{rTxAccessMode, rCheckMetadataConsistency}  ->
     case rTxAccessMode of
       Q.ReadOnly  -> False
-      Q.ReadWrite -> True
+      Q.ReadWrite -> maybe True id rCheckMetadataConsistency
```
lexi-lambda: `maybe x id` can (and should!) be abbreviated to `fromMaybe x`. I’ll make the change here myself, since it’s such a tiny thing, but just so you know!

codingkarthik: Oh, I didn't know about `fromMaybe`. Thanks for the fix and the suggestion 🙂

codingkarthik:
When `rCheckMetadataConsistency` is `True`, the schema cache is just built once in the current hasura instance.

If you're referring to https://github.com/hasura/graphql-engine/blob/85b7c552afb9b3737614a376dd82873f2d253788/server/src-lib/Hasura/Server/Query.hs#L208 

for the 2nd schema cache reload, it doesn't reload the schema cache in the current instance but creates a new event in `hdb_catalog.hdb_schema_update_event` which triggers a schema cache build in other hasura instances, which I think is the correct behaviour.

lexi-lambda:
Ah, you are right! I forgot that `queryModifiesSchemaCache` no longer triggers a cache reload in the current process. (But I shouldn’t have forgotten it, because I’m the one who changed it! 😄)

Issue #3793, opened 2020-01-29
optimise count query in browse rows page to account for really large tables


rikinsk:
The row count query currently being made on browse rows page can be quite slow to run in case a table has miliions of rows. A more optimised query needs to be used.

rikinsk:
As a step 1. The count query and the get page results query can be decoupled. ie. display the page results even though the count query takes a long time.

wawhal:
Let's implement what @rikinsk suggests in the above comment. Should be good enough for now.
The counter itself can be a separate component that maintains its state and can show a spinner while the query is in progress.

beerose:
What about using [count estimate](https://wiki.postgresql.org/wiki/Count_estimate)?
Table plus, for example, uses this instead of the exact count. 

Here go some comparison:

Number of rows | count(*) | SELECT reltuples | Notes
-- | -- | -- | --
1 000 000 | 47ms | 1ms |  
2 000 000 | 66ms | 1ms |  
5 000 000 | 136ms | 1ms |  
10 000 000 | 316ms | 1ms | Estimated ~7 000 000, still fast in console
20 000 000 | 461ms | 1ms | Estimated ~18 000 000, still pretty fast in console
30 000 000 | 850ms | 1ms | Estimated ~25 000 000
35 000 000 | 12005ms | 3ms | Estimated ~32 000 000
50 000 000 | 2379ms | 7ms | Estimated ~48 000 000

rikinsk:
The ideal behaviour I would like is that for small tables the count is accurate but for larger tables we display an estimate. 

Can we confirm how accurate the count estimate is for small tables (say less than 1,000,000 rows) and if there is a smart way to switch between the two queries. e.g. check if count is > a million and if so use the estimate query instead of exact count.

Also we will have to think what to do about pagination when we use the estimate count.

tirumaraiselvan:
Something that I felt was needed in our SQL APIs (and graphql APIs) was using `statement_timeout`: https://www.postgresql.org/docs/9.4/runtime-config-client.html

This way if an execution fail with timeout then you can retry it with cheaper query.

PR #4269, opened 2020-04-01
console: decouple count query from fetch data query

beerose:
### Description

This PR decouples count query from fetch data query. 

### Changelog

### Affected components

- [x] Console

### Related Issues
https://github.com/hasura/graphql-engine/issues/3793


beerose:
@rikinsk It should be fixed now.

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -113,6 +101,50 @@ const vMakeRequest = () => {
...
 };
 
+const vMakeCountRequest = () => {
+  return (dispatch, getState) => {
+    const state = getState();
+    const url = Endpoints.query;
+    const originalTable = getState().tables.currentTable;
+    dispatch({ type: V_COUNT_REQUEST_START });
+
+    const requestBody = {
```
wawhal, line 111: Move to v1QueryUtils.

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -57,29 +61,14 @@ const vMakeRequest = () => {
     dispatch({ type: V_REQUEST_PROGRESS, data: true });
 
     const requestBody = {
```
wawhal, line 63: Move to v1QueryUtils as `generateSelectQuery = (tableDef, where, limit, offset, orderBy) => { ...`

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -113,6 +101,50 @@ const vMakeRequest = () => {
...
+    };
+    const options = {
+      method: 'POST',
+      body: JSON.stringify(requestBody),
+      headers: dataHeaders(getState),
+      credentials: globalCookiePolicy,
+    };
+    return dispatch(requestAction(url, options)).then(
+      data => {
+        if (originalTable === getState().tables.currentTable) {
```
wawhal, line 129: What is the reason behind this check?
Isn't `originalTable` defined as `getState().tables.currentTable` in the same scope?

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -113,6 +101,50 @@ const vMakeRequest = () => {
...
+            showErrorNotification('Count query failed!', error.error, error)
+          ),
+          dispatch({ type: V_COUNT_REQUEST_ERROR }),
+        ]);
+      }
+    );
+  };
+};
+
 const fetchManualTriggers = tableName => {
```
wawhal, line 148: Move the query to v1QueryUtils.

console/src/components/Services/Data/TableBrowseRows/ViewRows.js
```diff
@@ -763,9 +764,7 @@ const ViewRows = ({
           <b className={styles.padd_small_right}>Selected:</b>
           {selectedRows.length}
           <button
```
wawhal, line 766: Use the common Button component.

console/src/components/Common/TableCommon/DragFoldTable.js
```diff
@@ -3,7 +3,10 @@
 import React, { Component } from 'react';
 import ReactTable, { ReactTableDefaults } from 'react-table';
 import 'react-table/react-table.css';
+
 import FoldableHoc from './foldableTable';
+import Spinner from '../Spinner/Spinner';
+import styles from '../Common.scss';
 
 Object.assign(ReactTableDefaults, {
```
wawhal, line 11: Sorry I missed this in the `browse rows improvements` PR, but can you remove these defaults and pass them to the ReactTable component directly where it's used?

Just to make the code declarative. There's no way anybody will read these defaults if they are reading the component code. Do you agree?

console/src/components/Common/TableCommon/DragFoldTable.js
```diff
@@ -87,7 +90,11 @@ class DragFoldTable extends Component {
             })
           }
           folded={this.state.folded}
+          showPagination={!this.props.countLoading}
```
wawhal, line 93: Is this needed?

beerose: When the count is still loading there's no pagination displayed. Instead, there's a spinner.

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
beerose: It's prev logic.

console/src/components/Services/Data/TableBrowseRows/ViewRows.js
beerose: Isn't it duplicated work, since we'll be soon integrating buttons from the design system?

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
beerose: The idea is that user may change table while request is pending. And there might be race condition with another fetch data query.

console/src/components/Common/TableCommon/DragFoldTable.js
beerose: I removed it because:
1) defaultPageSize is already 10,
2) minRows wasn't enforced in react-table anyway.

rikinsk:
Also noticed one other thing. The bulk selected rowws should be reset when we change pages.

beerose:
1. I implemented the solution with an estimated count (show estimate and then change to the exact count), but it feels kinda odd when the number of pages changes. 

2. I added resetting bulk select on page change.

rikinsk:
The count changing does feel a little weird. I think I have a solution though (this should be the final change for this PR hopefully). 
Instead of displaying the estimated count and then replacing it with the actual count, lets display a spinner in place of the count till the actual count loads. Lets use the estimated count to only set the estimated number of pages. This way when the count finally loads, the spinner will get replaced by the count (giving an appearance that some new information was loaded) and the number of pages changing will not feel weird.

beerose:
Just to confirm, by not displaying the count you mean removing the whole `of X` part like below?

Screenshot 2020-04-04 at 17 13 56

Btw, I read the comment before the edit, so I'll leave here how spinner would look like just for a reference.

rikinsk:
Not exactly. What I meant was this. 

- When count is not loaded (but we have estimated count)

  Count display:
  Screenshot from 2020-04-04 20-59-14

  Pagination: 
  Screenshot from 2020-04-04 21-00-05

- When count loads

  Count display:
  
  Screenshot from 2020-04-04 20-58-51

  Pagination:
  Screenshot from 2020-04-04 20-59-25

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
rikinsk: This was buggy indeed. Fixed it now

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -89,13 +86,12 @@ const vMakeRequest = () => {
     };
     return dispatch(requestAction(url, options)).then(
       data => {
-        const currentTable = getState().tables.currentTable;
-        if (originalTable === currentTable) {
+        if (originalTable === tables.currentTable) {
```
wawhal, line 89: This will always be true?

console/src/components/Services/Data/TableBrowseRows/FilterActions.js
```diff
@@ -108,7 +108,7 @@ const runQuery = tableSchema => {
       delete newQuery.order_by;
     }
     dispatch({ type: 'ViewTable/V_SET_QUERY_OPTS', queryStuff: newQuery });
-    dispatch(vMakeRequest());
+    dispatch(vMakeRequest()).then(() => dispatch(vMakeCountRequest()));
```
wawhal, line 111: This is done twice, here and once in `ViewTable` as well.
Do you think there should be a function called `vMakeTableRequests` that abstracts out this part? Something like:

```js
const vMakeTableRequests = () => (dispatch, getState) => {
  dispatch(vMakeRequest()).then(() => dispatch(vMakeCountRequest()));
};
```
You can then call this one function everywhere. Later, if somebody is iterating some part of this code, they might not know to make the change in both places.

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -51,34 +60,22 @@ const vSetDefaults = () => ({ type: V_SET_DEFAULTS });
 
 const vMakeRequest = () => {
   return (dispatch, getState) => {
-    const state = getState();
+    const { tables } = getState();
```
wawhal, line 63: ### Suggestion

I think it would make sense to extract the relevant parts here itself rather than rocking the `tables` object everywhere. This is because, in large schemas, the `tables` object could be **huge**.

You can just:

```js
const {
  currentTable: originalTable,
  currentSchema
} = getState().tables;
```

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -113,27 +109,50 @@ const vMakeRequest = () => {
   };
 };
 
+const vMakeCountRequest = () => {
```
wawhal, line 112: Same review as the `vMakeRequest` function.

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
rikinsk: Nope. currentTable is derived from state again in the fn and can be different as it async because it is in `then()` of requestAction.

rikinsk: Sorry. this seems like another instance of the issue I fixed

console/src/components/Common/utils/v1QueryUtils.js
```diff
@@ -279,3 +279,38 @@ export const resetMetadataQuery = {
...
+    columns,
+    where,
+    limit,
+    offset,
+    order_by,
+    table: tableDef,
+  },
+});
+
+export const getFetchManualTriggersQuery = tableName => ({
```
wawhal, line 299: @beerose @rikinsk This should ideally accept a table definition rather than just a table name. Can we assign this to somebody ASAP? 5 min fix.

PR #4270, opened 2020-04-01
console: support materialized views (fix #91)

### LINKED ISSUE #91: materialised views are not listed in console

nicuveo:
### Description
This PR add support for materialized views, fixing #91. It touches a lot of things, so let's break it down.

The first big change is in [Data/utils.js](https://github.com/nicuveo/graphql-engine/blob/b0b02bc7f2554699912cc17d1d00a3e1e7d5fecd/console/src/components/Services/Data/utils.js); we have to change the way we load the schema from the backend, as the `information_schema` views do NOT contain information about materialized views. To fix this, we replace both `information_schema.views` and `information_schema.columns` by a simplified version of how they were implemented in [postgres 9,5](https://github.com/postgres/postgres/blob/REL9_5_STABLE/src/backend/catalog/information_schema.sql#L635), while extending them to also include materialized views.

Obviously, this has a major drawback: we areadding postgres implementation details to the frontend. This was however deemed okay, for now, as the query already used some of those implementation details such as `pg_class`, and this allows us to have materialized views. A ticket has been filed for the work that needs to be done on the server side.

(This PR also temporarily includes a subset of #4250 (until it is merged), as expanding the sql request uncovered a bug in the metadata checker.)

Then, this PR add support for a new "table type" in the frontend: MATERIALIZED VIEW. For the sake of simplicity, most of the code treats is as a view; and since the code made the difference between tables and views by checking [whether the object was or wasn't a table](https://github.com/nicuveo/graphql-engine/blob/b0b02bc7f2554699912cc17d1d00a3e1e7d5fecd/console/src/components/Common/utils/pgUtils.js#L57), most of the structure remains the same.

We do however need to distinguish between views and materialized views when we want to modify the object; for instance, materialized views do not support `CREATE OR REPLACE`, which is why we have to inject a `DROP` first when [re-creating the request](https://github.com/nicuveo/graphql-engine/blob/b0b02bc7f2554699912cc17d1d00a3e1e7d5fecd/console/src/components/Services/Data/TableModify/ModifyActions.js#L988). Same thing for renaming, commenting, or deleting a materialized view.

The main change, beneath all this, is that we now have the `table_type` as a property in the `ModifyView` [component](https://github.com/nicuveo/graphql-engine/blob/b0b02bc7f2554699912cc17d1d00a3e1e7d5fecd/console/src/components/Services/Data/TableModify/ModifyView.js#L275).

This adds [cypress tests](https://github.com/nicuveo/graphql-engine/blob/b0b02bc7f2554699912cc17d1d00a3e1e7d5fecd/console/cypress/integration/data/materialized-views/test.js) for all common operations on materialized views. They are a straight copy of the tests on views.

### Changelog

### Affected components

- [ ] Console
- [ ] Docs
- [ ] Tests

### Screenshots

Materialized views are displayed almost exactly like views. They show up in the list of untracked objects:

matview_1

and when tracked, they display the same tabs and same buttons as a view. The query is updated, however to reflect the fact that it is indeed a materialized view.

matview_2


nicuveo:
> @nicuveo could you please add to the `CHANGELOG.md` and explain the feature from a user perspective?

Done! Sorry for pushing the PR out without it: but I suspect there is going to be quite a lot that needs to be changed, and I wanted it out as quickly as possible.

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -715,12 +715,11 @@ const setUniqueKeys = keys => ({
   keys,
 });
 
-const changeTableName = (oldName, newName, isTable) => {
+const changeTableName = (oldName, newName, isTable, objectType) => {
```
beerose, line 718: Name `objectType` might be a bit confusing. What about `tableType`?

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -959,11 +978,21 @@ const fetchViewDefinition = (viewName, isRedirect) => {
...
-      'DROP VIEW ' + '"' + currentSchema + '"' + '.' + '"' + viewName + '"';
+      'DROP ' +
+      viewType +
+      ' "' +
+      currentSchema +
+      '"' +
+      '.' +
+      '"' +
+      viewName +
+      '"';
```
beerose, line 995: Since this is pretty big, string literal might be a better fit:
```js
const sqlDropView = `DROP ${viewType} "${currentSchema}"."${viewName}".
```

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -959,11 +978,21 @@ const fetchViewDefinition = (viewName, isRedirect) => {
   };
 };
 
-const deleteViewSql = viewName => {
+const deleteViewSql = (viewName, viewType) => {
   return (dispatch, getState) => {
     const currentSchema = getState().tables.currentSchema;
+    const property = viewType.toLowerCase();
+    const capitalizedProperty = property[0].toUpperCase() + property.slice(1);
```
beerose, line 985: Could you add it to `jsUtils` as `capitalize` function?

console/src/components/Services/Data/TableModify/ModifyTable.js
```diff
@@ -228,7 +228,7 @@ class ModifyTable extends React.Component {
             <TableCommentEditor
               tableComment={tableComment}
               tableCommentEdit={tableCommentEdit}
-              isTable
+              tableType={'TABLE'}
```
beerose, line 231:
```suggestion
              tableType="TABLE"
```

console/src/components/Services/Data/TableModify/ModifyView.js
```diff
@@ -319,12 +320,30 @@ ModifyView.propTypes = {
...
 
+const findViewType = (currentSchema, viewName, allItems) => {
+  for (let index = 0; index < allItems.length; index++) {
+    const item = allItems[index];
+    if (item.table_schema === currentSchema && item.table_name === viewName) {
+      return item.table_type;
+    }
+  }
+  return 'VIEW';
+};
```
beerose, line 331:
```suggestion
const findViewType = (currentSchema, viewName, allItems) => {
 for (let item of allItems) 
    if (item.table_schema === currentSchema && item.table_name === viewName) {
      return item.table_type;
    }
  }
  return 'VIEW';
};
```

console/src/components/Services/Data/TableModify/ModifyView.js
```diff
@@ -319,12 +320,30 @@ ModifyView.propTypes = {
...
+  const tableName = ownProps.params.table;
+  const schemaName = state.tables.currentSchema;
+  const tableType = findViewType(
+    schemaName,
+    tableName,
+    state.tables.allSchemas
+  );
   return {
-    tableName: ownProps.params.table,
+    tableName: tableName,
```
beerose, line 342:
```suggestion
    tableName,
```

console/src/components/Services/Data/TableModify/ModifyView.js
```diff
@@ -319,12 +320,30 @@ ModifyView.propTypes = {
...
+    tableType: tableType,
```
beerose, line 343:
```suggestion
    tableType,
```

console/src/components/Services/Data/TableModify/ModifyActions.js
nicuveo: Done!

console/src/components/Services/Data/TableModify/ModifyActions.js
nicuveo: Done!

console/src/components/Services/Data/TableModify/ModifyActions.js
nicuveo: Done! Fixed the few other call sites I could find.

console/src/components/Services/Data/TableModify/ModifyTable.js
nicuveo: Done.

console/src/components/Services/Data/TableModify/ModifyView.js
nicuveo: Done.

console/src/components/Services/Data/TableModify/ModifyView.js
nicuveo: Done!

console/src/components/Services/Data/TableModify/ModifyView.js
nicuveo: Done!

console/cypress/integration/data/materialized-views/spec.js
```diff
@@ -0,0 +1,457 @@
...
+  validateCT,
+  createView,
+  validateColumn,
+  validateView,
+} from '../../validators/validators';
+import { setPromptValue } from '../../../helpers/common';
+
+const userId = 5555;
+
+export const Createtable = (name, dict) => {
```
beerose, line 18: I'd stick to the camel case for functions and use `createTable` instead.

console/cypress/integration/data/materialized-views/spec.js
```diff
@@ -0,0 +1,457 @@
...
+  cy.get(getElementFromAlias('delete-view')).click();
+  cy.window()
+    .its('prompt')
+    .should('be.called');
+  cy.wait(7000);
+  // cy.get('.notification-error');
+  validateView('author_average_rating_vt', 'failure');
+};
+
+export const Deletetable = name => {
```
beerose, line 434: Could you change to `deleteTable`?

console/cypress/integration/data/materialized-views/spec.js
nicuveo: Done!

console/cypress/integration/data/materialized-views/spec.js
nicuveo: Done!

nicuveo:
@marionschleifer Done! It's mostly the same as it was...
@rikinsk Quick question about the UI: I have kept it almost exactly the same, which means all the buttons say "view" for materialized views too. Only the popup messages have been changed. Does that sound ok to you?

CHANGELOG.md
```diff
@@ -35,6 +35,10 @@ The order, collapsed state of columns and page size is now persisted across page
 
 ## `v1.2.0-beta.3`
 
+### console: support for Postgres [materialized views](https://www.postgresql.org/docs/current/rules-materializedviews.html)
+
+Postgres materialized views are views that are persisted in a table-like form. They are now supported in the Hasura Console, in the same way as views. They will appear on the 'Schema' page, under the 'Data' tab, in the 'Untracked tables or views' section.
```
tirumaraiselvan, line 40: this needs to be moved to `Next release` section

PR #4280, opened 2020-04-02
cli: link cli-ext plugin with cli version (close #4105)

### LINKED ISSUE #4105: link cli-ext with hasura cli
Currently, `cli-ext` plugin manifest is placed under the plugins index repo which points to the latest version.

We need to link the `cli-ext` with the hasura cli version.

arvi3411301:
### Description
This PR adds the support to force install `cli-ext` plugin with the same version of cli.

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [x] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues
#4105 

### Solution and Design
If `CLISemVer` is set to nil and the plugin is not installed, CLI will try to install latest version from the plugin index.
If `CLISemVer` is set to nil and the plugin is already installed, CLI will return `ErrIsAlreadyInstalled`
If `CLISemVer` is a valid semver and the plugin is already installed with a different version, CLI will try to install plugin with `CLISemVer`


shahidhk:
@arvi3411301 If the plugin is being installed in background, will a corresponding log line be shown?

PR #4282, opened 2020-04-02
Produce coverage report for 'dev.sh graphql-engine' too

jberryman:
I find this useful during development.


jberryman:
https://github.com/hasura/graphql-engine/pull/4360 instead I guess

PR #4283, opened 2020-04-03
run_sql modifies the schema cache if the query is an DDL one

codingkarthik:
* If check_metadata_consistency is set, it will override it

### Description

Previously, when a `run_sql` is set without `check_metadata_consistency`,  `queryModifiesSchemaCache` was set to `true`, this PR check if whether the SQL query is an DDL one and if it is then only set `queryModifiesSchemaCache` to true
### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Server/Query.hs
```diff
@@ -259,10 +259,10 @@ queryModifiesSchemaCache (RQV1 qi) = case qi of
   RQAddCollectionToAllowlist _    -> True
   RQDropCollectionFromAllowlist _ -> True
 
-  RQRunSql RunSQL{rTxAccessMode, rCheckMetadataConsistency}  ->
+  RQRunSql RunSQL{ .. }  ->
     case rTxAccessMode of
       Q.ReadOnly  -> False
-      Q.ReadWrite -> fromMaybe True rCheckMetadataConsistency
+      Q.ReadWrite -> fromMaybe (containsDDLKeyword rSql) rCheckMetadataConsistency
```
lexi-lambda, line 265: It seems like a lot of logic for this is now duplicated between `runRunSQL` and `queryModifiesSchemaCache`, which is mildly unfortunate. It’s also a little sad because it means that `Hasura.Server.Query` has to know about `containsDDLKeyword`, which is really an implementation detail of `Hasura.RQL.DDL.Schema`.

Could we factor out this logic into a common function that encapsulates all the rules about whether or not `run_sql` should rebuild the schema cache, then export that function instead of `containsDDLKeyword`? Then we’d have a single source of truth for the rules, since both `runRunSQL` and `queryModifiesSchemaCache` could use that function to make the choice.

codingkarthik: Yes, it will make more sense to write a common function which decides whether a schema cache rebuild is required.

 I have made a common function `isSchemaCacheBuildRequiredRunSQL` and I'm referencing it from the appropriate places.

PR #4286, opened 2020-04-03
console: keep entire graphiql in the viewport

rikinsk:
### Description

Ensure that entire graphiql is within the viewport even with 2 headers (as we would typically always have content-type and admin-secret)

### Changelog

### Affected components

- [x] Console

### DIFF
--- a/console/src/components/Common/Common.scss
+++ b/console/src/components/Common/Common.scss
@@ -1328,7 +1328,7 @@ code {
 
 // keep graphiql in viewport
 .graphQLHeight {
-  height: calc(100vh - 360px);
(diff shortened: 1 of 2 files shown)


beerose:
It feels weird, how it ends, doesn't it?
Screenshot 2020-04-03 at 14 03 31

beerose:
This is just a suggestion -- but what about adding box-shadow? Below is an example, it can look different.
Screenshot 2020-04-03 at 14 16 37

rikinsk:
@beerose added box shadow and reduced height by 10px more to better fit

PR #4290, opened 2020-04-03
console: set allowJs compiler flag for type checker

beerose:
### Description

This PR is setting `allowJs` compiler option to `true` for `ForkTsCheckerWebpackPlugin`. It allows the type checker to transpile JS files. In other words, if we want to import JS module to TS file, we don't have to provide typings. The type checker will get them.
Especially useful when migrating from JS to TS.

This change only affects the TypeScript experience. 

Also, in order to enable this flag, I needed to install typings for `lodash` and `ws` since they are not in peer dependencies of the their parent libs. I also provided declaration for `SvgPanZoom` because of the lack of proper typings  in `graphql-voyager` ([issue](https://github.com/APIs-guru/graphql-voyager/issues/91)).

### Affected components

- [x] Console


rikinsk:
No UX changes. @wawhal Will merge post your approval

Issue #4142, opened 2020-03-18
console: incorporate filter and check for update permissions


tirumaraiselvan:
Filter and check for update permissions were introduced in #3804

wawhal:
- We'll have a tabbed layout at the update permissions
- The conditions will  be called `pre-update` and `post-update
- While referring to these permissions while setting other permissions, it'll display (`same as pre-update` and `same as post-update`)

PR #4313, opened 2020-04-06
console: allow setting filter and check for update permissions

beerose:
Solves https://github.com/hasura/graphql-engine/issues/4142

### Description

This PR incorporates server changes introduced in https://github.com/hasura/graphql-engine/pull/3804.
It adds tabs `Post update`(filter) and `Pre update`(check) for update permissions. 

Screenshot 2020-04-21 at 13 01 20

Note: I tried to keep this PR as small as possible.

### Changelog

### Affected components

- [x] Console


beerose:
> we can put the help texts only rows satisfying the check will be updatable and rows must satisfy the check after update as tooltips instead of just text. (This does make the UI look better but I am not sure of it is the best UX. What are your thoughts? You can chose to do this or ignore this based on your feel. I will be asking others for feedback about this later anyway.)

@rikinsk I think it makes sense since `pre/post-update check` labels are already quite self-explanatory. I will add tooltips.

console/src/declaration.d.ts
```diff
@@ -6,3 +6,5 @@ declare module '*.scss' {
 declare namespace SvgPanZoom {
   export interface Instance {}
 }
+
+declare module 'react-bootstrap/lib/Tooltip';
```
wawhal, line 10: Can you just add the react-bootstrap from definitelyTyped?

beerose: Yeah, I tried. But typings are not exported properly. Will investigate later.

PR #4316, opened 2020-04-07
cli: clarify flags on migrate create

scriptonist:
### Description
Some changes to the description of flags in hasura migrate create.

### Affected components

- [ ] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


scriptonist:
Oops! thanks for pointing that out @shahidhk, sending a fix!

shahidhk:
@scriptonist Instead of adding `--schema` to every other flag, can we add the usage to examples?

scriptonist:
Makes sense to me @shahidhk. This was originally suggested by @tirumaraiselvan , Let's see what he thinks too.

So @tirumaraiselvan what @shahidhk is suggesting is we include the usage in examples, instead of doing the same in flag description. Which will look something like this, and will show up in docs [here](https://hasura.io/docs/1.0/graphql/manual/hasura-cli/hasura_migrate_create.html#examples)

```
$ hasura migrate create --help                                                *(docker-desktop/default)
Create sql and yaml files required for a migration

Usage:
  hasura migrate create [migration-name] [flags]

Examples:
  # Setup migration files for the first time by introspecting a server:
  hasura migrate create "init" --from-server

  # Use with admin secret:
  hasura migrate create --admin-secret "<admin-secret>"

  # Setup migration files from an instance mentioned by the flag:
  hasura migrate create init --from-server --endpoint "<endpoint>"

  # Take pg_dump of schema and hasura metadata from server while specifying schemas to include
  hasura migrate create init --from-server --schema myschema1,myschema2

  # Take pg_dump from server and save it as a migration and specify which schemas to include
  hasura migrate create init --sql-from-server --schema myschema1,myschema2

Flags:
      --from-server                 take pg_dump of schema and Hasura metadata from the server
  -h, --help                        help for create
      --metadata-from-file string   path to a hasura metadata file to be used for up actions
      --metadata-from-server        take metadata from the server and write it as an up migration file
      --schema strings              name of Postgres schema to export as a migration. provide multiple schemas with a comma separated list e.g. --schema public,user (default [public])
      --sql-from-file string        path to an sql file which contains the SQL statements
      --sql-from-server             take pg_dump from server and save it as a migration
```

tirumaraiselvan:
LGTM. Can we say `take pg_dump of schema (default: public) ...` though?

Issue #4314, opened 2020-04-06
console: migrate to TypeScript


beerose:
As we merged https://github.com/hasura/graphql-engine/pull/3902 recently we plan to gradually migrate frontend code to TypeScript.

Console codebase is pretty large, so we would love your help! 💪

### How to contribute?

1. Convert file from `.js` to `.ts` or `.tsx`. 
2. Add types, fix all type errors. 
3. Submit a PR! 🎉

[Here you can find a guide on how to contribute to Hasura console](https://github.com/hasura/graphql-engine/blob/master/console/README.md#contributing-to-hasura-console).

We recommend you to take a look at the codebase and find something that is still in JS and seems interesting for you to migrate. If you find something, just let us know in the comment and go ahead! 😎 

<details>
  <summary>Click to see the list of easy and small modules that are a good fit for the start</summary>

- [ ] [AlertBox component](https://github.com/hasura/graphql-engine/tree/master/console/src/components/UIKit/atoms/AlertBox) @harishm72 https://github.com/hasura/graphql-engine/pull/4721

- [ ] [Tabs component](https://github.com/hasura/graphql-engine/tree/master/console/src/components/UIKit/atoms/Tabs) @sensiblefolk https://github.com/hasura/graphql-engine/pull/4536
- [ ] [Tooltip component](https://github.com/hasura/graphql-engine/tree/master/console/src/components/UIKit/atoms/Tooltip) @ngprnk https://github.com/hasura/graphql-engine/pull/4524

- [ ] [AceEditor](https://github.com/hasura/graphql-engine/tree/master/console/src/components/Common/AceEditor) @sensiblefolk https://github.com/hasura/graphql-engine/pull/4536

- [ ] [Headers component and utils](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Common/Headers) @harishm72 
- [ ] [Routes utils](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Common/utils/routesUtils.j) @ngprnk

- [ ] [Error boundary component](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Error/ErrorBoundary.js) @marekdano 
https://github.com/hasura/graphql-engine/pull/4542
- [ ] ~[SDL utils](https://github.com/hasura/graphql-engine/blob/master/console/src/shared/utils/sdlUtils.js)~ @mubaidr (can't be merged until cli-ext supports TS)
[...]

PR #4325, opened 2020-04-07
chore: convert theme files to typescript

alioguzhan:
### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

Implements`Theme` item at #4314


beerose:
Hi! Thank for your PR. Would you like to also convert this file: https://github.com/hasura/graphql-engine/blob/master/console/src/components/UIKit/theme/index.js as a part of your PR? In the migrate issue link pointed to the parent directory, but the theme object is in `/theme` folder.

alioguzhan:
> Hi! Thank for your PR. Would you like to also convert this file: https://github.com/hasura/graphql-engine/blob/master/console/src/components/UIKit/theme/index.js as a part of your PR? In the migrate issue link pointed to the parent directory, but the theme object is in `/theme` folder.

Ah, I just noticed that I need to convert also that file while converting `AlertBox`. Sending it now!

alioguzhan:
Hey @beerose ,

This file is a kind of base file for all other UIKit components. Most of them import something from this file. And I did not convert some parts of this file because they are related to some other components (`AlertBox` and others). I will complete those parts when I start to convert other ones.

alioguzhan:
But if you want this file completed, I can send those missing parts now. They are ready.

alioguzhan:
@beerose , Thanks for the great feedback.

> Another thing is that we may want to use those types (like radii type, fontSizes type, and whole theme type) in other places, so it would make sense to export interface Theme from this file. Could you add it?

By saying that, you mean a one, single, giant `interface Theme {}`  that defines all the types we used in this file?

beerose:
Yes, I meant one interface.  It can be build using `typeof theme`. Unfortunately doing
```ts
interface Theme extends typeof theme {}
``` 
is not a valid TypeScript, so we can do
```ts
type _Theme = typeof theme;
export interface Theme extends _Theme {}
```
instead.

console/src/components/UIKit/theme/index.ts
```diff
@@ -1,6 +1,29 @@
...
+  | 'original'
+  | 'primary'
+  | 'secondary'
+  | 'hover'
+  | 'light'
+  | 'border'
+  | 'text'
+  | 'tab';
+
+type Color = Partial<Record<ColorType, string>> | any;
```
alioguzhan, line 24: I couldn't find a proper way to handle this with `Union Types` or `Partial`.  Because there are color objects as both `record` and `string`.  What do you think ?

console/src/components/UIKit/theme/index.ts
```diff
@@ -287,3 +377,6 @@ export const theme = {
   alertBox,
   icon,
 };
+
+type _Theme = typeof theme;
+export interface Theme extends _Theme {}
```
alioguzhan, line 382: I didn't quite understand the use case of this.  We could have export all those types above individually as inline.

Or, I misunderstood you. Can you elaborate on this a bit? Thanks! @beerose

console/src/components/UIKit/theme/index.ts
beerose: Generally, I'd skip proving typings manually for `colors`, `icon`, `alertbox`, `button` -- TypeScript can infer those types on its own, and by typing it manually, we may accidentally break something. 

However, for parts of the theme, that have aliases (eg `radii`) things are more complicated, and we do need typings. 

So for those first four (`colors`, `icon`, `alertbox`, `button`) I'd remove typings; the rest looks okay. 

One suggestion: 
Could you change `FontWeight` -> `FontWeightKey`,  `Element` -> `FontSizeKey`, `Space` -> `SpaceKey`, etc. ?
That way, they won't be confused with the types of actual objects.

console/src/components/UIKit/theme/index.ts
beerose: Yes. So I think it's good to have type of the whole `theme` exported. As we may want to do for example:
```ts
const newTheme: Theme = {...}
```

console/src/components/UIKit/theme/index.ts
alioguzhan: I removed types from the first four objects. But we may need some of them as typed once we start to convert other components. I will let you know if something like that occurs. 

Apart from that, I think this is ready.

console/src/components/UIKit/theme/index.ts
```diff
@@ -224,42 +251,60 @@ const space = [0, 4, 6, 8, 10, 12, 14, 16, 18, 20, 32, 64];
...
 
-const lineHeights = [1.33, 1.5];
-
 /* line-height aliases
  * body: 1.5
  * explain: 1.3 ~ Explainer Text
  */
 
-lineHeights.body = lineHeights[1];
+type lineHeightKey = 'body' | 'explain';
```
beerose, line 284:
```suggestion
type LineHeightKey = 'body' | 'explain';
```

console/src/components/UIKit/theme/index.ts
```diff
@@ -150,38 +148,53 @@ const radii = [0, 2, 4, 8, 12, 16];
  * circle: 1000px
  */
 
-radii.xs = radii[1];
+type BorderRadiusSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'circle';
```
beerose, line 151:
```suggestion
type BorderRadiusKey = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | 'circle';
```

console/src/components/UIKit/theme/index.ts
beerose: If we need them, for example, the type of `colors` object, we could do this:

Screenshot 2020-04-09 at 13 54 47

alioguzhan:
Done @beerose

alioguzhan:
Ah, it is failed. I think it is not related to this branch, right?  @beerose

PR #4329, opened 2020-04-08
use only required session variables in multiplexed queries for subscriptions

0x777:
### TODO
1. Tests
2. Changelog

This will reduce the load on Postgres when the result of a subscription is not dependent on the session variables of the request. For example, if a permisson for 'user' role  on 'article' table is `{"is_public": true}`, the result of a subscription on 'article' table for 'user' role is not dependent on any session variables, but a 'user' role is most likely used with a session variable such as `x-hasura-user-id`. In such cases, the multiplexed query will now return a single row which is sent to all the subscribers (as opposed to one row for every subscription with a unique `x-hasura-user-id`)

### Changelog

- [ ] TODO `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [x] Server
- [x] Tests

### Solution and Design
A multiplexed query returns a row for each 'cohort', which is identified by `(session variables, query variables, prepared arguments)`. Previously, `session variables` would include all the session variables from the request, this is now changed to only use the `session variables` that are needed for the multiplexed query, this will result in lesser number of cohorts for a subscription and hence lesser load on Postgres. 

### Steps to test and verify
TODO: The `explain` output for a subscription should tell us the session variables that are used. In the above described permission rule, we shouldn't see any session variables.


server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -281,24 +348,29 @@ reuseLiveQueryPlan
...
+         mkCohortVariables _rlqpRequiredSessionVariables
+         sessionVars validatedVars _rlqpSyntheticVariableValues
 
 data LiveQueryPlanExplanation
   = LiveQueryPlanExplanation
-  { _lqpeSql  :: !Text
-  , _lqpePlan :: ![Text]
+  { _lqpeSql       :: !Text
+  , _lqpePlan      :: ![Text]
+  , _lqpeVariables :: !CohortVariables
```
0x777, line 362: I'm wondering if we should also include the parametrized sql query here.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -1,7 +1,9 @@
 {-# LANGUAGE UndecidableInstances #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE DerivingVia #-}
```
lexi-lambda, line 3: Let’s just turn this on everywhere.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -117,11 +95,20 @@ newCohortId = CohortId <$> liftIO UUID.nextRandom
...
   { _cvSessionVariables   :: !UserVars
+  -- ^ Only the session variables that are required for this query. For example,
+  -- if a permisson for 'user' role  on 'article' table is @{"is_public": true}@,
+  -- the result of a subscription on 'article' table for 'user' role is not dependent on
+  -- any session variables. But a 'user' role is most likely accompanied with
+  -- a session variable such as @user_id@, so if we include the 'user_id' as part
+  -- of the cohort's variables, we'll end up with multiple cohorts retrieving the
+  -- same result. Hence, only the session variables that will be referenced in the
+  -- generated SQL should be part of the cohort
+  --
```
lexi-lambda, line 106: This is a very helpful comment, but I admit I had to read it several times before I understood it, I think because it is something of a wall of text. Here is my attempt at reformatting it; feel free to incorporate only the parts of my suggestion you like.

```suggestion
-- ^ A set of session variables, pruned to the minimal set actually used by
-- this query. To illustrate the need for this pruning, suppose we have the
-- following query:
--
-- > query {
-- >   articles {
-- >     id
-- >     title
-- >   }
-- > }
--
-- If the select permission on @articles@ is just @{"is_public": true}@, we
-- just generate the SQL query
--
-- > SELECT id, title FROM articles WHERE is_public = true
--
-- which doesn’t use any session variables at all. Therefore, we ought to be
-- able to multiplex all queries of this shape into a single cohort, for quite
-- good performance! But if we don’t prune the session variables, we’ll
-- needlessly split subscribers into several cohorts simply because they have
-- different values for, say, @X-Hasura-User-Id@.
--
-- The 'mkCohortVariables' smart constructor handles pruning the session
-- variables to a minimal set, avoiding this pessimization.
```

server/src-lib/Hasura/RQL/Types/Permission.hs
```diff
@@ -66,6 +68,13 @@ newtype UserVars
   = UserVars { unUserVars :: Map.HashMap SessVar SessVarVal}
   deriving (Show, Eq, FromJSON, ToJSON, Hashable)
 
+-- | We need this to transform the underlying session variable map
+-- as the UserVars constructor is not exposed
+mapUserVars
+  :: (Map.HashMap SessVar SessVarVal -> Map.HashMap SessVar SessVarVal)
+  -> UserVars -> UserVars
+mapUserVars f = UserVars . f . unUserVars
```
lexi-lambda, line 76: This seems like it somewhat defeats the purpose of having the smart constructor, since one can trivially defeat the intended invariants by writing `mapUserVars (const myEvilHashMap) (mkUserVars [])`. Of course, I do not expect anyone would actually do that, but they might end up violating the invariants by mistake in a more complicated way.

Perhaps define a more restrictive `filterUserVars`, instead? Or use `mkUserVars` to reconstruct the result instead of the `UserVars` constructor?

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
lexi-lambda: Isn’t that what `_lqpeSql` is?

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -211,15 +227,61 @@ validateVariables pgExecCtx variableValues = do
...
+-- | Internal: Used to collect information about various parameters
+-- of a subscription field's AST as we resolve them to SQL expressions.
+data QueryParametersInfo
+  = QueryParametersInfo
+  { _qpiReusableVariableValues     :: !GV.ReusableVariableValues
+  , _qpiSyntheticVariableValues    :: !(Seq (WithScalarType PGScalarValue))
+  , _qpiReferencedSessionVariables :: !(Set.HashSet SessVar)
+  -- ^ The session variables that are referenced in the query root fld's AST.
+  -- This information is used to determine a cohort's required session
+  -- variables
```
lexi-lambda, line 243: I think it would be nice to pull the Haddock comment on `_cvSessionVariables` out into a `Note` and reference it in three places: the definition of `_cvSessionVariables`, the definition of `mkCohortVariables`, and the definition of `_qpiReferencedSessionVariables` (here). That way it’s much clearer what this is all about, but you don’t have to repeat yourself.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -130,54 +117,83 @@ data CohortVariables
...
-    J.object ["session" J..= sessionVars, "query" J..= queryVars, "synthetic" J..= syntheticVars]
+    J.object [ "session" J..= sessionVars
+             , "query" J..= queryVars
+             , "synthetic" J..= syntheticVars
+             ]
 
 -- These types exist only to use the Postgres array encoding.
-newtype CohortIdArray = CohortIdArray { unCohortIdArray :: [CohortId] }
+newtype CohortIdArray
+  = CohortIdArray { unCohortIdArray :: [CohortId] }
```
lexi-lambda, line 157: Why do this? The original line was under 80 characters!

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -130,54 +117,83 @@ data CohortVariables
...
+  -> ValidatedQueryVariables
+  -> ValidatedSyntheticVariables
+  -> CohortVariables
+mkCohortVariables requiredSessionVariables sessionVariableValues =
+  CohortVariables requiredSessionVariableValues
+  where
+    requiredSessionVariableValues =
+      mapUserVars
+      (Map.filterWithKey (\k _ -> Set.member k requiredSessionVariables))
+      sessionVariableValues
```
lexi-lambda, line 146: Does having this in a separate binding really add much? (Especially if you move the details into a `Note` as suggested above and reference it here.)

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
```diff
@@ -211,15 +227,61 @@ validateVariables pgExecCtx variableValues = do
...
+        modifying qpiReusableVariableValues $ Map.insert varName colVal
+        pure ["query", G.unName $ G.unVariable varName]
+      Nothing -> do
+        syntheticVarIndex <- use (qpiSyntheticVariableValues . to length)
+        modifying qpiSyntheticVariableValues (|> colVal)
+        pure ["synthetic", T.pack $ show syntheticVarIndex]
+    pure $ fromResVars (PGTypeScalar $ pstType colVal) varJsonPath
+  GR.UVSessVar ty sessVar -> do
+    modifying qpiReferencedSessionVariables (Set.insert sessVar)
+    pure $ fromResVars ty ["session", T.toLower sessVar]
```
lexi-lambda, line 269: I wonder if this really needs to be done statefully. It makes me a little sad to do it this way, because there’s really no reason this needs to be imperative. The assignment of synthetic variables *does* need to be done statefully, since we’re assigning unique indices, but that is not true for this.

What if we did this separately in `buildLiveQuery` plan in a more direct fashion? I think something like this would work:

```haskell
let referencedSessionVars = astUnresolved ^.. traverseQueryRootFldAST._UVSessVar._2
```

where `(^..)` is our generalized variant from `Control.Lens.Extended`.

0x777: > It makes me a little sad to do it this way, because there’s really no reason this needs to be imperative

Yup! Collecting variables too, need not be done statefully. I took this approach so that I don't have to worry about whether those three traversals will result in a single traversal in the generated code.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
0x777: It must have happened because of muscle memory - that's the code style I'm used to (and most of the current code is in this form if I'm not mistaken)!

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs
lexi-lambda: That seems highly unlikely to have a meaningful impact on performance given everything else going on in the pipeline. If you can measure *any* difference between the two approaches, I will eat my words! 🙂 But otherwise, I’d rather do the thing that’s easier to understand.

lexi-lambda:
For posterity: this is currently blocked on a strategy for writing regression tests for this optimization.

tirumaraiselvan (approved):
TODO: Changelog

lexi-lambda (approved):

maxpain:
@0x777 @lexi-lambda when will it be merged to the stable branch?

PR #4333, opened 2020-04-08
allow reusing Postgres scalars in custom types (close #4125)

### LINKED ISSUE #4125: Reuse Hasura postgres scalars with actions
Allow using Hasura postgres scalars for Actions, which otherwise leads to a lot of unnecessary type duplication on a codegen-ed client.

rakeshkky:
### Description

Allow reusing Postgres scalars in custom types. Now actions GraphQL schema can reuse Postgres base types as scalars. 
For example, I can set the type of field in the action output type as `geography` (from postgis extension) which is defined in Postgres database.

```graphql
# define a custom scalar
scalar Date

# The action output type
type User {
  id: ID!
  name: String!
  location: geography # Postgres scalar
  created_at: Date! # Custom scalar
}
```

### Changelog

### Affected components

- [x] Server
- [ ] Docs
- [x] Tests

### Related Issues

close #4125 

### Solution and Design

Fetch all available base types from postgres along with catalog metadata. Use the same for validating custom types and publish the reused scalars in the schema.

### Steps to test and verify

Use `set_custom_types` `/v1/query` API to define custom types. Use any Postgres scalar type to declare the type of any field in `objects` or `input_objects` without explicitly defining them in `scalars`. 

N/A

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The order and collapsed state of columns is now persisted across page navigation
 
 ### Bug fixes and improvements
 
+- server: reusing Postgres scalars in custom types is allowed (close #4125)
```
lexi-lambda, line 14: This is probably subjective, but I think this would read more positively if written in [the active voice](https://en.wikipedia.org/wiki/Active_voice)! Something like this:

```suggestion
- server: Support reusing Postgres scalars in custom types (close #4125)
```

server/src-lib/Hasura/RQL/DDL/CustomTypes.hs
```diff
@@ -23,12 +24,13 @@ import           Hasura.GraphQL.Schema.CustomTypes (buildCustomTypesSchemaPartia
 
 validateCustomTypeDefinitions
   :: (MonadValidate [CustomTypeValidationError] m)
-  => TableCache -> CustomTypes -> m ()
-validateCustomTypeDefinitions tableCache customTypes = do
+  => TableCache -> CustomTypes -> [PGScalarType] -> m [PGScalarType]
```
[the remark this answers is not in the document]
rakeshkky: I've documented the changes in source code in the latest commit.

docs/graphql/manual/actions/types/index.rst
```diff
@@ -141,6 +141,11 @@ a scalar called ``Date``, you can define it like.
 These scalars can be used as arguments of the mutation or as fields of object
 types and input types.
 
+.. admonition:: Postgres scalars
+
+   The base types from Postgres can be reusable as custom scalars without explicitly declaring them.
```
marionschleifer, line 146:
```suggestion
   The Postgres base types are reusable as custom scalars without explicitly declaring them.
```

docs/graphql/manual/actions/types/index.rst
```diff
@@ -141,6 +141,11 @@ a scalar called ``Date``, you can define it like.
...
+   For example, the ``uuid`` type from ``pgcrypto`` extension can be used to resolve a UUID data.
```
marionschleifer, line 147:
```suggestion
   For example, the ``uuid`` type from the ``pgcrypto`` extension can be used to resolve a ``uuid`` data field.
```

marionschleifer: @rakeshkky I was not sure what you meant by "resolve a uuid data". I added a suggestion, please check for correctness 🙂

server/src-lib/Hasura/RQL/DDL/CustomTypes.hs
```diff
@@ -15,20 +15,26 @@ import qualified Database.PG.Query                 as Q
...
 
 import           Hasura.GraphQL.Schema.CustomTypes (buildCustomTypesSchemaPartial)
 
+-- | Validate the custom types and return any reused Postgres base types (as scalars)
 validateCustomTypeDefinitions
   :: (MonadValidate [CustomTypeValidationError] m)
-  => TableCache -> CustomTypes -> m ()
-validateCustomTypeDefinitions tableCache customTypes = do
+  => TableCache -- ^ The Postgres tables
+  -> CustomTypes -- ^ Custom types subject to validation
```
lexi-lambda, line 29: I think the role of these two arguments is self-explanatory given their types alone (in combination with the function name), so I’d drop the comments here (since they’re just basically just restating existing information). The other two comments are great, though, thank you for them!

```suggestion
  => TableCache
  -> CustomTypes
```

server/src-lib/Hasura/RQL/Types/Catalog.hs
```diff
@@ -139,6 +140,20 @@ instance NFData CatalogFunction
...
 
+data CatalogCustomTypes
+  = CatalogCustomTypes
+  { _cctCustomTypes :: !CustomTypes
+  , _cctPgScalars   :: ![PGScalarType]
+  -- ^ Base types from Postgres required for validating custom types
+  -- to check if any one of them uses the Postgres base type name
+  -- and publish them in generated GraphQL schema as Scalars.
+  -- These won't be tracked in the metadata.
+  -- See https://github.com/hasura/graphql-engine/issues/4125
```
lexi-lambda, line 151: > `Base types from Postgres required for validating custom types to check if any one of them uses the Postgres base type name and publish them in generated GraphQL schema as Scalars.`

I’m afraid I don’t understand this sentence. 🙁 Perhaps it could be reworded as something like this:

```haskell
-- ^ All Postgres base types, which may be referenced in custom type
-- definitions. When we validate the custom types (see
-- 'validateCustomTypeDefinitions'), we record which base types were referenced
-- so that we can be sure to include them in the generated GraphQL schema.
```

> `These won't be tracked in the metadata.`

Why not? Presumably because we fetch them on-demand from the types in the GraphQL database, but then it would be good to explicitly say so!

```haskell
-- These are not actually part of the Hasura metadata --- we fetch them from
-- @pg_catalog.pg_type@ --- but they’re needed when validating the custom type
-- metadata, so we include them here.
```

> `See https://github.com/hasura/graphql-engine/issues/4125`

Generally, I like linking relevant issues! But in this case, #4125 doesn’t really include any useful information in its description (which is unfortunate, but a separate problem), so I’d just drop the link.

server/src-rsr/catalog_metadata.sql
```diff
@@ -173,6 +173,15 @@ from
...
       ) fi on 'true'
   ) as computed_field,
+  (
+    select
+      json_build_object(
+        'custom_types',
+         coalesce((select custom_types from hdb_catalog.hdb_custom_types), '{}'),
+        'pg_scalars',
+         coalesce((select json_agg(typname) from pg_catalog.pg_type where typtype = 'b'), '[]')
+      ) as item
```
lexi-lambda, line 183: This is the sort of situation where [GHC’s `Note`s](https://github.com/hasura/graphql-engine/blob/995e198d0af405a67415ef3d2a178a4c5bda1d54/server/STYLE.md#out-of-line-notes) really shine. It’s not immediately clear from reading this code why we’re fetching things from `pg_type` here, and we *could* figure it out by diligently following the control flow until we find the comment on the `CatalogCustomTypes` definition. But to make it easier on future readers, we could also add a `Note`:

```haskell
{- Note [Postgres scalars in custom types]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
It’s very convenient to be able to reference Postgres scalars in custom type
definitions. For example, we might have a type like this:

    type User {
      id: uuid!
      name: String
      location: geography
    }

The uuid and geography types are Postgres scalars, not separately-defined
GraphQL types. To support this, we have to take a few extra steps:

  1. The set of Postgres base types is not fixed; extensions like PostGIS add
     new ones, and users can even define their own. Therefore, we fetch the
     currently defined base types from the @pg_type@ system table as part of
     loading the metadata.

  2. It’s possible for a custom type definition to use a type that doesn’t
     appear elsewhere in the GraphQL schema, so we record which base types were
     referenced while validating the custom type definitions and make sure to
     include them in the generated schema explicitly. -}
```

You could put that somewhere in `Hasura.RQL.DDL.CustomTypes`, perhaps next to the `validateCustomTypeDefinitions` function, and then you could add a comment in `catalog_metadata.sql` that points to it:

```sql
      json_build_object(
        'custom_types',
         coalesce((select custom_types from hdb_catalog.hdb_custom_types), '{}'),
        'pg_scalars', -- see Note [Postgres scalars in custom types]
         coalesce((select json_agg(typname) from pg_catalog.pg_type where typtype = 'b'), '[]')
      ) as item
```

You could also include references to the `Note` in the comment attached to `CatalogCustomTypes` and in the body of the `validateCustomTypeDefinitions` function. That way you can just state all the tricky details in one spot, and you don’t have to restate yourself in several different places! 🙂

server/src-lib/Hasura/RQL/DDL/CustomTypes.hs
```diff
@@ -15,20 +15,26 @@ import qualified Database.PG.Query                 as Q
...
+  -> [PGScalarType] -- ^ List of all Postgres base types
+  -> m [PGScalarType] -- ^ List of reused Postgres base types (as scalars) in custom types
+validateCustomTypeDefinitions tableCache customTypes allPGScalars = do
   unless (null duplicateTypes) $ dispute $ pure $ DuplicateTypeNames duplicateTypes
   traverse_ validateEnum enumDefinitions
-  traverse_ validateInputObject inputObjectDefinitions
-  traverse_ validateObject objectDefinitions
+  inputObjectReusedPGScalars <- concat <$> traverse validateInputObject inputObjectDefinitions
+  objectReusedPGScalars <- concat <$> traverse validateObject objectDefinitions
```
lexi-lambda, line 36: All this threading around of the collected scalars seems like it would be much easier to do using [a writer monad](https://hackage.haskell.org/package/mtl-2.2.2/docs/Control-Monad-Writer-Strict.html). And if you use a `HashSet PGScalarType` instead of a `[PGScalarType]`, you won’t have to worry about duplicates! So you could just write

```haskell
validateCustomTypeDefinitions tableCache customTypes allPGScalars = execWriterT do
  unless (null duplicateTypes) $ dispute $ pure $ DuplicateTypeNames duplicateTypes
  traverse_ validateEnum enumDefinitions
  traverse_ validateInputObject inputObjectDefinitions
  traverse_ validateObject objectDefinitions
```

and every time you find a reference to a scalar type, you could write `tell $ S.singleton reusedScalar`.

docs/graphql/manual/actions/types/index.rst
```diff
@@ -141,6 +141,11 @@ a scalar called ``Date``, you can define it like.
 These scalars can be used as arguments of the mutation or as fields of object
 types and input types.
 
+.. admonition:: Postgres scalars
+
+   Postgres base types are reusable as custom scalars without explicitly declaring them.
```
marionschleifer, line 146: @rakeshkky do you think we need to explain a little more what this means? Like that the PG base types will be exposed as GraphQL scalars by Hasura? Since I don't understand this in depth, it's hard for me to make specific suggestions. Maybe @lexi-lambda can help?

rakeshkky: @marionschleifer I tried to improve the documentation in the latest commit.

marionschleifer: Oh, didn't see that, sorry!

wawhal:
This also needs changes on the CLI because we now need the introspection schema to parse the SDL.

console/src/shared/utils/deriveAction.js
```diff
@@ -170,7 +177,10 @@ const deriveAction = (
       name: v.variable.name.value,
     };
     const argTypeMetadata = getAstTypeMetadata(v.type);
-    if (!inbuiltTypes[argTypeMetadata.typename]) {
+    if (
+      !inbuiltTypes[argTypeMetadata.typename] &&
+      !isHasuraScalar(argTypeMetadata.type)
```
wawhal, line 182:
```suggestion
      !isHasuraScalar(argTypeMetadata.typename)
```

wawhal: Also @beerose, the `isScalarType` function handles `undefined` and `null` as well. So you can directly use `isScalar(allHasuraTypes[name])` in the `isHasuraScalar` function.

PR #4334, opened 2020-04-08
type is not required for jwk_url in JWT config

tirumaraiselvan:
### Description

`type` is not required for JWT config if configured with `jwk_url`

The `type` key is only used to parse the type of `HS` algorithms so that we can throw a short key error. Nothing else. So it might be worth considering removing it alltogether (in future).

### Changelog

- [ ] `CHANGELOG.md` is updated with user-facing content relevant to this PR. (TODO)

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs (TODO)
- [ ] Community Content
- [ ] Build System
- [x] Tests 
- [ ] Other (list it)


server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -400,13 +400,14 @@ instance J.FromJSON JWTConfig where
       (Nothing, Nothing) -> fail "key and jwk_url both cannot be empty"
       (Just _, Just _)   -> fail "key, jwk_url both cannot be present"
       (Just rawKey, Nothing) -> do
-        key <- parseKey keyType rawKey
+        keyType' <- onNothing keyType $ fail "type of key not provided"
+        key <- parseKey rawKey keyType'
```
lexi-lambda, line 404: In the PR description, you write:

> The `type` key is only used to parse the type of `HS` algorithms so that we can throw a short key error. Nothing else. So it might be worth removing it alltogether.

So it sounds like we only do anything with it right here. In that case, it sounds like we might as well drop it from the `JWTConfig` type, and we can move the `keyType <- o J..: "type"` statement directly into this nested `do` block:

```haskell
(Just rawKey, Nothing) -> do
  keyType <- o J..: "type"
  key <- parseKey rawKey keyType
```

Then we don’t have to bother with `J.:?` only to assert its presence later in this branch.

tirumaraiselvan: That does sound like it reduces some code but it also reduces visibility of the what the config can have at the top level. I personally feel it's more readable if everything is at one place.

So, we can remove the `type` field from JWTConfig but let's parse it at the top?

lexi-lambda: I disagree. The purpose of aeson’s monadic `Parser` abstraction is to give you the ability to use control flow to make decisions about parsing. Taking advantage of that is the idiomatic thing to do here; parsing something outside of a branch that’s only used inside the branch would be more confusing to me than parsing it locally!

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -369,12 +368,12 @@ verifyJwt ctx (RawJWT rawJWT) = do
...
 
 instance J.ToJSON JWTConfig where
-  toJSON (JWTConfig ty keyOrUrl claimNs aud claimsFmt iss) =
+  toJSON (JWTConfig keyOrUrl claimNs aud claimsFmt iss) =
     case keyOrUrl of
          Left _    -> mkObj ("key" J..= J.String "<JWK REDACTED>")
          Right url -> mkObj ("jwk_url" J..= url)
     where
-      mkObj item = J.object [ "type" J..= ty
+      mkObj item = J.object [ "type" J..= J.String "<TYPE REDACTED>"
```
tirumaraiselvan, line 376: This is bit clunky. It should be ideally with the `Left` pattern match right above. Hints on how to accomplish that?

lexi-lambda: I’m afraid I don’t think I understand your question—could you elaborate?

tirumaraiselvan: The `toJSON` would ideally not have `type` if `jwk_url` is provided. Currently, it was being serialized to have `type` for `jwk_url` as well. So now I need to send either two items ("key" and "type" on `Left`) or one item ("jwk_url" on Right).

lexi-lambda: Perhaps something like this?

```haskell
instance J.ToJSON JWTConfig where
  toJSON (JWTConfig keyOrUrl claimNs aud claimsFmt iss) =
    case keyOrUrl of
      Left _    -> J.object $ [ "type" J..= J.String "<TYPE REDACTED>"
                              , "key" J..= J.String "<JWK REDACTED>"
                              ] ++ sharedFields
      Right url -> J.object $ ("jwk_url" J..= url) : sharedFields
    where
      sharedFields = [ "claims_namespace" J..= claimNs
                     , "claims_format" J..= claimsFmt
                     , "audience" J..= aud
                     , "issuer" J..= iss
                     ]
```

lexi-lambda: …or perhaps this would be slightly nicer:

```haskell
instance J.ToJSON JWTConfig where
  toJSON (JWTConfig keyOrUrl claimNs aud claimsFmt iss) =
    J.object (jwkFields ++ sharedFields)
    where
      jwkFields = case keyOrUrl of
        Left _    -> [ "type" J..= J.String "<TYPE REDACTED>"
                     , "key" J..= J.String "<JWK REDACTED>" ]
        Right url -> [ "jwk_url" J..= url ]
      sharedFields = [ "claims_namespace" J..= claimNs
                     , "claims_format" J..= claimsFmt
                     , "audience" J..= aud
                     , "issuer" J..= iss
                     ]
```

tirumaraiselvan: Haha. Thanks. I guess there ain't no magic to do this with one or two symbols xD

PR #4365, opened 2020-04-10
accept a new argument `claims_namespace_path` in JWT config

### LINKED ISSUE #4349: introduce `claims_namespace_path` in JWT config
Currently, we have `claims_namespace` which takes a string value. This represents the top-level key in JWT payload inside of which hasura claims are to be found. This is an optional field and if not provided, the default `https://hasura.io/jwt/claims` is used. See:  https://hasura.io/docs/1.0/graphql/manual/auth/authentication/jwt.html#claims-namespace

The problem with this is that this is not customizable i.e. we cannot look for hasura claims either at the top level or even at a more nested level.

Hence, introduce a new optional field called `claims_namespace_path` which takes a JSON path value to the hasura claims. So a value of `$` would mean top-level, `$.other.hasura` would mean 2-levels of nesting and so on.

If both `claims_namespace_path` and `claims_namespace` are given, we can choose the value of `claims_namespace_path`. This way we can deprecate `claims_namespace` in the future.

Example:

```
{
  "sub": "1234567890",
  "name": "John Doe",
  "admin": true,
  "iat": 1516239022,
  "x-hasura-allowed-roles": ["editor","user", "mod"],
  "x-hasura-default-role": "user",
  "x-hasura-user-id": "1234567890",
  "x-hasura-org-id": "123",
  "x-hasura-custom": "custom-value"
}
```

codingkarthik:
fixes #4349 

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Server/Utils.hs
```diff
@@ -227,3 +229,17 @@ makeReasonMessage errors showError =
     [singleError] -> "because " <> showError singleError
     _ -> "for the following reasons:\n" <> T.unlines
          (map (("  • " <>) . showError) errors)
+
+executeJSONPath :: JSONPath -> Value -> IResult Value
+executeJSONPath jsonPath = iparse (valueParser jsonPath)
+  where
+    valueParser path value = case path of
+      []                      -> pure value
+      [pathElement]           -> parseWithPathElement pathElement value
```
rakeshkky, line 238: This case match isn't required :smiley: . The next match handles this case too. Feel free to delete this line.

server/src-lib/Data/Parser/JSONPath.hs
```diff
@@ -64,6 +61,36 @@ parseJSONPath = parseResult . parse parseElements
...
+    isIdentifierKey []     = False
+    isIdentifierKey (x:xs) = isAlpha x && all isAlphaNum xs
+
+    escapeKey :: String -> String
+    escapeKey = concatMap escapeChar
+
+    escapeChar :: Char -> String
+    escapeChar '\'' = "\\'"
+    escapeChar '\\' = "\\\\"
+    escapeChar c    = [c]
```
rakeshkky, line 96: We have `encodeJSONPath` function in `Hasura.RQL.Types.Error` module. How it is different from `formatPath`? If the later is more robust than the former, then please reuse `formatPath` in place of `encodeJSONPath`.

server/src-lib/Hasura/Server/Utils.hs
codingkarthik: If I omit the `[]` case, it will give a non-exhaustive pattern match warning, so that's why I have included it

server/src-lib/Data/Parser/JSONPath.hs
codingkarthik: I didn't know about the `encodeJSONPath` function, I think the `formatPath` function seems more robust so I'll use that

server/src-lib/Hasura/Server/Utils.hs
rakeshkky: No No, I mean 
```haskell
 [pathElement]           -> parseWithPathElement pathElement value
```
this line.

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -130,6 +130,7 @@ JSON object:
      "jwk_url": "<optional-url-to-refresh-jwks>",
      "claims_namespace": "<optional-key-name-in-claims>",
      "claims_format": "json|stringified_json",
+     "claims_namespace_path":"<optional-JSON-Path-to-the-claims>",
```
marionschleifer, line 133:
```suggestion
     "claims_namespace_path":"<optional-json-path-to-the-claims>",
```

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -168,6 +169,35 @@ https://tools.ietf.org/html/rfc7517.
 This is an optional field. You can also provide the key (certificate, PEM
 encoded public key) as a string - in the ``key`` field along with the ``type``.
 
+``claims_namespace_path``
+^^^^^^^^^^^^^^^^^^^^^^^^^
+An optional JSON path value to the hasura claims in the JWT token (e.g. ``$.hasura.claims``)
```
marionschleifer, line 174:
```suggestion
An optional JSON path value to the Hasura claims in the JWT token (e.g. ``$.hasura.claims``).
```

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -168,6 +169,35 @@ https://tools.ietf.org/html/rfc7517.
...
+
+For example, the JWT token should in this format if the ``claims_namespace_path`` is
```
marionschleifer, line 176:
```suggestion
The JWT token should be in this format if the ``claims_namespace_path`` is
```

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -168,6 +169,35 @@ https://tools.ietf.org/html/rfc7517.
...
+          "x-hasura-default-role": "user",
+          "x-hasura-user-id": "1234567890",
+          "x-hasura-org-id": "123",
+          "x-hasura-custom": "custom-value"
+       }
+     }
+  }
+
+
+This is an optional field, if the value is not provided, then the ``claims_namespace`` value
```
marionschleifer, line 198:
```suggestion
This is an optional field. If the value is not provided, then the ``claims_namespace`` value
```

server/tests-py/test_jwt.py
```diff
@@ -42,23 +42,33 @@ def mk_claims(conf, claims):
...
+        claims = mk_claims(hge_ctx.hge_jwt_conf, {
             'x-hasura-user-id': '1',
             'x-hasura-allowed-roles': ['contractor', 'editor'],
             'x-hasura-default-role': 'contractor'
         })
+        claims_namespace_path = hge_ctx.hge_jwt_claims_ns_path
+        if claims_namespace_path is None:
+            self.claims['https://hasura.io/jwt/claims'] = claims
+        elif claims_namespace_path == "$.hasuraClaims":
+            self.claims['hasuraClaims'] = claims
```
rakeshkky, line 71: This logic is repeated lot of places in the same file. If possible, can we abstract this into a new function and reuse it?

server/tests-py/conftest.py
```diff
@@ -39,6 +39,10 @@ def pytest_addoption(parser):
         "--hge-jwt-conf", metavar="HGE_JWT_CONF", help="The JWT conf", required=False
     )
 
+    parser.addoption(
+        "--hge-jwt-claims-ns-path", metavar="HGE_JWT_CLAIMS_NS_PATH", help="The JWT conf claims namespace path", required=False
```
rakeshkky, line 43: Isn't `claims_namespace_path` present in `--hge-jwt-conf`?

server/src-lib/Hasura/Server/Config.hs
```diff
@@ -10,13 +10,16 @@ import           Hasura.Prelude
...
+import           Data.Text                                (pack)
 
 import qualified Hasura.GraphQL.Execute.LiveQuery.Options as LQ
 
 data JWTInfo
   = JWTInfo
-  { jwtiClaimsNamespace :: !Text
-  , jwtiClaimsFormat    :: !JWTClaimsFormat
+  { jwtiClaimsNamespace     :: !(Maybe Text)
+  , jwtiClaimsNamespacePath :: !(Maybe Text)
```
rakeshkky, line 21: Both claims namespace and namespace path are orthogonal and only one should present at a time. Right? Then you could model these into a sum type.

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -369,14 +394,19 @@ verifyJwt ctx (RawJWT rawJWT) = do
...
         Left _    -> [ "type" J..= J.String "<TYPE REDACTED>"
                      , "key" J..= J.String "<JWK REDACTED>" ]
         Right url -> [ "jwk_url" J..= url ]
-      sharedFields = [ "claims_namespace" J..= claimNs
-                     , "claims_format" J..= claimsFmt
+
+      claimsNsFields = case claimNs of
+        ClaimNsPath nsPath ->
+          ["claims_namespace_path" J..= (encodeJSONPath nsPath)]
+        ClaimNs ns -> ["claims_ns" J..= J.String ns]
```
rakeshkky, line 407: Shouldn't `"claims_ns"` be `"claims_namespace"`?

server/src-lib/Hasura/Server/Auth/JWT.hs
```diff
@@ -389,23 +419,42 @@ instance J.FromJSON JWTConfig where
...
-    case (mRawKey, jwkUrl) of
-      (Nothing, Nothing) -> fail "key and jwk_url both cannot be empty"
-      (Just _, Just _)   -> fail "key, jwk_url both cannot be present"
-      (Just rawKey, Nothing) -> do
-        keyType <- o J..: "type"
-        key <- parseKey rawKey keyType
-        return $ JWTConfig (Left key) claimNs aud isStrngfd iss
-      (Nothing, Just url) ->
-        return $ JWTConfig (Right url) claimNs aud isStrngfd iss
+    let hasuraClaimsNs :: Either String JWTConfigClaims
```
rakeshkky, line 428: You can directly `parseJSONPath` and `fail` in single line
```haskell
     maybeClaimNsPath <- (o J..:? "claims_namespace_path")  >>= mapM (either failJSONPathParsing pure . JSONPath.parseJSONPath)
```
in `where`
```haskell
failJSONPathParsing err = fail $ "invalid JSON path claims_namespace_path error: " ++ err
```
Compute `hasuraClaimsNs` without any error handling

```haskell
hasuraClaimsNs = case maybeClaimNsPath of
  Just claimsNsPath, -> pure $ ClaimNsPath claimsNsPath
  Nothing -> pure $ ClaimNs $ fromMaybe defaultClaimNs claimNs
```

Now code is much read-able and pretty. :slightly_smiling_face:

server/tests-py/test_jwt.py
codingkarthik: Yes, I have done it in the new commit

server/src-lib/Hasura/Server/Config.hs
codingkarthik: Both of the fields can be set at the same time, if both are set then we prefer the `claims_namespace_path` value. Initially, I thought we can display both the values but then I changed the logic of the claims namespace which will discard the `claims_namespace` value if `claims_namespace_path` is present

server/tests-py/conftest.py
codingkarthik: Yes, I have modified the logic to parse the `claims_namespace_path` from `hge-jwt-conf`

server/src-lib/Hasura/Server/Config.hs
tirumaraiselvan: I am kinda leaning towards the behaviour that we throw an error if both are set at the same time.

tirumaraiselvan: If at all someone has a use-case where they want such fall-back behaviour, we can introduce this.

The main reason is to keep the config tight and obvious in reasoning.

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -168,6 +169,35 @@ https://tools.ietf.org/html/rfc7517.
...
+          "x-hasura-default-role": "user",
+          "x-hasura-user-id": "1234567890",
+          "x-hasura-org-id": "123",
+          "x-hasura-custom": "custom-value"
+       }
+     }
+  }
+
+
+This is an optional field. If the value is not provided, then the ``claims_namespace`` value
```
tirumaraiselvan, line 198: I believe this is no longer valid.

codingkarthik: I'll update this

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -92,7 +92,8 @@ etc.) JWT claims, as well as Hasura specific claims inside a custom namespace
 (or key) i.e. ``https://hasura.io/jwt/claims``.
 
 The ``https://hasura.io/jwt/claims`` is the custom namespace where all Hasura
-specific claims have to be present. This value can be configured in the JWT
+specific claims have to be present. This value can be configured using
```
tirumaraiselvan, line 95: This feels is misleading (also in the paragraph above). It should say `https://hasura.io/jwt/claims` something along the lines of `https://hasura.io/jwt/claims` is the default namespace when `claims_namespace` or `claims_namespace_path` are not provided.

tirumaraiselvan: NVM, this was explaining the example JWT above.

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -103,6 +104,14 @@ mandatory, while the rest of them are optional.
    All ``x-hasura-*`` values should be of type ``String``, they will be converted to the
    right type automatically.
 
+.. note::
+
+   The JWT config should either have the ``claims_namespace`` or ``claims_namespace_path``
```
tirumaraiselvan, line 109: This note should come in the below section. It feels too much to be in the example section.

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -130,6 +139,7 @@ JSON object:
      "jwk_url": "<optional-url-to-refresh-jwks>",
      "claims_namespace": "<optional-key-name-in-claims>",
      "claims_format": "json|stringified_json",
+     "claims_namespace_path":"<optional-json-path-to-the-claims>",
```
tirumaraiselvan, line 142: why not put `claims_namespace_path` right below `claims_namespace` as they are so similar?

docs/graphql/manual/auth/authentication/jwt.rst
codingkarthik: I'll update it

docs/graphql/manual/auth/authentication/jwt.rst
codingkarthik: I'll update it

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -256,6 +222,38 @@ inside which the Hasura specific claims will be present, e.g. ``https://mydomain
...
+          "x-hasura-user-id": "1234567890",
+          "x-hasura-org-id": "123",
+          "x-hasura-custom": "custom-value"
+       }
+     }
+  }
+
+.. note::
+
+   The JWT config can only have one of claims_namespace or claims_namespace_path values set.
```
marionschleifer, line 254:
```suggestion
   The JWT config can only have one of ``claims_namespace`` or ``claims_namespace_path`` values set.
```

docs/graphql/manual/auth/authentication/jwt.rst
```diff
@@ -256,6 +222,38 @@ inside which the Hasura specific claims will be present, e.g. ``https://mydomain
...
+   If neither keys are set, then the default value of
+   claims_namespace i.e. https://hasura.io/jwt/claims will be used.
```
marionschleifer, line 256:
```suggestion
   ``claims_namespace`` i.e. https://hasura.io/jwt/claims will be used.
```

PR #4372, opened 2020-04-12
server: create hlint config, apply most hints

nicuveo:
### Description
This PR only applies minor changes, and creates a slightly better-than-default hlint config.

I was exploring the idea of automating some of our [style guide](https://github.com/hasura/graphql-engine/blob/master/server/STYLE.md), by either using stylish-haskell as a pre-commit hook, instrumenting hlint, or trying perhaps other formatters. Regard hlint in particular: while in the past it seems we used to enforce [strict hlint use](https://github.com/hasura/graphql-engine/blame/master/server/CONTRIBUTING.md#L178), the codebase currently has a lot of hlint warnings, a lot of which are false positive. Bluntly activating it would be a bad idea; however, we could choose to make a very selective use of it: disable almost all regular errors, and write our own custom rules.

What this PR does is a step towards that: create a hlint config which disables a lot of the false positive messages. Furthermore, it fixes most of the actually actionable suggestions, and applies changes to match the style guide on some of the file. I've only applied changes that I believe to be uncontroversial, and that should be deemed improvements even if we don't end up using hlint at all. 

### Affected components

- [ ] Server

### DIFF
--- a/server/.hlint.yaml
+++ b/server/.hlint.yaml
@@ -1,0 +1,72 @@
+# HLint configuration file
+# https://github.com/ndmitchell/hlint
+##########################
+
+# This file contains a template configuration file, which is typically
+# placed as .hlint.yaml in the root of your project
+
+
+# Specify additional command line arguments
+#
+# - arguments: [--color, --cpp-simple, -XQuasiQuotes]
+
+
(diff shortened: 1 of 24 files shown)


abooij:
I like this and it would be great if we could get this merged.

PR #4375, opened 2020-04-12
server: run stylish-haskell on all files

nicuveo:
### Description
A small PR that simply updates our stylish-haskell config, and runs it on all files, which does nothing more than remove whitespace and align imports. Two changes worth noting in the config:
- newline style is switched from 'native' to 'lf', so that whichever OS one contributor uses won't suddenly change all newlines
- the tabs step is now enabled, as our style guide explicitly recommends spaces

### Affected components
- [ ] Server

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Validate/Field.hs
+++ b/server/src-lib/Hasura/GraphQL/Validate/Field.hs
@@ -144,7 +144,7 @@ withDirectives dirs act = do
       when (isJust $ _aivVariable val) markNotReusable
       case _aivValue val of
         AGScalar _ (Just (PGValBoolean v)) -> return v
-        _ -> throw500 "did not find boolean scalar for if argument"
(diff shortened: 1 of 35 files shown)


server/src-lib/Hasura/GraphQL/Validate/Field.hs
```diff
@@ -144,7 +144,7 @@ withDirectives dirs act = do
       when (isJust $ _aivVariable val) markNotReusable
       case _aivValue val of
         AGScalar _ (Just (PGValBoolean v)) -> return v
-        _ -> throw500 "did not find boolean scalar for if argument"
+        _                                  -> throw500 "did not find boolean scalar for if argument"
```
0x777, line 147: @nicuveo This kinda looks awkward. Is this a setting that's configurable? To not align if (longest - shortest) > some meaningful value?

Issue #4215, opened 2020-03-27
console: suspicious migrations created after column edition


vhiairrassary:
I am pretty sure the console creates suspicious migrations. Here a the reproductible steps:
- Start with a fresh Hasura instance & a fresh database as well;
- Set up a project directory & Open the console from the CLI (as show in the [documentation](https://hasura.io/docs/1.0/graphql/manual/migrations/new-database.html#migrations-new-db));
- Create a table `test` with a `ts` column, selecting `Timestamp` as the type (and not `Time`, `time without time zone` nor `timestamp without time zone`). Use `ts` as the PK. You shoud see the following migration now:
```sql
CREATE TABLE "public"."test"("ts" timestamptz NOT NULL, PRIMARY KEY ("ts") );
```
- Edit the `ts` column, change its `GraphQL field name` & save. Now you will see the following suspicious migrations:
```sql
ALTER TABLE "public"."test" ALTER COLUMN "ts" TYPE timestamp with time zone;
ALTER TABLE "public"."test" DROP CONSTRAINT "test_ts_key";
```

Hope this report will lead to a better product!

karthikvt26:
@vhiairrassary Is this `timestamp with time zone` vs `timestamptz` the problem? if not could you please explain why you think so?

0x777:
@karthikvt26 I think @vhiairrassary's point is that this migration shouldn't even be generated because only the custom field name of the column is updated?

karthikvt26:
I agree and atleast with v1.2 and above this should just add an entry in `tables.yaml` inside `metadata`.

karthikvt26:
Ok looks like a way forward is to identify any change to be either `metadata only` or `schema only` or `both`

vhiairrassary:
> @karthikvt26 I think @vhiairrassary's point is that this migration shouldn't even be generated because only the custom field name of the column is updated?

Exactly! For information, I am using version `1.2.0-beta.2`.

rikinsk:
This happens only with the timestamp column. there seems to be an inconsistency while comparing the type of a timestamp column during edit which causes console to believe the type of the column has changed and hence it generates an alter column type query

rikinsk:
related #3345

slimlime:
> This happens only with the timestamp column. there seems to be an inconsistency while comparing the type of a timestamp column during edit which causes console to believe the type of the column has changed and hence it generates an alter column type query
> 
[...]

PR #4393, opened 2020-04-13
console: don't generate unnecessary migrations (close #4125)

### LINKED ISSUE #4125: Reuse Hasura postgres scalars with actions

beerose:
Solves https://github.com/hasura/graphql-engine/issues/4215

### Description

The issue with additional migrations being generated after modifying a column occurs also for other data types than `timestamp` or `varchar` (mentioned in https://github.com/hasura/graphql-engine/issues/4215#issuecomment-605013116).

The problem was that we were comparing it by `display name` (we compared `column.data_type` with `column_edit.type` from the screenshots below):

Screenshot 2020-04-13 at 15 24 12
Screenshot 2020-04-13 at 15 24 25

Also, while investigated this, I noticed that we are generating migrations while editing a column if it's a primary key. The reason is that the logic behind determining whether the original column was unique only checks `tableSchema.unique_constraints` and doesn't check `primary_key` of the given table.

This PR fixes this as well.

### Changelog

### Affected components

- [x] Console


console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -84,8 +84,8 @@ const REMOVE_PRIMARY_KEY = 'ModifyTable/REMOVE_PRIMARY_KEY';
 const RESET_PRIMARY_KEY = 'ModifyTable/RESET_PRIMARY_KEY';
 const SET_PRIMARY_KEYS = 'ModifyTable/SET_PRIMARY_KEYS';
 
-const SET_COLUMN_EDIT = 'ModifyTable/SET_COLUMN_EDIT;';
-const RESET_COLUMN_EDIT = 'ModifyTable/RESET_COLUMN_EDIT;';
```
slimlime, line 88: Whoops ;';

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -1438,18 +1438,25 @@ const saveTableCommentSql = isTable => {
 };
 
 const isColumnUnique = (tableSchema, colName) => {
```
wawhal, line 1440: Can you undo this change?
I understand you did this because primary key is technically unique.
But the distinction here is based on the constraints. And it is confusing to mix the two. Maybe you can change the function name to `hasUniqueKeyConstraint` and `hasPKConstraint` for it to semantically be correct.

beerose: From Postgres POV column is unique if it has either a unique constraint or is a primary key. So I don't think it is **_so_** semantically incorrect. 

Also, the logic we had before resulted in unnecessary migrations.

wawhal: The `select` labelled as `Unique` in the column editor maps on to the unique constraint. Now that `select` will be true even if there is no unique key constraint on the column.

wawhal: From Postgres POV, Postgres has two separate things called `unique constraints` and `primary key constraints`, I think we should adhere to that so that users can view and exploit the two independently.

wawhal: For fixing the migrations, I think we should do proper comparison at the place where we are generating up and down migrations, rather than changing the logic that renders the app.

beerose: But this function `isColumnUnique` is scoped inside ModifyActions and only used to determine whether or not to create additional migrations.

wawhal: Here: https://github.com/hasura/graphql-engine/blob/master/console/src/components/Services/Data/TableModify/ModifyActions.js#L1449

You can read the column from Redux state and determine whether it is PK or UK or both.

beerose: > rather than changing the logic that renders the app.

Where is any change in rendering the app?

wawhal: Screenshot from 2020-04-24 15-20-38

Here. The `select` maps on to unique key constraints. With this PR, the `select` is a function of both PK and UK.

beerose: For this `select` we have logic in `ColumnEditorList`:
```
isUnique:
        (columnPKConstraints[colName] && pkLength === 1) ||
        columnUniqueConstraints[colName]
          ? true
          : false,
```

beerose: Where `columnPKConstraints` and `columnUniqueConstraints` are:
```
const columnPKConstraints = {};
  if (tableSchema.primary_key) {
    pkLength = tableSchema.primary_key.columns.length;

    tableSchema.primary_key.columns.forEach(col => {
      columnPKConstraints[col] = tableSchema.primary_key.constraint_name;
    });
  }

  const columnUniqueConstraints = {};
  tableSchema.unique_constraints.forEach(uConst => {
    if (uConst.columns.length === 1) {
      columnUniqueConstraints[uConst.columns[0]] = uConst.constraint_name;
    }
  });
```

wawhal: Wow it was already like this. My bad. I'll open an issue.
Approving.

wawhal: @beerose Wait I have a question. How was the previous way causing unnecessary migrations? And how did this change fix it?

Because AFAIK, we don't allow changing primary keys from the column editors. So the presence of primary key constraints will never matter while generating the migrations.

Just curious.

wawhal: On the contrary, from what I understand, if we ever plan to start allowing users to add unique constraints to primary key columns (which is disabled now), which is disabled now, it'll start causing unnecessary migrations. This is because:

1. Consider a column that isn't unique
2. You change the `default` value of that column
3. While generating migrations, it'll think that the column was originally unique and it'll create migrations for removing the unique key constraint.

Please correct me if I'm wrong.

beerose: Let's say you are modifying column `id` that is a primary key. In a column editor, it is checked as unique, because primary keys are in fact unique. 
In migrations code we were checking if `originalColumn` (`id`) has a unique constraint. The answer is no because we only checked `unique_constraint`.
But what was passed to the actions was column with property `unique` set to true.
Thus we were generating migration for `id` because `uniqueness` changed.

wawhal: Yeah so the ideal fix is changing:
```
isUnique:
        (columnPKConstraints[colName] && pkLength === 1) ||
        columnUniqueConstraints[colName]
          ? true
          : false,
```

to 

```
isUnique:
        columnUniqueConstraints[colName]
          ? true
          : false,
```

Don't ever have to involve primary keys in unique key stuff ;-)

beerose: If the primary key column for some reason also had `unique` constraint and we would allow modifying unique property, then the logic would be like this:
* original column was unique 
* edited column is not unique (user selected `false` for unique in a scenario when it's possible) 
* we do generate drop unique constraint mutation, which is okay because user wanted to get rid of unique constraint. But now the whole logic for determining uniqueness for `select` component doesn't make sense. So it should be exactly what you suggested above (don't involve PK). 
On the other hand, if you have PK columns and see `unique: false` it may be slightly confusing.

wawhal: > we do generate drop unique constraint mutation, which is okay because user wanted to get rid of unique constraint

There would be no unique constraint to drop. It'll result in an error. I mean, we are fixing one fallacy by adding another fallacy. So I don't quite agree.

But this is only if we allow changing unique key constraints for primary key columns. Which we won't unless somebody asks about it someday. So it's okay I guess.

beerose: > There would be no unique constraint to drop.

**If the primary key column for some reason also had unique constraint**

**we do generate drop unique constraint mutation, which is okay because user wanted to get rid of unique constraint**

So there would be a constraint to drop. 

If we ever enable changing unique constraint, then we would need to get rid of PK logic in determining if the value in select should be true or false. So:
* no unique constraint: `unique: false`, user can add a constraint, we generate migration.
* unique constraint: `unique: true`, user can remove a constraint: we generate drop migration.

wawhal: I am talking of the latter scenario.

wawhal: Can I make a commit to this which will solve both problems?

beerose: You want to get rid of PK logic in both places?

wawhal: I'll open a PR to your branch. You can review.

wawhal: https://github.com/beerose/graphql-engine/pull/6

You can remove the `do-not-merge` label whenever you feel like

beerose: Makes sense. Merging

PR #4412, opened 2020-04-15
skip server tests in PRs if there are no server changes 

rikinsk:
### Changelog

### Affected components

- [x] Build System

### Solution and Design

- Check if `CIRCLE_PR_NUMBER` exists to check if build is for a PR 
- Use `CIRCLE_COMPARE_URL` to check if `server` folder has any changes

ref: https://discuss.circleci.com/t/does-circleci-2-0-work-with-monorepos/10378/14

Screenshot from 2020-04-15 15-50-38


.circleci/config.yml
```diff
@@ -10,6 +10,23 @@ refs:
...
+        if [ -z "$CIRCLE_PR_NUMBER" ]; then
+          echo "Not a PR build, continue test" && exit 0
+        fi
+        echo "Compare refs: ${CIRCLE_COMPARE_URL##http*/}" 
+        for folder in `git log --format="" --name-only ${CIRCLE_COMPARE_URL##http*/} | cut -d"/" -f1 | sort -u`; do
+          echo "contents in $folder were modified"
+          if [ "$folder" = "server" ]; then
+            echo "Server changes present, continue test" && exit 0
+          fi
+        done
```
lexi-lambda, line 27: This seems rather complicated. I think this would do the same thing in a much more direct way:

```sh
if git diff --quiet "${CIRCLE_COMPARE_URL##http*/}" -- server; then
  echo "No server changes present, skipping test"
  circleci-agent step halt
else
  echo "Server changes present, continuing test"
fi
```

More generally: this `CIRCLE_COMPARE_URL` environment variable is mysterious! Looking at the job output, it appears to be empty according to the “Preparing Environment Variables” step, but clearly it is not. Searching the internet for information proves somewhat unhelpful, but it seems to work, so I guess that’s fine.

rikinsk: Yes. Even the CircleCI docs don't seem to mention much about it other than its existence. But like you said it seems to work.

Will update the PR with your suggestions.

rikinsk: Done

PR #4416, opened 2020-04-15
cli: validate endpoint using ParseRequestURI (fix #4407)

### LINKED ISSUE #4407: no check to validate if endpoint is a valid URL
It is possible to create a project directory using `hasura init` with a non URL value for key `endpoint`.
We don't have a check in place to validate that the given endpoint is a valid URL or not. https://github.com/hasura/graphql-engine/blob/master/cli/commands/init.go#L185

ShraddhaAg:
### Description

This commit adds validation of endpoint URL in `hasura init` command and
replaces validation function to `ParseRequestURI` in [`cli.go`](https://github.com/hasura/graphql-engine/blob/master/cli/cli.go#L163).
### Changelog

### Affected components

- [X] CLI

### Related Issues

#4407

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -160,7 +160,7 @@ func (s *ServerConfig) GetVersionEndpoint() string {
 
 // ParseEndpoint ensures the endpoint is valid.
 func (s *ServerConfig) ParseEndpoint() error {
-	nurl, err := url.Parse(s.Endpoint)
+	nurl, err := url.ParseRequestURI(s.Endpoint)
 	if err != nil {
 		return err
 	}
--- a/cli/commands/init.go
+++ b/cli/commands/init.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -183,6 +184,9 @@ func (o *InitOptions) createFiles() error {
 		}
 	}
 	if o.Endpoint != "" {
+		if _, err := url.ParseRequestURI(o.Endpoint); err != nil {
+			return errors.Wrap(err, "error validating endpoint URL")
+		}
 		config.ServerConfig.Endpoint = o.Endpoint
 	}
 	if o.AdminSecret != "" {


cli/cli.go
```diff
@@ -137,37 +137,53 @@ func (c ConfigVersion) IsValid() bool {
...
+// Type returns a string that uniquely represents this flag's type.
+func (ef *EndpointFlag) Type() string {
+	return "string"
+}
+
+// EndpointURL returns Parsed URL type of endpoint
+func (ef EndpointFlag) EndpointURL() *url.URL {
+	nurl, _ := url.ParseRequestURI(ef.String())
+	return nurl
+}
```
ShraddhaAg, line 166: Parsed endpoint URL is needed in a couple of places, [here](https://github.com/hasura/graphql-engine/blob/master/cli/cli.go#L156) and [here](https://github.com/hasura/graphql-engine/blob/master/cli/migrate/util.go#L134) thus I added this function. Since any errors while parsing endpoint will be caught in `Set()` I didn't catch errors here but I'm not sure if this is the best way to do it.

ShraddhaAg:
Hey @shahidhk, I updated the PR with your suggestions. Please let me know if something seems amiss :smile:

cli/cli.go
shahidhk: Got it. In that case, we can define the endpoint flag as of type `url.URL` instead of `string`. Then, we will only need to have `Endpoint url.URL` [here](https://github.com/hasura/graphql-engine/blob/master/cli/cli.go#L141-L152) instead of having both `Endpoint` and `ParsedEndpoint`.

Wherever we need it as a string, we can use [`String()`](https://pkg.go.dev/net/url?tab=doc#URL.String) method of `url.URL`.

ShraddhaAg: I tried implementing that as well, but I wasn't sure [here](https://github.com/hasura/graphql-engine/blob/master/cli/cli.go#L530) how to convert type string to type `EndpointFlag`.

shahidhk: I think `viper` should take care of validating the flag if we use `v.Get` and then execute a type assertion over the interface value that is returned. Something like `(Endpoint)(v.Get("endpoint))`

PR #4422, opened 2020-04-15
Remove some dead code, using weeder

jberryman:
Mostly was just curious to try out this tool. I include an initial weeder config as well with some TODOs for when we move to ghc 8.10 and might want to use this more.

Tried not to not remove code that seemed like utilities that we might want to use later, or which just make sense to have.

I did a manual grep for these on Pro master, so hopeful no breakage there.


0x777:
Hey @jberryman, I'll go through the changes and let you know. Some of these can definitely be removed.

PR #4423, opened 2020-04-15
console: surround varchar, character (bpchar), char and name type value with single quotes (close #4371)

### LINKED ISSUE #4371: Default value does not accept HTTP URL (Postgres error)
Hello,

I tried to add a default value in Hasura as an URL but Postgres threw an error:

image

I guess that the URL should be between quote in this query because Postgres does not like the charactere `:`. :)

I do not have the knowledge of your project to know where to edit this and if it is the good way to fix that. 

Thank you for your awesome work ! :)

ShahAnuj2610:
### Description
- support values in SQL query for `varchar` type without explicitly writing single quotes

### Changelog

### Affected components
- [x] Console

### Related Issues
#4371 

### Solution and Design
- make **varchar** behave same as **text** type

### Steps to test and verify
- create column with type **varchar** and default value as `image URL`.

#### Breaking changes

- [x] No Breaking changes


ShahAnuj2610:
> Hi! Thank for your PR!
> 
> I have two things:
> 
> 1. I suggest `isColTypeString`(or something similar) instead of `isColTypeQuote`.
> 2. We should also handle `character` (`bpchar`),  `char`,  `name`, as they should be quoted as well.

@beerose I have done the changes and added support for other string types.

PR #4424, opened 2020-04-15
Downgrade tests only on 'git tag --merged' tags

jberryman:
It's not clear this is actually okay as some tags point to branches not e.g. not in master, afaict. see slack discussion


jberryman:
@shahidhk did you have thoughts on this? In slack I mentioned being a little confused about how versions are tagged (it looks like there are tagged commits not on `master` branch for instance, and so this change wouldn't properly test those)

shahidhk:
@jberryman According to our current release process, all stable releases are tagged in the `stable` bench and all beta releases are tagged in the `beta` branch. These tags (or rather the commit to which the tag is pointing to) will never make it back to master.

So, I'm afraid this won't pick up any of the recent releases.

jberryman:
@shahidhk gotcha, hmm. So if we alter the test to only work on tags on: `master` + `stable` + `beta` does that sound right? Are users expected to be able to downgrade from some `stable` version to a previous `beta` (I should check this)?

I'm having trouble wrapping my head around what we ought to be testing for here

PR #4433, opened 2020-04-16
console miscellaneous fixes

### LINKED ISSUE #3976: Adding Column presets also adds an empty key/value to migration file
When adding Column presets in permissions, you automatically add a new preset row which is nice in the console.
But as saving and check migration files, there is an empty key-value also added.
```
      localPresets:
      - key: user_id
        value: x-hasura-User-Id
      - key: ""
        value: ""
```

Not a huge deal as it does not do anything to the system.

rikinsk:
### Description

- format row count in data browser for readablity
- move pre-release notification tooltip msg to top
- remove extra localPresets key from migrations (close #3976)
- make nullable and unique labels for columns clickable in insert and modify
- fix row delete for relationships in data browser

### Changelog

### Affected components

- [x] Console


console/src/components/Services/Data/TableCommon/TableReducer.js
```diff
@@ -480,77 +478,45 @@ const modifyReducer = (tableName, schemas, modifyStateOrig, action) => {
...
-              { ...setKeyVal },
-              ...modifyState.permissionsState[
-                action.data.queryType
-              ].localPresets.slice(
-                updatedIndex + 1,
-                modifyState.permissionsState[action.data.queryType].localPresets
-                  .length
-              ),
-            ],
+            set: {
```
beerose, line 513:
```suggestion
            set: updatedSet,
```

console/src/components/Services/Data/TableCommon/TableHeader.js
```diff
@@ -19,6 +19,7 @@ import {
   getTablePermissionsRoute,
   getTableRelationshipsRoute,
 } from '../../../Common/utils/routesUtils';
+import {getReadableNumber} from "../../../Common/utils/jsUtils";
```
beerose, line 22: I see some inconsistent formatting in some places, for example, here. But not like badly formatted, but rather not formatted at all. In another place, there was `set:{` — no space between `:` and `{`.

 Is it possible that you disabled git hook?

rikinsk: Nope. My hook is active. This is odd.

PR #4435, opened 2020-04-16
server: avoid integer overflows (fix #576)

### LINKED ISSUE #576: fix parser to accept integer values of any length and throw error for relevant types
The GraphQL API does not throw an error if a number  > 32 bit is inserted into a Int32 field, instead, it inserts the overflown value.

But JSON API throws error saying this is not possible.

### LINKED ISSUE #4368: BigInts not inserting properly
I've encountered a strange error on hasura v1.1.0. I'm storing time information as bigints and found that the initial insert does not create the right number in the database.

For example:
15866102209964 becomes 493018540
1586610220996 becomes 1767288772

When updating, the bigints update as normal.

I don't know where these numbers come from and am wondering if anyone has encountered this before and how it can be solved.

An example query:

```
mutation {
  insert_domain_category(objects: {description: "New Category", deleted: false, color: "#9B5456", name: "Ah", update_time: 15866102209964, update_op: "*vbnhaunefuxt", id: "ah", create_time: 1586610220996, create_op: "*vbnhaunefuxt"}) {
    returning {
      id
      name
      color
      description
      create_op
      update_op
      create_time
      update_time
      deleted
    }
  }
}

```

produces:

```
{
  "data": {
    "insert_domain_category": {
      "returning": [
        {
          "id": "ah",
          "name": "Ah",
          "color": "#9B5456",
          "description": "New Category",
          "create_op": "*vbnhaunefuxt",
          "update_op": "*vbnhaunefuxt",
          "create_time": 1767288772,
          "update_time": 493018540,
          "deleted": false
        }
      ]
    }
  }
}
```

abooij:
Support inserting unquoted bigint, and generate an integer overflow error on insert (fix #576) (fix #4368)

### Description

Before, when a bigint was inserted unquoted, internally it was cast into an Int32, potentially causing integer overflows. Using the new `Scientific` and `Integer` parsing in hasura/graphql-parser-hs#27, we can now avoid this: the new translation from JSON to PG values triggers appropriate errors. So we can insert integers both quoted `"187461"` and unquoted `187461`.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Solution and Design

The implementation has two components:
- Upgrade graphql-parser-hs, which simplifies code in the input value validator.
- Add bounded JSON integer parsing, which triggers an error on integer/float overflow.

### Steps to test and verify

Unit tests are included.

Although this will allow you to tell when an `_inc`'d _value_ overflows, we cannot distinguish at query time whether this will overflow the stored value. More precisely: even if numbers `x` and `y` individually fit in an `int32`, `x+y` might not. This code does not check for this type of overflow as the addition is handled by postgresql.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - server: add support for `_inc` on `real`, `double`, `numeric` and `money` (fix #3573)
 - server: support special characters in JSON path query argument with bracket `[]` notation, e.g `obj['Hello World!']` (#3890) (#4482)
 - server: add graphql-engine support for timestamps without timezones (fix #1217)
-- server: support inserting unquoted bigint, and generate an integer overflow error on insert (fix #576) (fix #4368)
+- server: support inserting unquoted bigint, and throw an error if value overflows the bounds of the integer type (fix #576) (fix #4368)
 - console: change react ace editor theme to eclipse (close #4437)
 - console: fix columns reordering for relationship tables in data browser (#4483)
 - docs: add API docs for using environment variables as webhook urls in event triggers
@@ -51,7 +51,7 @@ The order, collapsed state of columns and page size is now persisted across page
 - server: `type` field is not required if `jwk_url` is provided in JWT config
 - server: add a new field `claims_namespace_path` which accepts a JSON Path for looking up hasura claim in the JWT token (#4349)
 - server: support reusing Postgres scalars in custom types (close #4125)
-  
+
 ## `v1.2.0-beta.3`
 
 ### console: manage Postgres check constraints
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ The order, collapsed state of columns and page size is now persisted across page
 - console: add undefined check to fix error (close #4444) (#4445)
 - console: change react ace editor theme to eclipse (close #4437)
 - docs: add One-Click Render deployment guide (close #3683) (#4209)
+- server: support inserting unquoted bigint, and generate an integer overflow error on insert (fix #576) (fix #4368)
 - server: reserved keywords in column references break parser (fix #3597) #3927
 - server: fix postgres specific error message that exposed database type on invalid query parameters (#4294)
 - server: manage inflight events when HGE instance is gracefully shutdown (close #3548)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - server: add support for `_inc` on `real`, `double`, `numeric` and `money` (fix #3573)
 - server: support special characters in JSON path query argument with bracket `[]` notation, e.g `obj['Hello World!']` (#3890) (#4482)
+- server: support inserting unquoted bigint, and generate an integer overflow error on insert (fix #576) (fix #4368)
 
 ## `v1.2.0-beta.4`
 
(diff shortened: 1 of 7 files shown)


abooij:
This currently seems to fail on some git tags testing issue, which I don't know how to resolve because I don't know what it intends to check.

server/src-lib/Hasura/SQL/Value.hs
```diff
@@ -127,14 +127,35 @@ parsePGValue ty val = case (ty, val) of
...
+    parseBoundedInt :: forall i. (Integral i, Bounded i) => Value -> AT.Parser i
+    parseBoundedInt val' = do
+      parsedM <-
+        withScientific
+          ("Integer expected for input type: " ++ show ty)
+          (pure . toBoundedInteger)
+          val'
+      case parsedM of
+        Just parsed -> return parsed
+        Nothing     -> fail $ "Integer does not fit.  Maybe it is a float,"
```
0x777, line 139: It would be more helpful to have a message which includes the provided value and the type, maybe something like this?
```
the value '....' exceeds the bounds of type 'int'
```

0x777: or `lies outside the bounds`

server/src-lib/Hasura/SQL/Value.hs
```diff
@@ -127,14 +127,35 @@ parsePGValue ty val = case (ty, val) of
...
+                       ++ " or is there integer overflow?"
+    parseBoundedFloat :: forall a. (RealFloat a) => Value -> AT.Parser a
+    parseBoundedFloat val' = do
+      parsedM <-
+        withScientific
+          ("Float expected for input type: " ++ show ty)
+          (pure . toBoundedRealFloat)
+          val'
+      case parsedM of
+        Left _       -> fail $ "Float does not fit; is there an overflow?"
```
0x777, line 149: Same here, [improved message](https://github.com/hasura/graphql-engine/pull/4435/files#r411331383).

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - server: add support for `_inc` on `real`, `double`, `numeric` and `money` (fix #3573)
 - server: support special characters in JSON path query argument with bracket `[]` notation, e.g `obj['Hello World!']` (#3890) (#4482)
 - server: add graphql-engine support for timestamps without timezones (fix #1217)
+- server: support inserting unquoted bigint, and generate an integer overflow error on insert (fix #576) (fix #4368)
```
tirumaraiselvan, line 10: @abooij Do you want to say something like `...., and generate an error if value overflows the bounds of the integer type`

abooij: Yes, that is what I intended to express. I'll rephrase it following your suggestion :)

PR #4449, opened 2020-04-16
require async globals before render; track runtime errors

wawhal:
### Description

The core purpose of this PR is to track event invoked runtime errors. Some architecture changes have been made and some parts have been refactored.

#### Flow

**Before**: The required variables (server config, server uuid etc) were fetched from the server asynchronously after the first mount of the root component through multiple requests and were stored in redux. This tied the the API logic very strongly to redux i.e. it was neither possible to make authenticated API requests to the server, nor to access the servers config/identity variables from components not connected to Redux. 

**After**: The required async variables are fetched through a single request before the first render. These variables are then set in a global context and passed to the entire application so that all components can directly access these variables.

#### Telemetry Changes

**Before**: Emitting a telemetry event was possible only by dispatching a redux action.

**After**: The telemetry logic has been extracted out of the redux analytics logger and now we can track anything and everything **along with** the redux actions. This PR also starts tracking the runtime errors invoked by the `onClick` handler of the common `Button` component. Similar can be easily implemented for all the interactive components.

#### Some refactor

1. The `uploadFile` function from `jsUtils` imported `showErrorNotification` function from `Services/Comon/Notification.js`. This caused a dependency loop. So this PR breaks this loop by disconnecting the `uploadFile` function from `showErrorNotification`.

2. All the `telemetry1 logic has been moved out of `src/client.js` and moved to the `telemetry` directory.

### Changelog

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


console/src/Globals.js
```diff
@@ -1,3 +1,4 @@
+console.log('Globals');
```
wawhal, line 1:
```suggestion
```

console/src/components/App/Actions.js
```diff
@@ -46,6 +48,17 @@ const showNotification = ({
   };
 };
 
+export const requireAsyncGlobals = ({ dispatch }) => {
+  return (nextState, finalState, callback) => {
+    Promise.all([
+      dispatch(loadConsoleOpts()),
+      dispatch(fetchServerConfig()),
+    ]).finally(() => {
```
beerose, line 56: What about just:
```suggestion
    ]).finally(callback)
```
?

console/src/components/App/App.js
```diff
@@ -1,80 +1,66 @@
...
       dispatch(telemetryNotificationShown());
       dispatch(showTelemetryNotification());
     }
+  }, [telemetry]);
+
+  let connectionFailMsg = null;
+  if (connectionFailed) {
+    connectionFailMsg = (
+      <div
+        style={{ marginBottom: '0px' }}
```
beerose, line 50: maybe a tiny little class for it? 😛

console/src/components/Common/Button/Button.tsx
```diff
@@ -37,8 +40,29 @@ const Button: React.FC<ButtonProps> = props => {
       extendedClassName += 'btn-default';
       break;
   }
+
+  const globals = React.useContext(GlobalContext);
```
beerose, line 44: What about something like `globalsState`?

console/src/telemetry/filters.ts
```diff
@@ -10,14 +10,14 @@ const filterEventsBlockList = [
   'RNS_REMOVE_ALL_NOTIFICATIONS',
 ];
 
-const filterPayloadAllowList = [];
+const filterPayloadAllowList: any[] = [];
```
beerose, line 13: It seems this variable is not used anywhere.

console/src/telemetry/index.ts
```diff
@@ -0,0 +1,108 @@
...
+      server_version: serverVersion,
+      event_type: actionType,
+      url,
+      console_mode: globals.consoleMode,
+      cli_uuid: globals.cliUUID,
+      server_uuid: getState().telemetry.hasura_uuid,
+      data: null,
+    };
+
+    const isLocationType = actionType === '@@router/LOCATION_CHANGE';
```
beerose, line 77: Move to constant?

console/src/telemetry/index.ts
```diff
@@ -0,0 +1,108 @@
...
+import globals from '../Globals';
+import { filterEventsBlockList, sanitiseUrl } from './filters';
+import { RUN_TIME_ERROR } from '../components/Main/Actions';
+
+type ReduxAction = {
+  type: string;
+  payload: {
+    pathname: string;
+  };
+  data: any;
```
beerose, line 11: Can we maybe provide a more specific type here or it can be literally anything?

console/src/telemetry/index.ts
```diff
@@ -0,0 +1,108 @@
...
+      return null;
+    }
+  }
+  return null;
+};
+
+let client = createClient();
+if (client) {
+  const onClose = () => {
+    client = createClient();
```
beerose, line 40: Why client is created again?

console/src/telemetry/index.ts
wawhal: It has to be `any` untill we have a type that unites all the redux actions of the app.

console/src/components/Common/Button/Button.tsx
wawhal: It's referred to as globals everywhere. I think `globals` is okay.

console/src/telemetry/index.ts
wawhal: It is for reopening WebSocket connection if it dies out.

console/src/telemetry/index.ts
```diff
@@ -0,0 +1,109 @@
...
+
+const sendEvent = (payload: any) => {
+  if (client && isTelemetryConnectionReady()) {
+    client.send(
+      JSON.stringify({ data: payload, topic: globals.telemetryTopic })
+    );
+  }
+};
+
+export const trackReduxAction = (action: ReduxAction, getState: any) => {
```
beerose, line 61: Maybe we can say at least that this `getState` is a function that returns an object?
`getState: () => Record<string, any>`

console/src/telemetry/index.ts
```diff
@@ -0,0 +1,109 @@
...
+  const onClose = () => {
+    client = createClient();
+    if (client) {
+      client.onclose = onClose;
+    }
+  };
+  client.onclose = onClose;
+}
+
+const isTelemetryConnectionReady = () => {
```
beerose, line 49: Can't it be a variable? There are no arguments anyway.

wawhal: No. The variable will be a constant and never changed. We want to get the connection ready state at the instance where this is being checked.

PR #4408, opened 2020-04-15
docs: add reference docs for cli (fix #4327)

ShraddhaAg:
### Description

This commit adds docs for CLI configuration file format and ENV
vars.
### Changelog

- [X] `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [X] Docs

### Related Issues

#4327

PR #4454, opened 2020-04-17
cli: add suport for .env file (fix #4129)

### LINKED ISSUE #4129: cli: read env vars from .env file
The scope of this issue has changed. See the comments below.
---
Hi there!

I don't want to load `endpoint` and `admin-secret` from `config.yaml`. I want to load them from `.env` so I created a `npm script` like this:

`"console": "env-cmd hasura --project app console --endpoint $HASURA_GRAPHQL_ENDPOINT",`

However it doesn't work without an empty `config.yaml`.

Even this command doesn't work.

`hasura console --project app --endpoint https://my-graphql-engine.com --admin-secret adminsecretkey`

So think we should fix this bug for not using `config.yaml` cases.

Thank you!

ShraddhaAg:
### Description

This commit adds support for loading ENV vars using .env file in the project root directory. It also introduces a global flag, `--envfile` to facilitate loading of named .env files. (fix #4129) 

### Changelog

- [X] `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [X] CLI
- [ ] Docs

Note: I'll update the docs required for the same here: #4408 

### Related Issues

#4129 

### Solution and Design

Used `godotenv.Load` function to load ENV vars from `.env` file at the project root. The `Load` func is called as soon at the value for `ExecutionDirectory` is set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,19 @@
 
 ## Next release
 
+### cli: add support for .env file
+
+ENV vars can now be read from .env file present at the project root directory. A global flag, `--envfile`, is added so you can explicitly provide the .env filename, which defaults to `.env` filename if no flag is provided.
+
+**Example**:
+
+```
+hasura console --envfile production.env
+```
+The above command will read ENV vars from `production.env` file present at the project root directory.
+
+(close #4129) (#4454)
+
 ### Bug fixes and improvements
 
 - server: add support for `_inc` on `real`, `double`, `numeric` and `money` (fix #3573)
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/viper"
+	"github.com/subosito/gotenv"
 	"golang.org/x/crypto/ssh/terminal"
 	"gopkg.in/yaml.v2"
 )
@@ -206,6 +207,8 @@ type ExecutionContext struct {
 
 	// ExecutionDirectory is the directory in which command is being executed.
 	ExecutionDirectory string
+	// Envfile is the .env file to load ENV vars from
+	Envfile string
 	// MigrationDir is the name of directory where migrations are stored.
 	MigrationDir string
 	// MetadataDir is the name of directory where metadata files are stored.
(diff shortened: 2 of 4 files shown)


cli/cli.go
```diff
@@ -400,6 +401,10 @@ func (ec *ExecutionContext) Validate() error {
 		return errors.Wrap(err, "validating current directory failed")
 	}
 
+	// load .env file
+	envfile := filepath.Join(ec.ExecutionDirectory, ".env")
+	_ = godotenv.Load(envfile)
```
arvi3411301, line 406: @ShraddhaAg can we capture the error and log it in debug level if the file is found?

cli/cli.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/hasura/graphql-engine/cli/telemetry"
 	"github.com/hasura/graphql-engine/cli/util"
 	"github.com/hasura/graphql-engine/cli/version"
+	"github.com/joho/godotenv"
```
arvi3411301, line 30: @ShraddhaAg @shahidhk  Instead of using godotenv, can we use https://github.com/subosito/gotenv (we already have this dependency due to viper).

ShraddhaAg: Replaced the dependency as suggested.

cli/cli.go
ShraddhaAg: If user defined name of .env file is given, error is returned else logged if the file is found.

ShraddhaAg:
I've also added a global flag `--envfile` to facilitate loading of named `.env` files (eg- `production.env`) as suggested by @arvi3411301

cli/cli.go
```diff
@@ -576,6 +584,24 @@ func (ec *ExecutionContext) Spin(message string) {
 	}
 }
 
+// loadEnvfile loads .env file
+func (ec *ExecutionContext) loadEnvfile() error {
+	ec.Envfile += ".env"
```
shahidhk, line 589: User should be able to give the complete file path instead of just the name and us adding the extension. And `.env` should work by default.

ShraddhaAg:
> Good work @ShraddhaAg smile +1
> 
> It'll be great if we can add this to the CLI reference docs.

Thanks @shahidhk! :relaxed: 
I've added it [here](https://github.com/hasura/graphql-engine/pull/4408).

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ The order, collapsed state of columns and page size is now persisted across page
 
 ### Bug fixes and improvements
 
+- cli: add support for .env file (close #4129) (#4454)
```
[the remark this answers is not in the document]
ShraddhaAg: Added a detailed explanation along with an example. Please let me know if it needs to be changed.

cli/cli.go
ShraddhaAg: Accounted for explicit `.env` filename provided by flag. I've also added an INFO level acknowledgement in case the `.env` file is read.

PR #4458, opened 2020-04-17
wip: add relay modern support

0x777:
We've started work on adding a Relay (Modern) interface to graphql-engine (finally!). We've put together this PR so that you can track progress and provide feedback as we work.

Our main challenges with the connection spec / cursors, that we’ve largely resolved: 
- Pushing down Relay style queries into Postgres to enable Hasura optimisations like JSON aggregations and subscriptions.
- Hasura orderBy clauses are incredibly powerful especially as they work across relationships and can use aggregation values as well. These clauses now need to be encoded into Relay cursors.

Todos:
- [x] Generate Relay compatible schema.
- [x]  Implement query generation logic for the Connection spec.
- [x] Separate Relay fields from `/v1/graphql` to `/v1/relay` endpoint.
- [ ] Implement camelcase and singular/plural inflection
- [x] Implement GUID spec
- [x] Implement node resolver / interface

One of the key open questions as we land this support is:
> How do we add support for the existing querying features such as aggregations in the Relay interface? 

Should we add it as part of the `Connection` object in addition to `pageInfo` and `edges` fields? If so, do Relay clients support such additional fields?


ro-savage:
I believe the Relay spec supports queries without the use of edges/connections/nodes, I believe it is only the cursor spec that requires them.

To use example from the relay docs

```
query RebelsShipsQuery {
  rebels {
    name,
    ships(first: 1) {
      edges {
        node {
          name
        }
      }
    }
  }
}
```

and 

```
query RebelsShipsQuery {
  rebels {
    name,
    ships {
      name
    }
  }
}
```

Are both valid, although one obviously returns all and one turns the first only.

Using the second way, I believe you could keep the way Hasura currently handles aggregations.

```
query RebelsShipsQuery {
  rebels {
    name,
    ships(limit: 1) {
      name
    }
  }
}
```

and

```
query RebelsShipsQuery {
  rebels {
    name,
    ships(where: {id: {_eq: <Global Object Identifier> }}) {
      name
    }
  }
}
```

And this would be valid with Relay and work with Hasura's engine converts the <Global Object Identifier> to the database id in the standard way of using tablename_id.

We have been using relay modern for 12+ months and do not use cursors, edges, connections or nodes.

It would be great to see both ways of querying be supported with the Relay implementation. 

Please let us know when there is a working implementation, and we'll happy set it up and give it a try and provide any feedback.

rikinsk:
The relay api toggle status on console should be persisted to LS

sanderhahn:
Some more exotic queries can generate invalid sql: ```graphql
{
  _order_connection {
    edges {
      node {
        id
        name
        one: orderlines_connection(first: 2, order_by: {name: asc}) {
          edges {
            node {
              id
              name
            }
          }
          pageInfo {
            hasNextPage
            startCursor
          }
        }
        two: orderlines_connection(first: 3, order_by: {name: asc}) {
          edges {
            node {
              id
              name
            }
          }
          pageInfo {
            hasNextPage
            startCursor
          }
        }
      }
    }
  }
}
```

File: console/src/components/Services/ApiExplorer/Actions.js [hunk 1]
```diff
@@ -44,6 +44,18 @@ const CREATE_WEBSOCKET_CLIENT = 'ApiExplorer/CREATE_WEBSOCKET_CLIENT';
 const FOCUS_ROLE_HEADER = 'ApiExplorer/FOCUS_ROLE_HEADER';
 const UNFOCUS_ROLE_HEADER = 'ApiExplorer/UNFOCUS_ROLE_HEADER';
 
+const LOADING = 'ApiExplorer/LOADING';
```
beerose, line 47: As this is the name of the action, could we rename it `SET_LOADING` or `TOGGLE_LOADING` or something like this to follow naming conventions we have in other places?

console/src/components/Services/ApiExplorer/ApiExplorer.scss
```diff
@@ -827,3 +827,8 @@ label {
     }
   }
 }
+
+.graphiqlModeToggle {
+  margin-top: -5px;
```
beerose, line 832: Can we instead decrease margin-top here: https://github.com/hasura/graphql-engine/blob/master/console/src/components/Common/Common.scss#L1264?
So that those two elements don't overlap.

console/src/components/Services/ApiExplorer/ApiRequest/ApiRequest.js
```diff
@@ -1,12 +1,12 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-
 import jwt from 'jsonwebtoken';
 
 import TextAreaWithCopy from '../../../Common/TextAreaWithCopy/TextAreaWithCopy';
 import OverlayTrigger from 'react-bootstrap/lib/OverlayTrigger';
 import Tooltip from 'react-bootstrap/lib/Tooltip';
 import Modal from '../../../Common/Modal/Modal';
+import TooltipCustom from '../../../Common/Tooltip/Tooltip';
```
beerose, line 9: As you are here, what about replacing bootstrap `Tooltip` and `OverlayTrigger` with our Tooltip component? It's used only in two places, it shouldn't be much work.

console/src/components/Services/ApiExplorer/GraphiQLWrapper/GraphiQLWrapper.js
```diff
@@ -232,4 +246,9 @@ GraphiQLWrapper.propTypes = {
   urlParams: PropTypes.object.isRequired,
 };
 
-export default GraphiQLWrapper;
+const mapStateToProps = state => ({
+  mode: state.apiexplorer.mode,
+  loading: state.apiexplorer.loading,
+});
+
+export default connect(mapStateToProps)(GraphiQLWrapper);
```
beerose, line 254: Can we do named default export?

pronevich:
Does this PR support relay persist queries?

zth:
I'm doing some testing of the Relay mode. Do you want issues and feedback posted in this issue, as separate issues, or somewhere else?

Thanks!

0x777:
@zth Thanks for testing it out! Please leave your feedback here on this PR :)

0x777:
> Does this PR support relay persist queries?

@pronevich Not yet.

zth:
Ok, here's a few things:

- `__typename` is missing on the `Node` interface (although maybe this commit solves that? https://github.com/hasura/graphql-engine/pull/4458/commits/5b00b81494a943f4ff8c093751ab10a9eea8c2a5)
- Relationships as arrays should probably be exposed as connections rather than limit/offset based pagination, which they were now as I was testing
- I got database query errors trying to query array relationships (I tried a few in a pretty simple setup). Is this a known issue? If not, I'll try and create a reproducible example

Thank you for your hard work, it's really exciting to test this and see all the progress you've made!

ro-savage:
Thanks for the great work. Can't wait to start using this.

Some feedback after a quick try.

**The 'id' on an item is now a UUID instead of the ID. But any relationship id continues to the be the database id.**

```
person {
 id: 'eyJ0YWJs....',
 role_id: 3
 role {
  id: 'eyJ0YWJs...'
}
```

**Querying on an ID only accepts the database id not the relay UUID**
```
query MyQuery {
  people_connection(where: {id: {_eq: "eyJ0YWJsZSIgOiB7InNjaGVtYSIgOiAicHVibGljIiwgIm5hbWUiIDogInBlb3BsZSJ9LCAiY29s\ndW1ucyIgOiB7ImlkIiA6IDF9fQ=="}}) {
    edges {
      node {
        id
        first_name
      }
    }
  }
}
```

```
{
  "errors": [
    {
      "extensions": {
        "path": "$",
        "code": "data-exception"
      },
      "message": "invalid input syntax for integer: \"eyJ0YWJsZSIgOiB7InNjaGVtYSIgOiAicHVibGljIiwgIm5hbWUiIDogInBlb3BsZSJ9LCAiY29s\ndW1ucyIgOiB7ImlkIiA6IDF9fQ==\""
    }
  ]
}
```

I wasn't able to expand much on that in the real world without getting the correct ids.

However, the aggregates etc seemed to be working properly, and our database structure appeared to come through fine.

It would be useful if you could get both the UUIDs and the IDs. 

In our case, in React we always use the UUIDs but when we link to standalone CRUD pages we use regular ids. For example `/user/1`

zth:
Exposing the underlying IDs would be really nice! Graphile does it by having a `rowId` that's just the db id in whatever format that is for the current table. Maybe that's worth being inspired by, or calling it `dbId` or similar.

rakeshkky:
@zth 
> Relationships as arrays should probably be exposed as connections rather than limit/offset based pagination, which they were now as I was testing

In relay API, we are exposing `<array_relation>_connection` in a table object along side with simple and aggregate array relationships. 

> I got database query errors trying to query array relationships (I tried a few in a pretty simple setup). Is this a known issue? If not, I'll try and create a reproducible example

I know a issue which gives database query error if connections and normal fields of array relationships are queried in the same table object selection set. It is fixed. Please try the latest image and see if the issue you're saying still persists.

Thanks a lot for trying and giving us your valuable feedback. :smiling_face_with_three_hearts:

zth:
@rakeshkky excellent, thank you! When I tested before I didn't get the relationship as a connection, but it seems to be working now. Sweet! 👍

console/src/components/Services/ApiExplorer/ApiExplorer.scss
wawhal: I don't understand. Can you elaborate?

beerose: Sure. A few points on negative margins:
* Layout is not clean — elements overlap 
Screenshot 2020-06-04 at 08 50 21

* When composing a layout we should stick to make it as simple as possible 

* Negative margins may lead to inconsistent results in different browsers (might not be the case for us, tho)

Sometimes negative margins are needed and there's no other way around, but in this case, we can compose the elements by reducing the margin of one of them. That's why I put this comment. 
This is not something super bad but the thing is they may be harmful, so it's good to avoid them when possible.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 ## Next release
 
+### Relay
+
+The Hasura GraphQL Engine serves [Relay](https://relay.dev/en/) schema for Postgres tables which has a primary key defined.
```
tirumaraiselvan, line 7: small grammar: ... Postgres table which **have** a primary key...

ro-savage:
@rakeshkky - Will you be looking at the issue of primary key ID vs UUID as mentioned in https://github.com/hasura/graphql-engine/pull/4458#issuecomment-637968056 ?

Or is this behaviour deliberate?

rakeshkky:
@ro-savage By the relay spec, each object in the relay schema should implement the [Node Interface](https://relay.dev/graphql/objectidentification.htm#sec-Node-Interface) for which the `id` field is reserved with scalar type `ID!`. Unfortunately, the `id` as a column name is most likely in databases. I'm not sure about the relay clients support if the `id` field overridden with some other name. I'll be very happy if you provide some details about it. 

You can always make use root `node` field resolver to fetch the object of any type with the UUID as input argument. However, we still kept the `id` in `where` clause to reference the database column (if present) to perform filters as in regular Hasura queries.

ro-savage:
@rakeshkky - 
Thanks for the information about the where clause. 

To me it would make since to have a `where id` being the UUID, and `where db_id` for the database id. And then include the `db_id` along with the uuid `id` on every object. Similar to postgraphile and graphql-ruby. However, it can be worked around and it's actually not something we use in your current relay queries.

More importantly, if Hasura is providing the `id` field as an UUID (correct) why are all the relationships ids still using the database ids. How can the user match up these relationships on the client?

For example

```
{ account: {
  people: [{ role_id: 1 }]
  roles: [{ id: 'eyJ0YWJs...' }]
}
```

`role_id: 1` should be be the UUID of the role `role_id: 'eyJ0YWJs...'` so that on the front end the data can be matched up and it can be requested by using the root node resolver.

nirvdrum:
Is there another term we can use for the Relay ID? I found this discussion a bit confusing to follow because the Relay IDs shown are not [UUIDs](https://en.wikipedia.org/wiki/Universally_unique_identifier), as far as I can tell. Moreover, PostgreSQL has a [uuid type](https://www.postgresql.org/docs/12/datatype-uuid.html) and using that for a PK is supported by the Hasura console independent of Relay support.

soneymathew:
`id` in base64 decodes to `{"id" : 123}`
can it include the table information to make it globally unique?
similar to how cursor decodes to 
```
{"table" : {"schema" : "public", "name" : "mytable"}, "columns" : {"id" : 123}}
```

soneymathew:
I go this error while using first and last together

```
"message": "\"first\" and \"last\" are not allowed at once"
```

I think its possible/valid to support them if we want to.

For a bounded set of 100 items, if I specify first 20 and last 10
As per the algorithm I get items 11 to 20

image

It's essential to support `first` & `last` together for building ui experiences that uses page numbers to paginate.

for page size 10
Page 1 can be reached with first=10 last=10
Page 2 can be reached with first=20 last=10
.
.
.
.
Page 10 can be reached with first:100 last=10

rakeshkky:
@ro-savage 
You can always make use of Hasura's [customizing column names](https://hasura.io/docs/1.0/graphql/manual/schema/custom-field-names.html#expose-columns-with-a-different-name-in-the-graphql-api) feature to rename a column name in GraphQL API. In that way you can rename a column with `id` name to something like `db_id`. Both the `id` and `db_id` are exposed as fields for a table object and `db_id` is used in `where` clause to filter rows. Hope this helps you solving the `id` name conflict.

rakeshkky:
@nirvdrum It is actually Globally unique id (GUID) for [object identification](https://graphql.org/learn/global-object-identification/). I mistook UUID for GUID in the comment. The Node `id` is a opaque value which is globally unique and used to retrieve the object which implements the Node interface. Thank you. :smiley:

rakeshkky:
> When I create a fresh table with an id field, it properly decodes to a globally unique id.
> 
> ```
> {"table" : {"schema" : "public", "name" : "freshtable"}, "columns" : {"id" : 1}}
> ```
> 
> My existing database already have an id field on `oldTable` and it decoded to `{"id" : 1}`
> I dunno if it's a bug as I already have metadata on this table using an older version of hasura engine.
> 
> Tested that cursor on `oldTable` decodes correctly to
> 
> ```
> {"table" : {"schema" : "public", "name" : "oldTable"}, "columns" : {"id" : 123}}
> ```

@soneymathew I couldn't understand the actual issue here. Is the encoded `id` value in the later not correct?

rakeshkky:
@soneymathew -- Regarding `first` and `last` arguments:

In the relay [spec](https://relay.dev/graphql/connections.htm#sec-Pagination-algorithm) for pagination it is noted that

```
Including a value for both first and last is strongly discouraged, as it is likely to lead to confusing queries and results.
```

AFAIK, the input arguments for a field in a GraphQL query is order agnostic. So, I think, it is not possible to determine whether it is "last of first" or "first of last" if both arguments are specified.

soneymathew:
> @soneymathew I couldn't understand the actual issue here. Is the encoded `id` value in the latter not correct?

@rakeshkky encoded `id` value is correct, but the encoded id was not globally unique, perhaps I missed a migration step? note that I was running an older version of hasura in the same database.

soneymathew:
@rakeshkky The spec gives priority to `first` So it is always interpreted as `last` of `first` edges between `after` and `before` 😅  

https://relay.dev/graphql/connections.htm#sel-FAJLJCDDAAHFDDFBGB_yN

rakeshkky:
@soneymathew 

> > @soneymathew I couldn't understand the actual issue here. Is the encoded `id` value in the latter not correct?
> 
> @rakeshkky encoded `id` value is correct, but the encoded id was not globally unique, perhaps I missed a migration step? note that I was running an older version of hasura in the same database.

The global unique id is [base64[(https://en.wikipedia.org/wiki/Base64) encoding of the JSON object which contains the details of the table name itself and primary key column values. The `id` is a primary key column whose value is encoded. 

Regarding `first` and `last` argument:

I blindly followed the note and made a decision not to allow both arguments. We can support the behavior you mentioned in the use case. Soon the next beta release (`1.3.0-beta.2`) will support it.

ro-savage:
@rakeshkky - I am must not be making myself clear.

The issue is that the relationship ids are different to the item ids.

In the database the ids look like
**Person** `{id: 1, role_id: 4}`
**Role** `{id: 4}`

However Hasura is returning is returned the base64 encoded table+id for the id, but not for the relationship id.

From current hasura query result
**Person**: `{id: 'cGVvcGxlLTE=', role_id: 4}`
**Role**: `{id: 'cm9sZXMtNA=='}`

Now there is no way to match role_id: 4 to the role, because the ids are different.

Hasura should be returning role_id as the base64 encoded table+id instead like this:
**Person**: `{id: 'cGVvcGxlLTE=', role_id: 'cm9sZXMtNA=='}`
**Role**: `{id: 'cm9sZXMtNA=='}`

This is how all other relay + sql implementations I have used work.

pronevich:
@rakeshkky currently "normal" ids are used in URL like https://my.app/company/h3ol3ne. Here is short id "h3ol3ne" which exact row id in table "company". Now in Relay API mode ids not the same and match longer. Is there a  right way to use previous ids?

rakeshkky:
@ro-savage Let me make it clear.

In a given table, the `id` column is replaced with `Node` interface's GUID. The GUID is [base64](https://en.wikipedia.org/wiki/Base64) encoding of the JSON object which contains the details of the table name itself and primary key column values (the table may have more than one primary key columns). 

In the use case you mentioned:
**Person**: `{id: 'cGVvcGxlLTE=', role_id: 'cm9sZXMtNA=='}`
**Role**: `{id: 'cm9sZXMtNA=='}`

I think there's no point in having `Role` table information in `Person` table's column (`role_id`). Let's say if we are querying just the data from `Person`, having opaque base64 encoded value for `role_id` doesn't make it clear. Also, it is always assumed that `role_id` should match with `Role`'s `id`. 

To avoid conflicts, you can customize the column name. If relay clients work with custom name for `Node` interface GUID field (other than `id`), let us know.

Please open an issue in github, specifying your concerns on the relay as the PR got merged to master. :smiley: 

> This is how all other relay + sql implementations I have used work.

Please give us the details of the other implementations.

rakeshkky:
> @rakeshkky currently "normal" ids are used in URL like https://my.app/company/h3ol3ne. Here is short id "h3ol3ne" which exact row id in table "company". Now in Relay API mode ids not the same and match longer. Is there a right way to use previous ids?

You can make use of [customize column name](https://hasura.io/docs/1.0/graphql/manual/schema/custom-field-names.html#expose-columns-with-a-different-name-in-the-graphql-api) feature to rename the `id` column field.

jmsegrev:
What happen with this task "Implement camelcase and singular/plural inflection"?. Even though it might not be explicitly in the graphql spec the camel case, is definitely a convention. In 1.3 beta for relay the connections cant be aliased like the table fields, also there is a mix of camel case and snake case, which makes it inconsistent. 

For my particular case, I'm considering hasura to replace an existing graphql api that is consumed by a relay client. The camel case is a big problem, since we would have to modify all the existing graphql queries and I want to do this with the least possible changes to the existing client code.

jmsegrev:
In another note, I think the generated global IDs are too big. Just consider the request payloads and response data size increase just because of this small detail. Also consider that this IDs could be used in URLs (any client/app). 

The base64 encoded data has too much information for an identification string `{"table" : {"schema" : "public", "name" : "test"}, "columns" : {"id" : 1}}`. Wouldn't some text like `schema:table:id`, result in a smaller encoded string, have enough information and would be easy/faster to parse?, or am I missing something?

ro-savage:
@rakeshkky - I have made an issue here #5035 #5036 #5037

@jmsegrev - I've also added an issue for your suggestion

tsnobip:
Huge thanks for adding relay support! I'm testing it right now, it works pretty well but I find something weird, `node`s are typed as nullable, but I can't find any example where they would indeed be null, but it's quite a pain to deal with with nullable in the front end. Is it something that could be changed?

zth:
@tsnobip Just speculation from my side, but this is sometimes an effect of how permissions are implemented. Ie "fetch all rows, check permission for each node". I dunno if that's the case here though, but just to provide some context on when `node` might be nullable in general.

rakeshkky:
@tsnobip @zth 

The rows are filtered by applying permissions, if any. So, each item should be non-nullable. I'll fix this in https://github.com/hasura/graphql-engine/pull/5013.
Thank you.

PedroBern:
@rakeshkky please how the id is generated?

> The global unique id is base64 encoding of the JSON object which contains the details of the table name itself and primary key column values. The id is a primary key column whose value is encoded.

Can you show the JSON object format, please? I would like to generate the id, but I need the JSON fields, for example:

```json
{
    "table_name": "my_table",
    "pk": "4ddd3e6c-7528-41d4-a052-84e57931b199"
}
```

What names are used in the `table_name` and `pk`? Is there any other information I need to know to generate the ids?

I need to generate the ids while #5658 is not fixed because I'm using a custom function in the regular endpoint, that should contain the relay id of the record.

tirumaraiselvan:
@PedroBern This answer might help: https://stackoverflow.com/questions/62685944/hasura-relay-how-to-work-with-base64-encoded-ids

PedroBern:
@tirumaraiselvan thanks for the response, also I saw the [Note Relay Node Id](https://github.com/hasura/graphql-engine/blob/7e970177c19169ffff4e06fe37fe9d0aafe04ff8/server/src-lib/Hasura/GraphQL/Schema/Select.hs#L1219), but still I didn't manage to build the ids, I guess I will be finding some records by pk instead of using the node id for now.

Issue #4011, opened 2020-03-03
Always provide session variables to queries


m4dc4p:
Currently, Hasura will provide session variables to mutations and "custom functions" created with 'track_function V2'.

I would really like Hasura to *always* pass session variables to postgres, whether running queries, mutations or custom functions.

My database is accessed by multiple clients, one of which is Hasura. My other clients use Postgres "configuration settings" (via `set_config`, `current_setting`) to provide authorization information used during queries. I'd like to leverage the same infrastructure when executing queries generated by Hasura.

A similar request was made in #3872.

tirumaraiselvan:
@0x777 Thoughts?

Can we provide a directive for reads which will set session variables in current_setting??

tirumaraiselvan:
Similar request here: https://github.com/hasura/graphql-engine/issues/3846

MichaelHadji-duplicate:
I've been commenting in https://github.com/hasura/graphql-engine/issues/1582 though this is probably the place to be.

MichaelHadji-duplicate:
Any updates relating to this or the related comments in #1582 ?

0x777:
As mentioned in https://github.com/hasura/graphql-engine/issues/1582#issuecomment-665737794, passing session variables using Postgres's GUC (`current_setting('hasura.user')`) would defeat our optimisations for subscriptions.

@m4dc4p From what I understand of your use-case based on the conversation in [#3872](https://github.com/hasura/graphql-engine/issues/3872#issuecomment-592103783) you seem to want to define permissions in one place and have all your clients use them. If you define the permissions in graphql-engine, your REST API server should be able to use graphql-engine as its ORM and use the permissions that are already defined. The idea is that you can mimic a user request from your REST server to graphql-engine by setting `x-hasura-role` and the accompanying session variables as headers (`x-hasura-admin-secret` is necessary for such privileged requests).

0x777:
@m4dc4p I'm closing this issue but if what I mentioned in my above comment doesn't apply for your use-case, please feel free to re-open this issue.

m4dc4p:
Well, that doesn't really work for me as I don't want to use Hasura as an ORM or rely on its permission system. I can use custom functions to mimic what I need, so it's not a big deal, but it is inconvenient. But thank you for taking a look.

MichaelHadji-duplicate:
Doesn't work for me either (similar to m4dc4p), please re-open this!

MichaelHadji-duplicate:
[...]

Issue #3772, opened 2020-01-24
Allow filtering by computed fields


mikyoengine:
If you have a computed field such as `full_name`, it would be very useful to be able to filter by this value. E.g.

```graphql
{
  users(where: { full_name: { _eq: "John Smith"}) {
  ...UserDetails
  }
}
```

MatthiasGrandl:
Yes exactly the same usecase for me. I output a datatable module and I want the user to be able to filter and sort by any value inside the datatable. Right now I have to disable those features for any computed field, which is really unintuitive.

barmaley443-sfdev:
+1
now receiving "field "computed" not found in type: 'table_order_by'" when trying to sort by computed field.
any chance to get this feature?

esseswann:
Interestingly on surface there are no visible limitations because this is working
```
SELECT public.calculate_something(entity) AS status FROM entity
WHERE public.calculate_something(entity) = 'something'
```

Kosai106:
Just to add in on this, it could be useful to use a computed field as primary key
Example: I generate a `slug` for my `projects` and I want to get them with a query like:

```gql
query {
  projects_by_pk(slug: "example") {
    id
    name
  }
}
```

Or

```gql
query {
  projects(where: {slug: { _eq: "example"}}) {
    id
    name
  }
}
```

MatthiasGrandl:
Also related, it would be neat if we could select computed columns for permission checks.

maxcan:
@marionschleifer @coco98 any chance of getting this into 1.3?

mkalayci35:
+1

MarcelloTheArcane:
> Just to add in on this, it could be useful to use a computed field as primary key
> Example: I generate a `slug` for my `projects` and I want to get them with a query like:
> 
> ```graphql
> query {
>   projects_by_pk(slug: "example") {
>     id
>     name
>   }
> }
> ```
> 
> Or
> 
> ```graphql
> query {
>   projects(where: {slug: { _eq: "example"}}) {
>     id
>     name
>   }
> }
> ```

@Kosai106  Does this actually work?

Kosai106:
@MarcelloTheArcane No, it's a request.

TheMrZZ:
It looks fairly simple to implement, even though I might be mistaken.

Currently, my workaround is to create a SQL View, with all my table's field, and the computed field itself.

[...]

PR #4486, opened 2020-04-20
server: allow computed fields to have access to Hasura's session variables (fix #3846)

### LINKED ISSUE #3846: accessing session variables in function via computed field
It would be really cool to be able to define a function like

```
CREATE OR REPLACE FUNCTION article_liked(article_row article, hasura_session json)
RETURNS boolean AS $$
  SELECT EXISTS (
    SELECT 1
    FROM liked_article A
    WHERE A.user_id = hasura_session ->> 'x-hasura-user-id' AND A.article_id = article_row.id
  );
$$ LANGUAGE sql STABLE;
```

and use it via a computed field without any args.

I know that in order to get `hasura_session` injected the function needs to be tracked, but tracked functions only support `SETOF` return types.

I can dive in a try to do a PR if there's interest in this kind of functionality.

abooij:
### Description

By introducing a `session_argument` setting, similar in style to the one for custom SQL functions, and to the `table_argument` setting, we enable SQL functions that underlie computed fields to have access to Hasura's session variable. This allows, for instance, to have a computed field containing all the articles that the viewing user liked: so we do not need to specify the user explicitly, but can rather get the current user ID from the session variable directly. (fix #3846)

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

This PR has the following elements:
- Amendment of the datatypes `ComputedFieldDefinition`, `ComputedFieldValidateError`, `ArgumentExp` and `ComputedFieldFunction`.
- Corresponding additional executing code.
- Explanation of the new feature in the changelog, docs, and API docs.
- New integration test and modified (see comment below) existing integration tests.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [ ] No
- [x] Yes
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


abooij:
Note that because of the change in `server/src-lib/Hasura/RQL/DDL/ComputedField.hs`, empty `table_argument`s are no longer printed in the metadata, so that this changes test results.

This is no longer printed to be consistent with `session_argument`s, which are also not printed when empty. Any consistent choice (either print both when empty or print neither when empty) would be incompatible with the test suite from a previous version. The only option that would pass the old test suite would be to print `table_argument` always, and `session_argument` only when non-empty. But this seemed confusingly inconsistent to me.

abooij:
This possibly also solves #4011, although the scope of that request is not so clear.

server/tests-py/queries/v1/metadata/setup.yaml
```diff
@@ -104,7 +104,7 @@ args:
 - type: run_sql
   args:
     sql: |
-      CREATE FUNCTION fetch_articles(search text, author_row author)
+      CREATE FUNCTION fetch_articles(search text, author_row author, session json)
```
0x777, line 107: Any reason why this test case has to be modified? We can let it be so that it tests the `session_argument` not being specified?

abooij: Ah, oops, it's because I first considered adding the test to `fetch_articles`, and then I changed my mind. I'll remove it again.

abooij: You're right, I'll separate it.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Next release
 
+### Allow access to session variables by computed fields (fix #3846)
+
+Sometimes it is useful for computed fields to have access to the Hasura session variables directly.  Support for this is now added through the `add_computed_field` API.
+
+Read more about actions in the [docs](https://hasura.io/docs/1.0/graphql/manual/api-reference/schema-metadata-api/computed-field.html).
```
tirumaraiselvan, line 9: it says `Read more about actions`?

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Next release
 
+### Allow access to session variables by computed fields (fix #3846)
+
+Sometimes it is useful for computed fields to have access to the Hasura session variables directly.  Support for this is now added through the `add_computed_field` API.
```
tirumaraiselvan, line 7: Can we give an example? One such example is available in the original issue.

CHANGELOG.md
abooij: Haha, oops, this is because I copy-pasted the syntax for a link :laughing:  Will fix!

CHANGELOG.md
abooij: Okay, I'll add a quick example here.

docs/graphql/manual/schema/computed-fields.rst
```diff
@@ -184,6 +184,76 @@ Adding a computed field to a table
      A computed field can be added to a table using the :ref:`add_computed_field <api_computed_field>`
      metadata API
 
+Accessing Hasura session variables in computed fields
+-----------------------------------------------------
+
+It can be useful to have access to the session variable from the SQL function defining a computed field.  For instance, suppose we are recording which users have liked which article, using a table ``article_likes`` specifying a many-to-many relationship between ``article``s and ``user``s.  Then it may be useful to know if the current user liked the current article, and expose this as a *Boolean* computed field on ``article``.
```
marionschleifer, line 190:
```suggestion
It can be useful to have access to the session variable from the SQL function defining a computed field.  For instance, suppose we want to record which users have liked which articles. We can do so using a table ``article_likes`` that specifies a many-to-many relationship between ``article`` and ``user``. In such a case it can be useful to know if the current user has liked a specific article, and this information can be exposed as a *Boolean* computed field on ``article``.
```

marionschleifer: When we talk about the many-to-many relationship, can we link to the respective docs page: https://hasura.io/docs/1.0/graphql/manual/schema/relationships/database-modelling/many-to-many.html?

marionschleifer: Also, I made the suggestion to remove the `s`s right after the table names, since it screws up the formatting: 
image

docs/graphql/manual/schema/computed-fields.rst
```diff
@@ -184,6 +184,76 @@ Adding a computed field to a table
...
+
+Use the :ref:`add_computed_field` API to add a function, and specify the name of the argument taking a session argument in ``session_argument``.  The session argument will be a JSON object where keys are session variable names (in lower case) and values are strings.  Use the ``->>`` JSON operator to fetch the value of a session variable as shown in the following example.
+
+.. code-block:: plpgsql
+
+      -- 'hasura_session' will be session argument
```
marionschleifer, line 196:
```suggestion
      -- 'hasura_session' will be the session argument
```

docs/graphql/manual/schema/computed-fields.rst
```diff
@@ -184,6 +184,76 @@ Adding a computed field to a table
...
+                    "id": "3",
+                    "liked_by_user": true
+                }
+             ]
+        }
+    }
+
+.. note::
+
+   The specified session argument will not be included in the ``<function-name>_args`` input object in the GraphQL schema.
```
marionschleifer, line 254:
```suggestion
   The specified session argument is not included in the ``<function-name>_args`` input object in the GraphQL schema.
```

docs/graphql/manual/schema/computed-fields.rst
```diff
@@ -184,6 +184,76 @@ Adding a computed field to a table
      A computed field can be added to a table using the :ref:`add_computed_field <api_computed_field>`
      metadata API
 
+Accessing Hasura session variables in computed fields
+-----------------------------------------------------
+
+It can be useful to have access to the session variable from the SQL function defining a computed field.  For instance, suppose we are recording which users have liked which article, using a table ``article_likes`` specifying a many-to-many relationship between ``article``s and ``user``s.  Then it may be useful to know if the current user liked the current article, and expose this as a *Boolean* computed field on ``article``.
+
+Use the :ref:`add_computed_field` API to add a function, and specify the name of the argument taking a session argument in ``session_argument``.  The session argument will be a JSON object where keys are session variable names (in lower case) and values are strings.  Use the ``->>`` JSON operator to fetch the value of a session variable as shown in the following example.
```
marionschleifer, line 192:
```suggestion
Use the :ref:`add_computed_field` API to add a function, and specify the name of the argument taking a session argument in ``session_argument``.  The session argument is a JSON object where keys are session variable names (in lower case) and values are strings.  Use the ``->>`` JSON operator to fetch the value of a session variable as shown in the following example.
```

abooij:
> @abooij I would probably move this section after permissions, since I have a feeling that users will need to know about permissions before dealing with session variables.
> 
> Left some small suggestions too slightly_smiling_face

Thanks, Marion! Implemented as suggested.

MatthiasGrandl:
This would be amazing in combination with https://github.com/hasura/graphql-engine/issues/3772 . Would make session based permissions a lot less boilerplatey.

server/src-lib/Hasura/RQL/DDL/ComputedField.hs
```diff
@@ -76,6 +78,11 @@ data InvalidTableArgument
   | ITANotTable !QualifiedTable !FunctionTableArgument
   deriving (Show, Eq)
 
+data InvalidSessionArgument
+  = ISANotFound !FunctionArgName
+  | ISANotJSON !FunctionSessionArgument
+  deriving (Show, Eq)
```
rakeshkky, line 84: Hi, it'd be nice to see the tests which fails with these new validation errors. If added, please ignore this comment and feel free to mark it as resolved.

abooij: Thanks, Rakesh - I have added two tests corresponding to the two errors.

server/src-lib/Hasura/RQL/Types/ComputedField.hs
```diff
@@ -44,6 +44,17 @@ instance ToJSON FunctionTableArgument where
...
+-- | The session argument, which passes Hasura session variables to a
+-- SQL function as a JSON object.
+data FunctionSessionArgument
+  = FunctionSessionArgument !FunctionArgName !Int
+  deriving (Show, Eq, Generic)
+instance Cacheable FunctionSessionArgument
+
+instance ToJSON FunctionSessionArgument where
+  -- TODO is this the right JSON to output?
+  toJSON (FunctionSessionArgument argName _) = object ["name" .= argName]
```
rakeshkky, line 56: I'm sort of missed this `TODO`. So, the `FunctionSessionArgument` carries the session argument name and it's ordinal position in the function input params. It is enough to just output the name directly without any nesting.
```suggestion
instance ToJSON FunctionSessionArgument where
  toJSON (FunctionSessionArgument argName _) = toJSON argName
```

Also, document the type `FunctionSessionArgument` like the following:
```haskell
= FunctionSessionArgument 
     !FunctionArgName -- ^ The argument name
     !Int -- ^ The ordinal position in the function input parameters
```

abooij:
Thanks, @rakeshkky! Suggestions on improving the comments are *more* than welcome. :-)

SidhNor:
Gentlemen, will this go into 1.3.0-beta.1 now?

tirumaraiselvan:
@SidhNor Yes. v1.3.0-beta.1 should be out in a week or so.

SidhNor:
@tirumaraiselvan  thank you for the info

brittanyroddy:
@tirumaraiselvan Hey! Any update on whether this has been released yet?

PR #4492, opened 2020-04-21
add few more rql queries to be access mode agnostic

tirumaraiselvan:
### Description

Few more RQL queries can be run in READ ONLY mode.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### DIFF
--- a/server/src-lib/Hasura/Server/Query.hs
+++ b/server/src-lib/Hasura/Server/Query.hs
@@ -292,7 +292,10 @@ getQueryAccessMode q = (fromMaybe Q.ReadOnly) <$> getQueryAccessMode' q
     getQueryAccessMode' (RQV1 q') =
       case q' of
         RQSelect _ -> pure Nothing
-        RQCount _ -> pure Nothing
+        RQCount _  -> pure Nothing
+        RQGetInconsistentMetadata _     -> pure Nothing
+        RQExportMetadata _              -> pure Nothing
+        RQDumpInternalState _           -> pure Nothing
         RQRunSql RunSQL {rTxAccessMode} -> pure $ Just rTxAccessMode
         RQBulk qs -> foldM reconcileAccessModeWith Nothing (zip [0 :: Integer ..] qs)
         _ -> pure $ Just Q.ReadWrite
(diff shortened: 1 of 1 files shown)


server/src-lib/Hasura/Server/Query.hs
```diff
@@ -292,7 +292,10 @@ getQueryAccessMode q = (fromMaybe Q.ReadOnly) <$> getQueryAccessMode' q
     getQueryAccessMode' (RQV1 q') =
       case q' of
         RQSelect _ -> pure Nothing
-        RQCount _ -> pure Nothing
+        RQCount _  -> pure Nothing
+        RQGetInconsistentMetadata _     -> pure Nothing
+        RQExportMetadata _              -> pure Nothing
+        RQDumpInternalState _           -> pure Nothing
         RQRunSql RunSQL {rTxAccessMode} -> pure $ Just rTxAccessMode
         RQBulk qs -> foldM reconcileAccessModeWith Nothing (zip [0 :: Integer ..] qs)
```
0x777, line 300: Use `indexedFoldM` from `RQL.Types.Error` so that you don't have to worry about constructing the path. So, it'll be
`withPathk "args" $ indexedFoldM ...`

server/src-lib/Hasura/Server/Query.hs
```diff
@@ -304,10 +307,7 @@ getQueryAccessMode q = (fromMaybe Q.ReadOnly) <$> getQueryAccessMode' q
             "incompatible access mode requirements in bulk query, " <>
             "expected access mode: " <>
             (T.pack $ maybe "ANY" show expectedMode) <>
-            " but " <>
-            "$.args[" <>
-            (T.pack $ show i) <>
-            "] forces " <>
+            " but " <> "$.args[" <> (T.pack $ show i) <> "] forces " <>
```
0x777, line 310: We don't have to manually construct the path, it'll be handled automatically with `withPathK` and `indexedFoldM` above, just the error message will be sufficient.

tirumaraiselvan:
Stale

PR #4501, opened 2020-04-21
workflow: add lint and format action

beerose:
Solves https://github.com/hasura/graphql-engine-internal/issues/380

### Description

Right now we can easily workaround lint and prettier and as the result push badly formatted code to master. This PR adds GitHub action that runs linter and prettier.

* Action is only triggered if there are changes in `console/`.  For example, in this PR I changes `package.json` so the action was triggered.
* First step is linting that leaves messages for errors (skips warnings):
Screenshot 2020-04-21 at 20 34 58

- The second step is prettier that checks if code is properly formatted. 

This PR also makes pre-commit git hook optional:
```js
if (process.env.HUSKY_PRE_COMMIT) {
  execSync('node_modules/.bin/lint-staged', { stdio: 'inherit' });
}
```

### Changelog


.github/workflows/lint.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    env:
+      working-directory: ./console
+
+    steps:
+      - uses: actions/checkout@v2
+
```
soorajshankar, line 15: I am not very sure, Will this cache the npm libraries if the lock file has not changed?

What do you think about a cache here? 
https://github.com/actions/cache/blob/master/examples.md#node---npm

https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows
@beerose

beerose: It's okay to go without cache for now.

Issue #4344, opened 2020-04-09
Unexpected exception raised in websocket


haohedev:
![image](https://user-images.githubusercontent.com/5556270/78859559-40287b80-7a62-11ea-8f16-3d5f5f16db67.png)

haohedev:
The Hasura version is v1.1.1

tirumaraiselvan:
@haohedev Thanks for the report.

Is there any repro you can provide?

haohedev:
![image](https://user-images.githubusercontent.com/5556270/78897134-4c352d00-7aa4-11ea-84c3-47c70edd5f5e.png)

Before April 2nd, the version I was using was 1.1.0. It can be seen from the chart that there will be a memory leak, but the container will automatically restart. This is similar to the #4077 problem. After I used version 1.1.1, it seems that the problem is solved. But today a process of containers turning into zombies emerged. Now I added a liveness probe, I hope it works.
The scenario used by Hasura is the subscription data change, which is displayed on the front end. There will be programs that write data directly to the database.
Sorry for my english, i use google translate

PR #4518, opened 2020-04-22
Migrate to GHC 8.10, upgrade dependencies. Closes #4517

### LINKED ISSUE #4517: Migrate to GHC 8.10
The main motivator to do this now is some stubborn space leaks seem to be resolved.

Checklist:

- [x] Get working with some minimal clean forks with PRs
- [x] get feedback from dev team (do IDEs work?) **EDIT**: not much feedback, but did my best!
- [x] upgrade CI (upgrade `haskell-docker-packager`?) , coordinate migration with devs
- [x] shepherd remaining forks through upstream

Things we decided we can do post-merge:

- check out Nizar's benchmarks, quicjk before and after (*)  **EDIT**: in progress
- Upgrade pro (**NOTE**: pro only depends on stable OSS hge releases, and stack supports 8.10 so we anticipate this being simple)

\* There are some other performance-related things we'd like to experiment on, e.g. the new GC

---

## EDIT: A few benchmarks

Performance is pretty comparable with what looks like some modest improvements. Cherry-picking a couple graphs. (Note: ghc 8.10 version here is compiled with `-O1` since unfortunately I'm now getting occasional OOM compiling locally / O2 was taking a very long time for me; I guess we should at least open up a ticket):

s1
s2
s3
s4

jberryman:
This also seems to squash a stubborn space leak we see with
subscriptions (linking to canonical #3388 for reference).

This may also fix some of the "Unexpected exception" websockets
exceptions we are now surfacing (see e.g. #4344)

Also: dev.sh: fix hpc reporting

----

Please install GHC 8.10 and give this a try and let me know if there are issues with your development setup that we can sort out.

You'll need cabal-install 3.2 (which you can upgrade with `cabal`; it will work with ghc 8.6.5 as well, but note the `-new` commands have now become default)

I install GHC from the bindists here: 

https://www.haskell.org/ghc/download_ghc_8_10_1.html

It will clobber the symlinks from the last version you installed, but otherwise multiple ghc's play nice together (you can switch between them with the `with-compiler` config setting in a `cabal.project` or the `-w` flag to cabal).


server/cabal.project
```diff
@@ -33,8 +35,10 @@ source-repository-package
 
 source-repository-package
   type: git
-  location: https://github.com/hasura/graphql-parser-hs.git
-  tag: 623ad78aa46e7ba2ef1aa58134ad6136b0a85071
+  -- TODO for ghc-8.10:
+  -- https://github.com/hasura/graphql-parser-hs/pull/28
+  location: https://github.com/jberryman/graphql-parser-hs.git
```
abooij, line 40: I believe hasura is still the canonical source of this repository, isn't it?

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 
 ### Bug fixes and improvements
 
-(Add entries here in the order of: server, console, cli, docs, others)
+- server: compile with GHC 8.10.1, closing a space leak with subscriptions. (close #4517) (#3388)
```
abooij, line 37: I propose to leave the instructions in place, and not to have a newline between the `server` entry and the `console` entry.

server/cabal.project.dev-sh.local
```diff
@@ -9,6 +9,9 @@
 
 package *
   documentation: true
+  -- build with DWARF support. Still fairly useless, but see: 
+  --   https://www.haskell.org/ghc/blog/20200403-dwarf-1.html
+  debug-info: 2
```
abooij, line 14: If it's fairly useless, shall we leave it out for now? With the intention of keeping the build plan as simple as possible.

server/cabal.project.freeze
```diff
@@ -100,255 +101,266 @@ constraints: any.Cabal ==2.4.0.1,
...
              hashable -examples +integer-gmp +sse2 -sse41,
-             any.hashtables ==1.2.3.1,
-             hashtables -bounds-checking -debug -portable -sse42 +unsafe-tricks,
-             any.haskell-lexer ==1.0.2,
-             any.hasql ==1.3.0.5,
-             any.hasql-pool ==0.5.0.2,
-             any.hasql-transaction ==0.7.1,
-             any.hedgehog ==0.6.1,
+             any.hashtables ==1.2.3.4,
+             hashtables -bounds-checking -debug -detailed-profiling -portable -sse42 +unsafe-tricks,
```
abooij, line 146: Is it necessary to compile this with detailed profiling, even for non-dev applications?

server/cabal.project.freeze
```diff
@@ -100,255 +101,266 @@ constraints: any.Cabal ==2.4.0.1,
...
+             any.lifted-async ==0.10.0.6,
              any.lifted-base ==0.2.3.12,
-             any.list-t ==1.0.3.1,
+             any.list-t ==1.0.4,
              any.loch-th ==0.2.2,
-             any.math-functions ==0.3.1.0,
-             math-functions -system-expm1,
-             any.memory ==0.14.18,
+             any.math-functions ==0.3.3.0,
+             math-functions +system-erf +system-expm1,
```
abooij, line 194: Out of interest, why is are these flags on `math-functions` relevant for us? Do we get a compilation error in another dependency otherwise? Surely we're not actually using these functions.

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -285,7 +286,14 @@ createServerApp (WSServer logger@(L.Logger writeLog) serverStatus) wsHandlers !p
...
                 -- Process all messages serially (important!), in a separate thread:
-                msg <- liftIO $ WS.receiveData conn
+                msg <- liftIO $ 
+                  -- Re-throw "receiveloop: resource vanished (Connection reset by peer)" :
+                  --   https://github.com/yesodweb/wai/blob/master/warp/Network/Wai/Handler/Warp/Recv.hs#L112 
+                  -- as WS exception signaling cleanup below. It's not clear why exactly this gets 
+                  -- raised occasionally; I suspect an equivalent handler is missing from WS itself.
+                  -- Regardless this should be safe:
+                  handleJust (guard . E.isResourceVanishedError) (\()-> throw WS.ConnectionClosed) $
+                    WS.receiveData conn
```
abooij, line 296: Thanks, I've also seen this error. The comment is very helpful.

server/src-lib/Hasura/Server/Telemetry.hs
```diff
@@ -189,11 +189,11 @@ computeActionsMetrics ac ao =
...
         queryActionsLen = length . filter ((==ActionQuery) . _adType . _aiDefinition) $ actions
 
-        outputTypesLen = length . nub . (map (_adOutputType . _aiDefinition)) $ actions
-        inputTypesLen = length . nub . concat . (map ((map _argType) . _adArguments . _aiDefinition)) $ actions
+        outputTypesLen = length . L.nub . (map (_adOutputType . _aiDefinition)) $ actions
+        inputTypesLen = length . L.nub . concat . (map ((map _argType) . _adArguments . _aiDefinition)) $ actions
         customTypesLen = inputTypesLen + outputTypesLen
 
-        typeRelationships = length . nub . concat . map ((getActionTypeRelationshipNames ao) . _aiDefinition) $ actions
+        typeRelationships = length . L.nub . concat . map ((getActionTypeRelationshipNames ao) . _aiDefinition) $ actions
```
abooij, line 196: This section looks like it could use some refactoring as it is very repetitive, but it's good that you didn't do that in this PR.

server/tests-py/queries/v1/update/basic/person_error_no_where_clause.yaml
```diff
@@ -3,7 +3,7 @@ url: /v1/query
 status: 400
 response:
   path: $
-  error: key "where" not present
+  error: key "where" not found
```
abooij, line 6: Out of interest: where do these changed error messages come from? Are they caused by the fact that you now raise these errors through `throw400` via `Monadfail`, rather than constructing the errors explicitly? In any case, to me the new messages look more informative.

.circleci/config.yml
```diff
@@ -1,5 +1,9 @@
 # anchor refs to be used elsewhere
 refs:
+  constants:
+  # TODO upload to hasura docker hub:
+  # - &server_builder_image hasura/graphql-engine-server-builder:2020-01-14
+  - &server_builder_image jberryman/graphql-engine-server-builder-8.10:2020-04-29
```
abooij, line 6: Would it be possible to upload to docker first? Or is there a reason for doing it this way?

server/cabal.project.freeze
jberryman: This file is just the output of `cabal freeze`; all these flags are the baked in defaults that the packaged ship with. Any fiddling we want to do with them in the future will live in `cabal.project` and have docs

server/tests-py/queries/v1/update/basic/person_error_no_where_clause.yaml
jberryman: These changes just come from a newer aeson I believe, glad you think they are an improvement

.circleci/config.yml
jberryman: @0x777 I don't have credentials (I don't think) to push to hasura's docker hub, and forgot to ping someone about this TODO. Do you want to do this before merge? What's the best way to get this into hasura namespace?

server/cabal.project.dev-sh.local
jberryman: Good point. I would like to leave it in, and added a better comment to justify why. Let me know if you strongly object.

.circleci/config.yml
shahidhk: @jberryman I have added your the hasura org on docker hub. I have also re-tagged and pushed the image to hasura.

jberryman:
(@shahidhk thanks for doing the docker image!)

PR #4527, opened 2020-04-23
fix intermittent errors occur when query actions used with variables

### LINKED ISSUE #4520: do not cache plan for action queries
From [Discord](https://discordapp.com/channels/407792526867693568/516590536875048990/701909454811037738):

```
I get intermittent validation errors on the variables:
"path":"$.variableValues","error":"unexpected variables: ticker","code":"validation-failed"

it works on and off so it's really hard to understand what's going on. It's a simple query with just one string variable. It works well without variables

Any ideas?
```

@rakeshkky said

```
These action queries, we are marking them as re-usable without capturing the variable value in the state as re-usable variables. This is because we ain't using the input arguments in action query field for generating the SQL statement, instead we are converting them to JSON and including it in action webhook POST request payload. The fix is, by design, we shouldn't cache action queries.
```

codingkarthik:
fixes #4520 

### Description

A query action when used with variables used to give intermittent errors when executed multiple times. This PR fixes the bug by not adding the query action to the plan cache.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Resolve.hs
+++ b/server/src-lib/Hasura/GraphQL/Resolve.hs
@@ -119,14 +119,19 @@ queryFldToPGAST fld actionExecuter = do
     QCAsyncActionFetch ctx ->
       QRFActionSelect <$> RA.resolveAsyncActionQuery userInfo ctx fld
     QCAction ctx -> do
-      let f = case jsonAggType of
-                DS.JASMultipleRows -> QRFActionExecuteList
-                DS.JASSingleObject -> QRFActionExecuteObject
+      -- query actions should not be marked reusable because we aren't
+      -- capturing the variable value in the state as re-usable variables.
+      -- The variables captured in non-action queries are used to generate
+      -- an SQL query, but in case of query actions it's converted into JSON
+      -- and included in the action's webhook payload.
       markNotReusable
       f <$> actionExecuter (RA.resolveActionQuery fld ctx (userVars userInfo))
       where
         outputType = _saecOutputType ctx
         jsonAggType = RA.mkJsonAggSelect outputType
+        f = case jsonAggType of
+             DS.JASMultipleRows -> QRFActionExecuteList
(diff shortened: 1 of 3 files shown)


server/src-lib/Hasura/GraphQL/Resolve.hs
```diff
@@ -119,13 +119,19 @@ queryFldToPGAST fld actionExecuter = do
...
+      -- an SQL query, but in case of query actions it's converted into JSON
+      -- and included in the action's webhook payload.
+      markNotReusable
+      f <$> actionExecuter (RA.resolveActionQuery fld ctx (userVars userInfo))
       where
         outputType = _saecOutputType ctx
         jsonAggType = RA.mkJsonAggSelect outputType
+        f = case jsonAggType of
+             DS.JASMultipleRows -> QRFActionExecuteList
+             DS.JASSingleObject -> QRFActionExecuteObject
```
rakeshkky, line 134: Move this to function body under `let`. The `f` is ambiguous and it is more readable if it is used as soon as declared.

codingkarthik: I thought it would be better to have all the bindings under `where`. I'll make the change

PR #4538, opened 2020-04-23
Use default values when reusing a query plan (fix #3773)

### LINKED ISSUE #3773: Query sometime fails when ID is marked as required with error "expected a value for non-nullable variable"
I ran into an issue when executing the same query multiple times (5-10 times) in a row. Sometimes it works and sometimes it returns the follow error saying "expected a value for non-nullable variable." There are no other custom permissions for the user on the table. The user has access to all of the columns.

I did notice if I mark the id params as not required it works and returns data without an issue.  Any thoughts on this one? 

I'm running against version 1.0

{
    "type": "http-log",
    "timestamp": "2020-01-24T23:10:10.147+0000",
    "level": "error",
    "detail": {
        "operation": {
            "user_vars": {
                "x-hasura-role": "admin-role",
                "x-hasura-user-id": "b729b676-96b5-488e-ab12-dc045b82a846"
            },
            "error": {
                "path": "$.variableValues.id",
                "error": "expected a value for non-nullable variable",
                "code": "validation-failed"
            },
            "request_id": "70be7e67-fab6-43b5-af00-96fc150d09d4",
            "response_size": 110,
            "query": {
                "operationName": "ListThemeTracksByThemeId",
                "query": "query ListThemeByThemeId($id: uuid! = \"9ae362ab-c094-42f0-b152-9f7914544505\") {\n\tTheme(where: {themeId: {_eq: $id}}) { \n\t\tid \n    themeId\n\t\tobjectPath \n\t\t} \n}"
            }
        },
        "http_info": {
            "status": 200,
            "http_version": "HTTP/1.1",
            "url": "/v1/graphql",
            "method": "POST",
            "content_encoding": null
        }
    }

abooij:
### Description

When reusing a cached query (a.k.a. a query plan or a reusable query), we don't yet make use of the default values of the query. This fixes that.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

The most important change is in `server/src-lib/Hasura/GraphQL/Validate.hs`, where `validateVariablesForReuse` now additionally takes the default values.

While developing this I realized that `subscription`-type queries might suffer from the same issue, and indeed they do. But their codepath is somewhat different. Fixing that is not part of this PR.

I don't know a reasonable way to test the caching mechanism thoroughly. More code sharing between the cached (planned) and fresh query handling would have probably prevented this bug.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


abooij:
As documented in #4577, this PR does not necessarily improve the situation. So I'll think about this in some more detail.

abooij:
Closed in favor of #4620

PR #4545, opened 2020-04-23
console: add trim to colname input values

### LINKED ISSUE #4526: Trailing spaces when adding a new column should be trimmed.
When creating a new table if any one of the columns has an extra space in the column name table creating fails with the following error, but it's difficult for the user to understand  what's wrong as the space is  not available in the UI: 
 
image

To be precise `column_name `  throws an exception since it has a trailing space but `column_name` doesn't. 

I suggest we should apply trim() on all column names before sending it to the Postgres and the user need not have to worry about anything.

If you agree that this should be fixed, I am willing to raise a PR for this. Please assign it on me.

gnyani:
### Description

Added trim() to input change handlers of the colName inputs, in two cases:
1. When creating a new table.
2. Adding a new column to the existing table.

Also, had thoughts about fixing editing the existing colName, but it seems  to be working without issues, So I didn't change it. Let me know if its needed.

### Changelog

### Affected components

- [ ] Server
- [X] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

fix #4526 

#### Catalog upgrade

- [X] No
- [ ] Yes

#### Metadata

- [X] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [X] No Breaking changes


gnyani:
@beerose can we merge this? Please let me know if there are any changes required.

rikinsk:
Yes. It wpuld be better to do the trim before the actual save and not at the time of typing.

gnyani:
> Yes. It would be better to do the trim before the actual save and not at the time of typing.

@rikinsk  thanks for the feedback. I will do the updates soon.

gnyani:
@beerose @rikinsk  I took a second look at this and I feel the current implementation might be better because the error is thrown by the validation here if regex gqlPattern mismatch:
https://github.com/hasura/graphql-engine/blob/1ee3f40ad490665da4cb86f96a2e7050ad1d71c7/console/src/components/Services/Data/Add/AddTable.js#L257

To trim the column name before saving, we need to change the validation to accept white spaces and trim it later before sending a request to DB, it contradicts the whole point of client validation IMO. 

Please let me know what do you guys suggest.

beerose:
One thing that comes to my mind is doing `trim` **only after** used clicks `Create table / Add column` and **before** the validation step.

In the case of creating the table, it could be done in `validateAndSubmit` function in `AddTable.js`.
For adding a new column it could be done in `onSubmit` function in `ColumnCreator.js`.

gnyani:
@beerose  thank you for the review, can you please take a look once again. I have addressed your comments.

PR #4551, opened 2020-04-24
server: multiple subscriptions && `_st_d_within` in subscriptions (fix #3239)

### LINKED ISSUE #3239: "postgres query error" when using subscriptions and "_st_d_within" expression to compare columns of type geography

abooij:
### Description

PR #3344 fixes #3239, which was an issue caused by an oversight in how we generate SQL for geographic input. The oversight was caused by the lack of testing of subscriptions. This PR is a rebase of #3344, so that #3239 is fixed, and additionally implements the suggested flag indicating that we are exploiting an internal feature (see below). 

### Changelog

### Affected components

- [x] Server
- [x] Tests

### Solution and Design

As in #3344, the actual fix is minimal (see `server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Plan.hs`). The remainder of the PR consists of allowing easy testing of subscriptions by adding a `subscription` "transport" type to the test suite that changes a `query` GraphQL query to a `subscription`. Many `query` type tests use multiple root fields, which are not allowed by the GraphQL standard. So, rather than splitting them up, the easiest way to re-use these tests is to simply support subscriptions with multiple root fields, thus departing from the GraphQL standard. To avoid users shooting themselves in the foot, this feature is only enabled if a `_multiple_subscriptions` flag is passed: so the test suite replaces `query` with `subscriptions` and adds `@_multiple_subscriptions` to the request. This internal flag is picked up by `validateGQ` in `server/src-lib/Hasura/GraphQL/Validate.hs`.

Example query:
```graphql
subscription mySub($arg : Int! = 2) @_multiple_subscriptions {
  article (where: {id: {_eq: $arg}}) {
    title
  }
  author {
    name
  }
}
```
### Steps to test and verify

The above implementation allowed an existing `query` test (see `TestGraphQLQueryBoolExpPostGIS` in `server/tests-py/test_graphql_queries.py`) to be re-used as a `subscription` test, so that we now have a regression test for this.

Users may shoot themselves in the foot if they somehow use the directive `@_multiple_subscriptions` for a different purpose. However, they normally shouldn't have access to directives, and our usage of them is very limited.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/GraphQL/Validate.hs
```diff
@@ -172,12 +171,16 @@ validateGQ (QueryParts opDef opRoot fragDefsL varValsM) = do
...
-          unless (null rst) $
-            throwVE "subscription must select only one top level field"
-          return $ RSubscription fld
+      case selSet of
+        Seq.Empty       -> throw500 "empty selset for subscription"
+        (_ Seq.:<| rst) -> do
+          -- As an internal testing feature, we support subscribing to
+          -- multiple selection sets.  First check if the
+          -- corresponding directive is set.
+          let multipleAllowed = elem (G.Directive "_multiple_subscriptions" []) (G._todDirectives opDef)
```
0x777, line 180: I would probably call it, `_multiple_top_level_fields`, your call though, I'm not particular about this. Formatting of the comment is as per the style guide (wrap it at 80 characters)?

abooij: Formatting of the comment is as per Emacs' `M-q` :smile:.

I'll fix the formatting and rename the directive to `_multiple_top_level_fields`.

PR #4566, opened 2020-04-26
cli: allow initialising project in current directory (fix #4560)

### LINKED ISSUE #4560: cli: allow initialising project in current directory
When we execute `hasura init .` so that we need the current directory to be used as a Hasura Project, CLI throws an error: `FATA[0002] directory '.' already exist`.

We should allow using the current directory and create `config.yaml`, metadata, migrations directories etc. in the same directory itself.

ShraddhaAg:
### Description

This PR allows users to initialize a Hasura project in pwd.
### Changelog

- [X] `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [X] CLI
- [X] Docs

### Related Issues

#4560 

### Solution and Design

I used `os.Getwd()` to get the pwd in order to set `ExecutionDirectory`.


cli/commands/init.go
```diff
@@ -90,50 +90,47 @@ type InitOptions struct {
...
+			p := promptui.Prompt{
+				Label:   "Name of project directory ",
+				Default: defaultDirectory,
+			}
+			r, err := p.Run()
+			if err != nil {
+				return handlePromptError(err)
+			}
+			if strings.TrimSpace(r) != "" {
+				o.InitDir = r
```
arvi3411301, line 115: @ShraddhaAg Can we also handle if the user enters `.` in the promptui

ShraddhaAg: Ah right. Missed it. Done!

cli/commands/init.go
```diff
@@ -102,38 +105,43 @@ func (o *InitOptions) Run() error {
...
 		} else {
-			dir = defaultDirectory
+			o.InitDir = defaultDirectory
 		}
-	} else {
-		dir = o.InitDir
 	}
-
-	if o.EC.ExecutionDirectory == "" {
+	if o.InitDir == "." {
```
scriptonist, line 113: @ShraddhaAg a nitpick, 
Instead of directly comparing to `.` do you think something similar to the following is better?
```
	cwdir, err := os.Getwd()
	if err != nil {
		return errors.Wrap(err, "error getting current working directory")
	}
	initPath, err := filepath.Abs(o.InitDir)
	if err != nil {
		return err
	}
	if initPath == cwdir {
             // do stuff
        }
```
This can also handle paths like `./`

ShraddhaAg: Makes sense. Updated accordingly :smile:

PR #4581, opened 2020-04-27
check if migrate is required in relaxed tx mode

tirumaraiselvan:
### Description

This is the final step in relaxing tx isolation levels.

1. The main change is to check whether a catalog migrate is required by comparing the catalog version in the db and the latest catalog version. This is done in RepeatableRead tx isolation level.

2. Another auxiliary change is to relax the tx mode in SchemaUpdate module  

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


server/src-lib/Hasura/Server/Migrate.hs
```diff
@@ -17,7 +17,7 @@
 module Hasura.Server.Migrate
```
tirumaraiselvan, line 17: There is definitely scope for more refactor here, especially in the case when migrate is required: we end up re-doing many queries like checking for version, schema, extensions, etc.

server/src-lib/Hasura/Server/Migrate.hs
```diff
@@ -84,6 +84,46 @@ data MigrationPair m = MigrationPair
...
+      WHERE schemaname = $1 AND tablename = $2
+    ) |] (schemaName, tableName) False
+
+isExtensionAvailable :: (MonadTx m) => SchemaName -> m Bool
+isExtensionAvailable schemaName =
+  liftTx $ (runIdentity . Q.getRow) <$> Q.withQE defaultTxErrorHandler [Q.sql|
+    SELECT EXISTS
+    ( SELECT 1 FROM pg_catalog.pg_available_extensions
+      WHERE name = $1
+    ) |] (Identity schemaName) False
```
lexi-lambda, line 110: Did you have any particular reason for floating these out of `migrateCatalog`? It doesn’t seem like it was necessary to me, and having the definitions more local seems better than pulling them out.

server/src-lib/Hasura/Server/Migrate.hs
```diff
@@ -84,6 +84,46 @@ data MigrationPair m = MigrationPair
...
+     )
+  => m Bool
+isCatalogMigrateRequired = do
+  doesSchemaExist (SchemaName "hdb_catalog") >>= \case
+    False -> pure True
+    True  -> doesTableExist (SchemaName "hdb_catalog") (TableName "hdb_version") >>= \case
+      False -> pure True
+      True  -> do
+        dbCatalogVersion <- liftTx getCatalogVersion
+        pure $ dbCatalogVersion /= latestCatalogVersionString
```
lexi-lambda, line 125: Instead of having a predicate that this module exports to guard calls to `migrateCatalog`, I think it would be nicer to move the logic into `migrateCatalog` itself. Then we don’t have to pollute the call site with extra logic that really only this module should have any reason to care about.

PR #4111, opened 2020-03-13
Rewrite GraphQL schema generation and query parsing (close #2801)

lexi-lambda:
**This PR is a work in progress. This description will be updated to fit the usual format once it is closer to done.**

## Overview

This pull request is a complete rearchitecting of the way we build GraphQL schemas and parse GraphQL queries. It implements some of the ideas in #2801, but the approach is rather different from the one described there (as that approach turned out to be too naïve in various ways).

The core API is built around a new type, `Parser`, which has the following interface:

```haskell
data Parser (k :: Kind) (m :: * -> *) (a :: *)

parserType :: Parser k m a -> Type k
runParser :: Parser k m a -> ParserInput k -> m a

instance Functor (Parser k m)
```

This may raise more question than it answers. How do you do anything with a `Parser` if it’s only a `Functor`? What is `ParserInput`? And what on *earth* is a `Kind`?

### Our goal

Before we answer those questions, let’s start with a recap of the Big Idea:

  * Currently, schema generation and query parsing are separate code paths, so they can get out of sync. We want a new approach that can do both at once.

  * We want to [parse, not validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), so there shouldn’t be a separate GraphQL validation pass. Checking consistency against the schema should happen naturally as part of query parsing.

  * We need to be able to reflect on the GraphQL schema without even necessarily having a query to execute, since we need to serve GraphQL introspection queries.

The `Parser` abstraction is designed to solve all of the above problems. It works like this:

  * A `Parser` is something that knows how to parse a piece of a GraphQL AST. (Note that it doesn’t have anything to do with parsing GraphQL query concrete syntax, that step is still handled by `graphql-parser-hs`.)

  * You build `Parser`s using parser combinators, just like other Haskell parsing libraries. Since a `Parser` parses a GraphQL AST, the primitive combinators in this library are things like `scalar`, `enum`, and `object`.

  * However, unlike a traditional parser combinator library, every `Parser` is associated with a GraphQL type. For example, if you have a `Parser` that parses an input object, you can reflectively get information about what fields it contains and what their types are.

### Challenges

This idea is simple on the surface, but subtle to implement. Broadly speaking, there are three significant sources of complexity:

  1. GraphQL ASTs are much more complicated than something like JSON. Field sets, fragments, nullability, and variables all add rules that need to be captured in the type system.

[...]

PR #4585, opened 2020-04-28
server: validation code simplifications

abooij:
### Description

Rather than defining a new monad class and constructing new instances
of it, just use the `StateT` that it is really based on. I am proposing this change for several reasons: it is shorter, we reduce the amount of code that can be incorrect, and most importantly, it is easier to understand. At first I found it difficult to understand `MonadReusability`. But its description as a state monad explains everything to me. However, this description is only present in its implementation.

This PR also gets rid of some dead code, and `Monoid` instances that aren't
actually monoids (such as the ones for `QueryReusability` and `UnionTyInfo`).

Also avoid an instance of prisms where this doesn't add much value. This has the added benefit of the code being more welcoming to programmers who are perhaps not so familiar with lenses/prisms.

### Affected components

- [x] Server

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Validate/Types.hs
+++ b/server/src-lib/Hasura/GraphQL/Validate/Types.hs
@@ -71,10 +71,10 @@ module Hasura.GraphQL.Validate.Types
   , ReusableVariableValues
 
   , QueryReusability(..)
-  , _Reusable
-  , _NotReusable
-  , MonadReusability(..)
+  , MonadReusability
   , ReusabilityT
+  , markNotReusable
+  , recordVariableUse
   , runReusabilityT
   , runReusabilityTWith
   , evalReusabilityT
@@ -94,8 +94,6 @@ import qualified Data.Text                     as T
 import qualified Language.GraphQL.Draft.Syntax as G
 import qualified Language.GraphQL.Draft.TH     as G
 import qualified Language.Haskell.TH.Syntax    as TH
-
-import           Control.Lens                  (makePrisms)
 
 import qualified Hasura.RQL.Types.Column       as RQL
 
@@ -247,9 +245,6 @@ instance EquatableGType ObjTyInfo where
     (G.NamedType, Set.HashSet G.NamedType,  Map.HashMap G.Name (G.Name, G.GType, ParamMap))
   getEqProps a = (,,) (_otiName a) (_otiImplIFaces a) (Map.map getEqProps (_otiFields a))
 
-instance Monoid ObjTyInfo where
-  mempty = ObjTyInfo Nothing (G.NamedType "") Set.empty Map.empty
-
 instance Semigroup ObjTyInfo where
   objA <> objB =
     objA { _otiFields = Map.union (_otiFields objA) (_otiFields objB)
@@ -303,14 +298,6 @@ instance EquatableGType IFaceTyInfo where
     (G.NamedType, Map.HashMap G.Name (G.Name, G.GType, ParamMap))
   getEqProps a = (,) (_ifName a) (Map.map getEqProps (_ifFields a))
 
-instance Monoid IFaceTyInfo where
-  mempty = IFaceTyInfo Nothing (G.NamedType "") Map.empty
-
-instance Semigroup IFaceTyInfo where
-  objA <> objB =
-    objA { _ifFields = Map.union (_ifFields objA) (_ifFields objB)
-         }
-
 fromIFaceDef :: G.InterfaceTypeDefinition -> TypeLoc -> IFaceTyInfo
 fromIFaceDef (G.InterfaceTypeDefinition descM n _ flds) loc =
   mkIFaceTyInfo descM (G.NamedType n) fldMap loc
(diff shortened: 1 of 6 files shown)


abooij:
In fact, arguably we should define this in terms of writers rather than state: we should never need to know the current list of validated variables.

abooij:
I now see that #4111 already contains similar simplification work on `MonadReusability`, where it is renamed to `MonadParse`, and combined with some elements of `MonadError`. So this PR can be ignored in lieu of that one.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
```diff
@@ -810,36 +789,28 @@ data QueryReusability
...
 
-class (Monad m) => MonadReusability m where
-  recordVariableUse :: G.Variable -> RQL.PGColumnType -> m ()
-  markNotReusable :: m ()
+type MonadReusability m = MonadState QueryReusability m
 
-instance (MonadReusability m) => MonadReusability (ReaderT r m) where
-  recordVariableUse a b = lift $ recordVariableUse a b
-  markNotReusable = lift markNotReusable
+type ReusabilityT m a = StateT QueryReusability m a
```
lexi-lambda, line 795: I normally recommend against doing this. Why? Two reasons:

1. You can’t use `MonadState` for something else when this transformer is somewhere in the stack.

2. It’s a weaker abstraction. Code that is polymorphic in the monad but written against `MonadReusability` can only call `recordVariableUse` and `markNotReusable`, which accumulate information monoidally. But using `MonadState`, they can do arbitrary other things, including dropping the state altogether.

Perhaps the above two points seem like small potatoes, and maybe they are. But I don’t think they’re inconsequential enough to justify switching away from the abstraction just to be a little bit simpler!

One improvement that *would* be nice would be to switch from `StateT` to `WriterT` as the underlying transformer, since `ReusabilityT` is fundamentally about accumulating information monoidally, and it doesn’t need the full power of `StateT`. Alas, even the strict variant of `WriterT` leaks space, so I avoid using it for anything that isn’t tiny.

Fortunately, recent versions of `transformers` include `Control.Monad.Trans.Writer.CPS`, which is a non-leaky implementation of `WriterT` (it uses `StateT` under the hood), but `mtl` doesn’t include support for that in the currently-released version. It *does* support it on master (because [I added support for it](https://github.com/haskell/mtl/commit/ee511c5b0edd83477552a1f951f9b5b6733c4423)—probably at around the same time I wrote `ReusabilityT`, in fact!), but it doesn’t seem to have made its way into a release yet. I should probably bother the maintainers about that.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
```diff
@@ -810,36 +789,28 @@ data QueryReusability
   = Reusable !ReusableVariableTypes
   | NotReusable
   deriving (Show, Eq)
-$(makePrisms ''QueryReusability)
-
-instance Semigroup QueryReusability where
-  Reusable a <> Reusable b = Reusable (a <> b)
-  _          <> _          = NotReusable
-instance Monoid QueryReusability where
-  mempty = Reusable mempty
```
lexi-lambda, line 819: You say this “isn’t actually a monoid,” but that’s just not true; this *is* a monoid. It satisfies all the monoid laws. It wouldn’t be a monoid if we defined `mempty = NotReusable`, but we don’t. As-written, it’s a perfectly cromulent monoid.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
abooij: Thanks for this background, Alexis. Of course I completely agree with the points you raise. I am not as well-versed in the space leakage of the various monad transformers, so what you wrote is very informative.

The reason I opened this PR nonetheless is that I think in this case the abstraction of `MonadReusability` has a cost. There is an argument to be made against abstraction when this is in conflict with interpretability, as it was for me. But I'm sure we can resolve this conflict in the refactored version by making sure there is good documentation, and by minimizing dependency on this stateful computation. (At first sight this seems to be much better in the refactored version, which is why above I proposed to ignore this PR in lieu of that work.)

server/src-lib/Hasura/GraphQL/Validate/Types.hs
abooij: Yes, having taken a second look, you are right. So it makes sense to treat `QueryReusability` as a monoid.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
lexi-lambda: FWIW I agree with you 100% that the “mtl-style” approach to effect composition is way, way too complicated, and we want something better. That is in fact precisely why I have written [eff](https://github.com/hasura/eff): I think monad transformers are the single largest obstacle to new Haskell programmers who want to get productive with the language, and I think algebraic effects are a better design.

Unfortunately, existing approaches to algebraic effects in Haskell aren’t very efficient, so eff is also a project in making them fast. And eff *is* fast… but it is fast because it depends on some modifications to GHC. I have [an open GHC proposal](https://github.com/lexi-lambda/ghc-proposals/blob/delimited-continuation-primops/proposals/0000-delimited-continuation-primops.md) to get those modifications merged, but until then we’re stuck with what we have.

So yes, less tangentially, I’m happy to include some more documentation wherever it would help. Please feel free to add some extra documentation wherever it would be helpful! (Or, for that matter, request for someone else to add some documentation if you don’t feel like you understand it enough to do it yourself.)

abooij:
Closing this for now, awaiting the results of #4111.

PR #4604, opened 2020-04-29
tests: tear down numerics values (fix #4602)

### LINKED ISSUE #4602: server: `_inc` integration test fails on CI because it is being run twice
There seems to be a test failing [here](https://app.circleci.com/pipelines/github/hasura/graphql-engine/9560/workflows/b431f78b-16a4-4b22-a7c6-ebe6bfb4f784/jobs/117601/steps).

The origin of the problem is that the relevant integration test changes data in the DB using `_inc`, and compares the new value with an expected new value. So the process is: setup -> test -> test -> teardown, rather than setup -> test -> teardown, as would be normal when no upgrades are taking place. Since the `_inc` operator is not idempotent, this causes the test to fail the second time it is being run.

In other words, an integration test is failing because the new version has triggered an integration test to be modified. The code itself seems to be functioning correctly.

abooij:
### Description

Add missing value teardown in the tests for the `_inc` operator.

### Affected components

- [x] Tests

### DIFF
--- a/.circleci/server-upgrade-downgrade/run.sh
+++ b/.circleci/server-upgrade-downgrade/run.sh
@@ -174,9 +174,12 @@ run_server_upgrade_pytest() {
 		set -x
 
 		# With --avoid-error-message-checks, we are only going to throw warnings if the error message has changed between releases
+ 		# FIX ME: Remove the deselect below after the next stable release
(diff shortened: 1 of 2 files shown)


abooij:
Thanks to @tirumaraiselvan and @nizar-m for tracing this back!

PR #4611, opened 2020-04-29
cli: add support for alternative configuration filename (fix #4561)

### LINKED ISSUE #4561: cli: add an alternate filename for config.yaml
Couple of our users have mentioned that `config.yaml` is too generic. We cannot remove the support for config.yaml, but we can introduce an optional new file `hasura-config.yaml` which can be selected by using a flag on the `init` command. 

Later we could probably make it a default too.

The function which validates a folder to be a Hasura Project will need to be updated accordingly.

ShraddhaAg:
### Description

This PR adds an additional flag `--hasura-config` to `hasura init` to allow creating a hasura project directory using config filename `hasura-config.yaml`. 
### Changelog

### Affected components

- [x] CLI
- [ ] Docs

### Related Issues

#4561 
### Solution and Design

Add a flag to `--hasura-config` to name the configuration file as `hasura-config.yaml`.
To make sure config file is picked up in all other commands: 
* Changed the `ValidateDirectory` func in `directory.go` to return error if neither of the configuration files, `config.yaml` and `hasura-config.yaml` are not found. 
* Added a `getConfigFilename` func in `cli.go` to return the name of the config file found. 

In case a directory contains a `config.yaml` and `hasura-config.yaml`, the latter will be used as the configuration file for hasura project.

### DIFF
--- a/cli/directory.go
+++ b/cli/directory.go
@@ -87,7 +87,7 @@ func ValidateDirectory(dir string) error {
 	}
 	// if neither config.yaml or hasura-config.yaml is found, return error
 	if len(notFound) > 1 {
-		return errors.Errorf("cannot validate directory '%s': [%s] not found", dir, strings.Join(notFound, ", "))
+		return errors.Errorf("cannot validate directory | configuration file not found")
 	}
 	return nil
 }
--- a/cli/directory.go
+++ b/cli/directory.go
@@ -49,9 +49,11 @@ func (ec *ExecutionContext) validateDirectory() error {
 }
 
 // filesRequired are the files that are mandatory to qualify for a project
-// directory.
+// directory. Either of the configuration files, config.yaml or hasura-config.yaml
+// validate the project directory.
 var filesRequired = []string{
 	"config.yaml",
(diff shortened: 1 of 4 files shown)


cli/directory.go
```diff
@@ -83,7 +85,8 @@ func ValidateDirectory(dir string) error {
 			notFound = append(notFound, f)
 		}
 	}
-	if len(notFound) > 0 {
+	// if neither config.yaml or hasura-config.yaml is found, return error
+	if len(notFound) > 1 {
 		return errors.Errorf("cannot validate directory '%s': [%s] not found", dir, strings.Join(notFound, ", "))
```
ShraddhaAg, line 90: I'm not sure if this is the best way to validate the project directory, when alternative filenames for configuration file are present. 
If it is, should I rename the error message here to `cannot validate directory | no configuration file found`? As we currently validate a project directory if only a configuration file is found.

Stefmore02:
Closing this PR as it's not relevant to the most up-to-date Hasura version.

PR #4620, opened 2020-04-30
server: cache default values (fix #3773)

### LINKED ISSUE #3773: Query sometime fails when ID is marked as required with error "expected a value for non-nullable variable"

### LINKED ISSUE #4577: server: the first actual value, rather than the default value, is used for obligatory variables with default values
When a query with an obligatory variable `$myId` with a default value, such as `1`, is cached, the value in the query plan that is saved is the value that is used at the time of saving (e.g. `2`, since a value for `$myId` may have been provided with the query). This means that any subsequent call to the same query, that does not specify `$myId`, incorrectly uses that same value saved at the time of the first call, i.e. `2` in this example, even though a default value of `1` was specified in the query.

To reproduce, launch graphql-engine with a single capability, e.g. as:
```bash
$ cabal new-run --project-file=cabal.project.dev-sh --RTS -- exe:graphql-engine +RTS -N1 -T -RTS --database-url=postgres://postgres:postgres@127.0.0.1:25432/postgres serve --enable-console --console-assets-dir /somewhere/change/this/.../graphql-engine/console/static/dist
```
Then create a `test` table with a single column `id` (integer auto-increment), and create two entries. Open the console and run:

```graphql
query MyQuery($myId :Int! = 1) {
  test(where: {id: {_eq: $myId}}) {
    id
  }
}
```
Query variables:
```json
{"myId": 2}
```
Observe that the item with `id` `2` is returned (as expected).

Now remove the query variables, and run the same query again, so that we expect to get the entry with `id` `1`. Instead,
```json
{
  "data": {
    "test": [
      {
        "id": 2
      }
    ]
  }
}
```

The origin of this problem lies in the fact that the query plan that gets saved (generated by `Hasura.GraphQL.Validate.validateGQ`) incorporates the provided values. I suspect we'll need to refactor these functions somewhat. I would prefer to separate the query validation from the variable validation much more strongly, so that we can validate and cache the query plan independently from the query variables. Such separation would have probably avoided #3773, and its unreported variant for subscriptions (the latter is briefly discussed in #4538).

abooij:
### Description

This is a really dirty way to fix #3773 and fix #4577.

### Changelog

### Affected components

- [x] Server

### Solution and Design

The origin of #3773 is that the "validation" stage not just validates correctness of input data, but also gives back updated values. In particular, the "validation" combines the GraphQL query specification with the argument values, so that any default values are overwritten. As a consequence, the default value was not accessible during the query planning stage.

As a dirty hack to this, this code adds a field to `AnnInpVal` that specifies a default value (if any). This can then be updated at a later stage, so that the default value is available during the query planning stage.

The real solution here would be:
1. to parse, rather than validate, the query,
2. to separate the parsing (or validation) of the GraphQL query from the parsing (or validation) of argument values, so that we do not risk overwriting default values, and
3. to model the query planning in a more type-safe way, so that we avoid the temptation of overwriting one value with another.

### Steps to test and verify

See #3773 and #4577.

This PR does NOT contain a regression test, as I do not know of an easy way to test this. So when #4111 is merged, we should pay attention that this bug does not re-appear.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ A scheduled trigger can be used to execute custom business logic based on time. 
 
 A cron trigger will be useful when something needs to be done periodically. For example, you can create a cron trigger to  generate an end-of-day sales report every weekday at 9pm.
 
-You can also schedule one-off events based on a timestamp. For example, a new scheduled event can be created for 2 weeks from when a user signs up to send them an email about their experience. 
+You can also schedule one-off events based on a timestamp. For example, a new scheduled event can be created for 2 weeks from when a user signs up to send them an email about their experience.
 
 <Add docs links>
 
@@ -51,6 +51,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 (Add entries here in the order of: server, console, cli, docs, others)
 - server: compile with GHC 8.10.1, closing a space leak with subscriptions. (close #4517) (#3388)
 
+- server: cache default variables for non-optional variables better (fix #3773)
 - server: avoid loss of precision when passing values in scientific notation (fix #4733)
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
 - console: avoid count queries for large tables (#4692)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,7 +51,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 (Add entries here in the order of: server, console, cli, docs, others)
 - server: compile with GHC 8.10.1, closing a space leak with subscriptions. (close #4517) (#3388)
 
-- server: cache default variables for non-optional variables better (fix #3773)
+- server: fix an issue with query plan caching where valid queries would intermittently fail (fix #3773)
 - server: avoid loss of precision when passing values in scientific notation (fix #4733)
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
 - console: avoid count queries for large tables (#4692)
--- a/server/src-lib/Hasura/GraphQL/Resolve/InputValue.hs
+++ b/server/src-lib/Hasura/GraphQL/Resolve/InputValue.hs
@@ -79,16 +79,11 @@ openOpaqueValue :: (MonadReusability m) => OpaqueValue a -> m a
 openOpaqueValue (OpaqueValue v isVariable) = when isVariable markNotReusable $> v
 
 asPGColumnTypeAndValueM
-  :: (MonadReusability m, MonadError QErr m)
+  :: forall m . (MonadReusability m, MonadError QErr m)
   => AnnInpVal
   -> m (PGColumnType, WithScalarType (Maybe (OpaqueValue PGScalarValue)))
 asPGColumnTypeAndValueM v = do
-  (columnType, scalarValueM) <- case _aivValue v of
-    AGScalar colTy val -> pure (PGColumnScalar colTy, WithScalarType colTy val)
-    AGEnum _ (AGEReference reference maybeValue) -> do
-      let maybeScalarValue = PGValText . RQL.getEnumValue <$> maybeValue
-      pure (PGColumnEnumReference reference, WithScalarType PGText maybeScalarValue)
-    _ -> tyMismatch "pgvalue" v
+  (columnType, scalarValueM) <- toPG (_aivValue v)
 
   for_ (_aivVariable v) $ \variableName -> if
     -- If the value is a nullable variable, then the caller might make a different decision based on
@@ -99,18 +94,20 @@ asPGColumnTypeAndValueM v = do
         Nothing -> do
           recordVariableUse variableName columnType ReusableNoDefault
         Just dv -> do
-          defValueM <- case dv of
-            AGScalar colTy val -> pure $ WithScalarType colTy val
-            AGEnum _ (AGEReference _ maybeValue) -> do
-              let maybeScalarValue = PGValText . RQL.getEnumValue <$> maybeValue
-              pure $ WithScalarType PGText maybeScalarValue
-            _ -> tyMismatch "pgvalue" v
+          (_ , WithScalarType _ defValueM) <- toPG dv
           let defVal = case defValueM of
-                (WithScalarType _ Nothing   ) -> ReusableNoDefault
-                (WithScalarType _ (Just val)) -> ReusableDefault val
+                Nothing  -> ReusableNoDefault
+                Just val -> ReusableDefault val
           recordVariableUse variableName columnType defVal
   let isVariable = isJust $ _aivVariable v
   pure (columnType, fmap (flip OpaqueValue isVariable) <$> scalarValueM)
+  where
+  toPG :: AnnGValue -> m (PGColumnType, WithScalarType (Maybe PGScalarValue))
+  toPG (AGScalar colTy val) = pure (PGColumnScalar colTy, WithScalarType colTy val)
+  toPG (AGEnum _ (AGEReference reference maybeValue)) =
+    let maybeScalarValue = PGValText . RQL.getEnumValue <$> maybeValue
+    in  pure (PGColumnEnumReference reference, WithScalarType PGText maybeScalarValue)
+  toPG _ = tyMismatch "pgvalue" v
 
 asPGColumnTypeAndAnnValueM
   :: (MonadReusability m, MonadError QErr m) => AnnInpVal -> m (PGColumnType, Maybe OpaquePGValue)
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 
 (Add entries here in the order of: server, console, cli, docs, others)
 
+- server: cache default variables for non-optional variables better (fix #3773)
```
0x777, line 39: Maybe a more user facing sentence like this?

> fixes an issue with query plan caching where valid queries would intermittently fail

server/src-lib/Hasura/GraphQL/Validate/Types.hs
```diff
@@ -781,10 +784,21 @@ stripTypenames = map filterExecDef
...
+data ReusableDefaultValue a
+  = ReusableNoDefault
+  | ReusableDefault a
+  deriving (Show, Eq)
+$(J.deriveToJSON J.defaultOptions{J.constructorTagModifier = drop 8} ''ReusableDefaultValue)
+
+data ReusableVariableInfo
+  = ReusableVariableInfo RQL.PGColumnType (ReusableDefaultValue PGScalarValue)
+  deriving (Show, Eq)
+$(J.deriveToJSON J.defaultOptions{J.constructorTagModifier = drop 8} ''ReusableVariableInfo)
```
0x777, line 796: Why do we need the extra `ReusableDefaultValue` type? Isn't this equally readable without losing any type safety?
```haskell
data ReusableVariableInfo
  = ReusableVariableInfo 
  { _rviType :: RQL.PGColumnType,
  , _rviDefault :: Maybe PGScalarValue
  } deriving (Show, Eq)
```

CHANGELOG.md
abooij: Yes, that is more helpful!

server/src-lib/Hasura/GraphQL/Validate/Types.hs
abooij: Because `Maybe` is subject a sort of Boolean blindness. In the current situation, we don't clearly separate `AnnInpVal`s originating from the query, and those originating from user-provided arguments. So we have to string the default values through our code, and this is error-prone, as this issue shows. The reason that we incorrectly cached these kinds of queries in the first place is that the provided value was simply `<|>`'d with the default value (in `validateVariables`), thus forgetting the default value. `ResuableDefaultValue` forces us to think explicitly about what to do with default vs user-provided values, at least in some places.

Put differently: suppose you get a `Maybe PGScalarValue` somewhere, what kind of value is this? Is this a user-provided argument? If so, has the default value been applied? If not, where would you get that from?

abooij:
This now has a regression test, conform the discussion in #4661.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
```diff
@@ -711,6 +713,7 @@ data AnnInpVal
   { _aivType     :: !G.GType
   , _aivVariable :: !(Maybe G.Variable)
```
0x777, line 714: Let's change `G.Variable` to `VariableInfo` where `VariableInfo` represents `(G.Variable, Maybe AnnGValue)`.

0x777: Currently, we seem to be adding this value [here](https://github.com/hasura/graphql-engine/pull/4620/files?file-filters%5B%5D=.hs&file-filters%5B%5D=.md&file-filters%5B%5D=.py&w=1#diff-10bf04471cd1721031b819698494806fL325), after this change we don't need to do that anymore, `validateVariables` itself can build this `VariableInfo` and add it to the `AnnInpVal`.

server/src-lib/Hasura/GraphQL/Resolve/InputValue.hs
```diff
@@ -95,8 +95,20 @@ asPGColumnTypeAndValueM v = do
...
-
+    | otherwise                 -> case _aivDefault v of
+        Nothing -> do
+          recordVariableUse variableName columnType ReusableNoDefault
+        Just dv -> do
+          defValueM <- case dv of
+            AGScalar colTy val -> pure $ WithScalarType colTy val
+            AGEnum _ (AGEReference _ maybeValue) -> do
+              let maybeScalarValue = PGValText . RQL.getEnumValue <$> maybeValue
+              pure $ WithScalarType PGText maybeScalarValue
```
0x777, line 106: We might as well skip the construction of `WithScalarType` if we are not using that information or better yet do you think we can abstract out the logic in L86-90 and reuse it here?

server/src-lib/Hasura/GraphQL/Validate/Types.hs
0x777: Right, I see what you mean. In that case wouldn't the standard pattern be a `newtype` wrapper on `PGScalarValue` like this?
```haskell
newtype ReusableDefaultValue = ReusableDefaultValue PGScalarValue
data ReusableVariableInfo
  = ReusableVariableInfo 
  { _rviType :: RQL.PGColumnType,
  , _rviDefault :: Maybe ReusableDefaultValue
  } deriving (Show, Eq)
```
i.e, why parameterize `ReusableDefaultValue` and why encompass `Maybe` too?

0x777:
Left a few comments otherwise the overall idea looks good. I see a lot of friction because of how `AnnInpVal` is modelled currently, ideally it should've been paramaterized with the source information - default, variables, query but given that we are getting rid of it soon, this seems like a reasonable compromise.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
abooij: With the `newtype` variant you now have to unpack two constructors to get to the value, as in:
```haskell
f : Maybe ReusableDefaultValue -> a
f (Just (ReusableDefaultValue pgScalarValue)) = _
f Nothing = _
```
So since data types are really cheap in Haskell, we might as well re-define Maybe to get a more convenient API:
```haskell
f : ReusableDefaultValue -> a
f (ReusableDefault pgScalarValue) = _
f ReusableNoDefault = _
```
Note that the two patterns directly give a lot of information about *why* we are doing pattern matching, in contrast with the `newtype` variant, where it's not quite clear, visually, what kind of value we are deconstructing. Indeed, in the second variant, the type signature is pretty much unnecessary to understand the code, whereas in the first variant one would need to be quite familiar with the codebase in order to be able to do without.

Anyway, this is not a critical point, so I'm happy to make this into a newtype if that allows this PR to get merged without further debate.

server/src-lib/Hasura/GraphQL/Resolve/InputValue.hs
abooij: Right, some code can be either simplified or unified here.

I have now fixed this by abstracting lines 86-90 as you suggest, but I think there is more work to be done here, as typing information now seems to be passed three times: once in the `PGColumnType`, once in the `WithScalarType`, and once in the inner `PGScalarValue`. This really needs to be reworked. I have made a note elsewhere about the idea of using type families for `WithScalarType` to avoid this duplication of type level information.

server/src-lib/Hasura/GraphQL/Validate/Types.hs
abooij: Your link doesn't work for me, unfortunately, so I am not sure what you are suggesting.

0x777:
Not relevant any more.

Issue #3097, opened 2019-10-11
`include`/`skip` directives can cause queries to be incorrectly cached


lexi-lambda:
Currently, the GraphQL validation pass handles `include` and `skip` directives by completely removing them from the query before passing the query to the resolve pass. Because these variable references are stripped, they are not counted as a reason not to cache the query, which is quite wrong.

To reproduce this, run `graphql-engine` with the `+RTS -N1 -RTS` option to use a single OS thread, which disables cache striping so there won’t be any spurious cache misses. Then execute a query with an `include` directive, such as this one:

```graphql
query ($flag: Boolean!) {
  user @include(if: $flag) {
    id
  }
}
```

The first time it is executed, it will succeed, but the second time it is executed, it will fail with the following error:

```json
{
  "errors": [
    {
      "extensions": {
        "path": "$.variableValues",
        "code": "validation-failed"
      },
      "message": "unexpected variables: flag"
    }
  ]
}
```

PR #4661, opened 2020-05-04
server: use reusability more, reuse introspection queries less (fix #4547)

### LINKED ISSUE #4547: Introspection-query error
Hello. I tried running an introspection-query of my environment using [elm graphql](https://www.npmjs.com/package/@dillonkearns/elm-graphql), but get this error:
```json
{"type":"http-log","timestamp":"2020-04-23T16:02:52.106+0000","level":"error","detail":{"operation":{"user_vars":{"x-hasura-role":"admin"},"error":{"path":"$.variableValues","error":"unexpected variables: includeDeprecated","code":"validation-failed"},"request_id":"ce53493c-fc37-42c0-9bf9-b7a5c7fe207a","response_size":104,"query":{"variables":{"includeDeprecated":true},"query":"query IntrospectionQuery($includeDeprecated: Boolean!) {\n    __schema {\n      queryType {\n        name\n      }\n      mutationType {\n        name\n      }\n      subscriptionType {\n        name\n      }\n      types {\n        ...FullType\n      }\n    }\n  }\n\n  fragment FullType on __Type {\n    kind\n    name\n    description\n    fields(includeDeprecated: $includeDeprecated) {\n      name\n      description\n      args {\n        ...InputValue\n      }\n      type {\n        ...TypeRef\n      }\n      isDeprecated\n      deprecationReason\n    }\n    inputFields {\n      ...InputValue\n    }\n    interfaces {\n      ...TypeRef\n    }\n    enumValues(includeDeprecated: $includeDeprecated) {\n      name\n      description\n      isDeprecated\n      deprecationReason\n    }\n    possibleTypes {\n      ...TypeRef\n    }\n  }\n\n  fragment InputValue on __InputValue {\n    name\n    description\n    type { ...TypeRef }\n    defaultValue\n  }\n\n  fragment TypeRef on __Type {\n    kind\n    name\n    ofType {\n      kind\n      name\n      ofType {\n        kind\n        name\n        ofType {\n          kind\n          name\n          ofType {\n            kind\n            name\n            ofType {\n              kind\n              name\n              ofType {\n                kind\n                name\n                ofType {\n                  kind\n                  name\n                }\n              }\n            }\n          }\n        }\n      }\n    }\n  }"}},"http_info":{"status":200,"http_version":"HTTP/1.1","url":"/v1/graphql","ip":"XXX.XX.XX.XXX","method":"POST","content_encoding":null}}}
```
If I reload my meta-data in the Hasura-console, this exact introspection-query succeeds, but every subsequent introspection fails :(

abooij:
### Description

Introspection queries accept variables, but we need to make sure to also touch the variables that we ignore, so that an introspection query is marked not reusable if we are not able to build a correct query plan for it. (fixes #4547)

### Changelog

### Affected components

- [x] Server

### Solution and Design

In `server/src-lib/Hasura/GraphQL/Resolve/Introspect.hs`, we ignored the argument `includeDeprecated` to `enumValues` and `fields`. This fixes that to simply use the `InputValue.hs` code to read that argument, so that if it's a variable, the query is marked unreusable.

A better solution here would be to deal with such unused variables correctly, so that more introspection queries become reusable.

An even better solution would be to type-safely track *how* to reuse which variables. This should help having such bugs in the first place. The problem with the current setup is that the reusage marking is separated from the query planning, so that it is easy to make mistakes (see also #3773).

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No
#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/GraphQL/Resolve/Introspect.hs
```diff
@@ -230,22 +230,39 @@ namedTypeR nt fld = do
...
+  -- Here we make sure to read the 'includeDeprecated' argument to the
+  -- 'fields' and 'enumValues' fields, to ensure that the query is
+  -- marked unreusable if necessary (fix #4547).
+  _ <- forM (toList (_fSelSet fld)) $ \subFld ->
+    case _fName subFld of
+      "fields"     -> readIncludeDeprecated subFld
+      "enumValues" -> readIncludeDeprecated subFld
+      _            -> return ()
+  -- Now fetch the required type information from the corresponding
+  -- information generator
```
0x777, line 247: We can parse the arguments for these fields in their respective type resolvers? In [`enumTypeR`](https://github.com/hasura/graphql-engine/pull/4661/files#diff-9539cee3bf7b3e5530f83d51b5ae1a0dL170) for example? And similarly in `objectTypeR` and `ifaceR'`?

0x777: We also don't seem to use these arguments to generate the response as required by the spec. In case of enums it is trivial to fix. In case of objects and interfaces, currently we don't seem to track whether a field is deprecated or not. So we might as well fix that.

0x777:
A bigger question is how do we test this? How do we test whether a query's plan should/should not be cacheable?

server/src-lib/Hasura/GraphQL/Resolve/Introspect.hs
abooij: I parse these arguments here, rather than in the individual fields, because the code would be the same in each of the fields.

> In case of objects and interfaces, currently we don't seem to track whether a field is deprecated or not. So we might as well fix that.

Where can we get deprecation information from?

abooij:
> A bigger question is how do we test this? How do we test whether a query's plan should/should not be cacheable?

My gut feeling tells me to rather focus on using more type safety and code reuse to avoid these kinds of bugs in the first place. However, that would of course require refactoring. I have to think about how to test these kinds of issues in the current setup.

abooij:
@0x777 One testing option is to have an additional query directive `@_force_from_cache` that first runs the query until it's cached, and then runs it again to get the result. #4547 would've been caught by that, as it would produce an error on the second run. It would also be possible to manifest #3773 by having such a `@_force_from_cache` query where the non-nullable variable with a default value is not set by the user-provided JSON values by the test. Although I'm not sure we would've intentionally written that: so I think at best we should see this as an approach towards regression testing. I don't feel confident that we would be able to catch, or even regression test, all caching-based bugs this way.

However, there is a disadvantage to this approach. It would mean that, on top of having a cached and non-cached code path, we would be introducing a *third* code path where both of these are run in succession. I see a lot of dangers that we introduce more bugs by adding a third code path.

And of course it should be added that this is fundamentally the wrong approach, since this directive would mean writing testing code directly inside the library, rather than having it external. From this point of view, it would make more sense to test these bugs by unit testing rather than integration testing, as we would be able to write this third code path outside of the graphql-engine library, thus at least avoiding the problem of being the wrong approach.

So, having written all of this, I am starting to think that we should test this as a Haskell-based unit test. If we would take that path, then, again, it would make most sense to start with a refactoring, so that we can test the smallest possible units, so that we can express desired behavior as properties that can be testing with quickcheck, and so that the Haskell types would give us a better view of where we can expect bugs.

server/src-lib/Hasura/GraphQL/Resolve/Introspect.hs
0x777: > I parse these arguments here, rather than in the individual fields, because the code would be the same in each of the fields.

Aren't we better off just writing it in the respective type handler? For example we seem to be reusing `objectTypeR` in `unionR`. 

> Do we currently store deprecation information anywhere?

`EnumValInfo` stores deprecation information in the `_eviIsDeprecated` field. This information seems to be missing in `ObjFldInfo`.

0x777: If we were to fix handling `includeDeprecated` argument correctly, the deprecation information in case of fields for object types introduced through remote schemas has to be propagated from the introspection query.

0x777:
Regarding testing query plans, I agree with you - I'm not sure how e2e tests can reliably capture this. However, maybe we can do this instead? 
1. Run graphql-engine with single capability with `developer` API enabled.
2. Run a query that we expect to be cached.
3. Ensure that this is cached or not-cached by checking the plan cache.
4. If we expect the query to be cached, then maybe run the query with different parameters?

The above approach still doesn't 'guarantee' that a query is executed using a cached plan even though we have ensured that an entry exists in the plan cache. This can be improved  
1. If we publish some metric in the response or a log that we can correlate with.
2. If we test the observable nature of a plan cache - lower latencies and high throughput

For now just having the above testing plan would improve things I feel. What do you think?

abooij:
@0x777 I have added functionality for outputting the enum values only when the `includeDeprecated` flag is set to true. I'm not sure adding the same functionality for objects and interfaces (`ObjFldInfo`) is worth the effort, since it would require quite some additional effort - I leave this judgment up to you for now.

> Aren't we better off just writing it in the respective type handler? For example we seem to be reusing objectTypeR in unionR.

We need to parse the `includeDeprecated` argument irrespective of what kind of thing we're actually looking at. If the user has selected a scalar type, we need to open the `includeDeprecated` argument to mark the query as unreusable (or register this variable by some other method), even though the value is irrelevant. So there is no "*the*" handler: all 6 handlers need to read this field.

> Run graphql-engine with single capability with developer API enabled.

How would you do this? Do you want to add a second python test suite that re-starts graphql-engine with these settings? Or do you want to run all integration tests under a single capability? I worry the latter would not result in realistic test conditions.

0x777:
> We need to parse the includeDeprecated argument irrespective of what kind of thing we're actually looking at. If the user has selected a scalar type, we need to open the includeDeprecated argument to mark the query as unreusable (or register this variable by some other method), even though the value is irrelevant. So there is no "the" handler: all 6 handlers need to read this field.

Good catch! I was thinking about `__Type` as a union of types but that isn't the case. Does this also mean that our use of `objectTypeR` in other locations is erroneous, i.e, it should be using `namedTypeR'`? 

> @0x777 I have added functionality for outputting the enum values only when the includeDeprecated flag is set to true. I'm not sure adding the same functionality for objects and interfaces (ObjFldInfo) is worth the effort, since it would require quite some additional effort - I leave this judgment up to you for now.

We can leave this for now, let's track it with an issue. 

> How would you do this? Do you want to add a second python test suite that re-starts graphql-engine with these settings? Or do you want to run all integration tests under a single capability? I worry the latter would not result in realistic test conditions.

Looks like we have different sets of tests testing different invocations of graphql-engine. See `.circleci/test-server.sh`!

abooij:
> > We need to parse the includeDeprecated argument irrespective of what kind of thing we're actually looking at. If the user has selected a scalar type, we need to open the includeDeprecated argument to mark the query as unreusable (or register this variable by some other method), even though the value is irrelevant. So there is no "the" handler: all 6 handlers need to read this field.
> 
> Good catch! I was thinking about `__Type` as a union of types but that isn't the case. Does this also mean that our use of `objectTypeR` in other locations is erroneous, i.e, it should be using `namedTypeR'`?

I think this is correct, since the GraphQL spec says:
> `possibleTypes` returns the list of types that can be represented within this union. They must be object types.

Continuing:

> > @0x777 I have added functionality for outputting the enum values only when the includeDeprecated flag is set to true. I'm not sure adding the same functionality for objects and interfaces (ObjFldInfo) is worth the effort, since it would require quite some additional effort - I leave this judgment up to you for now.
> 
> We can leave this for now, let's track it with an issue.

Okay, I'll make one.

> > How would you do this? Do you want to add a second python test suite that re-starts graphql-engine with these settings? Or do you want to run all integration tests under a single capability? I worry the latter would not result in realistic test conditions.
> 
> Looks like we have different sets of tests testing different invocations of graphql-engine. See `.circleci/test-server.sh`!

Ah, so you are proposing to only run this in CI? (Which I would agree would be good enough.)

In fact, there already seems to be one test running on graphql-engine with a single capability, namely `test_graphql_queries.py::TestGraphQLQueryCaching` - see lines 465-468 and #3097. Okay, given the fact we can run graphql-engine with a single capability, I think it would be sufficient to regression test this bug rather naively (just check that the query runs and re-runs successfully), even without checking whether anything is cached, as, under a single capability, this thing used to cache very reliably. (Of course, after this PR is merged, it won't be anymore.) And the caching-or-not of this query is not a bug in itself: the bug is caused by *incorrect* caching.

0x777:
> Okay, given the fact we can run graphql-engine with a single capability, I think it would be sufficient to regression test this bug rather naively (just check that the query runs and re-runs successfully), even without checking whether anything is cached, as, under a single capability, this thing used to cache very reliably.

Makes sense. 

>> Does this also mean that our use of `objectTypeR` in other locations is erroneous, i.e, it should be using `namedTypeR'`?
> 
> I think the current code is correct, since the GraphQL spec says:

If we don't use `namedTypeR'` when handling `possibleTypes`, we wouldn't end up reading the `includeDeprecated` argument of `fields` and we won't end up marking the plan as not reusable?

abooij:
> If we don't use `namedTypeR'` when handling `possibleTypes`, we wouldn't end up reading the `includeDeprecated` argument of `fields` and we won't end up marking the plan as not reusable?

Oh, right. So yes, under the current design, we should probably touch that argument, for example via `objectTypeR`.

0x777:
I feel like there needs to be a comment explaining this question: 
> Why are we looking to read `fields` and `enumValues` for all types?

Something along these lines:

> This is because `__Type` is not defined as a `union` of `__Scalar`, `__Object` ... in the GraphQL spec, so every type handler has to handle all fields.

If this can be [added as a note](https://github.com/hasura/graphql-engine/blob/master/server/STYLE.md#out-of-line-notes) and referenced in each type handler, it would be great. Otherwise LGTM.

server/src-lib/Hasura/GraphQL/Resolve/Introspect.hs
```diff
@@ -230,22 +270,25 @@ namedTypeR nt fld = do
...
   -> m J.Object
-namedTypeR' fld = \case
-  TIScalar colTy        -> scalarR colTy fld
-  TIObj objTyInfo       -> objectTypeR objTyInfo fld
-  TIEnum enumTypeInfo   -> enumTypeR enumTypeInfo fld
-  TIInpObj inpObjTyInfo -> inputObjR inpObjTyInfo fld
-  TIIFace iFaceTyInfo   -> ifaceR' iFaceTyInfo fld
-  TIUnion unionTyInfo   -> unionR unionTyInfo fld
+namedTypeR' fld tyInfo = do
+  -- Now fetch the required type information from the corresponding
```
0x777, line 278: This comment [may not be needed](https://github.com/hasura/graphql-engine/blob/master/server/STYLE.md#inline-comments) but I don't feel strongly about it, so your call.

abooij: This comment is here to help any future readers of the code who may be confused, as I was initially, about what is going on here. I like to document things that are obvious in hindsight :-)

abooij:
> LGTM, one minor change is needed: [#4661 (comment)](https://github.com/hasura/graphql-engine/pull/4661#issuecomment-629035969)

I added a GHC-style note, hopefully explaining what is going on.

CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 (Add entries here in the order of: server, console, cli, docs, others)
 - server: compile with GHC 8.10.1, closing a space leak with subscriptions. (close #4517) (#3388)
 
+- server: avoid an error when calling a cached introspection query (fix #4547)
```
tirumaraiselvan, line 54:
```suggestion
- server: fixes an issue where introspection queries with variables would fail because of caching (fix #4547)
```

per @0x777

PR #4667, opened 2020-05-05
Refactor the WAI shutdownHandler to use an MVar

paf31:
This makes it possible to run the server without setting up the default graceful shutdown handler (which listens for SIGTERM). Instead, the caller can pass in their own shutdown latch in the form of an `MVar`.

### Description

There should be no user-facing changes. This just makes it possible to control the shutdown of the existing server in new ways.

### Changelog

### Affected components

- [x] Server

### Steps to test and verify

The existing behavior can be verified using `killall graphql-engine`. The `graceful shutdown initiated` message will show in the logs.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No
#### GraphQL
#### Breaking changes
- [x] No Breaking changes


server/src-exec/Main.hs
```diff
@@ -26,7 +27,16 @@ runApp (HGEOptionsG rci hgeCmd) =
   withVersion $$(getVersionFromEnvironment) case hgeCmd of
     HCServe serveOptions -> do
       (initCtx, initTime) <- initialiseCtx hgeCmd rci
-      runHGEServer serveOptions initCtx initTime
+      latch <- liftIO newShutdownLatch
+      -- Catches the SIGTERM signal and initiates a graceful shutdown. Graceful shutdown for regular HTTP
+      -- requests is already implemented in Warp, and is triggered by invoking the 'closeSocket' callback.
+      -- We only catch the SIGTERM signal once, that is, if the user hits CTRL-C once again, we terminate
+      -- the process immediately.
```
lexi-lambda, line 34: This is just a style comment, but [our style guide now states comments should be wrapped to 80 characters](https://github.com/hasura/graphql-engine/blob/1211c23a2fa0cefa81c1258e7f8071c7da5c6bbf/server/STYLE.md#line-length).

server/src-lib/Hasura/App.hs
```diff
@@ -193,8 +208,9 @@ runHGEServer
   -> InitCtx
   -> UTCTime
   -- ^ start time
+  -> ShutdownLatch
```
lexi-lambda, line 211: This seems like it should probably go in the `InitCtx` with all the other odds and ends.

Issue #1558, opened 2019-02-04
Allow me to throw config.yaml in the trash


corysimmons:
To run `$ hasura console` I've needed to have a `config.yaml` present.

It _only_ had the graphql endpoint in it.

If I delete this file `$ hasura console` fails.

If I delete this file and use the `--endpoint` flag (effectively replacing the only data `config.yaml` was providing), `$ hasura console` _still_ fails. I'd expect it to work.

My use-case is wanting to do something like `$ yarn dev:staging` and `$ yarn dev:production` where these scripts just have different `--endpoint`s.

letientai299:
My issue (#2039) is closed as duplicated to this one, so I will plug my suggestion here. 

- Beside flag `--endpoint`, we can also have flag `--admin-secret`, and that will free us from `config.yaml` completely.

- The folder `migration` is not needed for following commands: 
    - `metadata version`
    - `metadata clear`
    - `metadata export` (the folders can be created on the fly)
    - `console` (with the help of 2 new flags)

HarryVasanth:
> My issue (#2039) is closed as duplicated to this one, so I will plug my suggestion here.
> 
> * Beside flag `--endpoint`, we can also have flag `--admin-secret`, and that will free us from `config.yaml` completely.
> * The folder `migration` is not needed for following commands:
>   
>   * `metadata version`
>   * `metadata clear`
>   * `metadata export` (the folders can be created on the fly)
>   * `console` (with the help of 2 new flags)

Did you manage to find a workaround for this @letientai299?

PR #4668, opened 2020-05-05
[RFC] cli: remove config.yaml

ShraddhaAg:
### Description

RFC for removing `config.yaml` from cli. Read at https://github.com/ShraddhaAg/graphql-engine/blob/rfc-remove-config.yaml/rfcs/remove-config-file.md
### Changelog

### Affected components

- [X] RFC

### Related Issues

#1558 and #4561

### DIFF
--- a/rfcs/remove-config-file.md
+++ b/rfcs/remove-config-file.md
@@ -1,0 +1,56 @@
+Problem
+-----------
+As mentioned in #1558, users don't want be restricted by a specific project directory structure. Removing `config.yaml` will enable us in doing that.
+
+Context
+-----------
+
+**Purpose of `config.yaml`:**
+1. Provide configuration key values.
+2. Get execution directory path (by validating project directory, which in turn, enables us to run commands from anywhere inside the hasura project directory tree.) 
+
+**Current alternatives available:**  
+* For 1, all the key/value pairs in `config.yaml` can be provided using CLI flags / ENV vars / .env file (user defined or default `.env`). `config.yaml` is completely replaceable in this regard.
+
+* For 2, we have a global flag `--project` to define the execution directory. In case we remove complete support for `config.yaml`, the flag will need to be provided with every command we run, which might not be the best user experience. 
+
+Proposed Solutions
+--------------------------    
+
+**Define a global flag `--config-file` and make `config.yaml` optional.** 
+
+In this case, depending on how config is handled: 
+
+* **User defined config file** ie `--config-file` flag provided: execution directory derived from config file provided (unless `--project` is provided). Commands can be run from anywhere inside the project root but flag needs to be provided. 
+```
+# initializing hasura project, here the given file is created 
+# (syntax: mandatory flag, path: no change) 
+hasura init [directory-name] --config-file path/to/config/file
+
+# running other commands, (syntax: mandatory flag added, path: no change)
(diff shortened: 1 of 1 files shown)


scriptonist:
I'm listing down my initial thoughts here and expect this to have many rough edges.

## Objectives
1. Make `config.yaml` optional for running CLI
    - If a user is running the command outside a project directory and providing all the required configuration values via flags or environment variables things should just work.
2. All v1.x projects should maintain the same behaviour AKA to be backward compatible.
    - The behaviour of CLI inside a project should not change in any of the 1.x CLI's. 
       ie things like running a Hasura command from anywhere inside a project directory should continue to work.

Respecting Objective 2, we may not have the liberty to discontinue using a config file, rather make it optional. Therefore I feel our solution should be a combination of everything @ShraddhaAg suggested.

## Use cases
The use cases that we want to satisfy are
### Use case 1
A user expecting Hasura commands to work the same as in the past and likes having a project directory setup.


### Use case 2
A user who doesn't want populate a config.yaml file just to do simple tasks [like opening a console](https://github.com/hasura/graphql-engine/issues/1558)

Even though this is a simple task, the CLI needs to know the locations to which migrations and metadata files are to be saved.
    
### Use case 3
A user with a config V2 project who'll not have a project directory as such, they will have a configuration file with all the necessary values filled in. This use case is only applicable to a config v2 since config v1 enforces a project structure.

### Rough Solution

##### Satisfying use case 1 and 2
`hasura migrate apply`

https://docs.google.com/drawings/d/1WmR6z4WSgk-dRFlZNrAdPutwWcTldkmDXVVsXrJEox4/edit?usp=sharing

##### Satisying use case 3
`hasura migrate apply --config-file /some/path`

https://docs.google.com/drawings/d/1zWCD7Un6qLgfs1r6zNQ6I-f6g6SWwboJyDdQuQSJaVU/edit?usp=sharing

I was thinking out loud here, so please feel free to point out any mistakes.

shahidhk:
I agree with @scriptonist. This looks good. @arvi3411301 Will you take a look too?

arvi3411301:
@shahidhk which config version we would assume if the config file is not found in case of use case 1. Currently cli assumes config version to be 1 in case version field is not found.

Other than that everything seems to be fine for me. 👍

scriptonist:
Whenever a config file is not in the picture I was thinking we assume Config V2, do you foresee any issues from this @arvi3411301

arvi3411301:
@scriptonist We need to inform the user about the default config version, otherwise user who is in config version `1` may end up corrupting his project directory since migrations and metadata are handled differently from config version `2`

cc @shahidhk

scriptonist:
@arvi3411301 I agree with you that we should explicitly let the user know what config version the CLI is using.

[If we cannot find a config file then it means -- we were not able find a project directory](https://docs.google.com/drawings/d/1WmR6z4WSgk-dRFlZNrAdPutwWcTldkmDXVVsXrJEox4/edit)
Then, what we'll do is get all config values from flags and env vars. This should be according to Config V2 since we have a strict requirement of a project directory for Config V1.

But I agree we have to be explicit in docs, command help, logs etc about the config version used in each case.

Stefmore02:
Closing this PR as it's outdated. Feel free to re-open if you consider otherwise.

PR #4675, opened 2020-05-05
docs: Add example for nested objects do not exist

bertyhell:
I kept getting stuck here because i was trying to invert the query like this:

    {
      author (
        where: {
          articles: {
            _not: {}
          }
        }
      ) {
        id
        name
        articles_aggregate {
          aggregate {
            count
          }
        }
      }
    }

instead of:

    {
      author (
        where: {
          _not: {
            articles: {}
          }
        }
      ) {
        id
        name
        articles_aggregate {
          aggregate {
            count
          }
        }
      }
    }

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


chrishawn:
is this the only way to achieve "where nested object(s) do not exist:" because with larger data sets the execution plan on this is unmanageable

```
Aggregate  (cost=18443218.52..18443218.53 rows=1 width=32)
  ->  Limit  (cost=166867.52..18443217.27 rows=100 width=32)
        ->  Nested Loop Left Join  (cost=166867.52..6900402715.70 rows=37755 width=32)
```

some other approaches that would be nice but don't seem to be supported while i know they are faster in sql
`left join where article.id is null`
`articles_aggregate_count=0`

chrishawn:
i know this is a doc related ticket but at minimum something should be included to say doing this can present some very inefficient results with larger data sets and should be reconsidered

PR #4684, opened 2020-05-06
console: convert settings/utils to typescript

poul-kg:
### Affected components

- [x] Console

### Related Issues
References #4314 

### Solution and Design
First try to convert .js to .ts need a feedback if something is not per expectations.


console/src/components/Services/Settings/utils.ts
```diff
@@ -5,18 +5,41 @@ export const permissionTypes = [
...
+    configuration: {
+      name: string;
+    }
+  };
+  table_name: string;
+  relationships: Array<{ rel_name: string }>;
+  permissions: Array<{ role_name: string }>
+};
+
+export const getTableNameFromDef = (def: TableDef|string): string => {
```
beerose, line 31: As TypeScript can infer this.
```suggestion
export const getTableNameFromDef = (def: TableDef | string) => {
```

console/src/components/Services/Settings/utils.ts
```diff
@@ -5,18 +5,41 @@ export const permissionTypes = [
   'delete_permission',
 ];
 
-export const getTableNameFromDef = def => {
-  if (def.constructor.name === 'Object') {
+export type TableDef = {
```
beerose, line 8: It looks like something used only in this particular use case (for `getTableNameFromDef`). And as in some other places, we have table definition represented as `{ table_name: string }` or `{ tableName: string }` or `{ name: string, other_attribute: Type }`etc, I wouldn't export this type. (At least not until we have all of these types unified).

What's more, I would just inline this type in the function definition:
```ts
export const getTableNameFromDef = (def: { name: string } | string) => {
```

console/src/components/Services/Settings/utils.ts
```diff
@@ -5,18 +5,41 @@ export const permissionTypes = [
...
 
 const filterInconsistentMetadataObject = (
-  objects,
-  inconsistentObject,
-  type
-) => {
+  objects: MetadataObject[],
+  inconsistentObject: MetadataObject,
+  type: string
+): MetadataObject[] => {
```
beerose, line 42: TypeScript can infer this properly.
```suggestion
) => {
```

poul-kg: I've removed type. But isn't it better to have type, so if later we modify method and return something different it will catch such issue?

poul-kg:
upgraded my forked `master` branch with current `graphql-engine/master`, and rebased my branch onto `master` so everything is in sync.

marionschleifer:
@poul-kg thanks a lot for your contribution 🙏 as a thank you, we'd love to send you some swag. If you're interested, please send an email to marion@hasura.io and we'll take it from there 🙂

Issue #4189, opened 2020-03-25
console: investigate and possibly update dependencies


beerose:
In the console, we have a lot of outdated dependencies, which may lead to some security vulnerabilities or issues with the library in the older version not being maintained anymore. From the TypeScript point of view, we may face problems like missing typings or not maintained typings for older versions. 

Updating dependencies will also require an investigation and further discussion with the team, because even if some of the updates will be relatively straightforward, some of them may introduce significant breaking changes. How can we go about that?
a) If we can benefit from the breaking change — we should go with an update and preferably create a new issue for that if this update requires rewriting major parts of the project.
b) If there are no benefits or there are some disadvantages for us coming with the new version (like entirely new API that we don't want to use), we don't have to update this dependency. We can leave it or think about removing this library from the project or replacing it with an alternative.

Example
We are using `react-table` in version v.6.x, but in the newest v7.x, the library API completely changed (from React components to hooks). If (after discussion) we are okay with the new API and see it as an improvement, then there should be an issue created for this describing how we go about the refactor — whether we want to use `react-table` in our design system or use it somehow else.

Otherwise, if we decide we don't want to go with the new `react-table` API, we can either leave it as it is (as long as it doesn't make any troubles) or get rid of it by a) using other similar solution, b) creating our abstraction on tables. The second option (getting rid of the lib) would also require creating a separate issue for it. 

---

The point of this issue is to investigate which libs are straightforward to update and update them. For the rest, decide whether we want to leave it or upgrade it. Choosing the second option requires creating a new issue for a particular lib.

lastnamearya:
also while updating move <code>prettier</code> from dependencies to devDependencies.

PR #4705, opened 2020-05-07
console: dependencies cleanup

beerose:
### Description

‼️ **This PR upgrades Node from 8.x to 12.x.**

**Updated major versions:**
* `prettier` (formatting was run after upgrading).
* `lint-staged` (removed `"git add"` from lint-staged config in package.json, since from v10 onwards the `git add` part is automatic and not necessary to include in configuration),
* `react-autosuggest`,
* `react-tabs`.

**Dependencies that require additional work:**
* `react-router-redux` — not maintained anymore, replace with `connected-react-router`.
* `react-bootstrap` — lots of breaking changes, needs to be discussed if we want to upgrade. Pros: `react-bootstrap` v1 includes TypeScript typings.
* `react-helmet` — upgrade to the newest version, adjust code to the new API.
* `react-table` — completely different API; we may want to stick to the older version for now.
* `react-tabs` — investigate if can be removed/replaced. Currently, it's only used by two components, but one of them is not rendered (and also can be removed). Also, we have multiple solutions for tabs (react-bootstrap, UIKit).
* `react-redux`, `react-notification-system-redux` — upgrading caused some hot reload issues. 
* `graphql` — upgrade from v14 -> v15 introduces breaking changes. 
* `history` — dependent on `react-router`.
* `react-redux` — newer version introduces new API.
* `react-router` — replace with `react-router-dom`.
* `less-loader` — needs investigation.
* `resolve-url-loader` — needs investigation.
* `sass-loader` — needs investigation.
* `webpack-isomorphic-tools` — deprecated.

**Dependencies removed**:
* `deep-equal`,
* `hoist-something`,
* `invariant`,
* `lru-memoize`,
* `map-props`,
* `multireducer`,
* `match-sorter`,
* `pretty-error`,
* `query-string`,
* `uuid`,
* `valid-url`,
* `react-a11y`,
* `react-addons-test-utils`,
* `sinon`,
* `timekeeper`,
* `unused-files-webpack-plugin`,
* `webpack-bundle-analyzer`.

**Unused babel plugins removed from `.babelrc` and dependencies list**:
* `@babel/plugin-syntax-dynamic-import`,
* `@babel/plugin-syntax-import-meta`,
* `@babel/plugin-proposal-json-strings`,
* `@babel/plugin-proposal-decorators`,
* `@babel/plugin-proposal-function-sent`,
* `@babel/plugin-proposal-export-namespace-from`,
* `@babel/plugin-proposal-numeric-separator`,
* `@babel/plugin-proposal-throw-expressions`,
* `@babel/plugin-proposal-logical-assignment-operators`,
* `@babel/plugin-proposal-optional-chaining`,
* `@babel/plugin-proposal-pipeline-operator`,
* `@babel/plugin-proposal-nullish-coalescing-operator`,
* `@babel/plugin-proposal-do-expressions`,
* `@babel/plugin-proposal-function-bind`.

---

This PR is created on top of https://github.com/hasura/graphql-engine/pull/4677.

### Changelog

### Affected components

- [x] Console

### Related Issues

https://github.com/hasura/graphql-engine/issues/4189


beerose:
I don't think we need to convert the whole codebase. There is nothing wrong
with class component. The only thing we should do is replace unsafe
functions.

wawhal:
@beerose I prefer functional components over class components due to the following:

1. It is easier to read effects atomically rather than scanning the lifecycles figure out what's happening
2. Stateful logic cannot exist out of the component in class components and hence cannot be made reusable. This is a big dealbreaker really.
3. Worrying about `this` is unnecessary
4. Functional components were introduced by the React team to solve problems of class components, which by design, make them the better choice.

Nothing's blocked on it so it's not high priority, but it'll make sense to keep this as a community task so that it keeps happening on the side. I can review these in case there's lack of bandwidth.

beerose:
Of course, functional components are better choice. There's no doubt in
that. What is disagree with is giving this to the community. Refactoring
class components to functional ones is error prone, thus I think it should
be done along with some other work on the particular part of the code. So
I'd rather have a task: "Refactor X" which would be about simplifying the
data flow, refactoring react flow, etc., *and* changing class components to
functional.

console/src/components/Services/Data/Function/Modify/ModifyCustomFunction.js
```diff
@@ -62,7 +62,7 @@ class ModifyCustomFunction extends React.Component {
     ]);
   }
 
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
soorajshankar, line 65: we should slowly get rid of this.

beerose: That's true

PR #4724, opened 2020-05-10
Convert `Container.js` to TypeScript

ryands17:
### Affected components

- [x] Console

### Related Issues
[#4314 (comment)](https://github.com/hasura/graphql-engine/issues/4314#issue-595358014)

### Solution and Design

- Convert [Settings Container](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Services/Settings/Container.js) to TypeScript

- The types for the state are best addressed in the [State.js](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Services/Settings/State.js) file directly on converting to TypeScript which is why currently I have provided only the base types.

#### Breaking changes

- [x] No Breaking changes


console/src/components/Services/Settings/Container.tsx
```diff
@@ -0,0 +1,49 @@
...
+  children,
+  metadata,
+}) => {
+  const helmet = 'Settings | Hasura';
+
+  const sidebar = <Sidebar location={location} metadata={metadata} />;
+
+  const childrenWithProps = React.Children.map(
+    children,
+    (child: React.ReactElement<any>) => React.cloneElement(child, { metadata })
```
beerose, line 30: We can replace `any` with `{metadata: Metadata}`.
```suggestion
    (child: React.ReactElement<{ metadata: Metadata }>) => React.cloneElement(child, { metadata })
```

PR #4742, opened 2020-05-12
console: restore partial support for foreign tables (fix #4714)

### LINKED ISSUE #4714: Tracking foreign data wrappers from another schema, v 1.2.1
* We used the Hasura 1.2.1 version and all of a sudden the previously tracked tables from two schemas disappeared
* The tables in the public schema remained tracked, but we were not able to retrack the lost tables (they were still in the database)
* After reloading the metadata, resetting everything we could, deleting all data and starting from scratch, but nothing worked
* We resolved this issue by downgrading Hasura to v. 1.1.1, but with the new version it's not working like it used to

nicuveo:
### Description
This PR fixes an oversight of #4270.

While the console never had explicit support for foreign tables, they used to be implicitly queried alongside tables and views, as foreign tables do appear in `information_schema.tables`. In #4270, we switched to a more direct approach, explicitly querying `pg_class` instead, and explicitly specifying which types of postgres items we wanted to return: tables, views, and materialized views, effectively dropping support for foreign tables.

This PR fixes this by adding foreign tables to the list of items that the console fetches. However, while this restores a previously lost functionality, the console _still doesn't have explicit support for foreign tables_: they are misclassified as views, and some of the features of the console won't work with them (for instance, the "delete view" (_sic_) button will fail to drop the foreign table).

Adding full support for them in the console is beyond the scope of this PR and will be added to the backlog for triage.

### Changelog

### Affected components

- [x] Console

### Steps to test and verify
This PR doesn't have tests. After discussing it internally, we agreed that the ROI of the effort it would take to add them wouldn't be worth it, especially considering foreign tables are only partially supported.

The simplest way to verify this is to do something along those lines, using a running postgres server as its own foreign server:
```
create table foo (id integer, name text);
create extension postgres_fdw;
create server foreign_test
  foreign data wrapper postgres_fdw
  options (host 'localhost', port '$PORT', dbname '$DB');
create user mapping for $USER server foreign_test options (user '$USER');
create foreign table foreign_foo (id integer, value text)
  server foreign_test
  options(schema_name 'public', table_name 'foo');
```


CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 - console: avoid count queries for large tables (#4692)
 - console: add read replica support section to pro popup (#4118)
 - console: allow modifying default value for PK (fix #4075) (#4679)
+- console: re-enable partial support for foreign tables
```
tirumaraiselvan, line 42: why is it "partial"?

nicuveo: Because there never was explicit support for foreign tables in the console. They were misclassified as views, meaning several features of the console never worked with them. It was kind of an "accidental support", in a way. This PR restores this, but doesn't go any further.

tirumaraiselvan: Ah I see.

tirumaraiselvan, line 42:
```suggestion
- console: re-enable foreign tables to be listed as views in console
```

PR #4761, opened 2020-05-13
console: handle generated and identity columns in console data section (close #4552, #4863)

### LINKED ISSUE #4552: error when inserting through data browser on Postgres 12 "Generated column" based table
Hey folks,

Today I decided to use Postgres 12 feature **Generated columns** (instead of hasura computed columns for efficiency), here is what went wrong:-

1. The Generated column name is "FLID"
2. It just concats 2 text columns in the same table. (No issue found while creating on Postgres)
3. When inserting through Hsaura console, this error was shown:-

Screenshot from 2020-04-23 18-17-40

It seems hasura thinks "FLID" is a normal column and tries to insert into it, while Postgres doesn't allow you to insert it into a generated column.

Note: I have tried to pass NULL and a random value into "FLID" on INSERT, still the same error.

Regards,
Khalid Bashir

**UPDATE 1:-**

1. Hasura Metadata was not synced with postgres, because I created the column on postgres.
2. After Syncing a new error comes into action:-

```json
{
    "statement": "WITH \"fl_centers_info__mutation_result_alias\" AS (INSERT INTO \"fl_centers\".\"info\" ( \"flcn_role\", \"FLID_zone_code\", \"flcn_name\", \"FLID_region_code\", \"flcn_active_status\", \"FLID\", \"flcn_customer_fulfillment_status\", \"created_at\" ) VALUES ($1, $2, $3, $4, $5, $6, $7, DEFAULT)  RETURNING * , CASE WHEN 'true' THEN NULL ELSE \"hdb_catalog\".\"check_violation\"('insert check constraint failed')  END ) SELECT  json_build_object('returning', (SELECT  coalesce(json_agg(\"root\" ), '[]' ) AS \"root\" FROM  (SELECT  row_to_json((SELECT  \"_1_e\"  FROM  (SELECT         ) AS \"_1_e\"      ) ) AS \"root\" FROM  (SELECT  *  FROM \"fl_centers_info__mutation_result_alias\" WHERE ('true')     ) AS \"_0_root.base\"      ) AS \"_2_root\"      ), 'affected_rows', (SELECT  COUNT(*)  FROM \"fl_centers_info__mutation_result_alias\"      ) )        ",
    "prepared": true,
    "error": {
        "exec_status": "FatalError",
        "hint": null,
        "message": "cannot insert into column \"FLID\"",
        "status_code": "42601",
        "description": "Column \"FLID\" is a generated column."
    },
    "arguments": [
        "(Oid 25,Just (\"PRIMARY\",Binary))",
        "(Oid 25,Just (\"LH\",Binary))",
        "(Oid 25,Just (\"New Muslim Town\",Binary))",
        "(Oid 25,Just (\"NMT\",Binary))",
        "(Oid 25,Just (\"OFFLINE\",Binary))",
        "(Oid 25,Nothing)",
        "(Oid 25,Just (\"OFFLINE\",Binary))"
    ]
}
```
3. The reason is still same, i.e, Hasura console INSERT ROW tab is trying to insert a generated column.
4. It does not happened when inserting through graphql API though, so this can be treated as minor issue currently, but it might be a quick fix also.

beerose:
Solves https://github.com/hasura/graphql-engine/issues/4552, #4863 

### Description

If `is_generated` is set to true, editing a column is disabled in the console in `Edit Row` and `Insert Row` sections. 

I also did a small refactor and moved code that was shared between `EditRow` and `InsertRow` to the separate component. 

### Changelog

### Affected components

- [x] Console


CHANGELOG.md
```diff
@@ -98,6 +98,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 - console: respect read-only mode in actions pages (fix #4656) (#4764)
 - console: allow configuring session_argument for custom functions (close #4499) (#4922)
 - console: fix listen update column config selection for event trigger (close #5042) (#5043)
+- console: handle generated columns in console data section (fix #4552) (#4761)
```
tirumaraiselvan, line 101:
```suggestion
- console: handle generated columns correctly in data browser (fix #4552) (#4761)
```

CHANGELOG.md
```diff
@@ -98,6 +98,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 - console: respect read-only mode in actions pages (fix #4656) (#4764)
 - console: allow configuring session_argument for custom functions (close #4499) (#4922)
 - console: fix listen update column config selection for event trigger (close #5042) (#5043)
+- console: handle generated columns correctly in data browser (fix #4552) (#4761)
```
beerose, line 101:
```suggestion
- console: handle generated columns correctly in the data browser (fix #4552, #4863) (#4761)
```

beerose:
@rikinsk yes, I missed slight difference in the behaviour between `always` and `by default`.  Should be fixed now.

PR #4762, opened 2020-05-13
Allowlist metadata Import bug

### LINKED ISSUE #4687: allowList property omitted during application of metadata snapshot
_I've prepared a [minimum reproduction](https://github.com/rhyslbw/hasura-allow-list-min-reproduction)
Docker image `hasura/graphql-engine:v1.2.1.cli-migrations`_

The `allowList` property from the provided snapshot is not applied on startup, as demonstrated by a subsequent export of the metadata with no changes:
## Starting with
``` json
{
  ...
  "query_collections": [
    {
      "name": "allowed-queries",
      "definition": {
        "queries": [
          {
            "name": "allowed",
            "query": "query allowed {\n  foo {\n    id\n  }\n}"
          }
        ]
      }
    }
  ],
  "allowlist": [
    {
      "collection": "allowed-queries"
    }
  ]
}
```
### After container starts
``` json
{
  ...
  "query_collections": [
    {
      "name": "allowed-queries",
      "definition": {
        "queries": [
          {
            "name": "allowed",
            "query": "query allowed {\n  foo {\n    id\n  }\n}"
          }
        ]
      }
    }
  ]
}
```
### Remove, then manually re-add via console
``` json
{
  ...
  "query_collections": [
    {
      "name": "allowed-queries",
      "definition": {
        "queries": [
          {
            "name": "allowed",
            "query": "query allowed {\n  foo {\n    id\n  }\n}"
          }
        ]
      }
    }
  ],
  "allowlist": [
    {
      "collection": "allowed-queries"
    }
  ]
}
```
:point_up: Matches the snapshot provided to the container

codingkarthik:
fixes #4687 

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Solution and Design

Earlier, while importing the metadata, the key expected for allow-list was `allow_list` and the key used while exporting is `allowlist`. Due to this mismatch, the allow list was not getting imported.

Solution: Change the key name from `allow_list` to `allowlist`.

### Steps to test and verify

1. Add a query to the allowed list
2. Export the metadata
3. Import the metadata (earlier, the allow-list was not being imported)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/tests-py/test_v1_queries.py
```diff
@@ -535,6 +535,10 @@ def test_export_replace(self, hge_ctx):
         }
         replace_code, replace_resp, _ = hge_ctx.anyq(url, replace_query, headers)
         assert replace_code == 200, replace_resp
+        # The export_resp and export_resp_1 should be the same
```
0x777, line 538: It will greatly help to have an explanation on **why** we are doing this check, the `assert export_resp == export_resp_1` [already conveys this comment](https://github.com/hasura/graphql-engine/blob/master/server/STYLE.md#inline-comments).

codingkarthik: I have added some comments to explain the test.

server/tests-py/test_v1_queries.py
```diff
@@ -527,14 +529,26 @@ def test_export_replace(self, hge_ctx):
...
         }
+        # we are replacing the metadata with the exported metadata from the
+        # `export_metadata` response.
         replace_code, replace_resp, _ = hge_ctx.anyq(url, replace_query, headers)
         assert replace_code == 200, replace_resp
+        # we are again exporting the metadata to assert that the metadata exported
+        # after creating the metadata through the metadata APIs and the metadata
+        # exported after importing the metdata using the `replace_metadata` to be
+        # the same. Doing this check, ensures that the metadata has been properly
+        # imported by the graphql-engine.
```
0x777, line 548: This is definitely better but if you think about it, this content
> we are again exporting the metadata to assert that the metadata exported
        after creating the metadata through the metadata APIs and the metadata
        exported after importing the metdata using the `replace_metadata` to be
        the same.

can again be inferred by reading the code easily. What is hard to understand is this part:

> Doing this check, ensures that the metadata has been properly
        imported by the graphql-engine

I spent a good few minutes trying to understand how this catches the current bug at hand, you are exporting, then importing and then exporting again, why would anything change? The 'why' is the most important part that should go into the comment: 

> This test catches incorrect key names in `export_metadata` serialization, for example if `allowlist` is misspelled as `allow_list`, the first export metadata would have `{..., "allow_list":, ...}` but when you import this and export the metadata again, this incorrect key will be absent because `replace_metadata` doesn't import the incorrectly serialized metadata and therefore there is nothing about allowlist that can be exported the second time and hence the two exports will differ.

codingkarthik: I have changed the comment a bit

> This test catches incorrect key names(if any) in the `export_metadata` serialization, for example, A new query collection is added to the allow list using the `add_collection_to_allowlist` metadata API. When
        the metadata is exported it will contain the `allowlist`. Now, when this
        metadata is imported, if the graphql-engine is expecting a different key
        like `allow_list`(instead of `allowlist`) then the allow list won't be imported. Now, exporting the metadata won't contain the `allowlist` key
        because it wasn't imported properly and hence the two exports will differ. 

Let me know, what you think

PR #4765, opened 2020-05-13
cli: make config.yaml optional

ShraddhaAg:
### Changelog

### Affected components

- [X] CLI
- [ ] Docs

### Related Issues

Issue: #1558
RFC: #4668

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -483,6 +483,14 @@ func (ec *ExecutionContext) Validate() error {
 		return errors.Wrap(err, "cannot read config")
 	}
 
+	ec.Logger.Debug("config file version: ", ec.Config.Version)
+
+	// validate execution directory if config is V1
+	err = ec.validateConfigV1()
+	if err != nil {
+		return err
+	}
+
 	// set name of migration directory
 	ec.MigrationDir = filepath.Join(ec.ExecutionDirectory, ec.Config.MigrationsDirectory)
 	if _, err := os.Stat(ec.MigrationDir); os.IsNotExist(err) {
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -465,10 +465,10 @@ func (ec *ExecutionContext) Validate() error {
 		return errors.Wrap(err, "ensuring codegen-assets repo failed")
 	}
 
-	// validate execution directory
-	err = ec.validateDirectory()
+	// set execution directory and config file
+	err = ec.setExecutionDirectory()
 	if err != nil {
-		return errors.Wrap(err, "validating current directory failed")
+		return errors.Wrap(err, "validating execution directory failed")
 	}
 
 	// load .env file
@@ -477,13 +477,16 @@ func (ec *ExecutionContext) Validate() error {
 		return errors.Wrap(err, "loading .env file failed")
 	}
 
-	// set names of config file
-	ec.ConfigFile = filepath.Join(ec.ExecutionDirectory, "config.yaml")
-
 	// read config and parse the values into Config
 	err = ec.readConfig()
 	if err != nil {
 		return errors.Wrap(err, "cannot read config")
+	}
+
+	// check config version and strictly validate directory if V1
+	err = ec.validateV1Directory()
+	if err != nil {
+		return errors.Wrap(err, "validating config V1 directory failed")
 	}
 
 	// set name of migration directory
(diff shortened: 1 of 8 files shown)


ShraddhaAg:
I have not added the check, if `--config-file` flag is provided and config version is V1 in the following commands:

1. all `actions` commands: since this command is supported only from V2, the check isn't needed
2. `hasura docs`: N/A
3. `hasura help`: N/A
4. all `plugin` commands: N/A
5. `hasura update cli`: N/A
6. `hasura version`: N/A

Initially, I took the approach to verify if `--config-file` flag is set within `ec.Validate()` (by checking if `ec.ConfigFile` is already set) but that results in failing [integeration tests for `ec.Validate`](https://github.com/hasura/graphql-engine/blob/master/cli/integration_test/cli_test.go#L56) (as `ec.ConfigFile` is already set in the [`init` command test](https://github.com/hasura/graphql-engine/blob/master/cli/integration_test/cli_test.go#L50), run right before [`ec.validate` test](https://github.com/hasura/graphql-engine/blob/master/cli/integration_test/cli_test.go#L56)). So now I'm doing the check in every command where it's applicable. Hope that makes sense.

cli/cli.go
```diff
@@ -584,17 +588,25 @@ func (ec *ExecutionContext) readConfig() error {
...
-		return errors.Wrap(err, "cannot read config from file/env")
+	if ec.ConfigFile != "" {
+		configBase := filepath.Base(ec.ConfigFile)
+		configName := strings.TrimSuffix(configBase, filepath.Ext(configBase))
+		configPath := filepath.Dir(ec.ConfigFile)
+		v.SetDefault("version", "1")
+		v.SetDefault("metadata_directory", "")
+		v.SetConfigName(configName)
+		v.AddConfigPath(configPath)
+		err := v.ReadInConfig()
```
scriptonist, line 606: Will it make sense if we use https://godoc.org/github.com/spf13/viper#Viper.SetConfigFile here ?

cli/directory.go
```diff
@@ -26,7 +34,61 @@ func (ec *ExecutionContext) validateDirectory() error {
...
+// is set relative to project flag.
+func (ec *ExecutionContext) validateConfigFile() error {
+	if len(ec.ExecutionDirectory) == 0 {
+		ec.ExecutionDirectory = filepath.Dir(ec.ConfigFile)
+	} else {
+		if !filepath.IsAbs(ec.ConfigFile) {
+			ec.ConfigFile = filepath.Join(ec.ExecutionDirectory, ec.ConfigFile)
+		}
+	}
+	ec.ConfigFile, _ = filepath.Abs(ec.ConfigFile)
```
scriptonist, line 84: Can we handle the error here?

cli/directory.go
```diff
@@ -26,7 +34,61 @@ func (ec *ExecutionContext) validateDirectory() error {
 		ec.ExecutionDirectory = cwd
 	}
 
-	ed, err := os.Stat(ec.ExecutionDirectory)
+	err := validateExecutionDirectory(ec.ExecutionDirectory)
+	if err != nil {
+		return err
+	}
+	ec.ExecutionDirectory, _ = filepath.Abs(ec.ExecutionDirectory)
```
scriptonist, line 41: Maybe handle the error?

scriptonist:
While we are at it, I was thinking if It'll be a good idea to move all config related things to a separate package, I've vaguely tried this out here (POC -- will not build)
https://github.com/scriptonist/graphql-engine/tree/config-pkg/cli/internal/config

We can be done with all config related logic in the root command: 
https://github.com/scriptonist/graphql-engine/blob/d465a9b0f7ad7cb8b9ea61fec8519ab7dff78325/cli/commands/root.go#L50

I think, doing so can nicely isolate config setup and makes it a bit more testable. But it requires a fair bit of exploration and validation. 

@ShraddhaAg can you think about this and if you find it worthy, spend some time validating it?

cli/cli.go
ShraddhaAg: Ah yes, this makes so much more sense. Updated :+1:

cli/directory.go
ShraddhaAg: Done

cli/directory.go
ShraddhaAg: Done

Stefmore02:
Closing this PR as is no longer relevant.

PR #4773, opened 2020-05-14
console: add note to request header

ManUtopiK:
### Description

When testing from many JWTs, it's hard to remember which token own which role, or which user...

hasura-request-header-before

Also, if we use the same console for many endpoints, we can set many `x-hasura-admin-secret`, but how to know which secret is for?

This pull request add one column "Note". It's allow user to add side note for each row.
This is only interface improvement, requests are not impacted.

### Changelog

### Affected components

- [x] Console

### Solution and Design

It's just an idea of design. Maybe there's something more elegant.
Key column width was reduced to give space for note column (keys generally aren't too long).

hasura-request-header-after

Maybe we should add an tooltip or change column name "Note" by "Note (only for you, no effect on header request)" to avoid mistake... 

### Steps to test and verify

Git merge and `npm run dev`.

### Catalog upgrade

- [x] No

### Metadata

- [x] No

### GraphQL

### Breaking changes
- [x] No Breaking changes

### Suggestions
File `ApiRequest.js` should be splited...


beerose:
Thanks for the PR @ManUtopiK! This is a really nice idea!

Regarding the design — `Note` column could be collapsible so that people who don't want to use this feature, can easily collapse this column and don't be bothered by it. Let me know what you think about it.

beerose:
Closing this PR due to lack of activity

PR #4775, opened 2020-05-15
console: allow entering big int values in the console

### LINKED ISSUE #3667: Cannot view or set bigint max value in console
I have a row with a bigint value of 9223372036854775807 ("`max_bigint`"). When I view the row in the console, the value displays as 9223372036854776000. When I try use the console to set the value to `max_bigint`, I get the following error: 

>Edit failed!
> Int64 is either floating or will cause over or underflow: 9.223372036854776e18

soorajshankar:
* convert bigint to string before JSON parse
* prevent auto round off and shows accurate value

### Description

v8 has auto round off to nearest safe value, this will appear as a bug in the UI. 
by converting the big int values to string, UI will render the values correctly.

### Changelog

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

close #3667 

### Solution and Design

convert bigint to string before JSON parse.

### Steps to test and verify

1. add a table in the UI with column with bigint type.
2. go to graphiql and insert a bigint value not ending with zeros (with zeros we cannot test the round of scenario)
3. go-to data table and verify the big int value is displayed correctly.

* Considering the solution involves a conditional type change, any arithmetic operations without pre-type check might get affected. 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


CHANGELOG.md
```diff
@@ -49,6 +49,8 @@ Read more about the session argument for computed fields in the [docs](https://h
 ### Bug fixes and improvements
 
 (Add entries here in the order of: server, console, cli, docs, others)
+- console: display big int values in the console (close #3667, #3697, #4548)
```
tirumaraiselvan, line 52: could you add this with the rest of the console entries here?

soorajshankar: Done!

soorajshankar:
I found one related issue here, making this draft PR

weyert:
Wouldn't you be able to use `reviver`-parameter of `JSON.parse` to make this transformation? 
See for example: https://usefulangle.com/post/112/json-parse-reviver-function

soorajshankar:
@weyert Thanks for the note!
I had tried that as well, actually, when we try using reviver function, js engine already parse and round off the value to a nearest safe integer. 
ie. the value argument which we are getting is not the same which we are getting from the server. if the value was a string, we could have parsed it into a bigInt or a string.

this is what js will do :)
image

console/src/components/Services/Data/TableInsertItem/InsertActions.js
```diff
@@ -40,8 +40,7 @@ const insertItem = (tableName, colValues) => {
       const colSchema = columns.find(x => x.column_name === colName);
       const colType = colSchema.data_type;
       if (Integers.indexOf(colType) > 0) {
-        insertObject[colName] =
-          parseInt(colValues[colName], 10) || colValues[colName];
```
soorajshankar, line 44: The flag `HASURA_GRAPHQL_STRINGIFY_NUMERIC_TYPES` will take care of the response but while sending we may need to avoid this parse.

soorajshankar:
@wawhal considering we have the server-side number stringify solution, removed response parser.

console/src/components/Services/Data/TableInsertItem/InsertActions.js
```diff
@@ -40,8 +40,7 @@ const insertItem = (tableName, colValues) => {
       const colSchema = columns.find(x => x.column_name === colName);
       const colType = colSchema.data_type;
       if (Integers.indexOf(colType) > 0) {
```
wawhal, line 42: Can remove this check entirely because this is what happens in the final `else`.

soorajshankar: Done

beerose:
@soorajshankar Can you check the failing test?

PR #4790, opened 2020-05-17
console: down migrations improvements (close #3503, #4988)

### LINKED ISSUE #3503: down migration of a renamed column fails
when a column is renamed,  the generated down migration tries to set the comment of the column with the old name before renaming it back.

ie. a column is renamed from "iconkey" to "iconKey" with camelcase

the generated down migration looks like this:

```
- args:
    sql: COMMENT ON COLUMN "public"."menu"."iconkey" IS E'null'
  type: run_sql
- args:
    sql: alter table "public"."menu" rename column "iconKey" to "iconkey";
  type: run_sql
```

and therefore fails.

after manually reversing the order it works

soorajshankar:
### Description

Down migration of a renamed column fails because of the order of down migration was not reversed.

### Changelog

### Affected components

- [x] Console

### Related Issues

close #3503, #4988

### Solution and Design

`schemaChangesDown` and `schemaChangesDown` were getting stacked on the same order. I have reversed `schemaChangesDown` just before calling `makeMigrationCall`.
Now even if the column is renamed, the down migration works correctly.

This PR will also, add SQL comments in down schema when there is no down schema changes 
```
-- Could not auto-generate a down migration.
-- Please write an appropriate down migration for the SQL below:
--   DROP TABLE orders;
```

* Refactored some of the code to give better code readability 

### Steps to test and verify

1. Go to the Hasura Data module.
2. Do multiple edits on a column (eg. try renaming it, changing the comment, and toggle nullable)
3. Click on save button.
4. Validate the order of the up and down migrations from the network tab.


soorajshankar:
Previous behavior.
image
1. Here the last up migration command is setting column name to `ttt` 
2. but first, down migration is trying to `SET NOT NULL` on the table with name test. (at that time revert has not completed the rename yet. 


** note the image shows up query at the bottom :)

rikinsk:
@soorajshankar A gentle reminder on the previous comment. it would be nice if this could be made consistent across the codebase.

I also noticed an issue while squashing migrations that if a migration does not have an equivalent down migration, that information gets lost once the migrations are squashed. So can we ensure that we add an sql comment `-- no down migration created for <migration name>` in case there are no down migrations passed in any case. (This happens in cases such as a drop table as it is very hard to create the equivalent create table statement required in the down migration due to the many dependent objects a table can have)

beerose:
> Also can we standardize the pattern of ensuring the down migrations statements order across all usages. I like this style of adding up and down migration statements in the same order and then reversing the order of the down migration statements at the end 

I think that all migration-related code can be extracted to its own module and maybe even unit tested. This way it'll be easier to enforce some behaviors and consistency. Also, some repeated parts like calling `getRunSqlQuery` could be _hidden_ inside of this module. For example, we could have smaller functions that are returning `{ upMigration, downMigration }` objects. 

@soorajshankar We can discuss the details and brainstorm some ideas on how to approach this.

console/src/utils/MigrationHelper.ts
```diff
@@ -0,0 +1,24 @@
+// todo use well typed interface after generating
+export interface RunSQLQueryType {
+  type: string;
+  args: object | [];
+}
+
+export default class MigrationHelper {
```
soorajshankar, line 7: @rikinsk @beerose please have a look at this approach.

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -1889,12 +1505,12 @@ const saveColumnChangesSql = (colName, column, onSuccess) => {
     };
     const customOnError = () => {};
 
-    if (schemaChangesUp.length > 0) {
+    if (upMigration.length > 0 || downMigration.length > 0) {
```
beerose, line 1508: We don't want to call this if `upMigration` is empty and `downMigration` is not. `upMigration` needs to be there.

console/src/index.d.ts
```diff
@@ -0,0 +1,10 @@
+// HGE common types
```
beerose, line 1: `.d.ts` files are for declarations for JavaScript modules. Please move this type to `types.ts` in the same file. 
Note for the feature: we should provide exhaustive typings for RQL queries and specify that `type` is a union of given string literals.

console/src/utils/migration/Migration.ts
```diff
@@ -0,0 +1,24 @@
+// todo use well typed interface after generating
+export interface RunSQLQueryType {
+  type: string;
+  args: object | [];
```
beerose, line 4:
```suggestion
  args: object | RunSQLQueryType[];
```

console/src/utils/migration/utils.ts
```diff
@@ -0,0 +1,283 @@
...
+  newName: newColumn.name.trim(),
+  currentSchema: newColumn.schemaName,
+  customFieldName: (newColumn.customFieldName || '').trim(),
+});
+
+const parseOldColumns = (oldColumn: OldColumnType) => ({
+  originalColType: oldColumn.udt_name, // "value"
+  originalData_type: oldColumn.data_type, // "value"
+  originalColDefault: oldColumn.column_default || '', // null or "value"
+  originalColComment: oldColumn.comment || '', // null or "value"
```
beerose, line 66: Are these comments relevant? Ones for `comment` and `column_default`

console/src/utils/migration/utils.ts
```diff
@@ -0,0 +1,283 @@
...
+  originalColNullable: oldColumn.is_nullable, // "YES" or "NO"
+});
+
+// utility to compare old & new columns and generate up &down migrations
+export const getColumnUpdateMigration = (
+  oldColumn: OldColumnType,
+  newColumn: NewColumnType,
+  allSchemas: [],
```
beerose, line 74: We should never use array type this way, because it means literally an empty array. Check out this: https://www.typescriptlang.org/play/#code/DYUwLgBAHgXBDaBdA3AKFVCBeBBGANAEz4DMiQA

Schedules Triggers PR introduces proper typings for it, for now, you can use `unknown[]` or `object[]`.

console/src/utils/migration/utils.ts
```diff
@@ -0,0 +1,283 @@
...
+  }
+  const migrationName = `alter_table_${currentSchema}_${tableName}_alter_column_${colName}`;
+  return {
+    migrationName,
+    upMigration: migration.upMigration,
+    downMigration: migration.downMigration,
+  };
+};
+export const getDownQueryComment = (upqueries: RunSqlType[]) => {
+  if (upqueries instanceof Array && upqueries.length >= 0) {
```
beerose, line 271: You provided typing for it being an array so either this check is redundant or type should say that it's an array or something else. 

If the second option, then let's use `Array.isArray`.

console/src/utils/migration/utils.ts
soorajshankar: I am removing those comments since we are on ts.

soorajshankar:
> @soorajshankar This looks good.
> 
> Can we use this PR to fix all such issues of incorrect down migrations statement ordering. There definitely could be more which havent been reported.
> 
> Also can we standardize the pattern of ensuring the down migrations statements order across all usages. I like this style of adding up and down migration statements in the same order and then reversing the order of the down migration statements at the end cc: @beerose

Had a discussion with @beerose, there are 53 places we use this function. If I modify everywhere this is gonna be a very huge PR. We will take this as a separate issue and PR.

soorajshankar:
Back to in progress, fixing #4988 along with this

webdeb:
close #4868

console/src/utils/migration/Migration.ts
```diff
@@ -0,0 +1,38 @@
...
+        getDownQueryComments([up as RunSqlType])[0], // reusing the method which works with array
+        ...this.downMigration,
+      ];
+  };
+
+  // this is called when there is only one migration for at that particular index
+  UNSAFE_add = (up?: RunSQLQueryType, down?: RunSQLQueryType) => {
+    if (up !== undefined && up !== null)
+      this.upMigration = [...this.upMigration, up];
+    if (down !== undefined && up !== null)
```
beerose, line 35: Did you mean
```suggestion
    if (down !== undefined && down !== null)
```
?

console/src/components/Services/Actions/ServerIO.js
```diff
@@ -218,8 +222,8 @@ export const createAction = () => (dispatch, getState) => {
   makeMigrationCall(
     dispatch,
     getState,
-    upQueries,
-    downQueries,
+    migration.upMigration,
+    migration.downMigration,
```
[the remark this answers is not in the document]
soorajshankar: yeah I was thinking this

console/src/components/Services/Data/Function/customFunctionReducer.js
```diff
@@ -7,20 +7,13 @@ import Endpoints, { globalCookiePolicy } from '../../../../Endpoints';
...
-import { handleMigrationErrors } from '../../../../utils/migration';
-
-import { showSuccessNotification } from '../../Common/Notification';
-
 import { fetchTrackedFunctions } from '../DataActions';
 
 import _push from '../push';
 import { getSchemaBaseRoute } from '../../../Common/utils/routesUtils';
 import { getRunSqlQuery } from '../../../Common/utils/v1QueryUtils';
+import { makeRequest } from '../../RemoteSchema/Actions';
```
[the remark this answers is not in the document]
soorajshankar: Thinking of removing this function if possible in the next PR, this is also a migration call, we could merge them into a single place and consider it as a migration service.

can we make this change in the next PR?

console/src/utils/migration/utils.ts
```diff
@@ -151,11 +152,15 @@ export const getColumnUpdateMigration = (
...
+    isColTypeString(colType) &&
+    !isPostgresFunction(colDefault) &&
+    !isTypeCast(colDefault)
       ? `'${colDefault}'`
       : colDefault;
   const originalColDefaultWithQuotes =
-    isColTypeString(colType) && !isPostgresFunction(originalColDefault)
+    isColTypeString(colType) &&
+    !isPostgresFunction(originalColDefault) &&
+    !isTypeCast(colDefault)
```
beerose, line 163:
```suggestion
    !isTypeCast(originalColDefault)
```

CHANGELOG.md
```diff
@@ -169,6 +169,9 @@ hasura seed apply --file 1234_add_some_seed_data.sql
 - server: fix importing of allow list query from metadata (fix #4687)
 - server: flush log buffer during shutdown (#4800)
 - server: fix edge case with printing logs on startup failure (fix #4772)
+- console: fix down migration of a renamed column fails (close #3503) (#4790)
+- console: provide option to cascade metadata on dependency conflicts on console (fix #1593)
+- console: fix enum tables reload data button UI (#4647)
```
beerose, line 174:
```suggestion
```

soorajshankar:
> * Noticed an odd behaviour,
>   On creating a table via the console, the `down.sql` has the following contents, I believe the comment shouldnt be present here.
> 
> ```
> -- Could not auto-generate a down migration.
> -- Please write an appropriate down migration for the SQL below:
> -- undefined;
> DROP TABLE "public"."author";
> ```
> 
> * Also seems like while creating a migration on Raw SQL page, a `down.sql` with a comment does not get created

@rikinsk I have added a check which generates down migration comments only for the `run_sql` queries. 
Please have a look at the same.

rikinsk:
> Also seems like while creating a migration on Raw SQL page, a down.sql with a comment does not get created

@soorajshankar I still dont see a down migration with a comment created when running sql from raw sql page.

rikinsk (changes requested):
down migration with comment not created for sql run in raw sql page

rikinsk (approved):

PR #4797, opened 2020-05-18
use printErrJExit instead of logger

tirumaraiselvan:
### Description

Logger only writes to stdout after buffer is full. This means that some log messages might not be flushed if the system exists (especially during startup failures)

### Changelog

### Affected components

- [x] Server

### DIFF
--- a/server/src-lib/Hasura/Server/App.hs
+++ b/server/src-lib/Hasura/Server/App.hs
@@ -36,9 +36,11 @@ import qualified System.Metrics.Json                    as EKG
 import qualified Text.Mustache                          as M
 import qualified Web.Spock.Core                         as Spock
 
+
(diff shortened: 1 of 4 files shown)


server/src-lib/Hasura/Server/App.hs
```diff
@@ -532,12 +534,11 @@ mkWaiApp isoLevel logger sqlGenCtx enableAL pool ci httpManager mode corsCfg ena
 
       ((migrationResult, schemaCache), lastUpdateEvent) <-
         initialiseResult `onLeft` \err -> do
-          L.unLogger logger StartupLog
+          printErrJExit StartupLog
```
tirumaraiselvan, line 537: This only prints partial log because it needs to be wrapped in `EngineLog`

PR #4800, opened 2020-05-18
flush logger on shutdown

tirumaraiselvan:
### Description

The buffer needs to be flushed on termination:  https://hackage.haskell.org/package/fast-logger-3.0.1/docs/System-Log-FastLogger.html#v:flushLogStr

Currently, if you don't flush the logger, the following line (and possibly other buffered logs) are not printed:
`{"type":"startup","timestamp":"2020-05-18T09:48:02.636+0000","level":"info","detail":{"kind":"server","info":"gracefully shutting down server"}}
`

### Changelog

### Affected components

- [x] Server

### DIFF
--- a/server/src-lib/Hasura/App.hs
+++ b/server/src-lib/Hasura/App.hs
@@ -28,6 +28,7 @@ import qualified Database.PG.Query                    as Q
 import qualified Network.HTTP.Client                  as HTTP
 import qualified Network.HTTP.Client.TLS              as HTTP
 import qualified Network.Wai.Handler.Warp             as Warp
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/App.hs
```diff
@@ -353,7 +354,9 @@ runHGEServer ServeOptions{..} InitCtx{..} initTime = do
         closeSocket
         shutdownApp
         logShutdown
+        flushLogger
```
0x777, line 357: We can just call `rmLoggerSet` which will flush and close the `LoggerSet`.

PR #4801, opened 2020-05-18
flush log buffer on exception in mkWaiApp ( fix #4772 )

### LINKED ISSUE #4772: Not writing errors to docker log
Hi version >1.1 doesn't write errors to the docker log or atleast `{"path":"$","error":"pgcrypto extension is required, but the current user doesn’t have permission to create it. Please grant superuser permission, or setup the initial schema via https://docs.hasura.io/1.0/graphql/manual/deployment/postgres-permissions.html","code":"postgres-error"}` isn't written.

Just wasted two hours debugging docker network before trying to use an older version.

tirumaraiselvan:
### Description

Currently, we are calling `exitFailure` in mkWaiApp (introduced in #3901 ). This quits the program but doesn't flush the log buffer and hence we have issues like: https://github.com/hasura/graphql-engine/issues/4772

### Changelog

### Affected components

- [x] Server


server/src-lib/Hasura/App.hs
```diff
@@ -213,7 +216,8 @@ runHGEServer ServeOptions{..} InitCtx{..} initTime = do
 
   authMode <- either (printErrExit . T.unpack) return authModeRes
 
-  HasuraApp app cacheRef cacheInitTime shutdownApp <-
+  let flushLogger = liftIO $ FL.flushLogStr $ _lcLoggerSet loggerCtx
+  HasuraApp app cacheRef cacheInitTime shutdownApp <- flip onException flushLogger $
```
lexi-lambda, line 220: Please add a small comment that references issue #4772 so that it’s clear why this is needed! Otherwise this LGTM.

PR #4810, opened 2020-05-18
console: fix visiting view modify page overwriting raw sql content

### LINKED ISSUE #4798: raw sql page content gets overwritten when modify page for a view is visited
the view definition should be stored in the component state instead of raw sql page redux and it should be set to the Raw SQL page content when Modify button is clicked

soorajshankar:
### Description

Preserve data in local storage to ensure the user-provided data in raw SQL editor. data will preserve even after going to view modify tab.

### Changelog

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

close #4798

### Solution and Design

Since `data/sql` is used commonly a URL query string has been implemented. this helps to identify the container is mounted with `modify mode` or `run sql mode`.
using this identity, local storage value/redux value is used in the editor.

### Steps to test and verify

1. Go to `Run SQL` page and type some text.
2. go to any `view` under tables section.
3. click on `modify` tab.
2. go back run SQL page

## Current Behaviour
1. Run SQL page shows the data from the lastly viewed `view` under table.

## Expected Behaviour
1. Run SQL page should show the user entered data from the text editor.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 - server: fixes an issue where introspection queries with variables would fail because of caching (fix #4547)
 - server: avoid loss of precision when passing values in scientific notation (fix #4733)
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
+- console: Raw SQL preserve editor values (#4798)
```
tirumaraiselvan, line 57: This is a regression right? Maybe something like this:


```suggestion
- console: fix regression with Raw SQL editor preserving values (#4798)
```

soorajshankar:
@rikinsk 
**Previous Behaviour**
previous

**This PR Behaviour**
current

console/src/components/Services/Data/RawSQL/molecules/NotesSection.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+/**
+ * # NoteSection React static component
+ *
+ * @returns
+ */
+const NotesSection = () => {
```
beerose, line 8: We want new files to be created in TypeScript. Could you change it to `.tsx` file?

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -24,7 +24,30 @@ import {
...
+ * @param lastSuccess,
+ * @param isModalOpen,
+ * @param isCascadeChecked,
+ * @param isMigrationChecked,
+ * @param isTableTrackChecked,
+ * @param migrationMode,
+ * @param allSchemas,
+ * @param location,
+ * @returns
+ */
```
beerose, line 50:
```js
/**
 * @typedef Props
 * @property {<insert type>} sql
 * @property {<insert type>} resultType
 * @property {<insert type>} result
 * @property {<insert type>} resultHeaders
 * @property {<insert type>} dispatch
 * @property {<insert type>} ongoingRequest
 * @property {<insert type>} lastError
 * @property {<insert type>} lastSuccess
 * @property {<insert type>} isModalOpen
 * @property {<insert type>} isCascadeChecked
 * @property {<insert type>} isMigrationChecked
 * @property {<insert type>} isTableTrackChecked
 * @property {<insert type>} migrationMode
 * @property {<insert type>} allSchemas
 * @property {<insert type>} location
 *
 * @param {Props}
 */
```

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -500,26 +449,31 @@ const RawSQL = ({
...
+          {getMetadataCascadeSection()}
+          {getMigrationSection()}
 
-            {getRunButton()}
-          </div>
+          <RunButton onClick={submitSQL} className={styles.add_mar_top} />
         </div>
-        <div className="hidden col-xs-4">{alert}</div>
+        <Alert
+          {...{
```
beerose, line 470: Could we stick to a common react way of passing props?

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -529,6 +483,56 @@ const RawSQL = ({
   );
 };
 
+/**
+ * # Alert
+ * @component
+ *
+ * @param {*} { ongoingRequest, lastError, lastSuccess, className }
+ */
+const Alert = ({ ongoingRequest, lastError, lastSuccess, className }) => (
```
beerose, line 492: What about moving it to a separate typescript file?

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -529,6 +483,56 @@ const RawSQL = ({
...
+  </div>
+);
+/**
+ * #RunButton
+ * @component
+ *
+ * @param {*} { onClick, className }
+ * @returns
+ */
+const RunButton = ({ onClick, className }) => {
```
beerose, line 521: Is a separate component needed? There is no additional logic in `RunButton`, we can just use `Button` instead. JSX will be a couple of lines bigger, but there will be fewer components.

console/src/components/Services/Data/RawSQL/molecules/NotesSection.js
soorajshankar: Sure Ill change it

console/src/components/Services/Data/RawSQL/RawSQL.js
soorajshankar: Yeah will change, I thought this will keep the main render method clean and much readable.

beerose:
@soorajshankar Can you take a look at the failing test?

rikinsk:
@beerose 

> Don't store SQL string in local storage when it's derived from view definition or custom function.

This might be just as confusing as I would lose any edits I make to the edit view SQL on the Raw SQL page if I navigate. Typically I just expect the last content on the RawSQl page to persist so this doesnt seem ideal.

> Decouple modifying custom functions and views from the Raw SQL page and inline it on Modify pages.

This would be the best solution but is a little tricky because we let the user type in Raw SQL and that can allow them to do unexpected things. e.g. if they drop the current view, the current page we are on will become a 404. If we can list all the possible end states and handle them or ensure that the user can only modify the current view (make the "create or replace view <name> as " part of the statement fixed) this can be a good option. 
 
> (My favorite) Create a history of executed queries in the Raw SQL page in a similar way as it is in the GraphiQL section (with a clear all button). Then we wouldn't need to worry about possible overrides and it would also be an extremely useful feature.

This seems generally useful even beyond this issue and surprising no one has asked for this yet. Maybe we can create an issue and see the interest it gets and then decide when to pick it up.

beerose:
@rikinsk I propose to keep the scope of this PR to only fix https://github.com/hasura/graphql-engine/issues/4798, and for those two ideas (decoupling edit and query history) create issues and decide whether to do them or not later depending on the users' interest.

rikinsk:
> @rikinsk I propose to keep the scope of this PR to only fix #4798, and for those two ideas (decoupling edit and query history) create issues and decide whether to do them or not later depending on the users' interest.

Definitely. I think the current UX is good to go.

console/cypress/integration/data/raw-sql/spec.js
```diff
@@ -18,69 +18,67 @@ export const openRawSQL = () => {
   // Match URL
   cy.url().should('eq', `${baseUrl}/data/sql`);
 };
-
+const clearText = () => {
+  cy.get('textarea').type('{selectall}', { force: true });
```
beerose, line 22: 👍

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -96,9 +117,6 @@ const RawSQL = ({
   );
 
   const submitSQL = () => {
```
beerose, line 119: We need to bring this logic back. The last ran SQL query should be stored in local storage. Users can for example close or refresh the page and current SQL won't be saved after running the query.

soorajshankar: @beerose This functionality is already taken care [here](https://github.com/soorajshankar/graphql-engine/blob/12a1968908b733ce188191aea60773d88cc8d7bf/console/src/components/Services/Data/RawSQL/RawSQL.js#L86) with `useEffect` hook.
I removed only the extra implementation.

beerose: This won't work on page refresh. 
2020-05-27 at 11 45 08 - Beige Frog

soorajshankar: Ohh okay got it!

beerose: `useEffect` would only set this on unmounting the component.

soorajshankar: one more point, if the user has not clicked `Run!` and if he refreshes the page the data will be lost. 
I hope that should be okay.

rikinsk: I am not sure if I understand correctly, but it should not be necessary for a user to run a query for it to be persisted. Any content I have in my editor should be persisted so that I can come back and continue.

beerose: The current logic in master persists SQL in two cases:
1. The user clicks `run`.
2. The user leaves the component. 

It doesn't persist SQL if the user refreshes the page while being on Raw SQL page.

beerose: This PR would leave this part of functionality as it was.

console/src/components/Services/Data/RawSQL/molecules/NotesSection.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+
+/**
+ * # NoteSection React static component
+ *
+ * @returns
+ */
```
beerose, line 7: Let's remove this comment

soorajshankar: Done

console/src/components/Services/Data/RawSQL/RawSQL.js
```diff
@@ -552,6 +523,7 @@ const mapStateToProps = state => ({
   currentSchema: state.tables.currentSchema,
   allSchemas: state.tables.allSchemas,
   serverVersion: state.main.serverVersion ? state.main.serverVersion : '',
+  location: state.routing.locationBeforeTransitions,
```
beerose, line 526:
```suggestion
```

soorajshankar: Good catch! This was the first approach!

soorajshankar: Done

PR #4820, opened 2020-05-19
cli-ext: add typescript support

beerose:
Solves https://github.com/hasura/graphql-engine-internal/issues/395

### Description
This PR adds TypeScript support to `cli-ext/`.

### Changelog

### Affected components

- [x] CLI


cli-ext/src/services/actions-codegen/template.js
```diff
@@ -1,8 +1,3 @@
-const { buildSchema, printSchema, parse } = require('graphql');
```
wawhal, line 1: These removals should be undone because these dependencies need to be in scope when we eval a the codegen script.

PR #4838, opened 2020-05-21
1.2.1 fix and fix explain queries with role permissions (fix #4816)

### LINKED ISSUE #4816: Analyze button does not show permission filters anymore
Hi,

Before when using the Analyze button I could see the permissions applied to the query.  

For example in version 1.0.0-beta-8 (not sure when this changed, this version is just an old one I had laying around), the following permission:
`{"id":{"_eq":"x-hasura-user-id"}}`
Will generate the following query in the Analyze button:
```SQL
SELECT
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_1_e"
          FROM
            (
              SELECT
                "_0_root.base"."id" AS "id"
            ) AS "_1_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public"."image"
        WHERE
          (
            (
              ("public"."image"."id") = (('582605c8-c888-4a06-bd78-908be1e4f1ec') :: uuid)
            )
            OR (
              (("public"."image"."id") IS NULL)
              AND (
                (('582605c8-c888-4a06-bd78-908be1e4f1ec') :: uuid) IS NULL
              )
            )
          )
      ) AS "_0_root.base"
  ) AS "_2_root"
```

But in the latest version of Hasura (v1.2.1), the same permission will generate the following query in the Analyze button:
```SQL
SELECT
FROM
  (
    SELECT
      row_to_json(
        (
          SELECT
            "_1_e"
          FROM
            (
              SELECT
                "_0_root.base"."id" AS "id"
            ) AS "_1_e"
        )
      ) AS "root"
    FROM
      (
        SELECT
          *
        FROM
          "public"."image"
        WHERE
          ('true')
      ) AS "_0_root.base"
  ) AS "_2_root"
```
The query results will be corrrect, just the generated query that won't.

Was this intentional? If it was, is there a way to show the permissions in the Analyze button?  For more complicated queries and permissions I use the Analyze to try to optimize a few things.  But if the applied permissions are not there, it will be hard to change the permissions or the database structure to optimize my queries.

Thanks

rakeshkky:
**Note to Reviewers**:-
No need to review 1.2.1 fix which is already reviewed and released. Only review the fix related to #4816 bug which is in [this commit](https://github.com/hasura/graphql-engine/pull/4838/commits/b470089f4e222cee0ae84dcfd8b8594e029791e9). 

### Description

Include v1.2.1 release fix in master. Also fixes minor bug related to GraphQL explain queries with role permissions which is originally caused via v1.2.1 fix. 

### Changelog

### Affected components

- [x] Server
- [x] Console
- [x] CLI
- [x] Docs
- [x] Tests

### Related Issues

Fix #4816 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/tests-py/test_graphql_queries.py
+++ b/server/tests-py/test_graphql_queries.py
@@ -1,5 +1,5 @@
 import pytest
-from validate import check_query_f
+from validate import check_query_f, check_query, get_conf_f
 from context import PytestConf
 
 # Mark that all tests in this module can be run as server upgrade tests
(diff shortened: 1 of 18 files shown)


server/tests-py/test_graphql_queries.py
```diff
@@ -632,24 +632,19 @@ def dir(cls):
...
-            }
-        check_query(hge_ctx, conf, add_auth=False)
-
-    def skip_if_pg_9_5(self, hge_ctx):
-        pg_version = hge_ctx.pg_version
-        # skip tests for PG version < 9.6 since the plan response varies in the same
-        if pg_version < 90600:
-            pytest.skip('Skip explain query tests for Postgres version < 9.6')
+            headers['X-Hasura-Admin-Secret'] = hge_ctx.hge_key
+        status_code, resp_json, _ = hge_ctx.anyq(conf['url'], conf['query'], headers)
```
0x777, line 646: A comment would be good here, explaining why we are only testing the `sql` key and not the plan.

Issue #4818, opened 2020-05-19
console: wrap voyager in error boundary 


beerose:
It will be a partial solution to https://github.com/hasura/graphql-engine/issues/3023. The idea is to wrap the voyager component in error boundary so that any voyager's errors won't crash the whole console, and users still will be able to use the application without page refresh. 

In the error message, we should display information that the error comes from Voyager. 

Voyager component: https://github.com/hasura/graphql-engine/blob/master/console/src/components/Services/VoyagerView/VoyagerView.js

kolharsam:
@beerose I think I'd like to work on this. Is anyone else working on it?

I also had a couple of questions.

Also, what is the fallback for the ErrorBoundary? 
Are there other places where this ErrorBoundary can be used?

beerose:
@kolharsam I'll assign you to this issue, you can work on it 😉

We have error boundary defined for the whole app here: https://github.com/hasura/graphql-engine/blob/master/console/src/components/Error/ErrorBoundary.js
And the fallback is either `PageNotFound` or `RuntimeError` component. The idea here is to wrap the Voyager component into an additional error boundary so that the top menu is still visible. 

> Also, what is the fallback for the ErrorBoundary?

Fallback will be a similar component to the [`PageNotFound`](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Error/PageNotFound.js) or [`RuntimeError`](https://github.com/hasura/graphql-engine/blob/master/console/src/components/Error/RuntimeError.js) components (styling wise), but with the information that Voyager error occurred.

> Are there other places where this ErrorBoundary can be used?

No, we want to cover this specific case and wrap only the Voyager component. The whole application is already wrapped in the error boundary.

kolharsam:
@beerose I guess the error boundary is ready, I've tested it & seems to be working as expected. 
For the styling, however, I've followed the style present on the `RuntimeError` component

<img width="1439" alt="Screenshot 2020-05-27 at 9 47 49 PM" src="https://user-images.githubusercontent.com/6604943/83047347-df7eed80-a065-11ea-8908-b16287a0d06f.png">

NoriSte:
The Voyager is not linked anymore in the Console, I close this PR.

Issue #3023, opened 2019-10-05
Error on voyager view


edsonbonfim:
I got this message when I trying to open voyager view in my hasura app with heroku.

```
Error: abort(7). Build with -s ASSERTIONS=1 for more info.
    at Array.n.listeners.<computed> (https://graphql-engine-cdn.hasura.io/console/assets/channel/beta/v1.0/vendor.js.gz:1:3828515)
    at Worker.<anonymous> (https://graphql-engine-cdn.hasura.io/console/assets/channel/beta/v1.0/vendor.js.gz:1:3828346)
```

EdsonOnildoJR:
![image](https://user-images.githubusercontent.com/8020047/69923463-c95fee00-1483-11ea-9ef9-be5b070c7100.png)

This message is all that I have. I've deploy my app with heroku.

stefan1968:
I get the same issue.

mnlbox:
I have the same issue in big projects with many tables. For small projects every thing is working fine.

PeterDraex:
I have a similar issue. Also occurs on voyager view, but different row numbers.

```
Error: abort(7). Build with -s ASSERTIONS=1 for more info.
    at Array.n.listeners.<computed> (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/vendor.js.gz:1:3832520)
    at Worker.<anonymous> (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/vendor.js.gz:1:3832351)
```

I'm running Hasura in Docker, which is connected to PostreSQL outside of docker, which is connected to MySQL (via Foreign Data Wrappers). GraphQL queries are working fine.

DevSnap:
Same error on a project with 40+ tables.

```
Error: abort(7). Build with -s ASSERTIONS=1 for more info. 
at Array.n.listeners.<computed> (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/vendor.js.gz:1:3832520) 
at Worker.<anonymous> (https://graphql-engine-cdn.hasura.io/console/assets/channel/stable/v1.0/vendor.js.gz:1:3832351)
```

I tried to upload the introspection query result on [https://apis.guru/graphql-voyager](https://apis.guru/graphql-voyager) and it works (it takes 10 seconds to render).

tpurschke:
same issue here with hasura running in docker and 60 tables with very little data. 
Error:
```
[...]

PR #4856, opened 2020-05-23
[WIP] Resolve[#4818] : Error Boundary for VoyagerView component

kolharsam:
### Changelog

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3023, #4818 

### Solution and Design

The design is the same as the `RuntimeError` component, as suggested in #4818  The solution is a Error Boundary specifically for the `VoyagerView` component

### Steps to test and verify

- First make 40+ tables on the `GraphQL console`
- You should see the fallback mentioned in the `VoyagerErrBoundary` component if an error does occur. 

For the state `errorInfo` I've annotated it with the type of `any` which isn't a clean solution. This because of the `componentStack` property I require from the `errorInfo` object. I'd probably need some help to clean this up.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/console/src/components/Services/VoyagerView/VoyagerView.tsx
+++ b/console/src/components/Services/VoyagerView/VoyagerView.tsx
@@ -2,8 +2,8 @@ import React, { Component } from 'react';
 import { Connect } from 'react-redux';
 import { GraphQLVoyager } from 'graphql-voyager';
 import fetch from 'isomorphic-fetch';
-
 import Endpoints from '../../../Endpoints';
+import VoyagerErrBoundary from './../../Error/VoyagerErrBoundary';
 import '../../../../node_modules/graphql-voyager/dist/voyager.css';
 import './voyagerView.css';
 
@@ -35,10 +35,12 @@ class VoyagerView extends Component<Props, State> {
 
   render() {
     return (
-      <GraphQLVoyager
-        introspection={this.introspectionProvider}
-        workerURI="https://cdn.jsdelivr.net/npm/graphql-voyager@1.0.0-rc.27/dist/voyager.worker.min.js"
-      />
+      <VoyagerErrBoundary>
+        <GraphQLVoyager
+          introspection={this.introspectionProvider}
+          workerURI="https://cdn.jsdelivr.net/npm/graphql-voyager@1.0.0-rc.27/dist/voyager.worker.min.js"
+        />
+      </VoyagerErrBoundary>
     );
   }
 }
--- a/console/src/components/Error/VoyagerBoundary.js
+++ b/console/src/components/Error/VoyagerBoundary.js
@@ -1,0 +1,72 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import globals from '../../Globals';
+
+// #4818 Issue
+
+class VoyagerBoundary extends Component {
+  constructor() {
+    super();
+    this.state = {
+      error: false,
+      errorInfo: null,
+    };
+  }
+
+  componentDidCatch(error, errorInfo) {
+    this.setState({
+      error,
+      errorInfo,
+    });
+  }
+
+  render() {
+    const errorImage = `${globals.assetsPath}/common/img/hasura_icon_green.svg`;
+    const styles = require('./ErrorPage.scss');
+
+    if (this.state.errorInfo) {
+      return (
+        <div className={styles.viewContainer}>
+          <div className={'container ' + styles.centerContent}>
+            <div className={'row ' + styles.message}>
+              <div className="col-xs-8">
+                <h1>Error</h1>
+                <br />
+                <div>
+                  Something went wrong.
+                  <details>
+                    <summary>Error Summary</summary>
+                    {this.state.error && this.state.error.toString()}
+                    <br />
+                    <div>
+                      <pre className={styles.errorStack}>
+                        {this.state.errorInfo.componentStack}
+                      </pre>
+                    </div>
+                  </details>
+                </div>
+                <br />
+              </div>
+              <div className="col-xs-4">
+                <img
+                  src={errorImage}
+                  className="img-responsive"
+                  name="hasura"
+                  title="Something went wrong!"
+                />
+              </div>
+            </div>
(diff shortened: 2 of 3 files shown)


soorajshankar:
@kolharsam Please fix the conflicts

Also, can we make ErrorBoundaryComponent more generic?
maybe by passing the title and description as props and renaming the component with a generic name.
I see that currently, we use this only for voyager view but maybe we can use this somewhere else in the future

console/src/components/Services/VoyagerView/VoyagerView.tsx
```diff
@@ -42,12 +42,36 @@ class VoyagerView extends Component<Props, State> {
       body: JSON.stringify({ query }),
     });
 
+  loadWorker = async () => {
+    const url =
+      'https://cdn.jsdelivr.net/npm/graphql-voyager@1.0.0-rc.30/dist/voyager.worker.min.js';
```
soorajshankar, line 47: can we put hard string URLs in some constant file? may be endpoints.js

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+                  </pre>
+                </div>
+                <br />
+              </div>
+              <div className="col-xs-4">
+                <img
+                  src={errorImageSrc}
+                  className="img-responsive"
+                  title={this.props.title}
+                  alt=""
```
soorajshankar, line 55: Add some meaningful alt value

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { Component, ReactNode } from 'react';
+import globals from '../../../Globals';
+import styles from '../../Error/ErrorPage.scss';
+
+const errorImageSrc = `${globals.assetsPath}/common/img/hasura_icon_green.svg`;
+
+interface ErrorState {
+  error: Error;
+  hasError: boolean;
```
beerose, line 9: It's redundant. You can specify `error` ad `error: Error | null` or `error?: Error`

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+type Props = {
+  children: ReactNode;
```
beerose, line 13: You don't need to specify this in `Props`. It will be included as your component extends `React.Component`.

Btw. maybe we can convert it to functional component?

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+  title: string;
+  message: string;
+};
+
+class ComponentErrorBoundary extends Component<Props, ErrorState> {
+  state: ErrorState = {
+    hasError: false,
+    error: new Error(''),
```
beerose, line 21:
```suggestion
    error: null,
```

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  componentDidCatch(error: Error) {
+    this.setState({
+      hasError: true,
+      error,
+    });
+  }
+
+  render() {
+    if (this.state.hasError) {
```
beerose, line 32:
```suggestion
    if (this.state.error) {
```

console/src/components/Services/VoyagerView/VoyagerView.tsx
```diff
@@ -42,12 +42,35 @@ class VoyagerView extends Component<Props, State> {
       body: JSON.stringify({ query }),
     });
 
+  loadWorker = async () => {
+    const url = Endpoints.voyagerWorker;
+    const response = await fetch(url);
```
beerose, line 47: Shouldn't we handle the case when `jsdelivr` is down or we can't fetch the resource?

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { Component, ReactNode } from 'react';
```
beerose, line 1: Let's call this file `ErrorBoundary`. `Component` is redundant as it is, in fact, a component.

console/src/components/Services/Common/ComponentErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+  hasError: boolean;
+}
+
+type Props = {
+  children: ReactNode;
+  title: string;
+  message: string;
+};
+
+class ComponentErrorBoundary extends Component<Props, ErrorState> {
```
beerose, line 18:
```suggestion
class ErrorBoundary extends Component<Props, ErrorState> {
```

console/src/components/Services/VoyagerView/VoyagerView.tsx
kolharsam: I'm not sure, of how we can handle it. Should we have alternate addresses for fetching the worker?

console/src/components/Services/Common/ComponentErrorBoundary.tsx
kolharsam: we need it to be a class component for the `componentDidCatch` lifecycle method.

Issue #4256, opened 2020-03-31
Delete button in History list


eaubin:
It would be helpful to have a delete or remove button in the history list of the api explorer to clear away old queries.

gnyani:
It looks like adding a clear button to graphiQL is still in progress.
https://github.com/graphql/graphiql/issues/1388

thiscompsciguy:
any updates. I get its not top priority, but would a be a very nice to have.

sauzzon:
I started working on this issue.

criveratro:
Any news? Really looking forward to this.

PR #4652, opened 2020-05-03
console: add code exporter to graphiql (close #4531)

ShahAnuj2610:
fixes #4531 

### Description
- user can export `query` using code-exporter

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [x] Console

### Solution and Design
- use https://www.npmjs.com/package/graphiql-code-exporter

### Steps to test and verify
- tested locally

### Limitations, known bugs & workarounds
for now have used inbuilt snippets that supports only nodejs and react-apollo.
- [x] NodeJS fetch
- [x] React Apollo
- [x] Typescript

PR #4867, opened 2020-05-24
console: add clear history button to graphiql

gnyani:
### Description

Added a clear history button to graphiql to clear the history of old queries.

### Affected components

- [X] Console
### Related Issues

#4256


gnyani:
@rikinsk  any progress on this?

rikinsk:
@gnyani Thanks for raising this PR. It's quite a useful feature! 

I am just a little concerned about the increasing number of buttons on the GraphiQL pane. We will soon be having a Code exporter button added as well via [this PR](https://github.com/hasura/graphql-engine/pull/4652). 

Ideally it would have been amazing to have the clear button be added into the History panel itself. But I am not sure how feasible that is to do. Maybe you could take a quick look and let us know. Or else we'll go ahead with this itself.

gnyani:
@rikinsk  unfortunately not, we have almost zero control to tweak the History pane of the GraphiQL wrapper. I understand your concern and I agree that this should be fixed. Also, there are many changes happening at the GraphiQL editor side as well. Probably we might be able to fix then or better we have to make a new module for ourselves on top of GraphiQL but I am pretty sure its a big task.  

What do you think about this @beerose ?

rikinsk:
Thanks @gnyani. I suppose we can leave this as is and figure out a solution for the number of buttons separately. We'll probably merge this after the code exporter PR gets merged.

gnyani:
@rikinsk @beerose can we make a decision on this? Now that code generator is merged: https://github.com/hasura/graphql-engine/pull/4652

gnyani:
@martin-hasura  does this PR still makes sense? If so, can we do some progress on this and get this out?

gnyani:
closing this out as there was no response and its stale now.

Issue #4736, opened 2020-05-11
Improvements to Hasura security testing


0x777:
1. Flesh out the E2E test suite with more negative tests for authentication
2. Refactor the auth subsystem to make it more self-contained and harder to misuse within the codebase.
3. Add property-based/fuzz testing for the auth subsystem
4. Engaging with the Hasura & security community via a bug-bounty program

These are the immediate action items. As we make more progress and flesh these issues out in more detail, we will be linking to appropriate issues.

hugbubby:
All good steps. Bug bounties are the most important signal to me, as it ensures the Hasura team has a financial stake in keeping the app secure and is willing to bet money on its procedures.

PR #4919, opened 2020-05-28
#4736 Improvements to auth testing

jberryman:
### Description
One minor user-facing change. The second commit is just a refactor for ease of unit testing (and as a consequence, better code organization hopefully)

### Changelog

### Affected components

- [x] Server
- [x] Docs
- [x] Community Content
- [x] Tests

### Related Issues
#4736  ... it's not clear if this is sufficient to close that issue

### Solution and Design

Simple factoring out of tricky bits to make unit tests of core auth logic possible. This would also be the first step towards more clever solutions like e.g. having a JWT transformer/effect, but I don't think that gives us much.

### Steps to test and verify
- Please read through test spec carefully, make sure actual behavior is desired
- make note of change mentioned in changelog 
- look closely at big block moves (with poor diff; sorry about that), to make sure they are in fact trivial refactors

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


jberryman:
@0x777 @rakeshkky bump for review

server/src-test/Hasura/Server/AuthSpec.hs
```diff
@@ -0,0 +1,408 @@
...
+    -- 'userRoleHeader').
+    --
+    -- This would harden a bit against bugs, makes the spec simpler, but
+    -- especially is better UX since in the current behavior the user can't be
+    -- sure which role their query is operating as (and in the worst case we
+    -- might e.g. delete rows the user didn't intend)
+    --
+    -- But this is a breaking change we need to think a little more about;
+    -- users might be relying on the behavior, e.g. just hardcoding a dev role
+    -- into clients.
```
abooij, line 320: (I think my question of "default" versus "unauthenticated" role applies to this discussion as well: there are a couple of related terms here which would be nice to have clarified in a central place.)

server/src-test/Hasura/Server/AuthSpec.hs
```diff
@@ -0,0 +1,408 @@
...
+  -- These are all various error cases, except for the AMNoAuth mode:
+  it "with no admin secret provided" $ do
+    setupAuthMode' Nothing       Nothing             Nothing              Nothing
+      `shouldReturn` (Right AMNoAuth)
+    -- We insist on an admin secret in order to use webhook or JWT auth:
+    setupAuthMode' Nothing       Nothing             (Just fakeJWTConfig) Nothing
+      `shouldReturn` Left ()
+    setupAuthMode' Nothing       (Just fakeAuthHook) Nothing              Nothing
+      `shouldReturn` Left ()
+    -- ...and we can't have both:
```
abooij, line 338: I just want to add here that I think the comments here are really useful, especially since it's often difficult to see if the output should be `Left` or `Right`. I can't say I'm convinced that all of these are right, but that is due to my brain not being able to parse such a long list of outputs. But at least this gives a clear reference that we can check against, together with motivations for each test.

server/src-test/Hasura/Server/AuthSpec.hs
abooij: (also, again, just to check: do you want to solve this FIXME?)

server/src-lib/Hasura/GraphQL/Explain.hs
```diff
@@ -103,6 +104,8 @@ explainField userInfo gCtx sqlGenCtx actionExecuter fld =
       resolvedAST <- RS.traverseQueryRootFldAST (resolveVal userInfo) unresolvedAST
       let (query, remoteJoins) = RS.toPGQuery resolvedAST
           txtSQL = Q.getQueryText query
+          -- CAREFUL!: an `EXPLAIN ANALYZE` here would actually *execute* this query. If `txtSQL` 
+          -- somehow contained a mutation this would result in privilege escalation:
```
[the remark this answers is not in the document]
jberryman: Hm you might be right, but did you have a specific issue in mind? Like information leaking to client somehow?

server/src-lib/Hasura/Server/Auth.hs
```diff
@@ -48,39 +54,75 @@ class (Monad m) => UserAuthentication m where
...
+--
+-- Although this exists only in memory we store only a hash of the admin secret
+-- primarily in order to:
+--     
+--     - prevent theoretical timing attacks from a naive `==` 
+--     - prevent misuse or inadvertent leaking of the secret
+--
+-- NOTE: if we could scrub memory of admin secret (from argv and envp) somehow,
+-- this would additionally harden against attacks that could read arbitrary
+-- memory, so long as the secret was strong. I'm not sure that's attainable.
```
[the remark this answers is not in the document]
jberryman: > This comment suggests that we are trying to protect the system against a threat model that we will never be able to protect agains

Good point!

server/src-test/Hasura/Server/AuthSpec.hs
jberryman: Added these references in comments:

-- [1]: https://hasura.io/docs/1.0/graphql/manual/auth/authentication/unauthenticated-access.html 
-- [2]: https://hasura.io/docs/1.0/graphql/manual/auth/authentication/jwt.html#the-spec

We have to punt on the FIXME for now. Requires more coordination. Vamshi is aware

jberryman:
Thanks for the review @abooij ! Addressed comments in latest

server/src-lib/Hasura/GraphQL/Explain.hs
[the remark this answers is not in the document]
abooij: Right, there may information leaking through `EXPLAIN ANALYZE` for `SELECT`s.

PR #4928, opened 2020-05-29
unlock scheduled events on graceful shutdown

codingkarthik:
- When shutting down gracefully, unlock the locked scheduled events

### Description

The graphql-engine locks the scheduled events when it attempts to process it. Earlier, if a shutdown process was initiated the locked scheduled events remain locked forever. This PR, will unlock the locked events when graceful shutdown is initiated. 

If the graphql-engine has crashed, then the scheduled events will remain locked. To solve that, on startup, we unlock all the locked scheduled events, so that they can be processed.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

Maintain a set of the scheduled event ID's in the cache that have been fetched from the database and are in-line to be processed and when the graceful-shutdown is initiated, then use the ID's to update the status of the locked scheduled events.

### Steps to test and verify

Create lots of scheduled events with the same `schedule_at` and when the graphql-engine is processing them, initiate a graceful-shutdown. Then, check the status of the scheduled events, none of the scheduled events should have status as 'locked'.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -404,18 +459,33 @@ processStandAloneEvents logger logEnv httpMgr pgpool = do
   where
     logInternalError err = L.unLogger logger $ ScheduledTriggerInternalErr err
 
+    saveLockedStandAloneEvents :: [StandAloneScheduledEvent] -> IO ()
+    saveLockedStandAloneEvents standAloneEvents =
+      liftIO $ atomically $ do
```
lexi-lambda, line 464: No need to use `liftIO` here if you’re just in plain old `IO`—`liftIO` is only necessary if you are in some other monad that wraps `IO`.

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -404,18 +459,33 @@ processStandAloneEvents logger logEnv httpMgr pgpool = do
...
+        lockedEvents <- readTVar lockedStandAloneEvents
+        let ids = map saseId standAloneEvents
+        writeTVar lockedStandAloneEvents $!
+          Set.union lockedEvents $ Set.fromList ids
+
+    removeEventFromLockedEvents :: ScheduledEventId -> IO ()
+    removeEventFromLockedEvents eventId =
+      liftIO $ atomically $ do
```
lexi-lambda, line 472: You can drop `liftIO` here as well.

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -345,24 +379,44 @@ processCronEvents logger logEnv httpMgr pgpool getSC = do
...
+        lockedEvents <- readTVar lockedCronEvents
+        let ids = map cepId partialCronEvents
+        writeTVar lockedCronEvents $!
+          Set.union lockedEvents $ Set.fromList ids
+
+    removeEventFromLockedEvents :: CronEventId -> IO ()
+    removeEventFromLockedEvents eventId =
+      liftIO $ atomically $ do
+        lockedEvents <- readTVar lockedCronEvents
+        writeTVar lockedCronEvents $! Set.delete eventId lockedEvents
```
lexi-lambda, line 400: There’s quite a lot of duplicate logic between `processEventQueue`, `processCronEvents`, and `processStandAloneEvents`. Would it be possible to share more of that logic? It seems like maybe there’s a potential “event processor” abstraction here that could simplify a lot of this code!

server/src-lib/Hasura/App.hs
```diff
@@ -396,6 +406,31 @@ runHGEServer ServeOptions{..} InitCtx{..} pgExecCtx initTime = do
...
+      lockedCronEvents <- readTVarIO lseCronEvents
+      lockedStandaloneEvents <- readTVarIO lseStandAloneEvents
+      liftIO $ do
+        when (not $ Set.null $ lockedCronEvents) $ do
+          res <- runTx pool (Q.ReadCommitted, Nothing) (unlockCronEvents $ toList lockedCronEvents)
+          case res of
+            Left err -> logger $ mkGenericStrLog
+                         LevelWarn "scheduled_triggers" ("Error in unlocking the cron events " ++ (show err))
+            Right count -> logger $ mkGenericStrLog
+                            LevelInfo "scheduled_triggers" ((show count) ++ " cron events were unlocked")
```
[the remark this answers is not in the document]
codingkarthik: Thanks for the suggestion, I have implemented this in the next commit!

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
codingkarthik: Yes, I have made a new module, which has common functions to share the locking/unlocking events logic

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@
 - server: few relay fixes (fix #5020, #5037, #5046) (#5013)
 - server: raise error on startup when `--unauthorized-role` is ignored (#4736)
 - server: fix bug which arises when renaming/dropping a column on a remote relationship (#5005, #5119)
+- server: unlock locked scheduled events on graceful shutdown
```
tirumaraiselvan, line 36: I think you should remove this changelog. It is not user-facing.

lexi-lambda: I think it’s worth keeping, for what it’s worth.

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -345,24 +363,44 @@ processCronEvents logger logEnv httpMgr pgpool getSC = do
...
+    --     lockedEvents <- readTVar lockedCronEvents
+    --     let ids = map cepId partialCronEvents
+    --     writeTVar lockedCronEvents $!
+    --       Set.union lockedEvents $ Set.fromList ids
+
+    -- removeEventFromLockedEvents :: CronEventId -> IO ()
+    -- removeEventFromLockedEvents eventId =
+    --   atomically $ do
+    --     lockedEvents <- readTVar lockedCronEvents
+    --     writeTVar lockedCronEvents $! Set.delete eventId lockedEvents
```
lexi-lambda, line 384: Please remove this commented-out code!

codingkarthik: Oh, I missed that. I'm sorry 😓

CHANGELOG.md
tirumaraiselvan: @codingkarthik Can you move it under `Next release` ?

PR #4932, opened 2020-05-29
remove event from the saved locked events after processing it

codingkarthik:
### Description

In #4214, we introduced unlocking of events when graceful shutdown is initiated. This is done by maintaining an in-memory set of locked events. However, after an event has been processed, it's not being removed from the set of locked events. This PR fixes that problem.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Eventing/EventTrigger.hs
```diff
@@ -189,6 +189,11 @@ processEventQueue logger logenv httpMgr pool getSchemaCache eeCtx@EventEngineCtx
         let newLockedEvents = Set.union lockedEvents (Set.fromList evtsIds)
         writeTVar _eeCtxLockedEvents newLockedEvents
 
+    removeEventFromLockedEvents :: EventId -> IO ()
+    removeEventFromLockedEvents eventId = do
+      liftIO $ atomically $ do
+        lockedEvents <- readTVar _eeCtxLockedEvents
+        writeTVar _eeCtxLockedEvents $ Set.delete eventId lockedEvents
```
jberryman, line 196: It's probably fine, but can you add a strictness annotation here?:

```
writeTVar _eeCtxLockedEvents $! Set.delete eventId lockedEvents
                                                    ^
```
I'm always a little worried about writing thunks to mutable variables

codingkarthik: Oh yes..I didn't realize that I'm writing a thunk to a mutable variable. I will also change it where we are saving the locked events.

PR #4948, opened 2020-05-31
docs: console / cli /  api workflows (close #3593)

### LINKED ISSUE #3593: add api / cli / console workflows for all use cases in docs
Use tabs: Console / API / CLI

Sample content: 
- Console: 
  Head to < path > on the console and click on < btn >: 
  < screeshot >
- API
  < functionality > can be achieved using the < api-ref-link > metadata API
- CLI 
  < functionality > can be achieved via the < command > cli command. Run < command > -h for more details

marionschleifer:
### Description

Add tabs for console / CLI / API workflows for all reference documentation.

### Changelog

### Affected components

- [x] Docs

### Related Issues

https://github.com/hasura/graphql-engine/issues/3593


marionschleifer:
@sezgi I'll update the `down.sql` files here first and I'll add the `hasura metadata reload` command to the CLI tabs. Also I solve the conflicts. So moving this back to "In progress" 🙂

sezgi:
### Part 2/2

Continuing with `/schema/` files:

`/schema/basics.rst` 
- Suggestion: Skitch has a blur effect that looks nicer than the scratch effect on images (and it has other effects like arrows, borders, etc).
- Otherwise LGTM

`/schema/computed-fields.rst` 
- Maybe mention that if you're not using an existing function, you need to create a new function first.
- Confusing user flow:
1. `/schema/custom-functions.html#create-sql-functions` says to check `Track this`, but there are two other checkboxes it doesn't mention: `Cascade metadata` and `This is a migration`. Do these need to be checked?
2. In `/schema/computed-fields.html#scalar-computed-fields`, when I try to create this function, I get an error: `the function does not return a \"COMPOSITE\" type\n  • the function does not return a SETOF\n  • the function does not return a SETOF table`.
3. When I uncheck `Track this` and `Cascade metadata` and try again, it seems to work.
4. Where do I find my function now, so I can delete/edit it? It's not listed under `Tables/Views/Functions`. I suppose #3 didn't really work, since not tracking/cascading doesn't make sense.
5. Found it under `Non trackable functions` in `Data` tab.
- I was able to verify the workflows for adding the field to a table, but not sure if I followed the expected user flow to create the function.

`/schema/custom-field-names.rst`

- Needs more context in the intro. Something like, "Hasura auto-generates GrapqhQL field names based on your database column names. If you'd like to change the default..." etc.
- `Console` tabs need more text instructions.
- Otherwise LGTM

`/schema/custom-functions.rst`
- See above for comments on `Console` tab.
- `CLI` tab: what should the `down.sql` file say? Also missing `4. hasura metadata apply`? Getting an error on that. The function shows up under `Non trackable functions` in `Data` tab. Similar to the issue I was having above.
- `API` tab: Same error as above on `track_function` call: `cannot be tracked for the following reasons:\n  • the function does not return a \"COMPOSITE\" type\n  • the function does not return a SETOF\n  • the function does not return a SETOF table\n"`

`/schema/data-validations.rst`
- `Using Postgres triggers`
-- `CLI` tab: what should the `down.sql` file say?
-- Couldn't verify `API` tab. My Postman had issues with the double quotes on `article` in `BEFORE INSERT OR UPDATE ON "article"`. Tried backslash, single quote, or removing quotes -- didn't work.
- `Using Hasura Actions`
-- `CLI` tab: Missing `hasura metadata apply`? Also, how do I set the handler and other settings?
-- `API` tab: The action definition is different from the one from the other tabs, so doesn't work with the same handler (missing `author:`). Would be good to be consistent to avoid confusion.
- Otherwise LGTM

`/schema/default-values/column-presets.rst`
- `API` tab: should say `user_id` instead of `id`
- Otherwise LGTM

`/schema/default-values/postgres-defaults.rst`
- `Console` tab needs more text.
- Otherwise LGTM

`/schema/default-values/sql-functions.rst`
- `CLI` tab: What should the `down.sql` say?
- Otherwise LGTM

`/schema/enums.rst`
- We need to track the enum table after creating it. Otherwise it doesn't show up in the console.
- `Using an enum table`: Not clear how to set this up. Is there supposed to be a `role` column to begin with? If so , what type?
- Skipping this file for @rikinsk to review.

`/schema/table-relationships/create.html`
- `Add foreign-key constraint`
-- `CLI` tab: In `down.sql`, getting a syntax error with `(article_author_id_fkey)` but not with `article_author_id_fkey`
- `Using manual relationships` should link to how to create a view.
- `Using manual relationships`: Should `column_mapping` be `id: id`?
- Otherwise LGTM

`/schema/table-relationships/rename.rst` LGTM

## Files that still need review (probably from @rikinsk)

`/schema/enums.rst`
`/schema/tables.rst`
`/schema/using-existing-database.rst`
`/schema/views.rst`

rikinsk (approved):

tirumaraiselvan (approved):
changelog

PR #4955, opened 2020-06-01
Disables optimizations on Hasura.Server.Init module

0x777:
When compiling graphql-engine binary with `-O2`, ghc-8.10 seems to be stuck at the module `Server.Init` while consuming `17G` of RAM (for 5 minutes at least before I forcefully terminated the compilation). With this pragma, ghc-8.10 now takes under `12G` to compile graphql-engine binary.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [x] Build System
- [ ] Tests
- [ ] Other (list it)

### DIFF
--- a/server/src-lib/Hasura/Server/Init.hs
+++ b/server/src-lib/Hasura/Server/Init.hs
@@ -1,5 +1,6 @@
 -- | Types and functions related to the server initialisation
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP    #-}
+{-# OPTIONS_GHC -O0 #-}
 module Hasura.Server.Init
   ( DbUid(..)
   , getDbId


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -1,5 +1,6 @@
 -- | Types and functions related to the server initialisation
-{-# LANGUAGE CPP #-}
+{-# LANGUAGE CPP    #-}
+{-# OPTIONS_GHC -O0 #-}
```
0x777, line 3: I tried `-O1` too, seems to consume less memory but still ghc doesn't move past this module (gave it a good 5 mins).

PR #4959, opened 2020-06-01
allow logging details of a live query polling thread

0x777:
Currently we don't log any metrics related to the polling that happens in the background to serve live queries. These metrics are captured and summarized internally using datatypes from ekg-core (which can be dumped by `/dev/subscriptions` endpoint). This PR changes this behaviour, the metrics can now be logged by enabling a new log type `livequery-poller-log`. They are not summarized internally anymore. This opens up options to pipe the poller logs into a metric collection system where they can be analyzed further. 

### Changelog

### Affected components

- [x] Server

### Steps to test and verify
Add `livequery-poller-log` to `HASURA_GRAPHQL_ENABLED_LOG_TYPES` and you'll see logs of type: 
```json
{"type":"livequery-poller-log","timestamp":"2020-06-01T20:58:58.565+0530","level":"info","detail":{"poller_id":"4ec1aab0-43e0-4978-bc33-36a4f642ae12","total_time":1.872028e-3,"snapshot_time":2.2264e-5,"batches":[{"pg_execution_time":1.71155e-3,"push_time":9.8706e-5}]}}
```

#### Catalog upgrade

- [x] No

## Notes to the reviewer:

Module dependency tree:

image

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
+++ b/server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
@@ -2,6 +2,7 @@
 module Hasura.GraphQL.Execute.LiveQuery.Poll (
   -- * Pollers
     Poller(..)
+  , PollerId(..)
   , PollerIOState(..)
   , pollQuery
 
@@ -23,6 +24,7 @@ module Hasura.GraphQL.Execute.LiveQuery.Poll (
   -- * Subscribers
   , Subscriber(..)
   , SubscriberId
+  , UniqueSubscriberId(..)
   , newSinkId
   , SubscriberMap
   , OnChange
@@ -33,25 +35,31 @@ module Hasura.GraphQL.Execute.LiveQuery.Poll (
 
 import           Hasura.Prelude
 
-import qualified Control.Concurrent.Async                 as A
-import qualified Control.Concurrent.STM                   as STM
-import qualified Control.Immortal                         as Immortal
-import qualified Crypto.Hash                              as CH
-import qualified Data.Aeson.Extended                      as J
-import qualified Data.ByteString                          as BS
-import qualified Data.ByteString.Lazy                     as BL
-import qualified Data.HashMap.Strict                      as Map
-import qualified Data.Time.Clock                          as Clock
-import qualified Data.UUID                                as UUID
-import qualified Data.UUID.V4                             as UUID
+import qualified Control.Concurrent.Async                    as A
+import qualified Control.Concurrent.STM                      as STM
+import qualified Control.Immortal                            as Immortal
+import qualified Crypto.Hash                                 as CH
+import qualified Data.Aeson.Casing                           as J
+import qualified Data.Aeson.Extended                         as J
+import qualified Data.Aeson.TH                               as J
+import qualified Data.ByteString                             as BS
+import qualified Data.ByteString.Lazy                        as BL
+import qualified Data.HashMap.Strict                         as Map
+import qualified Data.Time.Clock                             as Clock
+import qualified Data.UUID                                   as UUID
+import qualified Data.UUID.V4                                as UUID
+import qualified Database.PG.Query                           as Q
 import qualified ListT
-import qualified StmContainers.Map                        as STMMap
-import qualified System.Metrics.Distribution              as Metrics
+import qualified StmContainers.Map                           as STMMap
+import qualified System.Metrics.Distribution                 as Metrics
 
-import           Data.List.Split                          (chunksOf)
+import           Data.List.Split                             (chunksOf)
 import           GHC.AssertNF
 
-import qualified Hasura.GraphQL.Execute.LiveQuery.TMap    as TMap
+import qualified Hasura.GraphQL.Execute.LiveQuery.TMap       as TMap
+import qualified Hasura.GraphQL.Transport.WebSocket.Protocol as WS
+import qualified Hasura.GraphQL.Transport.WebSocket.Server   as WS
+import qualified Hasura.Logging                              as L
 
 import           Hasura.Db
 import           Hasura.EncJSON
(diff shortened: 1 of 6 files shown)


server/src-lib/Hasura/RQL/DDL/RemoteSchema.hs
```diff
@@ -156,7 +156,7 @@ runIntrospectRemoteSchema (RemoteSchemaNameQuery rsName) = do
         throw400 NotExists $
         "remote schema: " <> remoteSchemaNameToTxt rsName <> " not found"
       Just rCtx -> mergeGCtx (rscGCtx rCtx) GC.emptyGCtx
-      -- ^ merge with emptyGCtx to get default query fields
+      -- merge with emptyGCtx to get default query fields
```
0x777, line 159: Trips haddock and hence ghcide

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -64,7 +72,35 @@ import           Hasura.Session
 -- -------------------------------------------------------------------------------------------------
 -- Subscribers
 
-newtype Subscriber = Subscriber { _sOnChangeCallback :: OnChange }
+-- | Identifies a subscriber of a live query based on their (websocket id,
+-- operation id) - this is useful for collecting metrics and then correlating.
+-- The 'createdAt' to ensure that there is uniqueness even if the operationId
+-- is later repeated on the same websocket
```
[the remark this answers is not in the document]
lexi-lambda: Actually, upon further thought, this seems sketchy to me. Suppose the operation id *is* repeated on the same websocket:

  1. Presumably, if we already have an active subscription for that operation id, we should reject the new one with an error, right?

  2. If that isn’t the case, and it’s okay to have multiple simultaneous subscriptions with the same operation id, then using `getCurrentTime` isn’t necessarily enough of a uniquifier, since it introduces a race condition if two subscriptions with the same operation id are created at the same time.

If uniqueness is required, it seems to me that it should either be enforced in the logic explicitly from the websocket and operation ids alone, or a `SubscriberId` should remain a UUID, and the value it’s *associated* with should hold the metadata for reporting purposes. Relying on a timestamp like this is bogus.

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -263,83 +289,164 @@ dumpPollerMap extended lqMap =
...
-      , "min" J..= Metrics.min stats
-      , "max" J..= Metrics.max stats
-      ]
+-- | An ID to track unique 'Poller's, so that we can gather metrics about each
+-- poller
+newtype PollerId = PollerId { unPollerId :: UUID.UUID }
+  deriving (Show, Eq, Generic, J.ToJSON)
+
+-- | Execution information related to a cohort on a poll cycle
+data CohortExecutionDetail
```
lexi-lambda, line 311: Incredibly minor nitpick, but this type seems like it should be named `CohortExecutionDetails`, plural (and the same goes for `BatchExecutionDetail` and `PollDetail` below). It contains multiple details, not one detail!

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -263,83 +289,164 @@ dumpPollerMap extended lqMap =
...
+
+$(J.deriveToJSON (J.aesonDrop 3 J.snakeCase) ''PollDetail)
+
+{- Note [Minimal LiveQuery Poller Log]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We only want to log the minimal information in the livequery-poller-log as it
+could be expensive to log the details of every subscriber (all poller related
+information can always be retrieved by dumping the current live queries state)
+The reason why we capture a lot more detail in PollDetail and
+BatchExecutionDetail as other implementations such as pro can use them if they
```
lexi-lambda, line 375: This seems like a typo; I think “as” should be “is.”
```suggestion
BatchExecutionDetail is other implementations such as pro can use them if they
```

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -180,69 +210,65 @@ pushResultToCohort
   -> Maybe ResponseHash
   -> LiveQueryMetadata
   -> CohortSnapshot
-  -> IO ()
+  -> IO ([SubscriberId], [SubscriberId])
+  -- ^ (subscribers to which data has been pushed, subscribers which already
+  -- have this data)
```
lexi-lambda, line 215: I’d appreciate an explicit comment that notes this information is only used for reporting:
```suggestion
  -- ^ subscribers to which data has been pushed, subscribers which already
  -- have this data (this information is exposed by metrics reporting)
```

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Poll.hs
```diff
@@ -350,21 +457,25 @@ pollQuery metrics batchSize pgExecCtx pgQuery handler =
...
+          let -- No reason to use lazy bytestrings here, since (1) we fetch the
+              -- entire result set from Postgres strictly and (2) even if we
+              -- didn’t, hashing will have to force the whole thing anyway.
+              respBS = encJToBS result
+              respHash = mkRespHash respBS
+              respSize = BS.length respBS
+          -- TODO: currently we ignore the cases when the cohortId from
+          -- Postgres response is not present in the cohort map of this batch
+          -- (this shouldn't happen but if it happens it means a logic error and
+          -- we should log it)
```
lexi-lambda, line 478: We should probably just do this rather than leave a TODO comment, right? It doesn’t seem challenging.

PR #4987, opened 2020-06-04
console: add the ability to delete a role in permissions summary page (close #3353)

### LINKED ISSUE #3353: allow renaming or deleting roles globally in console
In the console, `Schema permissions summary` page, i'm allowed to create new roles by cloning permission data from an existing role. I would like a way to delete the old role once i have cloned it to a new one. Or even better if i'm allowed to rename an existing role.

Right now the only way to delete a role completely is to manually go to every table's permission page and remove them one by one.

kolharsam:
resolve #3353 

### Description
- The `Schema Permissions Summary` button/link is repositioned to be next to the `Create Schema` section
- Also, delete a role using the delete button on the schema summary page

### Changelog

### Affected components

- [x] Console

### Solution and Design
a delete icon was added to next to the `copy` button that is currently available for each role on the `Schema Permissions Summary` page.


console/src/components/Services/Data/TablePermissions/Actions.js
```diff
@@ -526,6 +526,73 @@ export const isQueryTypeBackendOnlyCompatible = queryType => {
...
+          );
+
+          permissionsUpQueries.push(deleteQuery);
+          permissionsDownQueries.push(createQuery);
+        }
+      });
+    });
+
+    // Apply migration
+    const migrationName = `delete_all_permissions_for_${roleName}`;
```
kolharsam, line 569: Not sure if the name is fine. Need reviewers help on this, to see if there is a set nomenclature for migration names.

kolharsam: Changed it simply: `delete_role_${roleName}`

console/src/components/Services/Data/PermissionsSummary/RolesHeader.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+      if (selectedFirst && isCurrRole) {
+        rolesHeaders.unshift(roleHeader);
+      } else {
+        rolesHeaders.push(roleHeader);
+      }
+    });
+  }
+
+  return <>{rolesHeaders}</>;
```
beerose, line 68: So I really don't like this imperative style, so I was thinking about how we can make it more idiomatic react.

This part: `(selectedFirst && isCurrRole)` is responsible for positioning a role in the headers array and thus for the order of roles displayed in the UI. But we don't have to do it by `shift` and `unshift` of the components! We can order the array before doing any components work. 

How it can be done:
```ts
 let roles = allRoles;
  if (selectedFirst) {
    roles = allRoles.reduce((acc, role) => {
      if (role === currentRole) return [role, ...acc];
      return [...acc, role];
    }, [] as string[]);
  }
```

And then the component could look like this:

```tsx

const RolesHeader: React.FC<RolesHeaderProps> = ({
  selectable = true,
  selectedFirst = false,
  allRoles,
  currentRole,
  onCopyClick,
  onDeleteClick,
  setRole,
}) => {
  let roles = [...allRoles];
  if (selectedFirst) {
    roles = allRoles.reduce((acc, role) => {
      if (role === currentRole) return [role, ...acc];
      return [...acc, role];
    }, [] as string[]);
  }

  return (
    <>
      {roles.length ? roles.map(role => (
          <Header
            content={role}
            selectable={selectable}
            isSelected={currentRole === role}
            onClick={() => setRole(role, currentRole === role)}
            actionButtons={[
              <IconButton icon="fa-copy" onClick={e => onCopyClick(e, role)} />,
              <IconButton
                icon="fa-trash"
                onClick={e => onDeleteClick(e, role)}
              />,
            ]}
          />
        ))
      ) :  <Header content="No roles" selectable={false} />}
    </>
  );
};

export default RolesHeader;

```

rikinsk:
@beerose @kolharsam 
> Schema permissions summary behaves as a link (redirects to a different page), but looks like a button. Should we make it look as a link or say something like Show permissions summary (so that it indicates it's an action.)?

Agreed. We can make the text `Show permissions summary`. We can also use this opportunity to change the schema create/delete buttons to use text. Also lets increase the margin betweeen the create/delete buttons and the schema permissions link to `20px`.

Screenshot from 2020-06-16 18-59-34


> These buttons feel inconsistent with the rest of the console. I remember we were talking about having buttons with text instead of icons. What about having Copy and Delete buttons here? On the other hand it will take more space.

Yes. I dont think we will be able to follow that philosophy everywhere. Some buttons (especially in tables) might need to use icons (with the text as a title). Regarding the font colour, we can make it consistent. 

Also I noticed that currently the UI is a bit messed up
Screenshot from 2020-06-16 18-27-57

kolharsam:
@rikinsk I have made the changes as per your suggestions. Could you please take another look?

kolharsam:
> Minor change, Can you make the title for the delete role button as `Delete role permissions`. It says `Copy permissions` currently

sorry I missed that!

console/src/components/Services/Data/PermissionsSummary/PermissionsSummary.js
```diff
@@ -247,66 +208,46 @@ class PermissionsSummary extends Component {
...
 
-          const roleHeader = getHeader(
-            role,
-            selectable,
-            isCurrRole,
-            setRole,
-            getCopyBtn()
-          );
+      const deleteConfirmed = getConfirmation(
+        `This will delete permissions for the role: ${role}?`,
```
beerose, line 230:
```suggestion
        `This will delete permissions for the role: ${role}`,
```

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 (Add entries here in the order of: server, console, cli, docs, others)
 
 - console: allow manual edit of column types and handle array data types (close #2544, #3335, #2583) (#4546)
+- console: add ability to delete a role globally (close #3353) (#4987)
```
tirumaraiselvan, line 11: Could you specify where this is available?

beerose, line 11:
```suggestion
- console: add the ability to delete a role in permissions summary page (close #3353) (#4987)
```


Add the ability to delete a role in permissions summary page.

PR #5013, opened 2020-06-08
relay fixes

### LINKED ISSUE #5020: Relay: Schema introspection isn't working
After installing `v1.3.0-beta.1`, when I toggle the relay switch, the explorer doesn't show any schema. I've also tried a few other graphql tools, none can introspect the relay schema. I'd be happy to provide any info you need to debug, just let me know.

Having said that, I can run graphql queries against the relay endpoint perfectly fine, since I happen to know the shape of the query.

rakeshkky:
### Description

More tests, bug fixes and reduce node id string length.

### Changelog

Change log not required since these changes don't affect any stable release.

### Affected components

- [x] Server
- [x] Tests

### Related Issues

Fix #5020, #5037, #5046 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/GraphQL/Resolve/Types.hs
```diff
@@ -262,12 +261,47 @@ data InputFunctionArgument
   | IFAUnknown !FunctionArgItem -- ^ Unknown value, need to be parsed
   deriving (Show, Eq)
 
+{- Note [Relay Node Id]
```
0x777, line 264: @rakeshkky Let's add version as the `0`th element of the array. The reason being that if these GUIDs end up in URLs or if apps depend on them to not change as their schema evolves, an encoding which uses `(schema_name, table_name)` is not good enough (but it is the best we have currently).

 In future we may want to provide some sort of an identifier that can be attached to each table and this identifier is used to encode GUIDs for that table rather than (schema_name, table_name). So to accommodate for such changes, we'll need a version field in the ID.

server/src-lib/Hasura/GraphQL/Resolve/Types.hs
```diff
@@ -277,14 +277,32 @@ Node id data:
...
 
-It is represented in the type @'NodeIdData'.
+It is represented in the type @'NodeIdData'. The 'version-integer' represents the JSON
+schema version to enable any backward compatibility if it is broken in upcoming versions.
 
 The stringified JSON is Base64 encoded and sent to client. Also the same
 base64 encoded JSON string is accepted for 'node' field resolver's 'id' input.
 -}
 
+data NodeIdVersion
```
0x777, line 289: We don't need to have a separate type for the version I think. Instead, something like this would model the id correctly?
```haskell
data NodeId
  = NodeIdV1 !V1NodeId
```

server/src-lib/Hasura/GraphQL/Resolve/Types.hs
```diff
@@ -297,11 +315,12 @@ instance J.FromJSON NodeIdData where
   parseJSON v = do
     valueList <- J.parseJSON v
     case valueList of
-      (schemaValue:(nameValue:(firstColumn:remainingColumns))) ->
+      (idVersion:(schemaValue:(nameValue:(firstColumn:remainingColumns)))) -> do
```
0x777, line 318: Here, you can just pattern match against `1`.

PR #5039, opened 2020-06-10
cli: add new flags to generate sql migrations (close #5026)

### LINKED ISSUE #5026: cli: allow passing sql by flag to migrate create
Add two flags `--up-sql` and `--down-sql` to the `hasura migrate create` command. The contents of this flag should be used to fill in the `up.sql` and `down.sql` files of the created migration.

When either one of them is not provided show a warning.

Eg: 
```
hasura migrate create <migration_name> --up-sql "CREATE TABLE article(id serial NOT NULL, title text NOT NULL, content text NOT NULL, rating integer NOT NULL, author_id serial NOT NULL, PRIMARY KEY (id));" --down-sql "DROP TABLE article;"
```

kolharsam:
### Description

A new set of flags are added to the `migrate` command on the CLI. The `up-sql` and `down-sql` flags can be used to create SQL migrations from the CLI. These flags are to be used in conjunction. Either being absent during execution results in an error.

### Changelog

### Affected components
- [x] CLI

### Related Issues
resolves #5026 

#### Breaking changes

- [x] No Breaking changes


cli/commands/migrate_create.go
```diff
@@ -43,9 +47,13 @@ func newMigrateCreateCmd(ec *cli.ExecutionContext) *cobra.Command {
 		Long:         "Create sql and yaml files required for a migration",
 		Example:      migrateCreateCmdExamples,
 		SilenceUsage: true,
-		Args:         cobra.ExactArgs(1),
+		Args:         cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			opts.name = args[0]
+			if len(args) > 0 {
```
scriptonist, line 52: I have a feeling that, expecting and enforcing a meaningful name from the user for migration would be better.

cli/commands/migrate_create.go
```diff
@@ -111,12 +123,21 @@ func (o *migrateCreateOptions) run() (version int64, err error) {
 	if o.flags.Changed("metadata-from-file") && o.sqlServer {
 		return 0, errors.New("only one sql type can be set")
 	}
+
 	if o.flags.Changed("metadata-from-file") && o.metaDataServer {
 		return 0, errors.New("only one metadata type can be set")
 	}
 
+	if o.flags.Changed("up-sql") && !o.flags.Changed("down-sql") {
```
scriptonist, line 131: May be wrap this condition in a variable, something like
```
isSQLFromFlags :=  o.flags.Changed("up-sql") && o.flags.Changed("down-sql")
```
and then operating on that, thoughts ?

cli/commands/migrate_create.go
```diff
@@ -175,13 +196,27 @@ func (o *migrateCreateOptions) run() (version int64, err error) {
 		}
 	}
 
-	if !o.flags.Changed("sql-from-file") && !o.flags.Changed("metadata-from-file") && !o.metaDataServer && !o.sqlServer && o.EC.Config.Version == cli.V1 {
+	// create pure sql based migrations here
+	if o.flags.Changed("up-sql") && o.flags.Changed("down-sql") {
+		errUpSQL := createOptions.SetSQLUp(o.upSQL)
```
scriptonist, line 201: Handling error as soon as it happens will probably be better, something like
```
err = createOptions.SetSQLUp(o.upSQL)
if err != nil {
    return 0, errors.Wrap(errUpSQL, "up migration with SQL string could not be created")
}
```

kolharsam: Cool! Will change that

cli/commands/migrate_create.go
kolharsam: Okay then I'll leave it as it previously was?

scriptonist: Yup, that'll do

cli/commands/migrate_create.go
```diff
@@ -175,13 +192,26 @@ func (o *migrateCreateOptions) run() (version int64, err error) {
 		}
 	}
 
-	if !o.flags.Changed("sql-from-file") && !o.flags.Changed("metadata-from-file") && !o.metaDataServer && !o.sqlServer && o.EC.Config.Version == cli.V1 {
+	// create pure sql based migrations here
+	if o.flags.Changed("up-sql") && o.flags.Changed("down-sql") {
+		err = createOptions.SetSQLUp(o.upSQL)
+		if err != nil {
+			return 0, errors.Wrap(err, "up migration with SQL string could not be created")
```
scriptonist, line 199:
```suggestion
			ec.Logger.Warn("you are creating an up migration without a down migration")
```

cli/commands/migrate_create.go
```diff
@@ -175,13 +192,26 @@ func (o *migrateCreateOptions) run() (version int64, err error) {
...
+		}
+
+		err = createOptions.SetSQLDown(o.downSQL)
+		if err != nil {
+			return 0, errors.Wrap(err, "down migration with SQL string could not be created")
```
scriptonist, line 204:
```suggestion
			ec.Logger.Warn("you are creating a down migration without an up migration")
```

cli/commands/migrate_create.go
```diff
@@ -111,12 +119,21 @@ func (o *migrateCreateOptions) run() (version int64, err error) {
 	if o.flags.Changed("metadata-from-file") && o.sqlServer {
 		return 0, errors.New("only one sql type can be set")
 	}
+
 	if o.flags.Changed("metadata-from-file") && o.metaDataServer {
 		return 0, errors.New("only one metadata type can be set")
 	}
 
+	if o.flags.Changed("up-sql") && !o.flags.Changed("down-sql") {
```
scriptonist, line 127: Hey, @kolharsam extremely sorry, I messed up the previous suggestion, I actually meant the suggestion to be in this line, looks like it was misplaced. The Idea was to, allow a user to create a migration with either an up migration or a down migration but show a warning. Can you please make this change?

kolharsam: @scriptonist got it. I'll revert the previous suggestions and make this less strict I guess, since we only have to display the warnings.

cli/commands/migrate_create.go
```diff
@@ -30,7 +30,11 @@ const migrateCreateCmdExamples = `  # Setup migration files for the first time b
   hasura migrate create init --from-server --schema myschema1,myschema2
 
   # Take pg_dump from server and save it as a migration and specify the schemas to include
-  hasura migrate create init --sql-from-server --schema myschema1,myschema2`
+  hasura migrate create init --sql-from-server --schema myschema1,myschema2
+  
+  # Create up and down migrations based on SQL statements
```
scriptonist, line 35:
```suggestion
  # Create up and down SQL migrations, providing contents as flags
```

PR #5064, opened 2020-06-11
console: reuse websocket connection in api explorer

### LINKED ISSUE #3918: websocket isn't closed after stopped subscription on console graphiql
In console's graphiql, after the user stopped subscription, the websocket is still alive. If he starts new subscription, new websocket will be created for it. It leaks websocket connections

#### Reproduce steps
- Open chrome/firefox dev tools - Tab Network - `ws` filter
- Start a subscription
- Stop it
- Start another subscription
- Look at the dev tool

websocket_bug

kolharsam:
### Description
As noticed in #3918, the WebSocket connection wasn't being reused and again as pointed out in the issue, can lead to leaking WebSocket connections. So this PR hopes to fix that and help in reusing the existing WebSocket connection.

### Affected components
- [x] Console

### Related Issues
fixes #3918


console/src/components/Services/ApiExplorer/Actions.js
```diff
@@ -163,15 +177,15 @@ const createWsClient = (url, headers) => {
...
-    connectionParams: {
-      headers: {
-        ...headersFinal,
-      },
-    },
-    reconnect: true,
-  });
-  return client;
+
+  if (!WEBSOCKET_SUBSCRIPTION_CLIENT) {
```
kolharsam, line 181: Tried using `instanceof` against the class `SubscriptionClient`, didn't work as expected.

console/src/components/Services/ApiExplorer/Actions.js
```diff
@@ -44,6 +44,20 @@ const CREATE_WEBSOCKET_CLIENT = 'ApiExplorer/CREATE_WEBSOCKET_CLIENT';
...
+    connectionParams: {
+      headers: {
+        ...headers,
+      },
+      lazy: true,
+    },
+    reconnect: true,
+  });
+};
+
```
kolharsam, line 60: Might seem a little over the top...will remove it if that is the case. Did this only because it makes the changes below more readable.

console/src/components/Services/ApiExplorer/Actions.js
```diff
@@ -44,6 +44,20 @@ const CREATE_WEBSOCKET_CLIENT = 'ApiExplorer/CREATE_WEBSOCKET_CLIENT';
 const FOCUS_ROLE_HEADER = 'ApiExplorer/FOCUS_ROLE_HEADER';
 const UNFOCUS_ROLE_HEADER = 'ApiExplorer/UNFOCUS_ROLE_HEADER';
 
+let WEBSOCKET_SUBSCRIPTION_CLIENT;
```
beerose, line 47: We usually use the upper case for some string constants (as above). Let's just use the camel case here.

Issue #5087, opened 2020-06-14
PG pool can cause memory to never be freed on an active server


jberryman:
libpq grows the buffers it uses for input and output, as needed,
exponentially, but it never shrinks them. This means that if we try to keep
connections open in a thread pool for as long as possible we should expect
RES to grow and grow as e.g. a new biggest result query comes in.

Keep in mind this is for _each connection_ in a pool of maybe hundreds, so a few megabytes here and there can add up. And it only takes one outlier huge result set to grow the buffer.

Here we can see libpq doubling the buffer as we expect:

![hasura-heaptrack](https://user-images.githubusercontent.com/210815/84600579-4b27cf80-ae48-11ea-87c7-807d4a9eacfe.png)

One proper solution is to simply shrink the buffers back down to some max
acceptable size after each transaction. User Ilya confirms they've been
using essentially this approach successfully in production since reporting
the first ticket below.

~Unfortunately it seems we can't simply create a binding and do this without vendoring postgres itself. So we'll first have a workaround and then see if we can do something more efficient.~
**EDIT**: actually it looks like we can, so long as we copy what `postgresql-libpq` does in their Setup.hs, and we depend on e.g. `postgresql-server-dev-9.6` (the version doesn't matter AFAICT)

See e.g. pqCheckInBufferSpace which is the source of the leak that we noticed:

  https://docs.huihoo.com/doxygen/postgresql/fe-misc_8c.html#ab9525b356ad08b90e9beb497203848f8 

References to others who have encountered this issue

  https://www.postgresql.org/message-id/15693-0df90da151425ff5@postgresql.org 
  https://groups.google.com/forum/#!topic/pgsql.general/DIA4g8mMzNM 

Linking: 
https://github.com/hasura/graphql-engine/issues/3388
https://github.com/hasura/graphql-engine/issues/4077

jberryman:
A little graph showing libpq shrinking after big result sets. Here with only one connection, so we're clamping down to 1MB. In real life we'd probably expect to see something smoother since we'll have 50 or 100 connections and smaller result sets, maybe.

![new](https://user-images.githubusercontent.com/210815/86183881-fb602e00-bb00-11ea-8196-cb8e0cd3001e.png)

jflambert:
[...]

Issue #4077, opened 2020-03-10
Hasura consuming massive amounts of memory


mrmurphy:
We recently ended up shutting down our production Hasura deployment, because we were only using it for data ingestion, and we ended up in a situation where Hasura was constantly using all available memory, and then crashing.

I don't have any logs to offer, unfortunately. But I was asked on Discord to create an issue, so here's the information I can provide about our setup:

- Deployed on Heroku
- Initially ran on a $7/mo box on v1alpha1 build 28 (We didn't watch it closely, and at some point realized that it was crashing regularly from running out of memory)
- Scaled the box to larger standard dynos, even up to 3 boxes with 2.5Gb of ram each. Even in that situation, Hasura used all and more than the available ram, and was killed by Heroku.
- Heroku's router measurements said we were getting about 10 reqs / second from our client devices.
- The client devices were doing mutations only. No queries. They were writing in telemetry events.
- Sometimes queries could be multiple megabytes in size (a lot of events were in the write backlog)
- We used an auth hook to our application server that responded usually within a few hundred milliseconds.
- Sometimes the auth hook would throw a 500 error.
- Right before crashing the Hasura logs showed mutation requests timing out (taking longer than 30 seconds)
- Rebooting the box helped for a few seconds, but the slowness almost immediately returned.
- We truncated our events table to make sure the database wasn't the bottleneck, but the memory issues on Hasura continued.

Sorry I can't offer more. Thank you!

tirumaraiselvan:
@mrmurphy Thanks for creating the issue!

Seems like you were using Event Triggers on the tables. It is possible that the memory buildup was happening because of our in-memory queue (especially if there were lots of unprocessed events) for holding events which has now been removed with this PR: https://github.com/hasura/graphql-engine/pull/4013

The next stable release i.e. `v1.2.0` should have this new change and is likely to solve your problem.

mrmurphy:
Oh, sorry about that. No, we weren’t using Hasura Events,we were using Hasura to record client-side telemetry events. So the event queue I mentioned was a client-side queue of events to send to the server.

tirumaraiselvan:
@mrmurphy Ah, got it. We will look into it. 

I would also recommend to upgrade to latest stable if possible.

mrmurphy:
Ah sorry, I forgot to mention that as well. As part of the debugging process we upgraded to 1.1.0 and had the same issue

0x777:
@mrmurphy 

[...]

PR #5089, opened 2020-06-14
5087 libpq pool leak

jberryman:
### Description
See https://github.com/hasura/pg-client-hs/pull/16 and https://github.com/hasura/graphql-engine/issues/5087

### Changelog

### Affected components

- [x] Server
- [ ] Docs (**TODO** we probably need to document the way that the connection idle time affects memory; if user sets idle time really high and server is idle, memory will be retained in those connections)

### Related Issues
#4077 #5087 

### Solution and Design

See https://github.com/hasura/pg-client-hs/pull/16

### Steps to test and verify
Run lots of queries with different result sizes, maintaining a continuous load. ensure that overall RES eventually shrinks

See discussion in ticket.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -142,6 +142,7 @@ hasura seed apply --file 1234_add_some_seed_data.sql
 - server: avoid loss of precision when passing values in scientific notation (fix #4733)
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
 - server: fix importing of allow list query from metadata (fix #4687)
+- server: add new `--conn-lifetime` and `HASURA_GRAPHQL_PG_CONN_LIFETIME` options for expiring connections after some amount of active time (#5087)
 - server: flush log buffer during shutdown (#4800)
 - server: fix edge case with printing logs on startup failure (fix #4772)
 - console: allow entering big int values in the console (close #3667) (#4775)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,7 @@ hasura seed apply --file 1234_add_some_seed_data.sql
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
 - server: fix importing of allow list query from metadata (fix #4687)
 - server: add new `--conn-lifetime` and `HASURA_GRAPHQL_PG_CONN_LIFETIME` options for expiring connections after some amount of active time (#5087)
+- server: shrink libpq connection request/response buffers back to 1MB if they grow beyond 2MB, fixing leak-like behavior on active servers (#5087)
 - server: flush log buffer during shutdown (#4800)
 - server: fix edge case with printing logs on startup failure (fix #4772)
 - console: allow entering big int values in the console (close #3667) (#4775)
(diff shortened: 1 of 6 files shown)


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -391,6 +392,13 @@ pgTimeoutEnv =
   , "Each connection's idle time before it is closed (default: 180 sec)"
   )
 
+pgStaleTimeoutEnv :: (String, String)
```
tirumaraiselvan, line 395: We could call it `HASURA_GRAPHQL_PG_LIFETIME` or similar as both pgBouncer and pgpool use the word `lifetime` for signifying this. But, I am okay with `HASURA_GRAPHQL_PG_STALE_TIMEOUT` as well.

https://www.pgbouncer.org/config.html
https://www.pgpool.net/docs/latest/en/html/runtime-config-connection-pooling.html

CHANGELOG.md
```diff
@@ -76,6 +76,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 - server: raise error on startup when `--unauthorized-role` is ignored (#4736)
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
 - server: fix importing of allow list query from metadata (fix #4687)
+- server: add new `--stale-timeout` and `HASURA_GRAPHQL_PG_STALE_TIMEOUT` connection pool options, to work around space leak caused by libpq buffers (#5087)
```
tirumaraiselvan, line 79: Do we still need "to work around space leak caused by libpq buffers" as I presume that is coming in a different PR ?

server/cabal.project
```diff
@@ -36,7 +36,8 @@ package graphql-engine
 source-repository-package
   type: git
   location: https://github.com/hasura/pg-client-hs.git
-  tag: 70a849d09bea9461e72c5a5bbde06df65aab61c0
+  -- TODO merge https://github.com/hasura/pg-client-hs/pull/16
```
tirumaraiselvan, line 39: this is merged?

server/src-lib/Hasura/Server/Init.hs
jberryman: Thanks, yep it seems good to reuse that language.

CHANGELOG.md
jberryman: This PR contains the libpq fix for buffer shrinking. I think it's worthwhile to keep the new "lifetime" feature (formerly workaround for the buffer issue): if there's a reason that it is in pgbouncer I suppose it might be useful for some reason we maybe haven't thought about to hasura users. Editing the changelog though

PR #4429, opened 2020-04-16
server: _inc numeric (fix #3573)

abooij:
Add support for `_inc` on `real`, `double`, `numeric` and `money` (fix #3573)

### Description

Previously, the atomic `_inc` operator only supported `smallint`, `integer` and `bigint`. This expands support to other numeric types.

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Solution and Design

There are three components to this change:
- Hasura support for the `money` type.
- Advertising `_inc` support on other numeric types.
- Generating an `rhsExpOp` with the type annotation looked up from the update context.

### Steps to test and verify

I have added a test in `server/tests-py/queries/graphql_mutation/update/basic/numerics_inc.yaml`.

### Limitations, known bugs & workarounds

None known

#### Catalog upgrade

Does this PR change Hasura Catalog version?
- [x] No
- [ ] Yes
  - [ ] Updated docs with SQL for downgrading the catalog 

#### Metadata

Does this PR add a new Metadata feature?
- [x] No
- [ ] Yes
  - Does `run_sql` auto manages the new metadata through schema diffing?
    - [ ] Yes
    - [ ] Not required
  - Does `run_sql` auto manages the definitions of metadata on renaming?
    - [ ] Yes
    - [ ] Not required
  - Does `export_metadata`/`replace_metadata` supports the new metadata added?
    - [ ] Yes
    - [ ] Not required

#### GraphQL
- [ ] No new GraphQL schema is generated
- [x] New GraphQL schema is being generated:
   - [x] New types and typenames are correlated
   
   

#### Breaking changes

- [x] No Breaking changes
- [ ] There are breaking changes:

  1. Metadata API

     Existing `query` types:
     - [ ] Modify `args` payload which is not backward compatible
     - [ ] Behavioural change of the API
     - [ ] Change in response `JSON` schema
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     - [ ] Change in any `NamedType`
     - [ ] Change in table field names
     

     Schema Resolve:-
     
     - [ ] Change in treatment of `null` value for any input fields 
     

  3. Logging

     - [ ] Log `JSON` schema has changed
     - [ ] Log `type` names have changed

PR #5104, opened 2020-06-16
server: upgrade pg-client-hs to avoid loss of precision (fix #5092)

### LINKED ISSUE #5092: [postgres] type float4 not correctly parsed for large numbers (rounded)
Hasura does not correctly retrieve value for Postgres `float4` large numbers (ex: 500000.5).

The value is rounded, instead of returning the real value.

However, if the postgres type is float8, no problem.

Is only `float8` type is supported?

The [documentation](https://hasura.io/docs/1.0/graphql/manual/api-reference/postgresql-types.html#float) mentions `float8`, but does not explicitly mention that `float4` is not supported.

abooij:
### Description

#5092 reported a loss of precision when reading floats from the database. This was caused by odd default behavior by old versions of PostgreSQL.

The origin is that in PostgreSQL 11 and older, by default, floating point values are returned with a fixed decimal precision - in the case of `float4`, this is 6 digits. Using the option [`extra_float_digits`](https://www.postgresql.org/docs/11/runtime-config-client.html#GUC-EXTRA-FLOAT-DIGITS) (important: compare with the [v12 docs](https://www.postgresql.org/docs/12/runtime-config-client.html#GUC-EXTRA-FLOAT-DIGITS), which seems to claim some changed behavior alongside a change in the default value of `extra_float_digits`!), we can output the float with full precision. However, by default it is set to 0, which in this case means loss of precision.

This PR sets `extra_float_digits` to 3 globally to avoid this problem. But this introduces some platform dependent behavior when printing floats, so we make sure to modify our CI test suite to be resilient against this, by picking smartly chosen values that test the precision of floats, while at the same time being independent of the PostgreSQL version.

Additionally, we ask `pg_dump` to output floats with adequate precision using `extra_float_digits`.

### Changelog

### Affected components

- [x] Server
- [x] Tests

### Related Issues

We've had a variety of number precision issues before, such as #4429 and #4435.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/cabal.project
+++ b/server/cabal.project
@@ -35,8 +35,8 @@ package graphql-engine
 
 source-repository-package
   type: git
-  location: https://github.com/hasura/pg-client-hs.git
-  tag: 70a849d09bea9461e72c5a5bbde06df65aab61c0
+  location: https://github.com/abooij/pg-client-hs.git
+  tag: 6b5443b8850586f67b5bc2af95635abe431075cd
 
 source-repository-package
   type: git
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 - server: compile with GHC 8.10.1, closing a space leak with subscriptions. (close #4517) (#3388)
 - server: fixes an issue where introspection queries with variables would fail because of caching (fix #4547)
 - server: avoid loss of precision when passing values in scientific notation (fix #4733)
+- server: avoid loss of precision when retrieving floats (fix #5092) (#5104)
 - server: raise error on startup when `--unauthorized-role` is ignored (#4736)
 - server: fix mishandling of GeoJSON inputs in subscriptions (fix #3239)
 - server: fix importing of allow list query from metadata (fix #4687)
--- a/server/src-lib/Hasura/Server/API/PGDump.hs
+++ b/server/src-lib/Hasura/Server/API/PGDump.hs
@@ -81,6 +81,7 @@ execPGDump b ci = do
       , "SET default_tablespace = '';"
       , "SET default_with_oids = false;"
       , "SET default_table_access_method = heap;"
+      , "SET extra_float_digits = 3;"
       , "CREATE SCHEMA public;"
       , "COMMENT ON SCHEMA public IS 'standard public schema';"
       ]
--- a/server/tests-py/queries/graphql_mutation/update/basic/numerics_inc.yaml
+++ b/server/tests-py/queries/graphql_mutation/update/basic/numerics_inc.yaml
@@ -1,4 +1,3 @@
-# NB: Don't test for the returned num_real, since the precision depends on the PG version.
 description: Updated numerics data using _inc operator
 url: /v1/graphql
 status: 200
@@ -11,8 +10,8 @@ query:
              num_smallint: -1
              num_integer: -1
              num_bigint: -1
-             num_real: -1.1
-             num_double: -1.1
+             num_real: -1.5
+             num_double: -1.5
              num_money: -1.1
              num_numeric: -1.1
           }
@@ -23,6 +22,7 @@ query:
             num_smallint
             num_integer
             num_bigint
+            num_real
             num_double
             num_money
             num_numeric
@@ -32,9 +32,10 @@ query:
   - query: |
       mutation {
         update_numerics(
-          where: {id: {_eq: 1}},
+          where: {id: {_eq: 2}},
           _inc: {
-             num_double: -3E-4
+             num_real: -15E-1
+             num_double: -15E-1
              num_money: -2E-2
              num_numeric: -1E-30
           }
@@ -42,6 +43,7 @@ query:
           affected_rows
           returning {
             id
+            num_real
             num_double
             num_money
             num_numeric
@@ -57,14 +59,16 @@ response:
           num_smallint: 12344
           num_integer: 12344
           num_bigint: '12344'
-          num_double: '12344.57'
+          num_real: 500000.5
+          num_double: '500000000000000.5'
           num_money: $12,344.57
           num_numeric: '12344.5700'
   - data:
       update_numerics:
         affected_rows: 1
         returning:
-        - id: 1
-          num_double: '12344.5697'
-          num_money: $12,344.55
-          num_numeric: '12344.5699999999999999999999999999990'
+        - id: 2
+          num_real: 500000.5
+          num_double: '500000000000000.5'
+          num_money: $12,345.65
+          num_numeric: '12345.6699999999999999999999999999990'
--- a/server/tests-py/queries/graphql_mutation/update/basic/values_setup.yaml
+++ b/server/tests-py/queries/graphql_mutation/update/basic/values_setup.yaml
@@ -35,7 +35,14 @@ args:
     - num_smallint: 12345
       num_integer: 12345
       num_bigint: 12345
-      num_real: 12345.67
-      num_double: 12345.67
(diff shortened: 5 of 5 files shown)


abooij:
To the reviewer: note that this switches `pg-client-hs` to one hosted by me until [the PR for `pg-client-hs`](https://github.com/hasura/pg-client-hs/pull/17) gets merged. So that should probably happen first.

server/cabal.project
```diff
@@ -35,8 +35,8 @@ package graphql-engine
 
 source-repository-package
   type: git
-  location: https://github.com/hasura/pg-client-hs.git
-  tag: 70a849d09bea9461e72c5a5bbde06df65aab61c0
+  location: https://github.com/abooij/pg-client-hs.git
```
rakeshkky, line 38: Let us merge your changes to `pg-client-hs` master and update the reference here to `hasura`. Please link to the PR in pg-client-hs repo.

abooij: Hi Rakesh! I've linked to this PR above, and I have now also requested you as a reviewer for that PR.

rakeshkky: Hi @abooij , I merged the PR in pg-client-hs repo. Please refer the new commit. 
Thank you.

tirumaraiselvan:
After reading the description and seeing that this is postgres version dependent, I tried this on PG 9.5:

```

create table test(f float4);
insert into test values (12311.1), (12311.12);
select * from test;
    f    
---------
 12311.1
 12311.1


set extra_float_digits=3;
select * from test;

     f      
------------
 12311.0996
 12311.1201
(2 rows)

```
I wonder if `extra_float_digits` should be a server flag rather than a fixed default as it could be breaking change for existing users.

abooij:
@tirumaraiselvan (also re: [your response on the other part of this PR](https://github.com/hasura/pg-client-hs/pull/17))

I agree that this changes behavior. This is desirable: currently we output floats with inadequate precision, and this PR fixes that. So, I would rather refer to it as a fixing change than a breaking change.

Arguably, anyone currently relying on equality of floats is Doing It Wrong (tm) for the usual concerns about floats, but I can imagine there are use cases of that out there. For me, returning floats with adequate accuracy (i.e. not losing any data) weighs more heavily than avoiding breaking changes. And since Postgres 11 and older do not seem to have a mechanism to return exactly the right number of decimal digits (which is variable even within a single float datatype), `set extra_float_digits = 3` seems unavoidable to me. (*)

Rather than making this individual setting into a flag, more generally we might have a kind of `postgres_initialization` field that allows running several such settings, after we run our defaults. The disadvantage is that we'd also have to plumb this through `pg-client-hs`, requiring another PR, and it introduces additional configuration complexity.

An alternative to this scheme is to simply recommend everyone who has so far depended on getting a specific number of digits, to simply round to that number of digits client-side from now on. We can add javascript instructions to the changelog, for instance. Would that be adequate?

(*) I'd like to add that it is beyond me how any widely used piece of software (and I'm referring to Postgres here) can have _bugs_, because that's what this is, in its default handling of floats. Floats are dangerous data types that are difficult to get right, and this is discussed whenever you encounter any literature on them. There is no excuse for loss of data.

tirumaraiselvan:
It would have been okay if we gave extra precision results. But as you can see from the results, it actually gives imprecise answers (PG fault, no doubt). That's why I am a bit concerned about making this change. Suppose someone uses our API to show something in the UI and it shows these extra digits (many of them supposedly wrong), then they have no way to change that behaviour.

abooij:
> It would have been okay if we gave extra precision results. But as you can see from the results, it actually gives imprecise answers (PG fault, no doubt). That's why I am a bit concerned about making this change. Suppose someone uses our API to show something in the UI and it shows these extra digits (many of them supposedly wrong), then they have no way to change that behaviour.

Sorry, I must have misunderstood your concern in my previous message. 12311.1 cannot be represented exactly in IEEE 754 floats, see e.g. [this float4 calculator](https://www.h-schmidt.net/FloatConverter/IEEE754.html). The results you posted are correct. Saving 12311.1 as a float and then retrieving the saved value *should* result in what you obtained. Now, of course, if we then go and round the values you obtained to the original 1 digit (respectively 2 digits) after the decimal period, we get 12311.1 (respectively 12311.12), as desired.

So, yes, saving 12311.1 as a float4 in a database and then retrieving it will no longer result in 12311.1. This is avoided by using a "smart number of digits", which is available in Postgres 12. But the idea that you should "get back what you put in" is really a myth, since floats are not able to do that in general. (Only for specific values.)

tirumaraiselvan:
I see what you mean. But wouldn't a non-opinionated solution (so we don't have to worry about what we are sending) be to just let this set by user?

abooij:
@tirumaraiselvan I agree that it would be beneficial for some users if we can support custom PG runtime configuration. Would you see that as a blocker for getting this merged?

tirumaraiselvan:
@abooij Hardcoding a PG runtime configuration is something we should never do considering the number of legacy and existing databases that are running with Hasura. Having sane defaults is something we cannot avoid. We can choose to change the default to `extra_float_digits=3` for sure but we cannot have it non-configurable. So, yes I would say we should park this till we solve it more generally (and revert this: https://github.com/hasura/pg-client-hs/pull/17) but I am happy to hear more voices on this matter.

abooij:
@tirumaraiselvan I started thinking about how we may combine a safe default with user configurability. One issue here is that `extra_float_digits` can be set through the dabatase connection URL, as I documented [here](https://github.com/hasura/graphql-engine/issues/5092#issuecomment-647415766). This means that any parameters we set after connecting would overwrite user configuration, which I think is undesirable. But I'd also be quite hesitant to modify the user's connection URL to insert this parameter in the right place, so that the user can overwrite it.

So this is leading me to consider two possibilities:

- If the user provides a database connection URL, we trust the user to do the right thing. If no `extra_float_digits` is set by the user, we don't set any value for this either. We add instructions to the manual saying that if a URL is used, the user should probably set `extra_float_digits`, especially if Postgres 11 or older is used.
- If the user connects by setting host, port, user, pasword, etc, then we automatically set `extra_float_digits` to 3. If this is not what the user wants, then the user should connect using a database URL instead, so that this parameter can be set manually.

This makes things very configurable. But it also invites users to make mistakes when configuring database URLs. And we would still be setting `extra_float_digits` to 3 globally in some cases, which is not necessary for Postgres 12.

Any better configuration designs are very much welcome.

tirumaraiselvan:
Voila. I would say using the PG connection string to get this behaviour is an adequate solution to solve #5092. Also, we only support connecting to PG using connection strings (and not individual parameters).

> If the user provides a database connection URL, we trust the user to do the right thing. If no extra_float_digits is set by the user, we don't set any value for this either. We add instructions to the manual saying that if a URL is used, the user should probably set extra_float_digits, especially if Postgres 11 or older is used.

Sounds about right to me!

abooij:
> Also, we only support connecting to PG using connection strings (and not individual parameters).

Is that right? The manual [specifies](https://hasura.io/docs/1.0/graphql/manual/deployment/graphql-engine-flags/reference.html#server-flags) how to connect using the individual host/user/password/db flags.

tirumaraiselvan:
> Is that right? The manual [specifies](https://hasura.io/docs/1.0/graphql/manual/deployment/graphql-engine-flags/reference.html#server-flags) how to connect using the individual host/user/password/db flags.

Ah ok, my miss. I assumed this because ppl will use ENVs than override the CMD directive of our docker packages (because that will involve knowing the binary name, etc).

PR #5151, opened 2020-06-20
server: fix updating of headers behaviour in the update cron trigger API and create future events immediately

codingkarthik:
Currently, when a cron trigger is updated, the pending events are deleted and the updated cron trigger will be hydrated with new events within a minute. 

This PR fixes that behaviour by creating the pending events immediately after deleting the pending cron events.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/RQL/DDL/ScheduledTrigger.hs
+++ b/server/src-lib/Hasura/RQL/DDL/ScheduledTrigger.hs
@@ -95,10 +95,11 @@ updateCronTriggerInCatalog CronTriggerMetadata {..} = liftTx $ do
         cron_schedule = $3,
         payload = $4,
         retry_conf = $5,
-        include_in_metadata = $6,
-        comment = $7
+        header_conf = $6,
+        include_in_metadata = $7,
+        comment = $8
     WHERE name = $1
-   |] (ctName, Q.AltJ ctWebhook, ctSchedule, Q.AltJ <$> ctPayload, Q.AltJ ctRetryConf
+   |] (ctName, Q.AltJ ctWebhook, ctSchedule, Q.AltJ <$> ctPayload, Q.AltJ ctRetryConf,Q.AltJ ctHeaders
       , ctIncludeInMetadata, ctComment) False
   -- since the cron trigger is updated, clear all its future events which are not retries
   Q.unitQE defaultTxErrorHandler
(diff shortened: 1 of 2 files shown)


tirumaraiselvan:
> Oh, also: I think it would be good to include a changelog entry for this, even though it’s minor.

Since this is between beta releases, it's okay to skip this. Especially because the next one is stable (i.e. even if you put a changelog it will be skipped in the actual v1.3.0 stable changelog)

codingkarthik:
It'll be trivial to add a test case, I'll add it now

lexi-lambda:
> (i.e. even if you put a changelog it will be skipped in the actual v1.3.0 stable changelog)

I’m afraid I don’t understand what you mean by this—what do you mean by “skipped,” and why would that happen here?

tirumaraiselvan:
> > (i.e. even if you put a changelog it will be skipped in the actual v1.3.0 stable changelog)
> 
> I’m afraid I don’t understand what you mean by this—what do you mean by “skipped,” and why would that happen here?

Current release is `v1.3.0-beta.2`. Next release is `v1.3.0`. When we write changelog of `v1.3.0`, it would only include all changelog entries from the perspective of v1.2.2 (last stable). Since scheduled triggers didn't exist in v1.2.2, it will only have one entry ("introducing scheduled triggers" in highlight section) and none of these tiny bugfixes/improvements made during beta.

This changelog entry would have been useful if we were releasing `v1.3.0-beta.3` though.

lexi-lambda:
Ah, I see! That makes sense to me, thank you for the explanation.

tirumaraiselvan:
Ofcourse, it's okay if you add a changelog. But since I am following this issue and the next release, I just wanted to point out that it will get dropped.

codingkarthik:
While writing the test, I discovered that there was a bug which did not allow the user to update the `header_conf` of a cron trigger. I have fixed that and added a test for it too.

PR #5167, opened 2020-06-22
console: handle renaming cron triggers (close #5160)

### LINKED ISSUE #5160: [v1.3.0-beta.2] Renaming Cron Trigger crashes Hasura console
1. create a cron trigger
2. attempt to rename it
3. type RENAME
4. Hasura console "crashes" -- it goes to a 404 really

The problem is that the redirect URL is still the old cron trigger name instead of the new one.

BONUS:
1. click "Home" on the 404 page
2. click "EVENTS" tab
3. another wonderful 404

beerose:
Solves https://github.com/hasura/graphql-engine/issues/5160

### Description

**Why there was a bug in the first place?**

Whenever trigger data was missing we were throwing an error. However, it is in fact missing for a moment when the user renames it as the component re-renders during the rename (updates aren't batched — re-renders between updates).

We also had a race condition with `window.location` replace — that's why it sometimes was working and sometimes it wasn't. 

**Solution**

I work around the bug by introducing internal state that defers error with an arbitrary timeout. I think we should remodel events state entirely. 

### Changelog

### Affected components

- [x] Console


console/src/components/Services/Events/CronTriggers/Container.tsx
```diff
@@ -39,8 +48,28 @@ const STContainer: React.FC<Props> = ({
...
+        window.clearTimeout(triggerPresence.timeoutHandle);
+        setTriggerPresence('not-missing');
+      }
+    } else if (triggerPresence === 'not-missing') {
+      const timeoutHandle = window.setTimeout(() => {
+        setTriggerPresence('error-not-found');
+      }, 1200 /* arbitrary value */);
+      setTriggerPresence({ type: 'timeout', timeoutHandle });
+    }
+  }, [currentTrigger]);
```
soorajshankar, line 69: `setTriggerPresence` is a dependency here right?

beerose: No, `setTriggerPresence` is stable.

PR #5169, opened 2020-06-22
cli: add logs for reading and sending migrations

kolharsam:
### Description
On using the `--log-level debug` flag while performing migrations should log every time a file is read for migration and before sending the query to hasura.

### TODO

- [ ] try batching the files and then showing a progress bar? 

### Affected components
- [x] CLI


purush7 (approved):

kolharsam:
@purush7 I was thinking this could be improved a lot more. It's not worth merging this in it's current state. With the current design or the implementation of applying migrations onto the database it's hard to implement a progress bar and even in the case we do it, there will be a drop in the performance but would let us implement this feature in a better way. What are your opinions? cc @scriptonist

PR #5187, opened 2020-06-23
Add instructions to fix loss of float precision in PostgreSQL <= 11

abooij:
### Description

#5092 reported a loss of precision when reading floats from the database. This was caused by odd default behavior by old versions of PostgreSQL.

The origin is that in PostgreSQL 11 and older, by default, floating point values are returned with a fixed decimal precision - in the case of `float4`, this is 6 digits. Using the option [`extra_float_digits`](https://www.postgresql.org/docs/11/runtime-config-client.html#GUC-EXTRA-FLOAT-DIGITS) (important: compare with the [v12 docs](https://www.postgresql.org/docs/12/runtime-config-client.html#GUC-EXTRA-FLOAT-DIGITS), which seems to claim some changed behavior alongside a change in the default value of `extra_float_digits`!), we can output the float with full precision. However, by default it is set to 0, which in this case means loss of precision.

A previously proposed fix for this, PR #5104, set `extra_float_digits` to 3 globally to avoid this problem. However, this results in user-visible changes is behavior. This was an intentional consequence of #5104: floats were otherwise being retrieved with inadequate precision, which is a bug.

However, there are two workarounds available: one is to simply update to PostgreSQL 12 or later, which has sane default behavior. Another is to explicitly fix the configuration by setting `extra_float_digits`. The existence of these workarounds makes such global change is behavior undesirable. And globally setting `extra_float_digits` to a value of our choice would additionally interact poorly with user-chosen parameters.

This PR updates the documentation to instruct users to consider setting this flag through either the database connection URL or the new `--pg-connection-options` command line flag, which this PR also adds, and which takes flags that are used to connect to PostgreSQL.

### Changelog

### Affected components

- [x] Server
- [x] CLI
- [x] Docs

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -532,6 +532,14 @@ parseRawConnInfo =
                   metavar "<DBNAME>" <>
                   help "Database name to connect to"
                 )
+
+    options = optional $
+      strOption ( long "options" <>
```
0x777, line 537: I would probably call this `pg-connection-options` or something along those lines, `options` is too generic.

abooij: Done.

Issue #5190, opened 2020-06-24
Improve memory usage from STACK in websockets


jberryman:
Tested using Gavin's benchmark tool, spinning up 2000 subscriptions, which then sit idle, compiling with `-O1`:

RES as reported by top:
- baseline, idle server (both):  ~**90M**
- at 2000 connections (master): ~**340M**
- at 2000 connections (with `-kc8K`): ~**210M**

And here's what `+RTS -h` (non-prof build) looks like, this :

![compare](https://user-images.githubusercontent.com/210815/85492651-82625300-b5a3-11ea-8785-4b9eb10f01cc.png)

PR #5191, opened 2020-06-24
 improve websockets idle memory usage, closing #5190

jberryman:
### Description

See #5190

I started trying to do away with the other two threads we launch per websocket but they didn't seem to be contributing to memory usage so I left a couple TODO notes instead.

### Changelog

### Affected components

- [x] Server

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
+++ b/server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
@@ -136,7 +136,6 @@ data WSConn a
   { _wcConnId    :: !WSId
   , _wcLogger    :: !(L.Logger L.Hasura)
   , _wcConnRaw   :: !WS.Connection
-  , _wcSendQ     :: !(STM.TQueue WSQueueResponse)
   , _wcExtraData :: !a
   }
 
@@ -160,12 +159,11 @@ closeConnWithCode wsConn code bs = do
     WSLog (_wcConnId wsConn) (ECloseSent $ TBS.fromLBS bs) Nothing
   WS.sendCloseCode (_wcConnRaw wsConn) code bs
 
--- writes to a queue instead of the raw connection
--- so that sendMsg doesn't block
-sendMsg :: WSConn a -> WSQueueResponse -> IO ()
-sendMsg wsConn = \ !resp -> do
-  $assertNFHere resp  -- so we don't write thunks to mutable vars
-  STM.atomically $ STM.writeTQueue (_wcSendQ wsConn) resp
+sendMsg :: L.Logger L.Hasura -> WSConn a -> WSQueueResponse -> IO ()
+sendMsg (L.Logger writeLog) WSConn{..} = \WSQueueResponse{..} -> do
+  liftIO $ WS.sendTextData _wcConnRaw _wsqrMessage
+  writeLog $ WSLog _wcConnId (EMessageSent $ TBS.fromLBS _wsqrMessage) _wsqrEventInfo
+
 
 type ConnMap a = STMMap.Map WSId (WSConn a)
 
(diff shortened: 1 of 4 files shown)


server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
```diff
@@ -160,12 +159,11 @@ closeConnWithCode wsConn code bs = do
     WSLog (_wcConnId wsConn) (ECloseSent $ TBS.fromLBS bs) Nothing
   WS.sendCloseCode (_wcConnRaw wsConn) code bs
 
--- writes to a queue instead of the raw connection
--- so that sendMsg doesn't block
-sendMsg :: WSConn a -> WSQueueResponse -> IO ()
-sendMsg wsConn = \ !resp -> do
-  $assertNFHere resp  -- so we don't write thunks to mutable vars
-  STM.atomically $ STM.writeTQueue (_wcSendQ wsConn) resp
+sendMsg :: L.Logger L.Hasura -> WSConn a -> WSQueueResponse -> IO ()
```
0x777, line 162: Wouldn't this make `sendMsg` blocking?

jberryman: Yeah, I saw the comment that the queue approach was so that `sendMsg` wouldn't block, but I couldn't find a reason why we'd want or need that property. I don't have totally clarity though; we discussed this briefly but it seemed easier to just make the change and discuss it.

My thought is with this change we get backpressure rather than just filling up an in-memory queue while continuing to service messages, if for some reason the client was sending but not receiving?

0x777: The subscriptions backend has logic along these lines:
```python
while True:
  batchedQuery <- getBatchedQuery activeSubscriptions
  batchedQueryResponse <- execute batchedQuery
  for response in batchedQueryResponse:
    mapConcurrently (sendMsg response) websocketConnections
  sleep x
```
If `sendMsg` is blocking, the entire poll cycle is blocked on one badly behaving ws connection which defeats the whole idea behind batching multiple subscription responses.

jberryman: @0x777 gotcha, I think you're referring to `pushResultToCohort` called from `pollQuery`. Hm, some thoughts:

the way things work now is tricky and fragile; we would rather want to be able to look at the body of `pushResultToCohort` and see that it doesn't block

If we just change the `mapConcurrently` to `mapM . forkIO` we can't actually guarantee FIFO, but we could maybe associate each connection with an `MVar Int` counter mutex, fork and forget, then each thread can simply drop its update if its count is lower than the counter value, something like:

```
pollQuery :: Int -> ...
pollQuery iteration ... = 
   ...
    -- instead of mapConcurrently: 
   for_ websocketConnections $ \conn -> forkIO $ do
       tryTakeMvar (counterMutex conn) >>= \case
         Nothing -> -- an older iteration thread is blocked on a slow client, drop this update
         Just prevIteration 
           | iteration < previousIteration -> -- we raced so drop this stale update
           | otherwise -> do 
                     sendMsg ...
                     putMVar (counterMutex conn) iteration
```

I think this is a better semantics too: a slow client will receive relatively timely updates rather than old stale updates. Note subscriptions aren't really a strict change stream since we're polling, e.g. if an int column is moving up and down we already make no promise that a subscription change event will ever be generated.
  
I'm not sure this is worth the effort right now, but let me know what you think.

jberryman:
@0x777 the RTS change has been shifted into its own PR here: https://github.com/hasura/graphql-engine/pull/5214

The change under discussion here has a much less significant impact on memory. Might just close this PR

server/src-lib/Hasura/GraphQL/Transport/WebSocket/Server.hs
0x777: >  could maybe associate each connection with an `MVar Int` counter mutex, fork and forget, then each thread can simply drop its update if its count is lower than the counter value
> 
> I think this is a better semantics too: a slow client will receive relatively timely updates rather than old stale updates. Note subscriptions aren't really a strict change stream since we're polling, e.g. if an int column is moving up and down we already make no promise that a subscription change event will ever be generated.

A client can subscribe to any number of live queries on a single websocket connection so this approach might not work. Maybe a fixed size buffer? We drop messages if the buffer is full?

PR #5193, opened 2020-06-24
Chore/migration module #455

### LINKED ISSUE #455: Feat: Define aggregate_fields on array_relationships so permissions can be centrally managed
Imagine in the  the example poll app that an adminstrator detected some voting abuse and decided to institute a quality control process.  The developer added a field to the vote called `is_valid` and wants only the valid votes to be visible to users.  So he adds a user permission { is_valid: true }
 
Unfortunately, the developer forgot that there was a poll_results view that accessed votes directly without any knowledge of Hasura's centrally managed permissions.  

```sql
CREATE VIEW public.poll_results AS
 SELECT poll.id AS poll_id,
    o.option_id,
    count(*) AS votes
   FROM (( SELECT vote.option_id,
            option.poll_id,
            option.text
           FROM (public.vote
             LEFT JOIN public.option ON ((option.id = vote.option_id)))) o
     LEFT JOIN public.poll ON ((poll.id = o.poll_id)))
  GROUP BY poll.question, o.option_id, poll.id;
```

Since the developer implemented aggregate fields using hand written views, the users vote total count includes both valid and invalid votes.

soorajshankar:
### Changelog

no change log required

### Affected components

- [x] Console

### Related Issues

close #455

### Solution and Design

makeMigrationCall is a generic method that has many arguments, this PR simplifies the approach of passing arguments as a single object (where the namespace simplifies this).
`makeMigrationCall` will now accept the migration object instead of accepting upqueries and downqueries seperately.

### Steps to test and verify
Do all Migrations and test up&down migrations


soorajshankar:
@rikinsk @beerose this PR contains changes from #4790

PR #5195, opened 2020-06-24
cli: Feature/disable update cli

### LINKED ISSUE #5152: add way to disable the update command
This flag should disable the update command and potentially the update checking code. Homebrew prefers to not have self-updating software and can manage the update cycle fairly well by itself.

An implementation adding a compile flag is suggested here - https://github.com/hasura/graphql-engine/issues/422#issuecomment-646506231_

kolharsam:
closes #5152 

### Description
(needs update) 

### Affected components

- [x] CLI

### Solution and Design
an `IsUpdateCommandEnabled` is added 

(needs update)

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -28,11 +28,21 @@ endif
 .PHONY: build
 build: export CGO_ENABLED=0
 build:
-	gox -ldflags '-X github.com/hasura/graphql-engine/cli/version.BuildVersion=$(VERSION) -X github.com/hasura/graphql-engine/cli/plugins.IndexBranchRef=$(PLUGINS_BRANCH) -s -w -extldflags "-static"' \
+	gox -ldflags '-X github.com/hasura/graphql-engine/cli/version.BuildVersion=$(VERSION) -X github.com/hasura/graphql-engine/cli/plugins.IndexBranchRef=$(PLUGINS_BRANCH) -X github.com/hasura/graphql-engine/cli/commands.IsUpdateCommandEnabled=true -s -w -extldflags "-static"' \
 	-rebuild \
 	-os="$(OS)" \
 	-arch="amd64" \
 	-output="$(OUTPUT_DIR)/$(VERSION)/cli-hasura-{{.OS}}-{{.Arch}}" \
+	./cmd/hasura/
+
+.PHONY: build-brew
+build-brew: export CGO_ENABLED=0
+build-brew:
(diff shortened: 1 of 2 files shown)


cli/Makefile
```diff
@@ -28,13 +28,23 @@ endif
...
 build:
-	gox -ldflags '-X github.com/hasura/graphql-engine/cli/version.BuildVersion=$(VERSION) -X github.com/hasura/graphql-engine/cli/plugins.IndexBranchRef=$(PLUGINS_BRANCH) -s -w -extldflags "-static"' \
+	gox -ldflags '-X github.com/hasura/graphql-engine/cli/version.BuildVersion=$(VERSION) -X github.com/hasura/graphql-engine/cli/plugins.IndexBranchRef=$(PLUGINS_BRANCH) -X github.com/hasura/graphql-engine/cli/commands.IsUpdateCommandEnabled=true -s -w -extldflags "-static"' \
 	-rebuild \
 	-os="$(OS)" \
 	-arch="amd64" \
 	-output="$(OUTPUT_DIR)/$(VERSION)/cli-hasura-{{.OS}}-{{.Arch}}" \
 	./cmd/hasura/
 
+.PHONY: build-brew
```
scriptonist, line 38: I feel like we don't need an entry in `Makefile` entry since the brew formula directly sets the [ld flags](https://github.com/leos/homebrew-core/blob/c675cc5efb6315de4f7b8f22cc76ce6a70a3cf55/Formula/hasura-cli.rb#L12)

kolharsam: Right, okay. I shall remove this entry. But the formula does not include the flag `IsUpdateCommandEnabled`. Should it be updated in the formula?

scriptonist: Cool, I think the formula change can be done later

kolharsam: Have removed this.

cli/commands/root.go
```diff
@@ -23,6 +23,11 @@ const hasuraASCIIText = `
 
 `
 
+var (
+	//IsUpdateCommandEnabled is an ldflag used to disable the automatic updates
+	IsUpdateCommandEnabled string
```
scriptonist, line 28: Can we have the default value as true so that, the actual build command in the `Makefile` remains unchanged ?

scriptonist, line 28: Can we initialize this value as true so that the `build` step in `Makefile` remains the same ?

kolharsam: Removed change on Makefile and initialized this variable

cli/commands/root.go
```diff
@@ -23,6 +23,11 @@ const hasuraASCIIText = `
 
 `
 
+var (
+	//IsUpdateCommandEnabled is an ldflag used to enable/disable the automatic updates
+	IsUpdateCommandEnabled = "true"
```
scriptonist, line 28: Can we also use this to disable, `update-cli` command ?

purush7 (approved):

kolharsam:
Note to self: Migrate to mono-repo

PR #5248, opened 2020-07-01
Cascade delete on Postgres dependency errors #5109

### LINKED ISSUE #5109: Cascade delete on Postgres dependency errors
In #4846 we added an option to cascade metadata on [dependency errors](https://hasura.io/docs/1.0/graphql/manual/api-reference/schema-metadata-api/index.html#error-codes). We can do the same thing for Postgres dependency errors — have a prompt with information about dropping dependent objects and on continue we can run the drop command with `cascade` (eg. `drop table <name> cascade`).

ref. https://github.com/hasura/graphql-engine/pull/4846#pullrequestreview-421129907

soorajshankar:
### Description
This PR adds a user prompt when the user gets postgress error  [`dependent_objects_still_exist-2BP01` ](https://www.postgresql.org/docs/12/errcodes-appendix.html)

### Changelog

### Affected components
- [x] Console

### Related Issues
closes #5109 

### Solution and Design
This PR will add an on error listener at `makeMigrationCall` which makes sure that 
> if there is a Postgres dependency error, the user will get a prompt that allows the user to make a decision whether to continue the operation with cascading dependent modules or to cancel the operation. 

### Steps to test and verify
* create a 2 tables which have foreign key relationships.
* Try deleting the table where the foreign key relationship.
* check the network response, if this contains postgress error '2BP01' now the user will see an error prompt.
* click on continue, now check the network query.
* compare with the previous query, this should contain the cascade in the up query.

* possibility of inconsistent down migration after the cascade operations.

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/console/src/components/Services/Data/DataActions.js
+++ b/console/src/components/Services/Data/DataActions.js
@@ -582,7 +582,7 @@ const makeMigrationCall = (
     }
     customOnSuccess(data, globals.consoleMode, currMigrationMode);
   };
-  const retryMigration = (err = {}, errMsg = '') => {
+  const retryMigration = (err = {}, errMsg = '', isPgCascade = false) => {
     dispatch(
       showNotification(
         {
@@ -603,7 +603,7 @@ const makeMigrationCall = (
               makeMigrationCall(
                 dispatch,
                 getState,
-                cascadeUpQueries(upQueries), // cascaded new up queries
+                cascadeUpQueries(upQueries, isPgCascade), // cascaded new up queries
                 downQueries,
                 migrationName,
                 customOnSuccess,
@@ -623,8 +623,10 @@ const makeMigrationCall = (
 
   const onError = err => {
     if (!isRetry) {
-      const dependecyError = getDependencyError(err);
-      if (dependecyError) return retryMigration(dependecyError, errorMsg);
+      const { dependencyError, pgDependencyError } = getDependencyError(err);
+      if (dependencyError) return retryMigration(dependencyError, errorMsg);
+      if (pgDependencyError)
+        return retryMigration(pgDependencyError, errorMsg, true);
     }
 
     dispatch(handleMigrationErrors(errorMsg, err));
--- a/console/src/components/Services/Data/utils.js
+++ b/console/src/components/Services/Data/utils.js
@@ -790,13 +790,21 @@ WHERE
 export const isColTypeString = colType =>
   ['text', 'varchar', 'char', 'bpchar', 'name'].includes(colType);
 
-export const cascadeUpQueries = (upQueries = []) =>
+const cascadePGSqlQuery = sql => {
+  let prefix;
+  if (sql[sql.length - 1] === ';') prefix = sql.substr(0, sql.length - 1);
+  // SQL might have  a " at the end
+  else if (sql[sql.length - 2] === ';') prefix = sql.substr(0, sql.length - 2);
+  return prefix + ' CASCADE;';
+};
+export const cascadeUpQueries = (upQueries = [], isPgCascade = false) =>
   upQueries.map((i = {}) => {
     if (i.type === 'run_sql' || i.type === 'untrack_table') {
       return {
         ...i,
         args: {
           ...i.args,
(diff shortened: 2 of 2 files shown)


console/src/components/Services/Data/utils.js
```diff
@@ -805,16 +813,26 @@ export const cascadeUpQueries = (upQueries = []) =>
...
-    if (actualError.code == ERROR_CODES.dependencyError.code) {
-      return { ...actualError, message: actualError.error };
-    }
+    if (actualError.code == ERROR_CODES.dependencyError.code)
+      return {
+        dependencyError: { ...actualError, message: actualError.error },
+      };
+    else if (
+      actualError.code === ERROR_CODES.postgresError.code &&
+      actualError?.internal?.error?.status_code === '2BP01' // pg dependent error > https://www.postgresql.org/docs/12/errcodes-appendix.html
```
beerose, line 831:
```suggestion
      actualError?.internal?.error?.status_code === '2BP01' // pg dependent error > https://www.postgresql.org/docs/current/errcodes-appendix.html
```

console/src/components/Services/Data/utils.js
```diff
@@ -805,16 +813,26 @@ export const cascadeUpQueries = (upQueries = []) =>
   });
 
 export const getDependencyError = (err = {}) => {
+  const dflt = {};
```
beerose, line 816: What is it? Seems like it's not used anywhere except for final return

console/src/components/Services/Data/utils.js
```diff
@@ -790,13 +790,21 @@ WHERE
 export const isColTypeString = colType =>
   ['text', 'varchar', 'char', 'bpchar', 'name'].includes(colType);
 
-export const cascadeUpQueries = (upQueries = []) =>
+const cascadePGSqlQuery = sql => {
```
beerose, line 793: A lot is going on here. Can you put curly brackets for readability?

What about:
```js
const cascadePGSqlQuery = sql => {
  if (sql[sql.length - 1] === ';') {
    return sql.substr(0, sql.length - 1) + 'CASCADE';
  }
  // SQL might have  a " at the end
  return sql.substr(0, sql.length - 2) + ' CASCADE';
};

```

console/src/components/Services/Data/utils.js
soorajshankar: I had some other thoughts when I started writing this :joy:
Changing now

console/src/components/Services/Data/utils.js
soorajshankar: Done

rikinsk:
@soorajshankar i just tried this out and wasnt able to see the expected behaviour. I've created `author` and `article` tables with a fkey between them. Deleting the [author table](https://hge-ci-pull-5248.herokuapp.com/console/data/schema/public/tables/author/modify) throws the PG dependency error but I do not see the option to cascade. 
Screenshot from 2020-08-06 18-28-47

rikinsk:
Seems to work now. But I think we can improve the UX here.

**Current master behaviour:**
Screenshot from 2020-08-11 14-30-58
Screenshot from 2020-08-11 14-31-05
 
The user can expand to see details on what are the dependent objects in the `description` field .

**Current PR behaviour:**
Screenshot from 2020-08-11 14-33-10

I get option to drop the dependent bjects but I cant see what they are.

**Proposal:**
In case we have a PG dependency error, we need to do either of the following
- we can show the `description` of the error along with the `message` as it contains the dependent objects and then allow to drop the dependant objects
OR 
- we keep the behaviour same as master and allow to drop dependent objects after clicking the Details button and the whole error is displayed

console/src/components/Services/Data/DataActions.js
```diff
@@ -689,6 +693,7 @@ const makeMigrationCall = (
                 successMsg,
                 errorMsg,
                 shouldSkipSchemaReload,
+                false,
```
soorajshankar, line 696: @beerose Could you please add this change for the very next update?
Related to https://github.com/hasura/graphql-engine/issues/5436

PR #5254, opened 2020-07-01
searchable select ux bug fix (close#5246)

### LINKED ISSUE #5246: `column_type` dropdown bugs
A few bugs in the `column_type` dropdown:

- Clicking a dropdown option sometimes doesn't update the field:

Screen Shot 2020-06-30 at 7 39 21 PM

- Blurring and re-focusing the field and typing sometimes doesn't update the dropdown results:

Screen Shot 2020-06-30 at 7 39 02 PM

- Deleting text in the field has no effect:

Screen Shot 2020-06-30 at 7 39 13 PM

soorajshankar:
### Description
This PR will fix the behavior of Searchable select component, this will ensure that the unwanted events are ignored and hence SearchInput, Select Dropdown, selected value state  

### Changelog

### Affected components
- [x] Console

### Related Issues
close #5246

### Solution and Design
`onMenuClose`, `onBlur`, and `onChange` were triggered almost at the same time, and hence the order of events was not working as expected. (corner cases mentioned in the issue).

This PR will optimize the order by using ref and only emits an event when the component goes to `onBlur` state.

### Steps to test and verify
* Previously selecting an item from the list for the first time was messy, do a comparison.
* Try selecting items multiple times and make sure it's all in sync.
* Leave different user input values and leave the component by clicking outside.

Maybe a bit of confusion here 
* When there is a selected value, and when the user clears the field, `react-select` shows the last selected value as the placeholder until the component goes to the blurred state.


EmrysMyrddin:
Hi, I was working on this issue, and it seems that the bug was introduced when adding the possibility to add custom type. You can see it here : #5244

I have found out that `react-select` already implements a Select component allowing to create new option from search input. Using it resolved the bugs and cleaned up the code.

I'm just stuck on the placeholder not being displayed at first mount, I don't really understand why.

beerose:
@EmrysMyrddin We did explore an option to use `CreatableSelect` instead of implementing this whole logic on our own, however, this component is pretty limited — for example, it doesn't allow to type after selecting a value (you need to start typing from start) which is quite inconvenient.

EmrysMyrddin:
Ok, i understand :-)

I will close my PR then.

rikinsk:
@sezgi Could you check if this resolves all the issues you noticed.

PR #5255, opened 2020-07-01
do not use prepared statements for mutations, close #2012

### LINKED ISSUE #2012: don't use prepared statements for inserts
We already generate one insert statement with multiple value expressions when there is more than one row to be inserted. Currently, we prepare this query. But this becomes useful only if the same number of rows are inserted again with similar (column, value) pairs. We also run into issues where the prepared statement exceeds the number of allowed parameters (see https://github.com/hasura/graphql-engine/issues/2000). 

Given the low returns (if any) I'm inclined to simplify insert mutations by not using prepared statements.

0x777:
This PR introduces changes that should reduce memory consumption (which could result in a leak) on both graphql-engine and Postgres for a mutation heavy workload. 

### Description
With mutations we end up with several 'single-use' prepared SQL statements, i.e, statements which cannot be reused for multiple mutations. The underlying reason for ending up with 'single-use' prepared statements is we embed values inside the statement instead of parameterizing them to reduce the complexity around handling relationships in `returning` clause. The problem with single-use prepared statements is that we end up leaking memory both on graphql-engine (we store a mapping of "SQL prepared statement" to "postgres's prepared statement id" in each `PGConn`) and on Postgres (no limits on number of prepared statements per connection).

### Changelog

### Affected components

- [x] Server

### Related Issues
#2012 
#3388
#4077 

### Solution and Design
Prepared statements are not used for mutations any more. We *might* potentially end up with slightly slower but the benefit of predictable memory usage outweighs the potential degradation. 

### Steps to test and verify
@tirumaraiselvan has been running some benchmarks with a mutation heavy workload. This PR will be updated with the findings from those benchmarks.

- [x] No

#### GraphQL

#### Breaking changes
- [x] No Breaking changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 
 - server: add new `--conn-lifetime` and `HASURA_GRAPHQL_PG_CONN_LIFETIME` options for expiring connections after some amount of active time (#5087)
 - server: shrink libpq connection request/response buffers back to 1MB if they grow beyond 2MB, fixing leak-like behavior on active servers (#5087)
+- server: disable prepared statements for mutations as we end up with single-use objects which result in excessive memory consumption for mutation heavy workloads (#5255)
 - console: allow configuring statement timeout on console RawSQL page (close #4998) (#5045)
 - docs: add note for managed databases in postgres requirements (close #1677, #3783) (#5228)
 - docs: add 1-click deployment to Nhost page to the deployment guides (#5180)
@@ -119,7 +120,7 @@ Read more about the session argument for computed fields in the [docs](https://h
 A new `seeds` command is introduced in CLI, this will allow managing seed migrations as SQL files
 
 #### Creating seed
-```                                                        
+```
 # create a new seed file and use editor to add SQL content
 hasura seed create new_table_seed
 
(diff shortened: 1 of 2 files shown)


tirumaraiselvan:
Ran the benchmark with nested mutations:

On this build:

Hasura: 215MB
PG: 135MB (right after benchmark)

On `v1.3.0-beta.3`:

Hasura: 600MB
PG: 2.2G (right after benchmark), 50MB (after 5 minutes)

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - server: add new `--conn-lifetime` and `HASURA_GRAPHQL_PG_CONN_LIFETIME` options for expiring connections after some amount of active time (#5087)
 - server: shrink libpq connection request/response buffers back to 1MB if they grow beyond 2MB, fixing leak-like behavior on active servers (#5087)
+- server: prepared statements are not used in mutations which should fix memory leaks for mutation heavy workloads.
```
tirumaraiselvan, line 11:
```suggestion
- server: disable prepared statements for mutations as they cause single-use prepared statements and bloat memory for mutation heavy workloads.
```

nirvdrum:
I have no familiarity with talking to PostgreSQL from Haskell or even how Hasura takes user input to pass on via SQL, so apologies in advance if this is a misguided question, but how are SQL injection attacks guarded against? I had assumed that prepared statements were being used to prevent SQL injection issues and so seeing them removed caught my eye.

Issue #5199, opened 2020-06-24
Request Feature: Add ability to run specific hasura mutation in cron triggers


ghyath5:
Thank you for the great cron triggers feature.
Please add ability to run specific mutation in this cron-triggers, it could be super great feature

tirumaraiselvan:
Closing this as dupe of #5050

PR #5262, opened 2020-07-01
Include additional fields in scheduled trigger webhook body

### LINKED ISSUE #5253: Cron trigger is sending only custom portion of payload
## Actual behavior

This is what I see in the console for invocations.

image

And this is what I receive in my ngrok tunnel.

image

## Expected behavior

Full `payload` should be sent including ID of invocation and the name.

codingkarthik:
fix #5253 

Earlier, the request to the webhook included only the payload specified by the user, now some additional metadata will also be included in the request, like the `scheduled_time`,`created_at`,`id` will also be included in the request.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes
The response format of the webhook request body has changed. Earlier, only the `payload` specified by the user was sent. 
For example, If the payload specified by the user were ```{"hello":"world"}```. Then, only ```{"hello":"world"}``` was sent to the webhook. 

Now, Additional fields are added to the webhook, like `created_at`,`scheduled_time`,`comment`(if any),`id`,`name`(in case of cron triggers) are included in the request body. For example:

```json
{
    "scheduled_time": "2020-07-01T20:34:00Z",
    "payload": {
        "hello": "world"
    },
    "name": "sample_cron_trigger",
    "created_at": "2020-07-01T20:34:45.910268Z",
    "id": "2dc5a935-b98e-4b82-91df-268996faa20f"
}
```
  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -211,6 +211,18 @@ data ScheduledEventType =
   -- so all the configuration is fetched along the scheduled events.
     deriving (Eq, Show)
 
+data ScheduledEventWebhookBody
+  = ScheduledEventWebhookBody
```
lexi-lambda, line 215: I’m a little bit curious: the original issue shows a request payload with these fields in the console. So these fields must come from somewhere in the code already, right? It seems like it would be useful to share the definition of those types in some way. But maybe I’m misunderstanding something.

tirumaraiselvan: I think it is coming from here: https://github.com/hasura/graphql-engine/blob/9b247d4ed79e6e82409d994c70b52c561c832230/server/src-lib/Hasura/Eventing/HTTP.hs#L69

tirumaraiselvan: I think we could potentially change `_rqPayload` -> `_rqBody` to stay consistent.

codingkarthik: The request that's being shown in the console are from the invocation logs. While logging the request and response, the earlier behaviour was to log everything related to the scheduled event, which included un-necessary data like the `retry_conf`,`header_conf`. 

In this PR, I have made the change in the request body and while logging the invocation, only the request that has been sent will be included, nothing more.

tirumaraiselvan:
Yeah, I agree that a changelog here should be good to have.

Also, would be nice to address the comment here: https://github.com/hasura/graphql-engine/pull/5262/files#r448772984

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
codingkarthik: Currently, the `WebhookRequest` is being used by Event triggers also, that's why I was hesitant to make the `_rqPayload`-> `_rqBody` change. I'll add a new type for scheduled triggers to use `body` instead of `payload`.

tirumaraiselvan: I wonder if changing the invocation log format for event triggers would be an invasive change. I would like to hear more opinions but I am in favour of changing the name to `body` for both. It is clearer for the user as well.

tirumaraiselvan:
I also realized that this makes one super cool trick unusable. 

Suppose you want to invoke a fixed mutation on hasura in your webhook business logic. Refer https://github.com/hasura/graphql-engine/issues/5199

Then you could just configure a standard graphql json payload:

```
{
  "query": " mutation { update_table { ... ] }",
  "operationName": "cron-mutation",
  "variables": { "myVariable": "someValue", ... }
}
```

And it just works. But now the body has metadata and payload is nested inside, this won't work. :(

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
lexi-lambda: I don’t really have any opinion about whether we go with `body` or `payload`, but I do think we should be consistent about it! And if we already use `body` in event triggers, it sounds like that sort of forces our hand due to backwards-compatibility constraints. So I would be in favor of sharing the datatype and request format if at all possible.

lexi-lambda: …wait, no, I think I misunderstood. It sounds like event triggers already use `payload`, right? Not `body`. In which case everything I said still applies, except that we’re forced to use `payload` instead of `body`.

lexi-lambda: Okay, from a conversation on Slack, we worked out the confusion: I was trying to talk about the event trigger request payload, but the event trigger request payload is meaningfully different, so there’s no sharing opportunities there. Given that, I’m fine with this PR.

jflambert:
Hi guys, I'm trying to wrap my head around this change. Why was it necessary? I can't seem to run_sql with cron triggers anymore (yes, I was using the beta to some extent, but not production)

```
{
    "type" : "create_cron_trigger",
    "args" : {
        "name": "sample_cron",
        "webhook": "http://localhost:8080/v1/query",
        "schedule":  "* * * * *",
        "payload": {"args":{"sql":"TRUNCATE hdb_catalog.event_invocation_logs; DELETE FROM hdb_catalog.event_log WHERE delivered = TRUE OR error = TRUE;"},"type":"run_sql"},
        "include_in_metadata":true,
        "comment":"sample_cron commment"
    }
}
```

This doesn't work anymore, I keep getting this error:

```
{
    "data": {
        "status": 400,
        "body": "{\"path\":\"$\",\"error\":\"key \\\"type\\\" not found\",\"code\":\"parse-failed\"}",
        "headers": [
            {
                "value": "chunked",
                "name": "Transfer-Encoding"
            },
            {
                "value": "Wed, 22 Jul 2020 00:52:11 GMT",
                "name": "Date"
            },
            {
                "value": "Warp/3.3.10",
                "name": "Server"
            },
            {
                "value": "application/json; charset=utf-8",
                "name": "Content-Type"
            }
        ]
    },
    "version": "1",
    "type": "webhook_response"
}
```

jflambert:
[I ended up logging a defect.](https://github.com/hasura/graphql-engine/issues/5438)

PR #5280, opened 2020-07-03
docs: improve getting started flow (close #5279)

### LINKED ISSUE #5279: improve getting started flow
- We want to improve the docs getting started flow, per Monday's meeting.
- The Learn courses should be discoverable from the landing page.
- Possibly a landing page with columns.

sezgi:
### Description

- We want to improve the docs getting started flow, per Monday's meeting.
- The Learn courses should be discoverable from the landing page.
- Possibly a landing page with columns.

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues
#5279

### DIFF
--- a/docs/graphql/manual/index.rst
+++ b/docs/graphql/manual/index.rst
@@ -26,16 +26,16 @@ The Hasura GraphQL engine connects to your databases & microservices and auto-ge
      - `Authentication <auth/authentication/index>`_
    * - `Schema <schema/index>`_
      - `Remote Schemas <remote-schemas/index>`_
-     - `Authorization <auth/authorization/index>`_
+     - `Auth Using Webhooks <auth/authentication/webhook>`_
    * - `Queries <queries/index>`_
      - `Event Triggers <event-triggers/index>`_
-     - `Access Control Basics <auth/authorization/basics>`_
+     - `Auth Using JWT <auth/authentication/jwt>`_
    * - `Mutations <mutations/index>`_
      - `Scheduled Triggers <scheduled-triggers/index>`_
-     - `Roles & Session Variables <auth/authorization/roles-variables>`_
+     - `Unauthenticated / Public Access <auth/authentication/unauthenticated-access>`_
    * - `Subscriptions <subscriptions/index>`_
      - 
-     - `Permission Rules <auth/authorization/permission-rules>`_
+     - `Authorization / Access Control <auth/authorization/index>`_
 
 .. list-table::
    :header-rows: 1
--- a/docs/graphql/manual/index.rst
+++ b/docs/graphql/manual/index.rst
@@ -12,12 +12,62 @@ Hasura GraphQL Engine Documentation
   :depth: 1
   :local:
 
-The Hasura GraphQL engine lets you set up a GraphQL server and event triggers over a Postgres database in minutes.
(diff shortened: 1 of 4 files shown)


docs/graphql/manual/index.rst
```diff
@@ -12,12 +12,62 @@ Hasura GraphQL engine documentation
...
+     - Auth
+   * - `Getting Started <getting-started/index>`_
+     - `Actions <actions/index>`_
+     - `Authentication <auth/authentication/index>`_
+   * - `Schema <schema/index>`_
+     - `Remote Schemas <remote-schemas/index>`_
+     - `Authorization <auth/authorization/index>`_
+   * - `Queries <queries/index>`_
+     - `Event Triggers <event-triggers/index>`_
+     - `Access Control Basics <auth/authorization/basics>`_
```
marionschleifer, line 32: @sezgi it might be a bit confusing to have `Authentication` and `Authorization` and then the other 3 on the same level even though they belong under `Authorization`. From the structure it looks like they are all on the same hierarchy. Maybe `Authentication` and `Authorization` is enough?

docs/graphql/manual/index.rst
```diff
@@ -12,12 +12,62 @@ Hasura GraphQL engine documentation
...
+
+.. list-table::
+   :header-rows: 1
+   :class: toc-table
+
+   * - Migrate & Deploy
+     - Reference
+     - Learn
+   * - `Basics <migrations/basics>`_
+     - `Hasura CLI <hasura-cli/index>`_
```
marionschleifer, line 48: Hasura CLI is more like its own thing rather than a reference. It's usually used in the context of migrations & metadata. Maybe it can go under `Migrate & Deploy`?

docs/graphql/manual/index.rst
sezgi: I thought of these more as shortcuts than as a hierarchical nav. Maybe we can refer to traffic data for this to see the most used ones (I'll check today).

sezgi: Also only two links would make the list too short visually

docs/graphql/manual/index.rst
sezgi: Thinking about this now. This would make the layout a bit lopsided. Maybe we keep it under reference? It actually is a kind of reference, with the list of commands and info on how they're used.

sezgi: Also easier to find there at the top

docs/graphql/manual/index.rst
sezgi: changed the auth section to highlight authn sections instead of authz sections (based on traffic data)

marionschleifer:
Thanks @sezgi 🙂 

Just now also checked the mobile view:

image

image

Is there any way to optimize for this? Usually I think it's not that important for most pages (especially if they're a reference section that's nested under other sections) but this will be the first one users come to when opening our docs. So might be a bit more important? What are your thoughts @sezgi?

sezgi:
@marionschleifer thanks for catching that! I didn't expect this idea to go from proposal to reality so soon, so I missed that 😆  

I completely agree. Will fix.

sezgi:
@marionschleifer please see update

PR #5303, opened 2020-07-05
Removed Dynamic action types and unwanted store configuration

soorajshankar:
### Description
Removed Dynamic action types and unwanted store configuration
As part of this cleanup, 
* some redux, and component cleanup is done
* Migrated 2 reducers and few files to TS.
* Reuse of HeaderConfEditor is more generic now, and it is used across Actions and Remote schemas.

### Changelog

### Affected components
HeadersConfig component (used in both RemoteSchema and Actions)

- [x] Console

### Solution and Design
* TS migration of Header Reducer
* Refactored dynamic action types to static action types 

> Currently both remote schema and Actions have headers configuration section, which is separate components with slight UI changes.
* In Remote schema section
image
 * In Actions 
image

Unifying this UI with the RemoteSchema UI for consistency.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


console/src/components/Services/RemoteSchema/Actions.ts
```diff
@@ -134,16 +147,19 @@ const listReducer = (state = listState, action) => {
...
 const makeRequest = (
-  upQueries,
-  downQueries,
-  migrationName,
-  customOnSuccess,
-  customOnError,
-  requestMsg,
-  successMsg,
-  errorMsg
+  upQueries: unknown[],
```
soorajshankar, line 150: @beerose These types will change when the schema down changes gets merged.

PR #5313, opened 2020-07-06
console: add statement timeout for count query

beerose:
### Description

Estimated count as we're doing it now doesn't work for partitioned and foreign tables. So no matter of the table size we would do normal count (`select count(*) from ...`) which may cause db perf issues for big tables. 

This PR adds statement timeout 2s for count queries. 

### Changelog

### Affected components

- [x] Console


console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -140,7 +141,9 @@ const vMakeCountRequest = () => {
       view.query.order_by
     );
 
-    const timeoutQuery = getRunSqlQuery(getStatementTimeoutSql(2));
+    const timeoutQuery = getRunSqlQuery(
+      getStatementTimeoutSql(2) + `select pg_sleep(5)`
```
beerose, line 145:
```suggestion
      getStatementTimeoutSql(2)
```

console/src/components/Services/Data/TableBrowseRows/ViewRows.js
```diff
@@ -950,6 +951,7 @@ const ViewRows = ({
           persistColumnOrderChange(curTableName, currentSchema, reorderData)
         }
         defaultReorders={columnsOrder}
+        showPagination={!shouldHidePagination}
```
soorajshankar, line 954: Can we use the same prop name throughout the tree, to avoid confusion? :see_no_evil:

console/src/components/Services/Data/TableBrowseRows/ViewActions.js
```diff
@@ -31,6 +33,7 @@ const V_EXPAND_ROW = 'ViewTable/V_EXPAND_ROW';
 const V_COLLAPSE_ROW = 'ViewTable/V_COLLAPSE_ROW';
 
 const V_COUNT_REQUEST_SUCCESS = 'ViewTable/V_COUNT_REQUEST_SUCCESS';
+const V_COUNT_REQUEST_ERROR = 'ViewTable/V_COUNT_REQUEST_SUCCESS';
```
nigelainscoe, line 37: Constant looks wrong to me. Surely should be  'ViewTable/V_COUNT_REQUEST_ERROR'

PR #5321, opened 2020-07-07
Note libkrb5-dev dependency on Debian

evertedsphere:
On Debian, failing to install `libkrb5-dev` causes `libpq-client` to fail to compile (it can't find the `<gssapi/gssapi.h>` include in one of the `cbits/` files), which causes the entire `graphql-engine` build to fail.

## Changelog

No user-facing changes.

### Affected components
- [x] Docs

### DIFF
--- a/server/CONTRIBUTING.md
+++ b/server/CONTRIBUTING.md
@@ -5,7 +5,7 @@ own machine and how to contribute.
 
 ## Pre-requisites
 
-- [GHC](https://www.haskell.org/ghc/) 8.6.5 and [cabal-install](https://cabal.readthedocs.io/en/latest/)
+- [GHC](https://www.haskell.org/ghc/) 8.10.1 and [cabal-install](https://cabal.readthedocs.io/en/latest/)
   - There are various ways these can be installed, but [ghcup](https://www.haskell.org/ghcup/) is a good choice if you’re not sure.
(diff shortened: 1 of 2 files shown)


server/CONTRIBUTING.md
```diff
@@ -11,11 +11,12 @@ own machine and how to contribute.
 - npm >= 5.7
 - [gsutil](https://cloud.google.com/storage/docs/gsutil)
 - libpq-dev
+- libkrb5-dev
 - python >= 3.5 with pip3
 
-The last two prerequisites can be installed on Debian with:
+The last two prerequisites can be installed on Debian or Ubuntu with:
```
Naveenaidu, line 17: Is writing `Ubuntu` necessary, since Ubuntu is Debian based? Any specific reason for adding it ?

evertedsphere: I think it doesn't hurt to write, and it's possible that someone (especially a newer dev) might not be aware that Debian and Ubuntu packages have the same names. Many more people use Ubuntu than Debian, so I thought it might be worth adding.

tirumaraiselvan:
While we are at this, could you also change `8.6.5` -> `8.10.1` in the first line :) ?

evertedsphere:
@tirumaraiselvan done! this should be ready to merge now :tada:

evertedsphere:
See #5377 instead.

PR #5327, opened 2020-07-08
console: remove obsolete files and update dependencies

kolharsam:
fixes https://github.com/hasura/graphql-engine-internal/issues/443

### Description

Removes certain `*.min.js` files that were used on the console and uses their npm packages instead for their usage on the console.


console/package.json
```diff
@@ -59,6 +61,7 @@
     "graphiql-explorer": "0.6.2",
     "graphql": "14.5.8",
     "graphql-voyager": "1.0.0-rc.29",
+    "highlight.js": "^9.15.8",
```
soorajshankar, line 64: Can we use specific versions?

console/package.json
```diff
@@ -90,6 +93,7 @@
     "redux-thunk": "2.3.0",
     "sanitize-filename": "1.6.3",
     "semver": "5.5.1",
+    "sql-formatter": "^2.3.3",
```
soorajshankar, line 96: @kolharsam we have the same issue here as well

console/package.json
```diff
@@ -50,6 +50,8 @@
   "dependencies": {
     "@graphql-codegen/core": "1.13.5",
     "@graphql-codegen/typescript": "1.13.5",
+    "@types/highlight.js": "^9.12.4",
```
soorajshankar, line 53: +1

PR #5332, opened 2020-07-09
Replace idle GC with a custom GC thread

jberryman:
The current idle GC settings seem never to cause idle GC to trigger.
The changes here at least help memory usage to look more reasonable when
running certain benchmarks, and speculatively could partially fix some
memory leaks users have reported.

See `ourIdleGC` for details.

Referencing canonical memory issue #3388

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#3388 etc

### Steps to test and verify

Try some of the repros Tiru has been looking at. Observe memory drops somewhat after a few seconds after closing a benchmark, matching behavior of `-I0.3`

I thought i would use `-Iw` but it is actually not in GHC 8.10.1, and in any case actually doesn't get us all the way to what we want. See implementation.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


tirumaraiselvan:
@jberryman Running this build with `+RTS -S`, this is what I am observing:

```  
  3762968   1973832 599000392  0.091  0.046  232.404  310.529    0    0  (Gen:  0)
  3706056   1900912 599039392  0.091  0.045  232.914  393.125    0    0  (Gen:  0)
  3100080   1874128 599038912  0.088  0.044  233.413  475.471    0    0  (Gen:  0)
  2943984   1874568 599039968  0.074  0.038  233.890  557.797    0    0  (Gen:  0)
  2860168   1873512 599038912  0.095  0.051  234.331  640.129    0    0  (Gen:  0)
  2865752   1873488 599038888  0.105  0.053  234.823  722.477    0    0  (Gen:  0)
```

I expected IdleGC to be invoked every 10s or every 60s but it feels like it's closer to 80s? Am I reading this right?

tirumaraiselvan:
Running the benchmarks, this is what I observed right after saturating all connections:

```
  2152056   1467712 493169176  0.045  0.022  295.455  814.474    0    0  (Gen:  0)
  2136088   1364976 493222168  0.045  0.022  295.511  814.538    0    0  (Gen:  0)
   619312 180317008 461511080  0.472  0.236  295.985  814.782    0    0  (Gen:  1)
{"type":"unstructured","timestamp":"2020-07-09T13:45:54.450+0530","level":"warn","detail":"Overdue for a major GC: forcing one even though we don't appear to be idle"}
  2134680   1210296 461601632  0.014  0.007  296.008  814.794    0    0  (Gen:  0)
```

Till the point the log line got printed, my RES was around 630MB. But right after the GC invocation log-line, the memory went to 820MB and stayed constant there henceforth. Is that expected?

The subsequent forced GC invocations didn't have any effect on RES (RES remained at 820MB). Just the first one spiked it.

After closing the bechmark the memory had a big drop to ~450MB.

This is the final report on quitting the server:

```
  21,859,717,808 bytes allocated in the heap
  15,951,097,048 bytes copied during GC
     463,080,184 bytes maximum residency (74 sample(s))
       5,370,152 bytes maximum slop
             667 MiB total memory in use (0 MB lost due to fragmentation)

                                     Tot time (elapsed)  Avg pause  Max pause
  Gen  0      7532 colls,  7532 par   339.189s  176.610s     0.0234s    0.0824s
  Gen  1        74 colls,    73 par    8.200s   4.179s     0.0565s    0.2568s

  Parallel GC work balance: 24.35% (serial 0%, perfect 100%)

  TASKS: 28 (1 bound, 26 peak workers (27 total), using -N4)

  SPARKS: 0 (0 converted, 0 overflowed, 0 dud, 0 GC'd, 0 fizzled)

  INIT    time    0.001s  (  0.001s elapsed)
  MUT     time  192.863s  (1054.078s elapsed)
  GC      time  347.389s  (180.789s elapsed)
  EXIT    time    0.003s  (  0.003s elapsed)
  Total   time  540.256s  (1234.870s elapsed)

  Alloc rate    113,343,434 bytes per MUT second

  Productivity  35.7% of total user, 85.4% of total elapsed
```

I guess the code is working as expected but any reason for the "spike"?

jberryman:
> Is that expected?

I really don't know... I suppose we shouldn't be surprised; maybe we had to allocate some new memory for the old generation in which to copy during GC?

tirumaraiselvan:
> > Is that expected?
> 
> I really don't know... I suppose we shouldn't be surprised; maybe we had to allocate some new memory for the old generation in which to copy during GC?

Right. FYI, with `-G1` this spike doesn't happen.

PR #5346, opened 2020-07-09
console: fix custom root field placeholder for insert one mutation (close #5345)

### LINKED ISSUE #5345: Placeholder text in Custom Root Fields editor showing incorrect root field for _one mutation
The default should actually be `insert_operators_one` not `insert_one_operators`.

Screen Shot 2020-07-10 at 9 25 07 AM

Took this chance to do my first PR for Hasura.

toddheslin:
### Description
Simply updating the placeholder text shown to the user to match the default root field name that Hasura provides out of the box.

### Changelog

### Affected components

- [X] Console

### Related Issues

#5345 

### Solution and Design

Copied the pattern used to replace the `select` keyword in the `getDefaultRootField` function. It could be written a bit cleaner if we had a lot of conditionals but this is readable.


console/src/components/Services/Data/Common/Components/RootFieldEditor.js
```diff
@@ -33,6 +33,9 @@ const RootFieldEditor = ({
     if (rfType.includes('select')) {
       return rfType.replace('select', tableName);
     }
+    if (rfType.includes('one')) {
+      return rfType.replace('one', tableName) + '_one'
```
beerose, line 37:
```suggestion
      return rfType.replace('one', tableName) + '_one';
```

PR #5360, opened 2020-07-12
console: add "identity" to frequently used columns (close #4279)

### LINKED ISSUE #4279: add "identity" column ID + PK in frequently used columns list
In Postgres 10, a feature called "identity columns" was released:
https://www.2ndquadrant.com/en/blog/postgresql-10-identity-columns

The difference is:
```sql
CREATE TABLE test_old (
    id serial PRIMARY KEY,
    payload text
);

CREATE TABLE test_new (
    id int GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
    payload text
);
```

The issue with "serial", is that it's expanded into this:
```sql
CREATE SEQUENCE test_old_id_seq;

CREATE TABLE test_old (
    id int NOT NULL PRIMARY KEY,
    payload text
);

ALTER TABLE test_old
    ALTER COLUMN id SET DEFAULT nextval('test_old_id_seq');

ALTER SEQUENCE test_old_id_seq OWNED BY test_old.id;
```

Besides that it creates a lot of unnecessary overhead as a side-effect, the permissions on sequences need to be managed separately/individually, and it makes for a huge pain when attempting to do something like this:

```sql
INSERT INTO test_old (payload) VALUES ('a'), ('b'), ('c') RETURNING *;

CREATE TABLE test_old2 (LIKE test_old INCLUDING ALL);
INSERT INTO test_old2 (payload) VALUES ('e') RETURNING *;
 id | payload
----+---------
  4 | e          <-- keeps the sequence from the parent table
```

Because Hasura doesn't currently have a button to copy an existing table structure, and `serial` expands into an owned-sequence so you can't use `LIKE some_table`, it becomes tedious to make variants of tables.

Now the tables are undesirably/unintentionally linked, and trying to fix your mistake becomes another pain-point:
```sql
=> DROP TABLE test_old;
ERROR:  cannot drop table test_old because other objects depend on it
DETAIL:  default for table test_old2 column id depends on sequence test_old_id_seq
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

It would be really nice to have `int GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY` be the default option (because I assume this is what most people would rather have) or at least a Frequently-used option :pray:

GavinRay97:
### Description

Note: The types for the component and the note mentioned below need to be checked by @beerose, I just took a best guess

This PR:
1. Adds the option to use Identity columns from the "frequently used columns" dropdown.
2. Refactors `FrequentlyUsedColumns.js` to Typescript/TSX. Additionally it provides WIP typings for Postgres column types (which aren't able to be used 100% properly here due to design limitation with the SQL generation, see note).

Identity columns are the ideal format for ID columns, they avoid issues with Serial columns and the owned sequences they create. However, they were only introduced in PG 10, so there's a disclaimer in the menu text.

See here for more discussion:
https://github.com/hasura/graphql-engine/issues/4279

_*Note: this is due to this bit in the table SQL generation:_
```js
for (let i = 0; i < currentCols.length; i++) {
  tableDefSql +=
    '"' + currentCols[i].name + '"' + ' ' + currentCols[i].type;
```

It's impossible to generate a type which has modifier text between `name` and `type`, as is the case for identity columns:
```sql
id int GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
```

We would need some sort of modifiers array which can insert text before the `NOT NULL`/`PRIMARY KEY`, etc.

**Demo:**

identity-column-new

### Changelog

### Affected components

- [x] Console

### Related Issues

close #4279 (add "identity" column ID + PK in frequently used columns list)

### Steps to test and verify

Pull the changes, run it and create a table. For easier debugging, edit the Redux action to log out the SQL text and stop the actual creation:

```js
/* console/src/components/Services/Data/Add/AddActions.js, line 390 */

console.log('TABLE SQL:', tableDefSql)

// makeMigrationCall(
//   dispatch,
//   getState,
//   upQuery.args,
//   downQuery.args,
//   migrationName,
//   customOnSuccess,
//   customOnError,
//   requestMsg,
//   successMsg,
//   errorMsg,
//   true
// );
```

It puts a pretty large string in the column description field, `int GENERATED BY DEFAULT AS IDENTITY`. This would be better as something else but it's because of the limitation above.

#### Breaking changes
- [x] No Breaking changes


beerose:
@rikinsk Do you think we can go with the current solution or should we check for Postgres version first and if it's < 10, filter out identity type from frequently used?

PR #5364, opened 2020-07-13
console: fix default option selected when cloning row in data browser (close #5340)

### LINKED ISSUE #5340: default option is selected while cloning row in data browser
If a column has a default value defined, then while cloning a row, the default radio-btn is selected instead of the input box radio-btn.

For example,

**Current behaviour**
Screenshot from 2020-07-09 20-25-14

**Expected behaviour**
Screenshot from 2020-07-09 20-29-11

ShahAnuj2610:
close #5340 

### Changelog

### Affected components

- [x] Console

### Solution and Design
- check if it's a clone action and remove default radio for fields which have values excluding timestamp, uuid and disabled fields (e.g. auto increment)

### Steps to test and verify
- tried cloning the row, inclusive of the above fields

- as per my knowledge, timestamp, uuid and disabled fields (auto increment) are the fields which should have default selected even when cloning.
- There could be other fields like above which I might have missed.


console/src/components/Services/Data/Common/Components/TableRow.tsx
```diff
@@ -7,7 +7,11 @@ import {
 import styles from '../../../../Common/TableCommon/Table.scss';
 import { TypedInput } from './TypedInput';
 
-const getColumnInfo = (col: TableColumn, prevValue?: unknown) => {
+const getColumnInfo = (
+  col: TableColumn,
+  prevValue?: unknown,
+  clone?: Record<string, any>
```
soorajshankar, line 13: can we have the following type here? and use this same type for `TableRowProps`
```suggestion
  clone?: Record<string, number | string | boolean>
  // or 
  clone?: Record<string, unknown>
```
we will try to avoid `any` as much as possible

soorajshankar (dismissed):
Thanks @ShahAnuj2610 for the PR!

Left a few code review comments, please fix that.

console/src/components/Services/Data/Common/Components/TableRow.tsx
beerose: @soorajshankar, we don't know what is there. There can be literally `any` type. It won't cause less type safety, we just allow it to be anything. We can have json/jsonb, array data types, etc, right? We don't much care what exactly this thing is. This is a value stored in a table row.

beerose: `unknown` will work here.

soorajshankar: changing my suggested code to only `unknown` suggestion

CHANGELOG.md
```diff
@@ -32,7 +32,7 @@
 - server: have haskell runtime release blocks of memory back to the OS eagerly (related to #3388)
 - server: unlock locked scheduled events on graceful shutdown (#4928)
 - server: disable prepared statements for mutations as we end up with single-use objects which result in excessive memory consumption for mutation heavy workloads (#5255)
-- server: include scheduled event metadata (`created_at`,`scheduled_time`,`id`, etc) along with the configured payload in the request body to the webhook. 
+- server: include scheduled event metadata (`created_at`,`scheduled_time`,`id`, etc) along with the configured payload in the request body to the webhook.
```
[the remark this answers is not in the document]
ShahAnuj2610: @soorajshankar I think it was due to `husky` pre-commit hooks. Lmk, if it's not needed, I can remove it.

console/src/components/Services/Data/Common/Components/TableRow.tsx
beerose: However, note that `unknown` is not always better than any. When we don't care about the type (type can be anything) and we annotate it with `unknown` then it may be very hard to work with it. You basically can't do anything with an unknown, so every time you want to perform something on the value annotated as unkown you'd need add types assertion `as any` anyway.

console/src/components/Services/Data/Common/Components/TableRow.tsx
```diff
@@ -24,6 +28,12 @@ const getColumnInfo = (col: TableColumn, prevValue?: unknown) => {
 
   let columnValueType;
   switch (true) {
+    case clone &&
+      clone[col.column_name] &&
+      !isDisabled &&
+      !['uuid', 'timestamptz'].includes(col.udt_name):
+      columnValueType = 'value';
+      break;
     case !isEditing && (isIdentity || hasDefault || isGenerated):
```
beerose, line 37: We can simplify it.
Instead of adding another `case` we can handle it along with the first case as:
```
case !isEditing && !clone && (isIdentity || hasDefault || isGenerated):
```
Notice added `!clone` part.

beerose:
Hi @ShahAnuj2610! I hope you're doing well. I just wanted to ask if you plan to continue working on this PR? 🙂

CHANGELOG.md
```diff
@@ -528,6 +528,7 @@ See [upgrade docs](https://hasura.io/docs/1.0/graphql/manual/migrations/upgrade-
 - console: misc actions fixes (#4059)
 - console: action relationship page improvements (fix #4062, #4130) (#4133)
 - console: add code exporter to graphiql (close #4531) #4652
+- console: fix default option selected when cloning row in data browser (close #5340) #5364
```
beerose, line 531:
```suggestion
```

beerose (approved):
Console changes.

tirumaraiselvan (approved):
no cl required

PR #5374, opened 2020-07-14
Pass environment variables around as a data structure, via @sordina

sordina:
### Description

Environment variables are currently often accessed through IO calls to getEnvironment, and lookupEnv, etc.

This causes issues:

* Running two or more GraphQL-Engine threads in the same process with different environments is impossible
* Redundant IO calls are performed in code that could easily otherwise be pure

This is a refactoring to only require an IO call at the main entry-point and then subsequent passing of environment through explicit arguments and implicit structures and monads.

### Affected components

- [x] Server

### Solution and Design

The solution was to work bottom-up:

* The creation of a new `Data.Environment` module that exposed an IO call to create the environment, then all subsequent functions to access and lookup the environment as pure operations on `Environment`
* Replacing existing IO environment calls with pure `Data.Environment` calls.
* Resolution of errors to pass the environment as an argument to these functions
* Refactoring the argument into existing datatypes and monads in scope where appropriate

### Steps to test and verify

* Existing test-suite should ensure no regressions are made for existing behaviour.
* Explicit testing of environment accessing interfaces such as Events, Actions, etc.

The introduction of environment arguments is sometimes cumbersome. Further scrutiny of the implementation may reveal opportunities for more elegant or concise implementation through reader context, etc.

### Server checklist

* Ensure test coverage is adequate for environment-related functionality testing

#### Catalog upgrade

- [X] No

#### Metadata

- [X] No

#### GraphQL

#### Breaking changes

* Library usage of GraphQL-Engine may now require explicit interaction with `Data.Environment`

No UI driven braking changes have been introduced.


server/src-lib/Hasura/Eventing/EventTrigger.hs
```diff
@@ -158,45 +158,56 @@ initEventEngineCtx maxT _eeCtxFetchInterval = do
...
-             link t
-
-        -- return when next batch ready; some 'processEvent' threads may be running.
-        wait eventsNextA
+          processEvent event
+            & withEventEngineCtx eeCtx
+            & flip runReaderT (logger, httpMgr)
+          -- removing an event from the _eeCtxLockedEvents after the event has
+          -- been processed
+          removeEventFromLockedEvents (eId event) leEvents
```
0x777, line 209: looks like the `link t` part is missing?

server/src-lib/Hasura/Logging.hs
```diff
@@ -108,6 +108,7 @@ data InternalLogTypes
  | ILTJwkRefreshLog
  | ILTTelemetry
  | ILTSchemaSyncThread
+ | ILTPollerLog
```
0x777, line 111: Shouldn't be present, probably a bad merge somewhere up the chain.

server/src-lib/Hasura/Eventing/EventTrigger.hs
sordina: Good catch... Strange that that got removed.

sordina: Ah actually, is there a thread to be linked?

sordina: Seems like this action is no longer performed async, so there's nothing to link.

server/src-lib/Hasura/Logging.hs
sordina: So remove ILTPollerLog references? There's also a case-match on line 126.

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -25,6 +25,13 @@ import qualified Data.Text                 as T
 import qualified Network.HTTP.Types        as HTTP
 import qualified Network.Wai.Extended      as Wai
 
+-- import qualified Data.ByteString.Lazy                   as BL
```
0x777, line 28: We can remove these?

sordina: Don't see why not. We can always reimport them if they're required.

server/src-lib/Hasura/Server/Auth.hs
```diff
@@ -1,4 +1,6 @@
-{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE DerivingStrategies         #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
```
0x777, line 2: `RecordWildCards` and `GeneralizedNewtypeDeriving` is already part of cabal's `default-extensions`.

sordina: Should I remove all references to these extensions? Or just from my PR?

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -120,22 +122,22 @@ runHasuraGQ
...
   (E.ExecutionCtx logger _ pgExecCtx _ _ _ _ _) <- ask
-  logQuery' logger
-  (telemTimeIO, respE) <- withElapsedTime $ liftIO $ runExceptT $ case resolvedOp of
-    E.ExOpQuery tx _genSql -> do
+  (telemTimeIO, respE) <- withElapsedTime $ runExceptT $ case resolvedOp of
+    E.ExOpQuery tx genSql _asts -> do
       -- log the generated SQL and the graphql query
-      -- L.unLogger logger $ QueryLog query genSql reqId
-      ([],) <$> runQueryTx pgExecCtx tx
+      logQueryLog logger query genSql reqId
```
0x777, line 136: @sordina Do you have any idea on what these changes are related to?

server/src-lib/Hasura/Eventing/EventTrigger.hs
tirumaraiselvan: This was a regression in an earlier master. `processEvent` needs to be async.

server/src-lib/Hasura/GraphQL/Execute/Query.hs
```diff
@@ -196,51 +202,61 @@ convertQuerySelSet
...
-  => HTTP.Manager
+  => Env.Environment
+  -> HTTP.Manager
   -> [N.Header]
   -> QueryReusability
   -> V.ObjectSelectionSet
   -> QueryActionExecuter
-  -> m (LazyRespTx, Maybe ReusableQueryPlan, GeneratedSqlMap)
-convertQuerySelSet manager reqHdrs initialReusability selSet actionRunner = do
+  -> m (tx EncJSON, Maybe ReusableQueryPlan, GeneratedSqlMap, [R.QueryRootFldUnresolved])
```
0x777, line 214: Is this related to caching?

server/src-lib/Hasura/Eventing/EventTrigger.hs
sordina: Ah ok. I'll try patching this.

server/src-lib/Hasura/GraphQL/Execute/Query.hs
sordina: This does appear to be related to caching. I think that maybe it was included accidentally when splitting the PR logically into project, env, and tracing/caching.

server/src-lib/Hasura/Eventing/EventTrigger.hs
sordina: I've put in an async call and link again:


```
      eventsNext <- LA.withAsync popEventsBatch $ \eventsNextA -> do
        -- process approximately in order, minding HASURA_GRAPHQL_EVENTS_HTTP_POOL_SIZE:
        forM_ events $ \event -> do
          t <- processEvent event
            & withEventEngineCtx eeCtx
            & flip runReaderT (logger, httpMgr)
            & LA.async
          -- removing an event from the _eeCtxLockedEvents after the event has
          -- been processed
          removeEventFromLockedEvents (eId event) leEvents
          LA.link t
        LA.wait eventsNextA
```

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
sordina: This is probably a leaked bit of the tracing implementation.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -340,19 +344,24 @@ onStart serverEnv wsConn (StartMsg opId q) = catchAndIgnore $ do
...
+      -> GQLReqUnparsed
+      -> UserInfo
+      -> E.ExecOp (LazyTx QErr)
+      -> ExceptT () m ()
     runHasuraGQ timerTot telemCacheHit reqId query userInfo = \case
-      E.ExOpQuery opTx genSql ->
-        execQueryOrMut Telem.Query genSql $ runQueryTx pgExecCtx opTx
+      E.ExOpQuery opTx genSql _asts ->
+        execQueryOrMut Telem.Query genSql . fmap snd $
+          ([],) <$> runLazyTx pgExecCtx Q.ReadOnly opTx
```
0x777, line 358:
```suggestion
        execQueryOrMut Telem.Query genSql $ runQueryTx pgExecCtx opTx 
```

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -120,22 +122,22 @@ runHasuraGQ
...
+      ([],) <$> runLazyTx pgExecCtx Q.ReadOnly tx
```
0x777, line 137:
```suggestion
      ([],) <$> runQueryTx pgExecCtx tx
```

PR #5383, opened 2020-07-14
Add MonadTrace and MonadExecuteQuery abstractions

paf31:
### Description

A combination of two changes:

* Adding `MonadExecuteQuery` to support switching out the code we use to actually execute compiled SQL code
* Adding `MonadTrace` and `HasReporter` to support tracing with a flexible tracing backend implementation

### Changelog

### Affected components

- [x] Server

### Related Issues

* TODO

### Solution and Design

Tracing interfaces and implementation are provided for adding the ability to perform fine-grained tracing throughout actions as they are performed by the application. Implementation can also be deferred so that library-use of GraphQL-Engine can implement different tracing mechanisms as required.

* The addition of two new classes in `Hasura.Tracing` - `HasReporter` and `MonadTrace` provide an interface for tracing actions
* An implementation via `TraceT` to allow for tracing to be performed

Caching foundations have been implemented to allow for data cahcing of queries and actions.

### Steps to test and verify

* Existing test-suites and integration testing should be performed to ensure that this has not introduced any regressions as current behaviour of the application should remain the same
* No impact on the user-interfaces in graphql-engine - Downstream codebases should be tested to verify that custom tracing implementations are executed as expected.

#### Catalog upgrade

- [x] No
#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No

### DIFF
--- a/server/src-lib/Hasura/Tracing.hs
+++ b/server/src-lib/Hasura/Tracing.hs
@@ -1,0 +1,237 @@
+{-# LANGUAGE RecordWildCards      #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Hasura.Tracing
+  ( MonadTrace(..)
+  , TraceT
+  , runTraceT
+  , runTraceTWith
+  , runTraceTWithReporter
+  , runTraceTInContext
+  , interpTraceT
+  , TraceContext(..)
+  , Reporter(..)
+  , noReporter
+  , HasReporter(..)
+  , TracingMetadata
+  , SuspendedRequest(..)
+  , extractHttpContext
+  , traceHttpRequest
+  ) where
+
+import           Hasura.Prelude
+import           Control.Monad.Trans.Control
+import           Control.Monad.Morph
+import           Control.Monad.Unique
+import           Data.String                 (fromString)
+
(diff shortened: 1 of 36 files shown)


server/src-lib/Hasura/Tracing.hs
```diff
@@ -0,0 +1,237 @@
...
+  let reqBytes = case HTTP.requestBody req of
+        HTTP.RequestBodyBS bs -> Just (fromIntegral (BS.length bs))
+        HTTP.RequestBodyLBS bs -> Just (BL.length bs)
+        HTTP.RequestBodyBuilder len _ -> Just len
+        HTTP.RequestBodyStream len _ -> Just len
+        _ -> Nothing
+  for_ reqBytes \b ->
+    attachMetadata [("request_body_bytes", fromString (show b))]
+  TraceContext{..} <- currentContext
+  let tracingHeaders =
```
paf31, line 230: We might want to consider using the B3 spec from OpenZipkin here https://github.com/openzipkin/b3-propagation

paf31: (I don't think this should block a merge though, it's easily changed later)

PR #5395, opened 2020-07-16
docs: fix indexing problem 

marionschleifer:
### Description

[This index change](https://github.com/hasura/graphql-engine/pull/5282) has led to indexing problems in production. So we'll revert it and find for another solution after that.

### Changelog

### Affected components

- [x] Docs

### DIFF
--- a/docs/_theme/djangodocs/layout.html
+++ b/docs/_theme/djangodocs/layout.html
@@ -24,7 +24,7 @@
 
 {% block linktags %}
 {{ super() }}
-  <link rel="canonical" href="https://hasura.io/docs/1.0/{{ pagename }}{{ file_suffix }}" />
+  <link rel="canonical" href="https://{{ BASE_DOMAIN }}/docs/1.0/{{ pagename }}{{ file_suffix }}" />
 {% endblock %}
 
 {% block document %}
--- a/docs/_theme/djangodocs/layout.html
+++ b/docs/_theme/djangodocs/layout.html
@@ -24,7 +24,7 @@
 
 {% block linktags %}
 {{ super() }}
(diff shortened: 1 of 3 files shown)


docs/_theme/djangodocs/layout.html
```diff
@@ -24,7 +24,7 @@
 
 {% block linktags %}
 {{ super() }}
-  <link rel="canonical" href="https://hasura.io/docs/1.0/{{ pagename }}{{ file_suffix }}" />
+  <link rel="canonical" href="https://{{ BASE_DOMAIN }}/docs/1.0/{{ pagename }}{{ file_suffix }}" />
```
praveenweb, line 27: Actually the canonical can still be hasura.io, hardcoded, irrespective of whether it is staging or production.

sezgi:
I don't have a holistic understanding of this. Replacing myself with @rikinsk as reviewer so as not to block it.

docs/conf.py
```diff
@@ -36,22 +36,24 @@
...
 # GraphiQL defaults
 GRAPHIQL_DEFAULT_ENDPOINT = "http://localhost:8080/v1/graphql"
+
 # Get from env if set
 if os.getenv("GRAPHIQL_DEFAULT_ENDPOINT"):
     GRAPHIQL_DEFAULT_ENDPOINT = os.getenv("GRAPHIQL_DEFAULT_ENDPOINT")
-
-sitemap_domain = "https://hasura-stg.hasura-app.io/docs/"
-if(BASE_DOMAIN == "production"):
+    sitemap_domain = "https://hasura-stg.hasura-app.io/docs/"
```
praveenweb, line 47: Is this intended to be inside the `if` condition?

PR #5409, opened 2020-07-17
Propagate the trace context to event triggers

paf31:
This enables the collection of traces which span a mutation and the event trigger(s) it invokes.

### Changelog

### Affected components

- [x] Server

### Solution and Design

- The trace context is written into a transaction-local variable `hasura.trace_context`
- This is picked up by the `insert_event_log` function and written into the `event_log` table
- If a event contains a trace context, that trace is extended. Otherwise, a new one is created.

### Steps to test and verify

- Create an event trigger
- Perform an insert mutation
- Observe the trace context stored on the event: `SELECT * FROM hdb_catalog.event_log WHERE payload -> 'trace_context' IS NOT NULL`

#### Catalog upgrade

- [x] Yes

#### Metadata

- [x] No

#### GraphQL
#### Breaking changes

- [x] No Breaking changes


server/src-rsr/initialise.sql
```diff
@@ -616,10 +617,16 @@ CREATE OR REPLACE FUNCTION
                   session_variables := NULL;
       END;
     END IF;
+    BEGIN
+      trace_context := current_setting('hasura.tracecontext');
```
0x777, line 621: Can you use the same `IF .. ELSE` pattern as used for `hasura.user`? (Trapping exceptions is slow in Postgres and since Postgres >= 9.6 supports a way to fetch a session variable's value without raising an exception)

paf31: Done, thanks!

server/src-rsr/migrations/37_to_36.sql
```diff
@@ -0,0 +1,41 @@
...
+    payload json;
+    session_variables json;
+    server_version_num int;
+    trace_context json;
+  BEGIN
+    id := gen_random_uuid();
+    server_version_num := current_setting('server_version_num');
+    IF server_version_num >= 90600 THEN
+      session_variables := current_setting('hasura.user', 't');
+      trace_context := current_setting('hasura.tracecontext');
```
0x777, line 16: You'll need to [explicitly pass `t`](https://www.postgresql.org/docs/current/functions-admin.html#FUNCTIONS-ADMIN-SET-TABLE) to the second argument, `missing_ok`.

server/src-rsr/initialise.sql
```diff
@@ -604,22 +604,30 @@ CREATE OR REPLACE FUNCTION
     payload json;
     session_variables json;
     server_version_num int;
+    trace_context json;
   BEGIN
     id := gen_random_uuid();
     server_version_num := current_setting('server_version_num');
     IF server_version_num >= 90600 THEN
       session_variables := current_setting('hasura.user', 't');
+      trace_context := current_setting('hasura.tracecontext');
```
0x777, line 613: Second argument should be `t`.

server/src-rsr/migrations/36_to_37.sql
```diff
@@ -0,0 +1,38 @@
+CREATE OR REPLACE FUNCTION hdb_catalog.insert_event_log(schema_name text, table_name text, trigger_name text, op text, row_data json)
```
0x777, line 1: Looks like the content of `36_to_37` and `37_to_36` is interchanged?

PR #5411, opened 2020-07-18
docs: move cloud docs

sezgi:
### Description

Because Core docs and Cloud docs share the same layout, styles, and behaviors, we are constantly duplicating code and effort whenever there's a change to any of these. This has been happening a few times a week, and will only become harder to maintain over time. There's a redesign (header & footer) coming up as well.

This PR brings the Cloud docs into the Core docs repo so that they can share code. Even though it shows **tons** of files changed, this is because I renamed parent directories (summarized below). So **it’s actually pretty quick to review this if you skip all the moved & copied files**.

Done:
- create `/cloud/` dir (copied over from `hasura.io` repo) as sibling dir to `/graphql/manual/`
- copy cloud images into `/img/cloud/`
- conditional code to hide/show sidebars and styles for active core/cloud tabs
- ~rename legacy `/graphql/manual/` to `/core/`~ <-- restored to `/graphql/manual/` for now
- ~rename `/img/graphql/manual/` to `/img/core/`~ <-- restored to `/img/graphql/manual/` for now

### Deploy notes

After this is deployed, we should:

- test algolia index
- test redirects
- add [this change](https://github.com/hasura/hasura.io/pull/1455) if we haven't already
- remove the cloud docs directory from the `hasura.io` repo

### Changelog

Although this is a big change on the dev side, there shouldn't be any user-facing changes other than URLs (and the fact that they can now edit cloud docs).

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)


sezgi:
@marionschleifer @rikinsk -- please review when you can (should be quick since most of the diff is moved/copied files). The WIP is just referring to algolia, sitemap, redirects, etc.

rikinsk (approved):

PR #5430, opened 2020-07-21
Fix latency buckets for telemetry data

jberryman:
These must have gotten messed up during a refactor. As a consequence
almost all samples received so far fall into the single erroneous 0 to
1K seconds (originally supposed to be 1ms?) bucket.

I also re-thought what the numbers should be, but these are still
arbitrary and might want adjusting in the future.


jberryman:
@abooij yeah putting a comment there instead of a type was silly. I think the `Fractional` in `Units.hs` work well as long as the type is fairly obvious or close by the literal. I think the problem with `RunningTimeBucket` is the lack of unit in the name so the literal is ambiguous. And also the numerical hierarchy doesn't give us exactly what we want, as you point out. Thanks for the quick review!

PR #5434, opened 2020-07-21
console and cli: fix error reporting while creating actions

### LINKED ISSUE #4666: [Bug] Action's types definition
Hi, 

I am trying to create an action to call one of my rest endpoints but it seems that Hasura does not support fully GraphQL specs. Here is the action definition:

```
type Query {
  searchAssets (search: String): [Asset]
}


#4579 Types definitions

type Asset {
  friendlyName: String!
  description: String!
  symbol: String!
  type: AssetType!
  logo: String
  metadata: AssetMetadata
}

enum AssetType {
  Stock
  Future
  Bond
  Securities
  InternationalStock
  Check
  Option
  TreasuryBills
  FixedIncomeSecurities
  Mocked
}

union AssetMetadata = CompanyMetadata | OtherMetadata

type CompanyMetadata {
  ceo: String!
  headQuarters: String!
  foundedAt: Int!
  employees: Int!
}

type OtherMetadata {
  other: String!
}
```

When I click on the Create button basically nothing is happening, there is not error or any feedback. So my guess was that probably union was not supported, even if it was part of GraphQL spec. So when I removed all the metadata related types and field the action was created. Then I tried to create just one metadata type with all optional types like this: 

```
type Query {
  searchAssets (search: String): [Asset]
}


Types definition

type Asset {
  friendlyName: String!
  description: String!
  symbol: String!
  type: AssetType!
  logo: String
  metadata: Metadata
}

enum AssetType {
  Stock
  Future
  Bond
  Securities
  InternationalStock
  Check
  Option
  TreasuryBills
  FixedIncomeSecurities
  Mocked
}

type Metadata {
  ceo: String
  headQuarters: String
  foundedAt: Int
  employees: Int
  other: String
}
```

I got this error:

validation for the given custom types failed because the type "Metadata" of the field "metadata" in the object type "Asset" is object type which isn't allowed

Which is strange because GraphQl also supports it, and seems a really important feature to be able to a couple already implemented APIs with Hasura Action. The first issue while I would love to see it is not a real stopper but the seconds one it is.

So the question if these are bug or current limitations, and when you are planning to implement them if that is the case.

Thanks

kolharsam:
### Description
(TODO)

### Affected components

- [x] Console
- [x] CLI

### Related Issues

resolves #4666 

### Solution and Design
 - TODO

### Steps to test and verify
 - try creating actions with and/or interfaces and see if an appropriate and intuitive error message shows up.

### DIFF
--- a/console/src/components/Services/Actions/ServerIO.js
+++ b/console/src/components/Services/Actions/ServerIO.js
@@ -127,20 +127,18 @@ export const createAction = () => (dispatch, getState) => {
     );
   }
 
-  const { types, error: typeDefError } = getTypesFromSdl(
-    rawState.typeDefinition.sdl
-  );
+  const typeInformation = getTypesFromSdl(rawState.typeDefinition.sdl);
 
-  if (typeDefError) {
+  if (typeInformation.error) {
     return dispatch(
-      showErrorNotification('Invalid Types Definition', typeDefError)
+      showErrorNotification('Invalid Types Definition', typeInformation.error)
     );
   }
 
   const state = {
     handler: rawState.handler,
     kind: rawState.kind,
-    types,
+    types: typeInformation.types,
     actionType,
     name: actionName,
     arguments: args,
--- a/console/src/shared/utils/sdlUtils.js
+++ b/console/src/shared/utils/sdlUtils.js
@@ -115,29 +115,31 @@ export const getTypeFromAstDef = astDef => {
         error: 'Interface types are not supported',
       };
     default:
-      return;
+      return {
+        error: `You seem to using an unsupported type definition.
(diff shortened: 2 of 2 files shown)


cli-ext/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A service to generate Hasura action scaffolds",
   "main": "src/server.js",
   "scripts": {
-    "get-shared-modules": "rm -rf src/shared && cp ../console/src/shared ./src/shared -r",
```
kolharsam, line 7: This was not working as the usage of the `cp` command was incorrect. 
I was running this on a macOS machine

PR #5441, opened 2020-07-22
configure header size limit for warp

tirumaraiselvan:
### Description

Warp's default total header limit size for HTTP 1.x is 50KB which turns out to be little low (reported in Intercom). See: https://hackage.haskell.org/package/warp-3.3.13/docs/src/Network.Wai.Handler.Warp.Settings.html#defaultSettings

This increases the header limit size to 100KB and makes it configurable.

### Changelog

### Affected components

- [x] Server

### Steps to test and verify

`curl -v -H @temp.log localhost:8080/v1/version`

where temp.log is a header file greater than 50KB. In prior versions, this will throw a bad request.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 (Add entries here in the order of: server, console, cli, docs, others)
 
+- server: configure total header size limit via `HASURA_GRAPHQL_MAX_TOTAL_HEADER_LENGTH` or `--max-total-header-length` (default: 100KB)
 
 ## `v1.3.0`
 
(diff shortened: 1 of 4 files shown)


0x777:
@tirumaraiselvan I don't think we should change the default. Make it customisable?

tirumaraiselvan:
@0x777 I have made it configurable now.

tirumaraiselvan:
Currently, waiting on some triage in warp bug: https://github.com/yesodweb/wai/issues/807

Vlix:
I've set up a PR for yesodweb/wai#807 issue.

Vlix:
yesodweb/wai#807 has been fixed with yesodweb/wai#838

The calculation of header length was wrong in such a way that it could trigger at different sizes (it might let 30kb pass, or it might throw an exception) depending on how large the header chunks are and how many of them there are.
It should now be fixed and actually accept headers with the size up to the maximum setting (default 50kb)

manuFL:
Really need this one... any ETA for the merge?

Vlix:
@tirumaraiselvan The `do not merge` label can be removed, I think.

tirumaraiselvan:
Hey folks, apologies for the tardiness on this. We are going to get this in soon.

pranshi06:
Closing this PR because the issue has been addressed in 3124c93673f42f461ddf7c09b165fb765077a446.

PR #5446, opened 2020-07-22
cli: APIs for migrations on console

kolharsam:
### Description
This PR adds new APIs to the CLI that will help support migrations from the console. 

## TODOs
- [ ] The APIs are yet to be documented and tested properly.
- [ ] Migration metadata updates need to be added.

### Affected components
- [x] CLI

### Related Issues
resolves https://github.com/hasura/graphql-engine-internal/issues/474

### Solution and Design
 - Adding 2 new endpoints `/migrate/apply` and `/migrate/console/` to support the commands `apply` and management of the migration files respectively.

### Steps to test and verify
(TODO)

### DIFF
--- a/cli/migrate/api/apply.go
+++ b/cli/migrate/api/apply.go
@@ -124,9 +124,9 @@ func ApplyMigrationAPI(c *gin.Context) {
 			return
 		}
 
-		// there is some metadata part within the NewMigrate method, I'd have to perform those actions here
-
 		t.SkipExecution = skipExecution
+
+		// there is some metadata part within the NewMigrate method, I'd have to perform those actions here
 
 		err = executeMigration(migrateType, t, step)
 
--- a/cli/migrate/api/apply.go
+++ b/cli/migrate/api/apply.go
@@ -1,0 +1,25 @@
+package api
+
+import "github.com/gin-gonic/gin"
+
+type ApplyMigrationRequest struct {
+	UpMigration    string `json:"up"`
+	DownMigration  string `json:"down"`
+	SkipExecution  bool   `json:"skip_execution"`
+	MigrateVersion string `json:"version"`
+	MigrateType    string `json:"type"`
+	GotoVersion    string `json:"goto"`
+}
+
+func ApplyMigrationAPI(c *gin.Context) {
+	// migratePtr, ok := c.Get("migrate")
+	// if !ok {
+	// 	return
+	// }
+	// // Get File url
+	// sourcePtr, ok := c.Get("filedir")
+	// if !ok {
+	// 	return
+	// }
+
+}
--- a/cli/migrate/api/apply.go
+++ b/cli/migrate/api/apply.go
@@ -1,7 +1,89 @@
 package api
 
-import "github.com/gin-gonic/gin"
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
 
+	"github.com/gin-gonic/gin"
+	"github.com/hasura/graphql-engine/cli/migrate"
+	mig "github.com/hasura/graphql-engine/cli/migrate/cmd"
+)
+
+func getMigrationTypeAndStep(upMigration, downMigration, versionMigration, migrationType, gotoVersion string, skipExecution bool) (string, int64, error) {
+	var flagCount = 0
+	var stepString = "all"
+	var migrationName = "up"
+	if upMigration != "" {
+		stepString = upMigration
+		flagCount++
+	}
+	if downMigration != "" {
+		migrationName = "down"
+		stepString = downMigration
+		flagCount++
+	}
+	if versionMigration != "" {
+		migrationName = "version"
+		stepString = versionMigration
+		if migrationType == "down" {
+			stepString = "-" + stepString
+		}
+		flagCount++
+	}
+	if gotoVersion != "" {
+		migrationName = "gotoVersion"
+		stepString = gotoVersion
+		flagCount++
+	}
+
+	if flagCount > 1 {
+		return "", 0, errors.New("only one migration type can be applied at a time (--up, --down or --goto)")
+	}
+
+	if migrationName != "version" && skipExecution {
+		return "", 0, errors.New("--skip-execution flag can be set only with --version flag")
+	}
+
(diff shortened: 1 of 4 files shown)


cli/migrate/api/console.go
```diff
@@ -0,0 +1,265 @@
...
+// FIXME: May need to change this as well
+func getFileType(ext string) string {
+	if ext == ".yml" || ext == ".yaml" {
+		return "yaml"
+	}
+	if ext == ".sql" {
+		return "sql"
+	}
+
+	// TODO: this should also be an error probably
```
kolharsam, line 58: change to provide error

cli/migrate/api/console.go
```diff
@@ -0,0 +1,265 @@
...
+
+func getValidFilePath(path string, migrationType string, extensions []string) (string, string) {
+	for _, ext := range extensions {
+		fullPath := path + migrationType + ext
+		exists := isFilePresent(fullPath)
+		if exists {
+			return fullPath, ext
+		}
+	}
+	// TODO: this should probably be an error
```
kolharsam, line 95: report error here

cli/migrate/api/console.go
```diff
@@ -0,0 +1,265 @@
...
+	sourcePtr, ok := c.Get("filedir")
+	if !ok {
+		return
+	}
+
+	// Get Logger
+	// loggerPtr, ok := c.Get("logger")
+	// if !ok {
+	// 	return
+	// }
```
kolharsam, line 146: remove unused code

cli/migrate/api/console.go
```diff
@@ -0,0 +1,265 @@
...
+func getMigrationInfo(path string, migrationName string, version int) (MigrationData, MigrationData, error) {
+	// returns the type of file yml, yaml or sql
+	// the respective contents of the file
+	// this function will only look for up.* and down.* for now
+
+	fullPath := path + "/" + migrationName + "/"
+	validExtensions := []string{".yaml", ".sql", ".yml"}
+
+	upSource, upExt := getValidFilePath(fullPath, up, validExtensions)
+	downSource, downExt := getValidFilePath(fullPath, down, validExtensions)
```
kolharsam, line 108: `validExtensions` can be made a constant

cli/migrate/api/console.go
```diff
@@ -0,0 +1,265 @@
...
+	// Get version
+	version := c.GetInt("version")
+
+	// Convert to url.URL
+	t := migratePtr.(*migrate.Migrate)
+	sourceURL := sourcePtr.(*url.URL)
+
+	migrationVersion := c.Param("migrationVersion")
+
+	// for reading the contents of the
```
kolharsam, line 157: complete the comment

cli/migrate/api/console.go
```diff
@@ -0,0 +1,265 @@
...
+
+		status, err := t.GetStatus()
+
+		migrStatus, ok := status.Read(convertToUInt64(migrationVersion))
+		if !ok {
+			// TODO: have to change this code for not a valid migration
+			c.JSON(http.StatusBadRequest, &Response{Code: "bad_request", Message: "This is not a valid migration version"})
+			return
+		}
+		// logger.Info("Here's some info", migrStatus.Name, "\n")
```
kolharsam, line 173: unused code

cli/migrate/api/apply.go
```diff
@@ -0,0 +1,142 @@
...
+			c.JSON(http.StatusBadRequest, &Response{Code: "bad_request", Message: "incorrect apply request"})
+			return
+		}
+
+		upMigration := applyMigrationReq.UpMigration
+		downMigration := applyMigrationReq.DownMigration
+		migrationVersion := applyMigrationReq.MigrateVersion
+		migrationType := applyMigrationReq.MigrateType
+		gotoVersion := applyMigrationReq.GotoVersion
+		skipExecution := applyMigrationReq.SkipExecution
```
kolharsam, line 118: a better way to do this? and also not end up using `applyMigration.UpMigration`

cli/migrate/api/apply.go
```diff
@@ -0,0 +1,142 @@
...
+		migrateType, step, err := getMigrationTypeAndStep(upMigration, downMigration, migrationVersion, migrationType, gotoVersion, skipExecution)
+
+		if err != nil {
+			c.JSON(http.StatusBadRequest, &Response{Code: "bad_request", Message: err.Error()})
+			return
+		}
+
+		t.SkipExecution = skipExecution
+
+		// there is some metadata part within the NewMigrate method, I'd have to perform those actions here
```
kolharsam, line 129: talk to @scriptonist about this.

cli/migrate/api/console.go
kolharsam, line 108: this would only work if the file names were `up.*` and `down.*`. Need to check if other names are permissible

cli/migrate/api/apply.go
kolharsam: remove comment. this taken care in the call to `NewMigrate` on `apiserver.go`

cli/migrate/api/console.go
kolharsam: Yes! this is case, we don't have to worry..:)

cli/migrate/api/apply.go
```diff
@@ -0,0 +1,140 @@
...
+
+// ApplyMigrationRequest is the request type for the apply API endpoint
+type ApplyMigrationRequest struct {
+	UpMigration    string `json:"up"`
+	DownMigration  string `json:"down"`
+	SkipExecution  bool   `json:"skip_execution"`
+	MigrateVersion string `json:"version"`
+	MigrateType    string `json:"type"`
+	GotoVersion    string `json:"goto"`
+}
```
kolharsam, line 94: Check the types here and change accordingly

kolharsam: Nope. Nothing has to change here

scriptonist:
> @beerose suggests that we should include another flag `--to` to the squash APIs and that makes sense. @scriptonist let me know what you think of the same for the CLI? would it be useful to add the same to the CLI?

@kolharsam This is a good suggestion, but I think we can tackle it separately.

PR #5451, opened 2020-07-22
Tracing: Simplify HTTP traced request

Naveenaidu:
### Description
Remove the Inversion of Control (SuspendRequest) and simplify the tracing of HTTP Requests.

### Affected components
- [x] Server

### Solution and Design
The current approach using [SuspendedRequest](https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Hasura/Tracing.hs#L197-L237) might be needlessly complicated, and we can probably replace it with a function, where there may be no need for the Inversion of Control.

The function could be something like:

```haskell
tracedHttpRequest 
  :: MonadTrace m 
  => Text
  -- ^ human-readable name for this block of code
  -> HTTP.Request
  -- ^ http request that needs to be made
  -> (HTTP.Request -> m a)
  -- ^ a function that takes the traced request and executes it
  -> m a
```


server/src-lib/Hasura/Tracing.hs
```diff
@@ -208,16 +206,15 @@ extractHttpContext hdrs = do
...
+tracedHttpRequest 
+  :: MonadTrace m 
+  =>  HTTP.Request
+  -- ^ http request that needs to be made
+  -> (HTTP.Request -> m a)
+  -- ^ a function that takes the traced request and executes it
   -> m a
-traceHttpRequest name f = trace name do
-  SuspendedRequest req next <- f
+tracedHttpRequest req f = trace (T.decodeUtf8 (HTTP.path req)) do
```
paf31, line 217: We should use `bsToTxt'` here, because `decodeUtf8` throws exceptions in the event of a decoding error. It's very unlikely, but we should guard against it.

Naveenaidu: Fixed it :)

PR #5464, opened 2020-07-27
Multiple PostgreSQL & GraphQL Engine instances on Docker

iosamuel:
### Description
If you wanna run locally multiple Hasura instances there is no clear documentation about it. So I put my finding for what works, so you can easily create multiple instances of the database and GraphQL server for Hasura.

### Affected components

- [X] docker-compose.yaml

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 (Add entries here in the order of: server, console, cli, docs, others)
 
 - console: update sidebar icons for different action and trigger types
+- docker: docker-compose.yaml example to create multiple instances of hasura inside docker
 
 ## `v1.3.0`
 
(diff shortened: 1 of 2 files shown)


iosamuel:
this version of docker-compose.yaml could be something like *docker-compose.yaml.example* to put an example of how you can set up multiple instances of hasura in docker.

marionschleifer:
@iosamuel thank you so much for your contribution. While this is very helpful for some people, it's probably not the typical use case, and we'd like to keep the official guides more general. 

However, we created a Stackoverflow question [here](https://stackoverflow.com/questions/63519514/how-are-multiple-instances-of-hasura-graphql-engine-and-postgres-run-locally-via/63519515#63519515) and answered it with this PR as an example. Please feel free to add any more information to the thread if you'd like.

I'm closing this PR, thanks again for contributing 🙂

PR #5470, opened 2020-07-28
server: pass http and websocket request to logging context

ecthiender:
### Description

Minor refactor to pass down the request body in HTTP API and websocket API to the logging context.

### Changelog

### Affected components

- [x] Server

### Related Issues

https://github.com/hasura/graphql-engine-pro/issues/416

### Solution and Design

Minor refactor to pass down the request body in HTTP API and websocket API to the logging context.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

####  Logging

The `ws-server` `debug` logs structure have changed.

Changes
1. When `detail.event.type` == `message_received` - `detail.event.detail` was string -> is changed to an object
2.  When `detail.event.type` == `message_sent` - `detail.event.detail` was string -> is changed to an object

Following is the new log type:

```json
{
  "type": "ws-server",
  "timestamp": "2020-07-29T17:15:01.758+0530",
  "level": "debug",
  "detail": {
    "event": {
      "type": "message_received",
      "detail": {
        "message": "{\"id\":\"1\",\"type\":\"stop\"}",
        "message_size": 24
      }
    },
    "websocket_id": "d859375f-2c41-4510-bd46-ceec1661db79"
  }
}
```

```json
{
  "type": "ws-server",
  "timestamp": "2020-07-29T17:15:06.761+0530",
  "level": "debug",
  "detail": {
    "event": {
      "type": "message_sent",
      "detail": {
        "message": "{\"type\":\"ka\"}",
        "message_size": 13
      }
    },
    "websocket_id": "d859375f-2c41-4510-bd46-ceec1661db79"
  }
}
```

The older log structure:

```json
{
  "type": "ws-server",
  "timestamp": "2020-07-29T17:15:01.758+0530",
  "level": "debug",
  "detail": {
    "event": {
      "type": "message_received",
      "detail": "{\"id\":\"1\",\"type\":\"stop\"}"
    },
    "websocket_id": "d859375f-2c41-4510-bd46-ceec1661db79"
  }
}
```

```json
{
  "type": "ws-server",
  "timestamp": "2020-07-29T17:15:06.761+0530",
  "level": "debug",
  "detail": {
    "event": {
      "type": "message_sent",
      "detail": "{\"type\":\"ka\"}"
    },
    "websocket_id": "d859375f-2c41-4510-bd46-ceec1661db79"
  }
}
```

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
+++ b/server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
@@ -307,7 +307,14 @@ onConn (L.Logger logger) corsPolicy wsId requestHead ipAddress = do
             <> "HASURA_GRAPHQL_WS_READ_COOKIE to force read cookie when CORS is disabled."
 
 onStart
-  :: forall m. (HasVersion, MonadIO m, E.MonadGQLExecutionCheck m, MonadQueryLog m, Tracing.MonadTrace m, MonadExecuteQuery m)
+  :: forall m.
+  ( HasVersion
+  , MonadIO m
+  , E.MonadGQLExecutionCheck m
+  , MonadQueryLog m
+  , Tracing.MonadTrace m
+  , MonadExecuteQuery m
+  )
   => Env.Environment -> WSServerEnv -> WSConn -> StartMsg -> m ()
 onStart env serverEnv wsConn (StartMsg opId q) = catchAndIgnore $ do
   timerTot <- startTimer
@@ -333,7 +340,7 @@ onStart env serverEnv wsConn (StartMsg opId q) = catchAndIgnore $ do
 
   reqParsedE <- lift $ E.checkGQLExecution userInfo (reqHdrs, ipAddress) enableAL sc q
   reqParsed <- either (withComplete . preExecErr requestId) return reqParsedE
-  execPlanE <- runExceptT $ E.getResolvedExecPlan env pgExecCtx
+  execPlanE <- runExceptT $ E.getResolvedExecPlan env logger pgExecCtx
                planCache userInfo sqlGenCtx sc scVer queryType httpMgr reqHdrs (q, reqParsed)
 
   (telemCacheHit, execPlan) <- either (withComplete . preExecErr requestId) return execPlanE
--- a/server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
+++ b/server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
@@ -39,7 +39,6 @@ import qualified Data.Environment                            as Env
 import           Control.Concurrent.Extended                 (sleep)
 import           Control.Exception.Lifted
 import           Data.String
-import           Data.Int                                    (Int64)
 import           GHC.AssertNF
 
 import           Hasura.EncJSON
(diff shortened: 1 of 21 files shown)


server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -517,24 +529,26 @@ onMessage
...
-                           wsConn authMode params
-      CMStart startMsg  -> onStart env serverEnv wsConn startMsg
-      CMStop stopMsg    -> liftIO $ onStop serverEnv wsConn stopMsg
-      -- The idea is cleanup will be handled by 'onClose', but...
-      -- NOTE: we need to close the websocket connection when we receive the
-      -- CMConnTerm message and calling WS.closeConn will definitely throw an
-      -- exception, but I'm not sure if 'closeConn' is the correct thing here....
-      CMConnTerm        -> liftIO $ WS.closeConn wsConn "GQL_CONNECTION_TERMINATE received"
+onMessage env authMode serverEnv wsConn msgRaw = do
+  L.unLogger logger $ WSMessageLog (BL.length msgRaw)
```
paf31, line 533: This will materialize the message in memory. I'm not sure if we care about this as a possible source of space leaks, since I know we do this sort of thing in plenty of other places already. I don't think this needs to block anything here. cc @jberryman what do you think?

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -247,20 +245,21 @@ mkHttpErrorLogContext
   -- ^ Maybe because it may not have been resolved
   -> RequestId
   -> Wai.Request
+  -> (BL.ByteString, Maybe Value)
+  -- ^ (request body, if parsed request)
```
paf31, line 249: typo?

server/src-lib/Hasura/Server/Logging.hs
```diff
@@ -269,8 +268,8 @@ mkHttpErrorLogContext userInfoM reqId req err query mTiming compressTypeM header
            , olResponseSize       = Just $ BL.length $ encode err
            , olRequestReadTime    = Seconds . fst <$> mTiming
            , olQueryExecutionTime = Seconds . snd <$> mTiming
-           , olQuery              = either (const Nothing) Just query
-           , olRawQuery           = either (Just . bsToTxt . BL.toStrict) (const Nothing) query
+           , olQuery              = parsedReq
+           , olRawQuery           = Just $ bsToTxt $ BL.toStrict reqBody
```
paf31, line 272: This changes the behavior of this field slightly, but I assume that's fine and that it being null in some cases is not meaningful?

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
jberryman: I think it's probably okay too, but also I wonder: lazy ByteStrings don't have to imply lazy IO, which is happening here? I assume that's what you meant about `BL.length`: that it would be forcing chunks into memory?

paf31: Yes, I suppose that's right, we're not doing lazy IO necessarily. I think this is fine, thanks!

ecthiender: Thanks for all your comments :)

server/src-lib/Hasura/Server/Logging.hs
ecthiender: Yeah this has a weird behaviour. So the `olQuery` and `olRawQuery` both are `Maybe`s. And previously this function would take in `Either ByteString Value`. The idea was if the request body is parsed (`Right Value`) then we set the `olQuery` field. If we don't have the parsed request body (`Left ByteString`) then we set the `olRawQuery` field. With these changes as we now always have the request body, we should ideally change `olRawQuery` to have always have the value. (And probably it makes sense to call it `olRequestBody` , but that would be a breaking change)

Confirming with @hgiasac (who worked on this), is this right?

I think we can simplify this

server/src-lib/Hasura/Server/Logging.hs
ecthiender: Yup, will fix

server/src-lib/Hasura/Server/Logging.hs
ecthiender: Sorry, I completely misunderstood your comment! I have restored the behaviour of the field.

Issue #4171, opened 2020-03-24
logging action webhook calls


tirumaraiselvan:
If log-type is set to `debug`, then the webhook call should be logged.

PR #5471, opened 2020-07-28
server: add logging for action handlers

ecthiender:
### Description

Added logs for action handlers. Right now it logs request and response sizes to and from the action handler.

A sample log line:
```json
{                                                                                                                                                      
  "type": "action-handler-log",                                                                                                                            
  "timestamp": "2020-07-28T22:10:35.982+0530",                                                                                                         
  "level": "info",                                                                                                                                     
  "detail": {                                                                                                                                          
    "response_size": 36,                                                                                                                               
    "request_size": 102                                                                                                                                
  }                                                                                                                                                    
}
```

### Changelog

### Affected components

- [x] Server

### Related Issues

https://github.com/hasura/graphql-engine-pro/issues/431

### Solution and Design

Add logging inside the `callWebhook` function in `Hasura.GraphQL.Resolve.Action` module. This currently logs the request and repsonse sizes to and from the action webhook. 

The simple change is in this module (`Hasura.GraphQL.Resolve.Action`), rest of the changes are required to pass the logger to appropriate functions.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 (Add entries here in the order of: server, console, cli, docs, others)
 
 - console: update sidebar icons for different action and trigger types
+- server: add logs for action handlers
 
 ## `v1.3.0`
 
--- a/server/src-lib/Hasura/Logging.hs
+++ b/server/src-lib/Hasura/Logging.hs
@@ -83,7 +83,7 @@ instance J.ToJSON (EngineLogType Hasura) where
     ELTQueryLog -> "query-log"
     ELTStartup -> "startup"
     ELTLivequeryPollerLog -> "livequery-poller-log"
-    ELTActionHandler -> "action-handler"
+    ELTActionHandler -> "action-handler-log"
     ELTInternal t -> J.toJSON t
 
 instance J.FromJSON (EngineLogType Hasura) where
@@ -94,7 +94,7 @@ instance J.FromJSON (EngineLogType Hasura) where
     "websocket-log" -> return ELTWebsocketLog
     "query-log" -> return ELTQueryLog
     "livequery-poller-log" -> return ELTLivequeryPollerLog
-    "action-handler" -> return ELTActionHandler
+    "action-handler-log" -> return ELTActionHandler
     _ -> fail $ "Valid list of comma-separated log types: "
          <> BLC.unpack (J.encode userAllowedLogTypes)
 
--- a/server/src-lib/Hasura/Logging.hs
+++ b/server/src-lib/Hasura/Logging.hs
@@ -108,6 +108,8 @@ data InternalLogTypes
  | ILTJwkRefreshLog
  | ILTTelemetry
  | ILTSchemaSyncThread
+ | ILTActionHandler
+ -- ^ TODO: should it be a non-internal log?
  deriving (Show, Eq, Generic)
 
 instance Hashable InternalLogTypes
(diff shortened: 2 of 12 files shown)


tirumaraiselvan:
@ecthiender Does this close this issue? https://github.com/hasura/graphql-engine/issues/4171

ecthiender:
> @ecthiender Does this close this issue? #4171

@tirumaraiselvan somewhat. It doesn't do the comprehensive logging outlined by #4171. Currently it only logs the request/response size. (As shown in the PR description). We can modify this to support all of the requirements of #4171 but that would increase the scope of our original issue wrt cloud-sprint. I'll let @paf31 and you decide. And if the extra work is required on this PR, I can do that.

tirumaraiselvan:
I think it's not required right now. #4171 can be treated separately as a `debug` log line.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 (Add entries here in the order of: server, console, cli, docs, others)
 
 - console: update sidebar icons for different action and trigger types
+- server: add logs for action handlers
```
tirumaraiselvan, line 10: When you fix the conflict in this file, could you also move the server entry to top.

server/src-lib/Hasura/Logging.hs
```diff
@@ -92,6 +94,7 @@ instance J.FromJSON (EngineLogType Hasura) where
     "websocket-log" -> return ELTWebsocketLog
     "query-log" -> return ELTQueryLog
     "livequery-poller-log" -> return ELTLivequeryPollerLog
+    "action-handler" -> return ELTActionHandler
```
tirumaraiselvan, line 97: for consistency, should we call this `action-handler-log`?

ecthiender:
If you're saying we can have two different log-types (when #4171 is merged), then I agree that's the most low-friction approach. But thinking about it a bit more, isn't it redundant that there would be two log-types which log seemingly similar information? But maybe not? They have two different functions?

Ideally, I think this behaviour makes sense - when log-level is `info` it logs request/response size (to/from the call to action handler), and when `log-level` is `debug` it does more comprehensive logging. But our current logging infra doesn't help with this. Currently one log-type is fixed to a specific log-level.

tirumaraiselvan:
> If you're saying we can have two different log-types (when #4171 is merged), then I agree that's the most low-friction approach. But thinking about it a bit more, isn't it redundant that there would be two log-types which log seemingly similar information? But maybe not? They have two different functions?
> 
> Ideally, I think this behaviour makes sense - when log-level is `info` it logs request/response size (to/from the call to action handler), and when `log-level` is `debug` it does more comprehensive logging. But our current logging infra doesn't help with this. Currently one log-type is fixed to a specific log-level.

Actually, I think the ideal behaviour would be to print `info` log line (as given in this PR) and another `debug` log line as given in #4171. So that log lines are always "additive".

PR #5473, opened 2020-07-28
server: add a 30 minute timeout for event trigger locks

paf31:
This gives event trigger locks a 30 minute timeout. In case the graceful shutdown of a server fails, this will automatically unlock in-process rows after 30 minutes at most. 

This PR is a work in progress and I haven't had chance to test it extensively yet, but we can use this as a place to discuss the approach.

### Changelog

### Affected components

- [x] Server

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

This changes the type of the `locked` column to be a `timestamp null` instead of a `boolean not null`.


server/src-lib/Hasura/Eventing/EventTrigger.hs
```diff
@@ -408,10 +408,11 @@ fetchEvents :: Int -> Q.TxE QErr [Event]
 fetchEvents limitI =
   map uncurryEvent <$> Q.listQE defaultTxErrorHandler [Q.sql|
       UPDATE hdb_catalog.event_log
-      SET locked = 't'
+      SET locked = NOW()
       WHERE id IN ( SELECT l.id
                     FROM hdb_catalog.event_log l
-                    WHERE l.delivered = 'f' and l.error = 'f' and l.locked = 'f'
+                    WHERE l.delivered = 'f' and l.error = 'f' 
+                          and (l.locked IS NULL or l.locked < NOW() - interval '30 minute')
```
jberryman, line 415:
```suggestion
                          and (l.locked IS NULL or l.locked < (NOW() - interval '30 minute'))
```

as a very non-expert on what implicit type conversions might be happening in places in postgres, this makes me more comfortable

server/src-lib/Hasura/RQL/DDL/EventTrigger.hs
```diff
@@ -158,7 +158,7 @@ fetchEvent :: EventId -> Q.TxE QErr (EventId, Bool)
 fetchEvent eid = do
   events <- Q.listQE defaultTxErrorHandler
             [Q.sql|
-              SELECT l.id, l.locked
+              SELECT l.id, l.locked IS NOT NULL AND l.locked >= NOW() - interval '30 minute'
```
jberryman, line 161:
```suggestion
              SELECT l.id, l.locked IS NOT NULL AND l.locked >= (NOW() - interval '30 minute')
```

server/src-rsr/initialise.sql
```diff
@@ -302,7 +302,7 @@ CREATE TABLE hdb_catalog.event_log
   error BOOLEAN NOT NULL DEFAULT FALSE,
   tries INTEGER NOT NULL DEFAULT 0,
   created_at TIMESTAMP DEFAULT NOW(),
-  locked BOOLEAN NOT NULL DEFAULT FALSE,
+  locked TIMESTAMP,
```
jberryman, line 305:
```suggestion
  /* when locked IS NULL the event is unlocked and can be processed */
  locked TIMESTAMP,
```

tirumaraiselvan:
@paf31 IIUC, we can now get rid of this? https://github.com/hasura/graphql-engine/blob/df51a8eb181cda2df5b3d5f852df6d20b046597d/server/src-lib/Hasura/App.hs#L371

server/src-rsr/initialise.sql
```diff
@@ -302,7 +302,8 @@ CREATE TABLE hdb_catalog.event_log
   error BOOLEAN NOT NULL DEFAULT FALSE,
   tries INTEGER NOT NULL DEFAULT 0,
   created_at TIMESTAMP DEFAULT NOW(),
-  locked BOOLEAN NOT NULL DEFAULT FALSE,
+  /* when locked IS NULL the event is unlocked and can be processed */
+  locked TIMESTAMP,
```
codingkarthik, line 306: Not sure if this will affect the behaviour, but I think it will be a good idea to use `TIMESTAMPZ` instead of `TIMESTAMP` because `now()` returns a `TIMESTAMPZ` and there may be issues when comparing a `TIMESTAMPZ` value with a `TIMESTAMP` value

paf31:
> @paf31 IIUC, we can now get rid of this?
> 
> https://github.com/hasura/graphql-engine/blob/df51a8eb181cda2df5b3d5f852df6d20b046597d/server/src-lib/Hasura/App.hs#L371

That code unlocks all events if I understand correctly. I think it might be good to keep it, so that we have a way to quickly unlock everything without having to wait 30 minutes.

tirumaraiselvan:
> That code unlocks all events if I understand correctly. I think it might be good to keep it, so that we have a way to quickly unlock everything without having to wait 30 minutes.

We are tracking events which are in-memory and unlocking them during graceful shutdown, right? https://github.com/hasura/graphql-engine/blob/2c397f9f4f7628d6eb893974e273ecf3884d8a17/server/src-lib/Hasura/App.hs#L462 This way a proper server restart should not wait 30 mins to retry a locked event.

Shouldn't this cover most cases? Otherwise, I am not really sure what is the ROI of changing the locked column from bool to timestamp. We are managing the same amount of state as before?

jberryman:
Mm, yeah I agree we shouldn't be doing `unlockAllEvents` on startup anymore. Formerly I guess that was the only way to ensure at-least-once delivery in the case of a hard shutdown. But now that case is taken care of by having the timeout state in durable storage.

While this is still open, do we want to think more about the timeout for (re)delivery? Maybe it should be much shorter or be relative to the http client timeout value we use? Just a detail.

paf31:
> While this is still open, do we want to think more about the timeout for (re)delivery?

Definitely. I wasn't sure about this so I just did the most naive thing possible to start with. Any ideas?

jberryman:
> Any ideas?

I'm not sure, and setting it too low is the real danger (sort of slow thundering herd situation... lumbering herd). Seems good for now

jberryman:
(sorry if this was awaiting another approval by me)

paf31:
> (sorry if this was awaiting another approval by me)

Sorry, it may have, but I'm also being a little hesitant to merge this, since it is a little difficult to test and be sure that it is the best change, for arguably little benefit. Since the original problem is (I think) now addressed, I'm thinking we can hold off until I can write some tests for this. What do you think?

tirumaraiselvan:
I have marked this for v1.4 release since we get a beta period to test this change.

codingkarthik (approved):
LGTM

jberryman:
(FWIW if you want to try using kodiak for auto-merge/update you can add the `auto-update-auto-merge` tag)

Issue #5363, opened 2020-07-13
bound the size of query plan cache


tirumaraiselvan:
This is one of the TODOs in #3530

> - Consider making the bounded cache the default since... 
>  - it is equally effective for the use case that the cache was designed for: a finite number of elements
>  - even a small size is similarly performant to unbounded when the domain is
>    unbounded with long tail (note in this scenario the unbounded cache is
>    _inappropriate anyway_ since it implies unbounded space usage)
>  - the bookkeeping in the LRU cache might allow us to collect important
    telemetry data from users (see above).

Also, it is impossible to switch off the query plan caching by setting `HASURA_GRAPHQL_QUERY_PLAN_CACHE_SIZE: 0`

From code:

```
 ( "HASURA_GRAPHQL_QUERY_PLAN_CACHE_SIZE"
  , "The maximum number of query plans that can be cached, allowed values: 0-65535, " <>
    "0 disables the cache. If this value is not set, there is no limit on the number " <>
    "of plans that are cached"
  )
```
When I set to 0, server doesn't start with 

```
Environment variable HASURA_GRAPHQL_QUERY_PLAN_CACHE_SIZE: cache size must be given as a number between 1 and 65535
```

----

**EDIT (Brandon)**:

- [x] analyze telemetry we've been collecting about cache to find an appropriate size, and any other insights. Have a back-of-the-envelope figure for how much memory a cache of size `N` is likely to consume.
- [x] support disabling cache entirely with 0
- [ ] make bounded the default and update documentation

jberryman:
I think the cache should be disabled, with a warning if 0 is passed (I've wanted this for debugging and it seems more consistent)

jberryman:
(WIP COMMENT)

According to [`GHC.DataSize`](https://hackage.haskell.org/package/ghc-datasize-0.2.2/docs/GHC-DataSize.html), for integration tests load, we get average size of plan cache item of ~**7kb** with ~**70kb** the maximum size (this number fudged a bit, but close enough).

PR #5474, opened 2020-07-28
Support only bounded plan cache, with default of 4000

jberryman:
### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues
#5363 
### Solution and Design

An unbounded cache will grow unbounded in memory (and make GCs more and more expensive). Also the hypothesis for the cache is that a relatively small number of plans will form the vast majority of queries; it might even be that each plan corresponds to a source location in some service (where variables are used) and we can manage to fit everything into the cache easily.

For these reasons we want the bounded cache to be default. Vamshi agreed the unbounded version can be additionally removed.

There is a lot in the design space here that we can play with going forward, e.g. different eviction policies, etc. 

#### Breaking changes

- [x] No Breaking changes

No real breaking changes but this will certainly have some performance impact on some users (hopefully for the better). For the case where someone has, say 8000 unique queries and their workload is to scan through them all in a row, this change will suddenly mean performance will be equivalent to no cache at all


server/src-lib/Hasura/Cache/Bounded.hs
```diff
@@ -24,16 +33,23 @@ import qualified Data.IORef         as IORef
 import qualified Data.Vector        as V
 import           GHC.Natural        (Natural)
 
+-- MISC TODO: 
+--  - benchmark and consider unsafeLookupIncreasePriority and unsafeInsertIncreasePriorityView
```
0x777, line 37: I can't find the definitions of these `unsafe` variants

jberryman: They are here: http://hackage.haskell.org/package/psqueues-0.2.7.2/docs/Data-HashPSQ.html#g:9

Not documented, but the underlying IntPSQ have some docs. I'd want to see whether there was test coverage for these too, and benchmark

tirumaraiselvan:
Later we should also audit if we are not accidentally creating single-use objects: both in query plan cache and connection pool (in prepared statements)

PR #5479, opened 2020-07-29
server: stop eventing subsystem threads when shutting down

ecthiender:
### Description

Stop all background threads of the eventing sub-system, inside the shutdown handler.

### Changelog

### Affected components

- [x] Server

### Related Issues

https://github.com/hasura/graphql-engine-pro/issues/433

### Solution and Design

Add the eventing sub-system threads into the existing list of immortal threads. So that they can also be stopped during graceful shutdown, in the shutdown handler.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/App.hs
+++ b/server/src-lib/Hasura/App.hs
@@ -410,15 +410,14 @@ runHGEServer env ServeOptions{..} InitCtx{..} pgExecCtx initTime shutdownApp pos
     else return Nothing
 
   -- all the immortal threads are collected so that they can be stopped when gracefully shutting down
-  let immortalThreads = [ Just schemaSyncListenerThread
-                        , Just schemaSyncProcessorThread
-                        , Just updateThread
-                        , Just asyncActionsThread
-                        , Just eventQueueThread
-                        , Just scheduledEventsThread
-                        , Just cronEventsThread
-                        , telemetryThread
-                        ]
+  let immortalThreads = [ schemaSyncListenerThread
+                        , schemaSyncProcessorThread
+                        , updateThread
+                        , asyncActionsThread
+                        , eventQueueThread
+                        , scheduledEventsThread
+                        , cronEventsThread
+                        ] <> maybe [] pure telemetryThread
 
   finishTime <- liftIO Clock.getCurrentTime
   let apiInitTime = realToFrac $ Clock.diffUTCTime finishTime initTime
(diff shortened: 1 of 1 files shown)


server/src-lib/Hasura/App.hs
```diff
@@ -381,35 +382,43 @@ runHGEServer env ServeOptions{..} InitCtx{..} pgExecCtx initTime shutdownApp pos
...
+  -- all the immortal threads are collected so that they can be stopped when gracefully shutting down
+  let immortalThreads = [ Just schemaSyncListenerThread
+                        , Just schemaSyncProcessorThread
+                        , Just updateThread
+                        , Just asyncActionsThread
+                        , Just eventQueueThread
+                        , Just scheduledEventsThread
+                        , Just cronEventsThread
+                        , telemetryThread
+                        ]
```
codingkarthik, line 421:
```suggestion
  -- all the immortal threads are collected so that they can be stopped when gracefully shutting down
  let immortalThreads = [ schemaSyncListenerThread
                        , Just schemaSyncProcessorThread
                        , updateThread
                        , asyncActionsThread
                        , eventQueueThread
                        , scheduledEventsThread
                        , cronEventsThread
                        ] <> (maybe [] pure $ telemetryThread)
```

server/src-lib/Hasura/App.hs
```diff
@@ -491,7 +500,7 @@ runHGEServer env ServeOptions{..} InitCtx{..} pgExecCtx initTime shutdownApp pos
     -- we want to control shutdown.
     shutdownHandler
       :: Loggers
-      -> [Immortal.Thread]
+      -> [Maybe Immortal.Thread]
```
codingkarthik, line 503:
```suggestion
      -> [Immortal.Thread]
```

server/src-lib/Hasura/App.hs
```diff
@@ -508,13 +517,13 @@ runHGEServer env ServeOptions{..} InitCtx{..} pgExecCtx initTime shutdownApp pos
         stopWsServer
         -- kill all the background immortal threads
         logger $ mkGenericStrLog LevelInfo "server" "killing all background immortal threads"
-        forM_ immortalThreads $ \thread -> do
+        forM_ immortalThreads $ \maybeThread -> forM_ maybeThread $ \thread -> do
```
codingkarthik, line 520:
```suggestion
        forM_ immortalThreads $ \thread -> do
```

codingkarthik:
Minor change suggested, otherwise LGTM

server/src-lib/Hasura/App.hs
paf31: You can use `maybeToList` in place of `maybe [] pure`.

PR #5517, opened 2020-08-03
server: Add B3 headers for tracing

Naveenaidu:
### Description
Use [B3 Propagation](https://github.com/openzipkin/b3-propagation) for tracing headers. 

### Affected components

- [x] Server

### Solution and Design

The Tracing headers which we use now `X-Hasura-` are not standard, though they help us trace requests within our systems but this will break/not work when we tend to use the other Trace Recorders. Hence it's better to refactor the code to allow using at least B3 headers for now.

Request Headers are changed from `X-Hasura-` to `X-B3-`

Tracing Context type is kept as  `Word64` since it is the lowest common denominator(_most common_) and create extra functions which will convert this Word64 into the format type we want to send. We use these functions in `injectHTTPContext` and `extractHTTP Context`, doing so will help us include other tracing formats in future.


server/src-lib/Hasura/Tracing.hs
```diff
@@ -76,9 +79,10 @@ instance HasReporter m => HasReporter (ExceptT e m) where
 -- the active span within that trace, and the span's parent,
 -- unless the current span is the root.
 data TraceContext = TraceContext
-  { tcCurrentTrace  :: Word64
-  , tcCurrentSpan   :: Word64
-  , tcCurrentParent :: Maybe Word64
+  { tcCurrentTrace   :: BS.ByteString
+  , tcCurrentSpan    :: BS.ByteString
+  , tcCurrentParent  :: Maybe BS.ByteString
+  , tcCurrentSampled :: Maybe Bool
```
Naveenaidu, line 85: Are we going to require this field.

The sampled header is used to describe if the particular trace request needs to be traced or not. The value for this is either 0 or 1. This is an optional header. Do we drop this?

If we don't drop this, then should the interface be changed to:
```haskell
-- | Perform HTTP request which supports Trace headers
tracedHttpRequest 
  :: MonadTrace m 
  =>  HTTP.Request
  -- ^ http request that needs to be made
  -> Bool
  -- ^ Should the trace request be traced or not  
  -> (HTTP.Request -> m a)
  -- ^ a function that takes the traced request and executes it
  -> m a

File: server/src-lib/Hasura/Tracing.hs [hunk 2]
```diff
@@ -76,9 +79,10 @@ instance HasReporter m => HasReporter (ExceptT e m) where
 -- the active span within that trace, and the span's parent,
 -- unless the current span is the root.
 data TraceContext = TraceContext
-  { tcCurrentTrace  :: Word64
-  , tcCurrentSpan   :: Word64
-  , tcCurrentParent :: Maybe Word64
+  { tcCurrentTrace   :: BS.ByteString
```
Naveenaidu, line 82: The encoding of the traced id's are in 16-character hex string. So should I keep the value as ByteString or change it to Text? 

IMO having the type as text would be better, since mentioning ByteString might mean that the filed can have values greater than UTF-8, which is not acceptable in request headers.

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,42 +202,64 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
   currentReporter = lift currentReporter
   attachMetadata = lift . attachMetadata
 
+-- | Create a Random 16 character hex string
+randomHex :: IO BS.ByteString
```
Naveenaidu, line 206: Is this a good way of creating a random hex value or should I use the Text.printf and convert the random Word64 using the `016x` ?

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,42 +202,64 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
 -- and create a new 'TraceContext'. The new context will contain
 -- a fresh span ID, and the provided span ID will be assigned as
 -- the immediate parent span.
 extractHttpContext :: [HTTP.Header] -> IO (Maybe TraceContext)
 extractHttpContext hdrs = do
-  freshSpanId <- liftIO Rand.randomIO
+  freshSpanId <- liftIO randomHex
   pure $ TraceContext
-    <$> (HTTP.parseHeaderMaybe =<< lookup "X-Hasura-TraceId" hdrs)
+    <$> (HTTP.parseHeaderMaybe  =<< lookup "X-B3-TraceId" hdrs)
```
Naveenaidu, line 233: This part doesn't work because there is no instance  of `HTTP.FromHttpApiData BS.ByteString` thus another reason for converting the type of TraceContext to Text?

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,42 +202,64 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
 
 -- | Inject the trace context as a JSON value, appropriate for 
 -- storing in (e.g.) an event trigger payload.
 injectEventContext :: TraceContext -> J.Value
 injectEventContext ctx =
-  J.object
-    [ "trace_id" J..= tcCurrentTrace ctx
-    , "span_id"  J..= tcCurrentSpan ctx
-    ]
+  case tcCurrentParent ctx of
```
Naveenaidu, line 242: This part doesn't work as well because there is no instance of `J.ToJSON BS.ByteString`

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,42 +202,64 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
+                      , "span_id"  J..= tcCurrentSpan ctx
+                      , "parent_id" J..= parentID
+                      ]
+  
 
 -- | Extract a trace context from an event trigger payload.
 extractEventContext :: J.Value -> IO (Maybe TraceContext)
 extractEventContext e = do
-  freshSpanId <- liftIO Rand.randomIO
+  freshSpanId <- liftIO randomHex
```
Naveenaidu, line 257: Ignore this function. Changes needs to be made here.

Naveenaidu, line 257: **Ignore this function**. Changes needs to be made here.

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,42 +202,64 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
+  return hexString
+
 -- | Inject the trace context as a set of HTTP headers.
 injectHttpContext :: TraceContext -> [HTTP.Header]
 injectHttpContext TraceContext{..} = 
-  [ ("X-Hasura-TraceId", fromString (show tcCurrentTrace))
-  , ("X-Hasura-SpanId", fromString (show tcCurrentSpan))
-  ]
+  case tcCurrentParent of
+    Nothing -> [ ("X-B3-TraceId", fromString (show tcCurrentTrace))
```
paf31, line 216: You don't want to use `show` here or you'll end up with quotes around the strings.

server/src-lib/Hasura/Tracing.hs
paf31: Given we might want to support other propagation formats, I'd say it's fine to use a `ByteString` to store the 8 bytes, but we should store them unencoded in the context value, and encode them in a separate function which can be called in `injectHttpContext`. We might also want to create some newtype wrappers for trace and span IDs.

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,42 +202,68 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
   currentReporter = lift currentReporter
   attachMetadata = lift . attachMetadata
 
+-- | encode Byte String to 16 character hex string
+encodeBstoHex :: BS.ByteString -> BS.ByteString
```
Naveenaidu, line 206: Should `encodeBstoHex` return back `Text` instead of ByteString?

Because I think Text type makes more sense, since having ByteString might suggest that we might have characters outside the UTF-8 range.

server/src-lib/Hasura/Tracing.hs
Naveenaidu: I made a new encoding function and kept the ByteString type for the trace contexts. But I don't understand who do you mean by newtype wrappers for trace and span IDs and why do we need them. Could you please help me understand this.

server/src-lib/Hasura/Tracing.hs
Naveenaidu: Having it return `Text` is the right thing.

server/src-lib/Hasura/Tracing.hs
Naveenaidu: Since `Sampled` header is Optional and will be sent/used by external monitoring engines and also since we won't be processing this header, it's better we leave it out for now.

server/src-lib/Hasura/Tracing.hs
Naveenaidu: As discussed, we think it's better to have the Tracing Context type to be a `Word64` since it is the lowest common denominator(_most common_) and create extra functions which will convert this Word64 into the format type we want to send.

For eg: We can have a `encodeHex` function to convert the `Word64` number to a 16 character hex string in the `injectHTTPContext` and`extractHTTPContext` functions.

This way we could support more formats in the future.

server/src-lib/Hasura/Tracing.hs
```diff
@@ -197,13 +200,30 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
   currentReporter = lift currentReporter
   attachMetadata = lift . attachMetadata
 
+-- | Encode Word64 to 16 character hex string
+word64ToHex :: Word64 -> Text
+word64ToHex randNum = bsToTxt $ Hex.encode numInBytes
+  where numInBytes = BL.toStrict (BIN.encode  randNum)
+
+-- | Decode  16 character hex string to Word64
+-- | TODO: Do we need to check if it has an invalid base16 sequence?
```
Naveenaidu, line 209: Opinion needed on this.

IMHO we should not need this, because we can be sure that the value we recieve from the HTTP headers is actually a proper encoded hex value.

server/src-lib/Hasura/Tracing.hs
```diff
@@ -198,13 +201,30 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
-  , ("X-Hasura-SpanId", fromString (show tcCurrentSpan))
-  ]
-  
+  case tcCurrentParent of
+    Nothing -> [ ("X-B3-TraceId", txtToBs $ word64ToHex tcCurrentTrace)
+               , ("X-B3-SpanId", txtToBs $ word64ToHex tcCurrentSpan)
+               ]
+    Just parentID ->  [ ("X-B3-TraceId", txtToBs $ word64ToHex tcCurrentTrace)
+                      , ("X-B3-SpanId", txtToBs $ word64ToHex tcCurrentSpan)
+                      , ("X-B3-ParentSpanId", txtToBs $ word64ToHex parentID)
```
paf31, line 225: I'm not sure what the difference between Span and ParentSpan is supposed to be in B3, and why it requires both. Do you know?

server/src-lib/Hasura/Tracing.hs
paf31: In general, we have no control over what's in the headers we receive, so we need to check. We can check that `snd (Hex.decode ...)` is empty, which means that the decoded string had no errors. But this function will need to return a `Maybe Word64`.

server/src-lib/Hasura/Tracing.hs
```diff
@@ -213,19 +233,26 @@ extractHttpContext :: [HTTP.Header] -> IO (Maybe TraceContext)
...
+injectEventContext TraceContext{..} =
+  case tcCurrentParent  of
+    Nothing -> J.object
+                [ "trace_id" J..= tcCurrentTrace 
+                , "span_id"  J..= tcCurrentSpan 
+                ]
+    Just parentID -> J.object
+                      [ "trace_id" J..=  tcCurrentTrace 
+                      , "span_id"  J..=  tcCurrentSpan 
+                      , "parent_id" J..=  parentID
```
paf31, line 253: We encode the parent ID here but never read it. I don't think it's necessary.

server/src-lib/Hasura/Tracing.hs
paf31: So the Zipkin docs say

> This is an optional ID that will only be present on child spans. That is the span without a parent id is considered the root of the trace.

So I assume this means it's useful just to know if this is a root span? I guess it doesn't hurt to include it.

Naveenaidu: Yup! That's totally right, It helps us to differentiate between root span and child span i.e when there is no `ParentSpanID` it means that it is the root of the trace tree. I think it's okay to have it since that's how the propagation is described in format.

server/src-lib/Hasura/Tracing.hs
Naveenaidu: Fixed it :)

server/src-lib/Hasura/Tracing.hs
Naveenaidu: Fixed it :)

server/src-lib/Hasura/Tracing.hs
```diff
@@ -198,13 +201,34 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
+-- | Decode 16 character hex string to Word64 
+-- | Hex.Decode returns two tuples: (properly decoded data, string starts at the first invalid base16 sequence)
+hexToWord64 :: Text -> Maybe Word64
+hexToWord64 randText = 
+  case invalidString  of
+    True -> Nothing
+    False -> Just decodedWord64
+  where hexStringTuple = Hex.decode $ txtToBs randText
+        invalidString  = BS.null $ snd hexStringTuple
+        decodedWord64 = BIN.decode $ BL.fromStrict $ fst hexStringTuple
```
paf31, line 218:
```suggestion
hexToWord64 randText = do
  let (decoded, leftovers) = Hex.decode $ txtToBs randText
      decodedWord64 = BIN.decode $ BL.fromStrict decoded
  guard (BS.null leftovers)
  pure decodedWord64
```

server/src-lib/Hasura/Tracing.hs
```diff
@@ -198,13 +201,33 @@ instance MonadTrace m => MonadTrace (ExceptT e m) where
...
-  ]
-  
+  case tcCurrentParent of
+    Nothing -> [ ("X-B3-TraceId", txtToBs $ word64ToHex tcCurrentTrace)
+               , ("X-B3-SpanId", txtToBs $ word64ToHex tcCurrentSpan)
+               ]
+    Just parentID ->  [ ("X-B3-TraceId", txtToBs $ word64ToHex tcCurrentTrace)
+                      , ("X-B3-SpanId", txtToBs $ word64ToHex tcCurrentSpan)
+                      , ("X-B3-ParentSpanId", txtToBs $ word64ToHex parentID)
+                      ]
```
paf31, line 229:
```suggestion
  ("X-B3-TraceId", txtToBs $ word64ToHex tcCurrentTrace)
  : ("X-B3-SpanId", txtToBs $ word64ToHex tcCurrentSpan)
  : [ ("X-B3-ParentSpanId", txtToBs $ word64ToHex parentID)
    | parentID <- maybeToList tcCurrentParent
    ]
```

paf31:
@0x777 Would you mind please reviewing this one?

@Naveenaidu Have you been able to test that propagation works with this approach?

server/src-lib/Hasura/Tracing.hs
```diff
@@ -36,6 +36,9 @@ import qualified Network.HTTP.Client         as HTTP
 import qualified Network.HTTP.Types.Header   as HTTP
 import qualified System.Random               as Rand
 import qualified Web.HttpApiData             as HTTP
+import qualified Data.Binary                 as BIN
```
evertedsphere, line 39: Small code style thing: the idiomatic way to do a qualified import here, if you want to call it "bin", would be `import qualified Data.Binary as Bin`. When I first read the code I was trying to think what module could possibly satisfy `B*.I*.N*` :sweat_smile:

Naveenaidu: Fixed it, Thanks :)

server/src-lib/Hasura/Tracing.hs
```diff
@@ -77,9 +80,9 @@ instance HasReporter m => HasReporter (ExceptT e m) where
 -- the active span within that trace, and the span's parent,
 -- unless the current span is the root.
 data TraceContext = TraceContext
-  { tcCurrentTrace  :: Word64
-  , tcCurrentSpan   :: Word64
-  , tcCurrentParent :: Maybe Word64
+  { tcCurrentTrace   :: Word64
```
ecthiender, line 83: Our [style guide](https://github.com/hasura/graphql-engine/blob/master/server/STYLE.md#data-types-1) recommends using strict fields for ADTs. We should add strictness annotations to these.

Naveenaidu: Fixed it, Thanks :)

Issue #3814, opened 2020-02-01
Database Failure on v1.1.0-beta.2 with large event payloads


flybayer:
I'm on `v1.1.0-beta.2` and my database keeps failing when I try to access the logs for an event. Specifically an event trigger on a row with a JSONB field that commonly has 1,000+ object records in that field.

Trying to view the logs page in the console essentially results in a time out with the DB failing.

Here's the log:

```
graphql-engine_1  | {"type":"schema-sync-thread","timestamp":"2020-02-01T00:22:50.289+0000","level":"info","detail":{"thread_type":"processor","info":{"processed_event":{"instance_id":"ee4679bf-5370-4395-becc-252a2ab6f949","occurred_at":"2020-02-01T00:22:51.104384Z"}}}}
graphql-engine_1  | {"type":"schema-sync-thread","timestamp":"2020-02-01T00:22:50.289+0000","level":"info","detail":{"thread_type":"processor","info":{"processed_event":{"instance_id":"ee4679bf-5370-4395-becc-252a2ab6f949","occurred_at":"2020-02-01T00:22:51.116028Z"}}}}
graphql-engine_1  | WARNING:  terminating connection because of crash of another server process
graphql-engine_1  | DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
graphql-engine_1  | HINT:  In a moment you should be able to reconnect to the database and repeat your command.
postgres_1        | 2020-02-01 00:24:02.473 UTC [1] LOG:  server process (PID 66) was terminated by signal 9: Killed
[...]

PR #4446, opened 2020-04-16
console: add columns selector (close #2168)

beerose:
Solves https://github.com/hasura/graphql-engine/issues/2168

### To be discussed 

* Do we want `select all` button?
* Do we want to keep selected columns in a local state of selector component and have `save` button that updated columns in global state and runs fetch query?
* Do we want to persist selected columns in LS?

(cc @rikinsk)

### Description

Add selector to limit columns fetched and displayed in browse rows.

columns_selector

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR.

### Affected components

- [x] Console

PR #5518, opened 2020-08-04
Load big jsonb value in console (close #5136)

### LINKED ISSUE #5136: Browser freezes when browsing rows containing large JSONB values in console
One of my tables has a jsonb field and size of the JSON (not compressed into binary) typically ranges from 0.2-1.5MB

Loading a single row typically makes the browser unresponsive in Chrome and freezes it altogether in Safari to the point where I have to force-quit

macOS 10.15.4
2.2 GHz 6-Core Intel Core i7

soorajshankar:
### Description

Load big jsonb value in console by showing first 5000 characters on the screen. showing huge value in DOM can freeze the browser.

### Changelog

### Affected components
- [x] Console

### Related Issues
close #5136

### Solution and Design
This PR will trim the raw value before displaying it on the DOM.

### Steps to test and verify
add a table with huge raw data and validate that the first 5000 characters are only displayed on the DOM
 
We need to think about some other UX (maybe a new tab which opens up raw data) to show the complete data.


tirumaraiselvan:
Possibly also closes: https://github.com/hasura/graphql-engine/issues/3814

soorajshankar:
@beerose @rikinsk please let me know if you have any other thoughts. 
I think the following can also be considered 
* Adding an option to view raw data on a new tab
* A download option (which is a already in progress)

rikinsk:
Tried this out with large data objects and console still becomes really slow. This doesnt seem to be a DOM rendering issue actually, rather it seems like the amount of data going into the redux state slows down the console. This is happening on the Raw SQL page as well when we fetch large data objects. 
We'll need to figure out a smarter way of dealing with large data going into state too resolve this

soorajshankar:
> Tried this out with large data objects and console still becomes really slow. This doesnt seem to be a DOM rendering issue actually, rather it seems like the amount of data going into the redux state slows down the console. This is happening on the Raw SQL page as well when we fetch large data objects.
> We'll need to figure out a smarter way of dealing with large data going into state to resolve this

@rikinsk the fix can be a combined effect of this PR and [this feature](https://github.com/hasura/graphql-engine/pull/4446) (by this user will be able to choose the columns that they want to query)

This PR will only add an additional layer that protects the console from crashing/ non-responsive behavior by reducing the DOM load.

franciscolourenco:
In my experience the table view in the console is considerably slower than querying the same data via grahpQLi, so it is probably not only about the amount of data.

tibotiber:
Hi team! Just checking, are there any plans to merge this PR? Or is another approach in the works to fix the underlying issue?

soorajshankar:
Thanks folks for all the comments and suggestions, closing this PR for now as we are working on new components on the UI to improve the overall experience (some changes are already gone to latest version)

valstu:
I also have biggish bytea columns which crashes the browser and makes the console ui useless if I need to fix some data on tables. I really hope there would be some quickfix where the data would be truncated or even some query params that would let you disable some columns. 👍

PR #5525, opened 2020-08-04
server: show method and complete URI in traced HTTP calls

paf31:
This changes the name of traced HTTP spans from something like `/` to `POST domain.com/path`.

### Changelog

### Affected components

- [x] Server
#### Catalog upgrade
- [x] No
#### Metadata

- [x] No

#### GraphQL
#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/Tracing.hs
+++ b/server/src-lib/Hasura/Tracing.hs
@@ -26,6 +26,7 @@ import           Control.Monad.Trans.Control
 import           Control.Monad.Morph
 import           Control.Monad.Unique
 import           Data.String                 (fromString)
+import           Network.URI                 (URI)
 
 import qualified Data.Aeson                  as J
 import qualified Data.Aeson.Lens             as JL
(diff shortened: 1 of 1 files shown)


server/src-lib/Hasura/Tracing.hs
```diff
@@ -237,22 +238,25 @@ extractEventContext e = do
...
-  for_ reqBytes \b ->
-    attachMetadata [("request_body_bytes", fromString (show b))]
-  ctx <- currentContext
-  let req' = req { HTTP.requestHeaders =
-                     injectHttpContext ctx <> HTTP.requestHeaders req
-                 }
-  f req'
+tracedHttpRequest req f = do
+  let method = bsToTxt (HTTP.method req)
+      uri = show @URI (HTTP.getUri req)
```
Naveenaidu, line 248: Please don't mind if my doubt is silly but Don't we also have to include the pragma `{-#  LANGUAGE TypeApplicatdons  #-}` to use the `@` character for giving explicit type arguments?

Also what benefits do we have for using this over ` show ( (HTTP.getUri req) :: URI`

Naveenaidu:
The procedure LGTM 👍

server/src-lib/Hasura/Tracing.hs
paf31: TypeApplications is turned on by default in the Cabal file so we don't need it here. I'm using it because `show` is very polymorphic and if the argument type changes, we'll end up with a different `Show` instance being used. This has happened to me a few times before, where a refactoring broke some code using `show` or a similar function, and types were not enough to catch the mistake. This makes sure we can't accidentally change the type without changing the explicit application.

Naveenaidu: Understood, Thanks for the explanation :)

PR #5547, opened 2020-08-06
server: fix #5542 - check for empty string in insert_event_log

### LINKED ISSUE #5542: RQL insert throws runtime error because of event triggers
The change introduced here https://github.com/hasura/graphql-engine/pull/5409 does not work well with RQL inserts (and probably updates and deletes).

Create a table `test (id serial primary key, name text)` and create an event trigger on insert.

With graphql mutation:
```
$ curl 'http://localhost:8080/v1/graphql' -d '{"query":"mutation MyMutation {\n  insert_test_one(object: {name: \"t\"}) {\n    id\n    name\n  }\n}\n","variables":null,"operationName":"MyMutation"}'

{"data":{"insert_test_one":{"id":6,"name":"t"}}}
```

With RQL mutation:
```
$ curl 'http://localhost:8080/v1/query' -d '{"type":"insert","args":{"table":{"name":"test","schema":"public"},"objects":[{"name":"a"}],"returning":[]}}'

{"path":"$.args","error":"invalid input syntax for type json","code":"data-exception"}
```

Note that you must run the graphql mutation first and then run the rql insert (hopefully, same connection is used for both).

This is probably caused because we set `hasura.tracecontext` in a graphql mutation but afterwards if you run a RQL insert (on the same PG connection), it finds that to be an empty string as it is not set during RQL execution (why it returns empty string instead of NULL is discussed in Slack and to summarize, it is a PG bug: https://www.postgresql.org/message-id/d2ac387b-e796-e3ef-10e5-e06fede7c973%40joeconway.com ).

The solution is perhaps to set `hasura.tracecontext` for RQL as well.

paf31:
I've just added an explicit check for the empty string. Not the most elegant solution, but I know exception handling would add a runtime cost, which is why we have two branches in the first place.

Note: I didn't fix this in the RQL code because we need to also be able to handle the case where a user performs an insert directly, without going via Hasura at all.

### Changelog

### Affected components

- [x] Server

### Related Issues

Fixes #5542

### Steps to test and verify
I tested this in `psql` as follows:

```
begin;
set local hasura.tracecontext = '';
select hdb_catalog.insert_event_log('public', 'test', '', '', '{}');
...
rollback;
```

Before this change, you will see the reported exception. After this change, the trigger data will just be set to `NULL`.

#### Catalog upgrade

- [x] Yes
#### Metadata
- [x] No
#### GraphQL
#### Breaking changes

- [x] No Breaking changes


tirumaraiselvan:
> Note: I didn't fix this in the RQL code because we need to also be able to handle the case where a user performs an insert directly, without going via Hasura at all.

When a user doesn't go through Hasura, `hasura.tracecontext` is never set and hence the code works fine for those cases (by returning or catching `NULL`). That is why we never had this issue with `hasura.user` in event triggers so far.

I would be in favour of adding tracing context in RQL for the sake of consistency (tracing, transaction variables) and keeping the trigger function simple.

@paf31 @0x777 WDYT?

tirumaraiselvan:
Also, 9.5 test failed: https://app.circleci.com/pipelines/github/hasura/graphql-engine/13128/workflows/e4b8fed2-b8d0-4409-ae69-c623eccb1a66/jobs/165430

paf31:
I'll make a separate PR.

PR #5555, opened 2020-08-07
server: set hasura.tracecontext in RQL mutations [#5542]

### LINKED ISSUE #5542: RQL insert throws runtime error because of event triggers

paf31:
This fix sets the `hasura.tracecontext` variable in websocket and RQL mutations.

### Changelog

### Affected components

- [x] Server

### Related Issues
Fixes #5542

### Steps to test and verify
Following the steps in the issue:

```
$ curl 'http://localhost:8080/v1/graphql' -d '{"query":"mutation MyMutation {\n  insert_test_one(object: {id: \"t\"}) {\n    id\n    \n  }\n}\n","variables":null,"operationName":"MyMutation"}'
{"data":{"insert_test_one":{"id":"t"}}}

$ curl 'http://localhost:8080/v1/query' -d '{"type":"insert","args":{"table":{"name":"test","schema":"public"},"objects":[{"id":"a"}],"returning":[]}}'
{"returning" : [{}], "affected_rows" : 1}
```

If CI passes, I can work on adding a test for this exact scenario.

It's still technically possible for the same connection to issue a query without `hasura.tracecontext`, which would cause us to again run afoul of the same postgres bug, but this should catch all the currently-possible ways in which that can happen.
 
#### Catalog upgrade

- [x] No
#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/App.hs
+++ b/server/src-lib/Hasura/App.hs
@@ -235,7 +235,7 @@ migrateCatalogSchema env logger pool httpManager sqlGenCtx = do
   let pgExecCtx = mkPGExecCtx Q.Serializable pool
       adminRunCtx = RunCtx adminUserInfo httpManager sqlGenCtx
   currentTime <- liftIO Clock.getCurrentTime
-  initialiseResult <- runExceptT $ peelRun adminRunCtx pgExecCtx Q.ReadWrite $
+  initialiseResult <- runExceptT $ peelRun adminRunCtx pgExecCtx Q.ReadWrite Nothing $
     (,) <$> migrateCatalog env currentTime <*> liftTx fetchLastUpdate
 
   ((migrationResult, schemaCache), lastUpdateEvent) <-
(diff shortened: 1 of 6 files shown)


tirumaraiselvan:
@paf31 The build is failing for some reason in CI (builds locally tho :thinking: )

paf31:
Not sure what's going on with `test_server_upgrade` now.

tirumaraiselvan:
Just re-ran it in CI, seems to be going okay so far.

PR #5560, opened 2020-08-10
server: Pass EKG Metrics Store as argument to runHGEServer

Naveenaidu:
### Description
We need to make sure OSS can surface metrics in Datadog via Pro, so that means sending the metrics store value into OSS during the call to `mkWaiApp`/`runHGEServer`.

### Affected components

- [x] Server

### Solution and Design

We could make the store in the `Main.hs` and pass it as an argument to `runHGEServer`


server/src-lib/Hasura/App.hs
```diff
@@ -308,8 +310,9 @@ runHGEServer
   -> IO ()
   -- ^ shutdown function
   -> Maybe EL.LiveQueryPostPollHook
+  -> IO EKG.Store
```
paf31, line 313: Let's pass this in as `Maybe EKG.Store`

paf31:
@0x777 Can you please review this one, or assign it to someone to review? Thanks!

PR #5576, opened 2020-08-12
cli: add `to` flag for migrate squash command

kolharsam:
### Description

Adds a `to` flag to the `migrate squash` command on the CLI.

### Changelog

### Affected components

- [x] CLI
- [x] Docs

### Related Issues

This was mentioned here on this PR #5446 and specifically this comment here: https://github.com/hasura/graphql-engine/pull/5446#issuecomment-666891658


cli/commands/migrate_squash.go
```diff
@@ -63,14 +69,21 @@ type migrateSquashOptions struct {
 
 func (o *migrateSquashOptions) run() error {
 	o.EC.Logger.Warnln("This command is currently experimental and hence in preview, correctness of squashed migration is not guaranteed!")
-	o.EC.Spin(fmt.Sprintf("Squashing migrations from %d to latest...", o.from))
+	if o.from > o.to && o.to != 0 {
+		return errors.Wrap(errors.New("incorrect version provided for to flag"), "the `to` flag has to have a version that is greater than that of `from` flag")
+	}
+	if o.to != 0 {
```
purush7, line 75: Hey @kolharsam, I think it would be better to  change the final squash version like
example the versions are 1 , 11 , 21 , 31, 41.
squash:  if from 1 to 31 then final 32 ,41

According to this pr it creates 41, 51  and user may get error if they performs `squash` and migrate `apply` (if user didn’t apply previous versions ).

in

```
if o.to != 0{
o.newVersion = o.to + 1
o.EC.Spin(fmt.Sprintf("Squashing migrations from %d to %d...", o.from, o.to))
 } 
```

This can work I guess and I think we can assume that next version number will be more than 1000 ( as it changes with nano seconds / milli seconds).

purush7 (changes requested):

PR #5580, opened 2020-08-12
Remove 'bulldozer' config, try 'kodiak' for auto-merge

jberryman:
see: https://github.com/chdsbd/kodiak

The main issue that bit us was not being able to auto update forked
branches, also:
palantir/bulldozer#66
palantir/bulldozer#145


jberryman:
Used admin merge. We still have to set up the kodiak github app.

PR #5603, opened 2020-08-17
Server: Don't throw validation error while creating remote relationship joining singular type with array type

### LINKED ISSUE #5133: [v1.3.0-beta.2] Error remote relationship: cannot validate remote relationship because expected type \"Int\" but got \"Int\"
Hi

Steps to reproduce:
1) Create table with remote Int Column
2) Add remote scheme with "Int" type
3) Try create remote relationship

```
[
    {
        "definition": {
            "remote_field": {
                "bot_directions": {
                    "arguments": {
                        "key": "$direction_key"
                    }
                }
            },
            "name": "dk",
            "hasura_fields": [
                "direction_key"
            ],
            "remote_schema": "hasura-backend",
            "table": {
                "schema": "public",
                "name": "bot_groups"
            }
        },
        "reason": "in table \"bot_groups\": in remote relationship\"dk\": cannot validate remote relationship because expected type \"Int\" but got \"Int\"",
        "type": "remote_relationship"
    }
]
```

codingkarthik:
Closes #5133 

The issue arises when the user tries to make a remote relationship type, joining a singleton type with an array type. For example, joining `Int` with `[Int]`.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ If you do have such headers configured, then you must update the header configur
 - server: add logs for action handlers
 - server: add request/response sizes in event triggers (and scheduled trigger) logs (#5463)
 - server: change startup log kind `db_migrate` to `catalog_migrate` (#5531)
+- server: don't throw validation error when creating remote relationships joining singleton type with array type with same base type (fixes #5133)
```
jberryman, line 24: Does this really close 5133? I don't see any mention of list types there..

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -407,16 +407,18 @@ validateType permittedVariables value expectedGType types =
...
-  (when
-     (isListType' actualType /= isListType' expectedType)
-     (Failure (pure $ ExpectedTypeButGot expectedType actualType)))
-  -- if list type then check over unwrapped type, else check base types
-  if isListType' actualType
-    then assertType (unwrapTy actualType) (unwrapTy expectedType)
-    else (when
-            (getBaseTy actualType /= getBaseTy expectedType)
-            (Failure (pure $ ExpectedTypeButGot expectedType actualType)))
+  -- The GraphQL spec says that, we can coerce a singleton value into an array
```
jberryman, line 410: This comment is great. While you are here can you give `assertType` a better name? I guess what we're actually doing here is asserting that `actualType` is coercible to `expectedType`?

jberryman: Also, is this deficient in other ways according to the spec?

CHANGELOG.md
codingkarthik: Yes, it does. The issue doesn't mention anything about the list type, because the error message thrown is incorrect. Ideally, the error message should have been 

```Error remote relationship: cannot validate remote relationship because expected type \"[Int]\" but got \"Int\" ```

instead of 

```Error remote relationship: cannot validate remote relationship because expected type \"Int\" but got \"Int\"```

The reason the list notation is omitted from the error message because of this [line](https://github.com/hasura/graphql-engine/blob/ea76e7840eaf2da68c7170a3d55cccf60083954b/server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs#L68)

This PR solves that issue.

Comment (codingkarthik) [reply CHANGELOG.md:24]: This error message can occur in another case too, while joining an `[Int]` type to a `Int` type, where the validation is justified but the error message will be the same 

```Error remote relationship: cannot validate remote relationship because expected type \"Int\" but got \"Int\"``` 

which is not very helpful. 

I'll fix the error message.

File: server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs (hunk 2 above)
Comment (codingkarthik) [reply server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs:410]: > I guess what we're actually doing here is asserting that actualType is coercible to expectedType?

Yes, that's what we're doing. But I'm not sure, what can be a better name here? Perhaps, `isTypeCoercible`?

File: CHANGELOG.md (hunk 1 above)
Comment (codingkarthik) [reply CHANGELOG.md:24]: > Also, is this deficient in other ways according to the spec?

Yes, it was. My patch would have allowed to make a remote relationship joining types `[[Int]]` and `[[[Int]]]`, which is illegal, as specified in the GraphQL spec.

I have fixed this in the next commit

File: server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs (hunk 2 above)
Comment (codingkarthik) [reply server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs]: > Also, is this deficient in other ways according to the spec?

Yes, it was. My patch would have allowed to make a remote relationship joining types `[[Int]]` and `[[[Int]]]`, which is illegal, as specified in the GraphQL spec.

I have fixed this in the next commit

Comment (jberryman) [reply server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs]: `isTypeCoercible` sounds good to me!

File: CHANGELOG.md (hunk 1 above)
Comment (tirumaraiselvan) [inline CHANGELOG.md:24]:
```suggestion
- server: allow remote relationships joining `type` column with `[type]` input argument as spec allows this coercion (fixes #5133)
```

server/tests-py/queries/remote_schemas/remote_relationships/setup_remote_rel_joining_singleton_with_array.yaml
```diff
@@ -0,0 +1,25 @@
...
+  args:
+    name: remoteUsersMultipleIds
+    table: profiles
+    hasura_fields:
+      - id
+    remote_schema: my-remote-schema
+    remote_field:
+      users:
+        arguments:
+          user_ids: ["$id"]
```
tirumaraiselvan, line 25: Was support for providing a mapping like this included the current PR? Or did it exist already?

codingkarthik: The support for the `user_ids: ["$id"]` already existed before this PR. I added a test case to make sure that the older behaviour remains the same.

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -405,19 +405,25 @@ validateType permittedVariables value expectedGType types =
...
-  -- if list type then check over unwrapped type, else check base types
-  if isListType' actualType
-    then assertType (unwrapTy actualType) (unwrapTy expectedType)
-    else (when
-            (getBaseTy actualType /= getBaseTy expectedType)
-            (Failure (pure $ ExpectedTypeButGot expectedType actualType)))
-  pure ()
+isTypeCoercible :: G.GType -> G.GType -> Validation (NonEmpty ValidationError) ()
+isTypeCoercible actualType expectedType =
+  -- The GraphQL spec says that, a singleton type can be coerced into  an array
```
tirumaraiselvan, line 410: I think technically "singleton" type is called `named` type but we can fix this comment later.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ If you do have such headers configured, then you must update the header configur
 - server: add logs for action handlers
 - server: add request/response sizes in event triggers (and scheduled trigger) logs (#5463)
 - server: change startup log kind `db_migrate` to `catalog_migrate` (#5531)
+- server: allow remote relationships joining `type` column with `[type]` input argument as spec allows this coercion (fixes #5133)
```
tirumaraiselvan, line 24: oops, can you push this to Next release section also?

codingkarthik: Sure

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -405,19 +405,25 @@ validateType permittedVariables value expectedGType types =
...
+  in
+  if actualBaseType == expectedBaseType
+  then if (actualNestingLevel == expectedNestingLevel || actualNestingLevel == 0)
+  -- The check of 'actualNestedCount == 0' is the case of coercing a singleton type
+  -- into an array type
+       then pure ()
+       else Failure (pure $ ExpectedTypeButGot expectedType actualType)
+       --  we cannot coerce two types with different nesting levels,
+       -- for example, we cannot coerce [Int] to [[Int]]
+  else Failure (pure $ ExpectedTypeButGot expectedType actualType)
```
rakeshkky, line 426: Can we simplify this code by using MultiwayIf
```haskell
if | actualBaseType /= expectedBaseType -> raiseValidationError
   | actualNestingLevel /= expectedNestingLevel && actualNestingLevel /= 0 -> raiseValidationError
   | otherwise -> pure ()
   where
     raiseValidationError = Failure $ pure $ ExpectedTypeButGot expectedType actualType
```
Of course with appropriate comments.

PR #5612, opened 2020-08-18
console: improve UI for nested remote relationships (close #4973)

### LINKED ISSUE #4973: console: remote relationship doesn't show the full tree in mapping
If I create a deeply nested remote relationship:

image

I still see the relationship definition as two-level deep:

image

**Expected**: one-graph -> stripe -> customer

Varun-Choudhary:
closes #4973

### Description

This PR improves the representation of remote relationship mapping on console UI. 

### Affected components

- [X] Console

It's gone from being like this : 

image

to something like this:

image


kolharsam:
@Varun-Choudhary could you share the link to the remote schema you used while developing this feature?

console/src/components/Services/Data/TableRelationships/RemoteRelationships/components/EditorCollapsed.tsx
```diff
@@ -10,6 +10,11 @@ const Collapsed: React.FC<Props> = ({ relationship }) => {
   if (!relationship) {
     return null;
   }
+  const parseRelationship = parseRemoteRelationship(relationship);
+  const relationMap = () => {
+    const fields = parseRelationship.remoteFields.map(field => field.name);
+    return ['', ...fields].join(' →');
```
kolharsam, line 16: We should probably add more spacing here. The deeper the nesting, the more congested it might look. See image below for context.
 
image

kolharsam (commented):
Code wise it looks good! 👍

console/src/components/Services/Data/TableRelationships/RemoteRelationships/components/EditorCollapsed.tsx
```diff
@@ -19,7 +24,7 @@ const Collapsed: React.FC<Props> = ({ relationship }) => {
         <i>
           {`- ${relationship.table_name} → ${
             relationship.definition.remote_schema
-          } . ${Object.keys(relationship.definition.remote_field)}`}
+          }  ${relationMap()}`}
```
kolharsam, line 27: Another problem I saw was this: @beerose or @rikinsk either of you could suggest the workaround here.
I think it's a very rare case and only a good to have.

image

beerose (changes requested):
It looks like the issue was solved, but I have one comment about the UI. 

We need something like:
`from_column ➡ remote_schema . _field . more_nested_field . even_more_nested_filed . final_field`,
so you'd need to use `.` instead of arrows (except for the first one that is pointing from database column to some remote schema.)

beerose (approved):
Console changes

tirumaraiselvan (approved):

Issue #5582, opened 2020-08-13
Fix stringified JSON in some logs sub-keys


dsandip:
* prepared arguments in `query-log` are printed as strings (stringified JSON).

tirumaraiselvan:
Note that prepared arguments are not always key-value pairs. For example,

Query:

```
query($id: Int!) {
  test(where: {id: {_eq: $id}}) {
    id
    name
  }
}
```

Variables:

```
{
  "id": 1
}
```

The `query-log` prints: 

```
"prepared_arguments":["{\"x-hasura-role\":\"admin\"}","1"]
```

BenHizak:
Hi,
Has this issue been addressed?

Naveenaidu:
Hey @BenHizak, Yes the issue has been addressed and fixed by this  https://github.com/hasura/graphql-engine/pull/5615

PR #5615, opened 2020-08-18
Fix stringified JSON for the prepared_argument in query_log

Naveenaidu:
Fix strigified JSON for the `prepared_argument` in `query_log`

### Description

The `prepared_argument` in the `query_log` gets outputted as a string instead of a JSON value.

The current output for `query_log` looks like
```json
  "type": "query-log",
      "test_query": {
        "prepared_arguments":["{\"x-hasura-role\":\"admin\"}","1"]
       }
```

This PR fixes it to look like:
```json

  "type": "query-log",
      "test_query": {
        "prepared_arguments": [
          {
            "x-hasura-role": "admin"
          },
          3
        ]
```

### Affected components

- [x] Server

### Related Issues
https://github.com/hasura/graphql-engine/issues/5582

### Solution and Design
 Fix the `toJson` instance of `PreparedSql` to use `pgScalarValueToJson` instead of `txtEncodedPGVal

### Steps to test and verify

Run the `graphql-engine` using the following command:
``` bash
cabal run  -v0 -- exe:graphql-engine \             
  --database-url=<postgres_url> \
  serve --enable-console --console-assets-dir=../console/static/dist --enabled-log-types="startup,http-log,query-log,websocket-log,webhook-log"\
 --log-level debug| jq --unbuffered 'select(.type=="query-log")'
```

In the GraphiQL:

Query Type:
```
query($id: Int!) {
  test(where: {id: {_eq: $id}}) {
    id
    name
  }
}
```

Variables:
```
{
  "id": 1
}

```

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Execute/Query.hs
+++ b/server/src-lib/Hasura/GraphQL/Execute/Query.hs
@@ -281,7 +281,7 @@ data PreparedSql
 instance J.ToJSON PreparedSql where
   toJSON (PreparedSql q prepArgs _) =
     J.object [ "query" J..= Q.getQueryText q
-             , "prepared_arguments" J..= map (txtEncodedPGVal . snd) prepArgs
+             , "prepared_arguments" J..= map (pgScalarValueToJson . snd) prepArgs
              ]
 
 -- | Intermediate reperesentation of a computed SQL statement and prepared


server/src-lib/Hasura/SQL/Value.hs
```diff
@@ -186,14 +187,19 @@ parsePGValue ty val = case (ty, val) of
...
+  hashWithSalt s (TELit v)           = s + (hash v)
+  hashWithSalt s (TEJSON (Q.JSON j)) = s + (hash j)
 
 instance ToJSON TxtEncodedPGVal where
   toJSON = \case
     TENull  -> Null
     TELit t -> String t
+    TEJSON (Q.JSON j) -> j
 
 instance FromJSON TxtEncodedPGVal where
```
Naveenaidu, line 204: I'm not sure how to write a `FromJson` instance for the type `(Q.JSON j)`.

server/src-lib/Hasura/RQL/DML/Mutation.hs
```diff
@@ -212,3 +214,6 @@ mkSelCTEFromColVals qt allCols colVals =
       TENull          -> S.SENull
       TELit textValue ->
         S.withTyAnn (unsafePGColumnToRepresentation colTy) $ S.SELit textValue
+      TEJSON (Q.JSON j) -> 
+        S.withTyAnn (unsafePGColumnToRepresentation colTy) $ S.SELit $ jsonText
```
Naveenaidu, line 218: Is getting a `stringified json` here okay?

I assumed that since we are just logging the SQL query here. Having a json as string doesn't harm. If not we would have to add a new constructor for `S.SELit` which takes json value

server/src-lib/Hasura/SQL/Value.hs
```diff
@@ -265,6 +269,7 @@ txtEncoder :: PGScalarValue -> S.SQLExp
 txtEncoder colVal = case txtEncodedPGVal colVal of
   TENull  -> S.SENull
   TELit t -> S.SELit t
+  TEJSON (Q.JSON j) -> S.SELit $ TL.toStrict $ AE.encodeToLazyText j
```
Naveenaidu, line 272: Is getting a stringified json here okay?

Same reasoning as above

server/src-lib/Hasura/SQL/Value.hs
```diff
@@ -186,14 +187,19 @@ parsePGValue ty val = case (ty, val) of
 data TxtEncodedPGVal
   = TENull
   | TELit !Text
+  | TEJSON !Q.JSON
   deriving (Show, Eq, Generic)
 
-instance Hashable TxtEncodedPGVal
+instance Hashable TxtEncodedPGVal where
+  hashWithSalt s (TENull)            = s + (hash Null)
```
paf31, line 194: If we can't derive this, I'd say we should delegate it to an existing hashing function, because this one looks like it would generate a lot of collisions. That is, I think it should be `hashWithSalt s = hashWithSalt s . _f` where `_f` is a function to some intermediate type we need to determine.

0x777:
@Naveenaidu The `ToJSON` instance of `PreparedSql` is at fault and this can be fixed as follows:

```haskell
-- | Required to log in `query-log`
instance J.ToJSON PreparedSql where
  toJSON (PreparedSql q prepArgs _) =
    J.object [ "query" J..= Q.getQueryText q
             , "prepared_arguments" J..= preparedArgumentMap
             ]
    where
      preparedArgumentMap :: Map.HashMap Int J.Value
      preparedArgumentMap =
        Map.fromList $ zip [1..] $ map (pgScalarValueToJson . snd) prepArgs
```
Changes the format from list to an object, but this is better in my opinion so we should make this change.

ecthiender:
@Naveenaidu we don't need to do any of this. We are using `txtEncodedPGVal` function by mistake. We should use `pgScalarValueToJson`. So the simple change is:

In: `src-lib/Hasura/GraphQL/Execute/Query.hs`

```diff
 instance J.ToJSON PreparedSql where
   toJSON (PreparedSql q prepArgs _) =
     J.object [ "query" J..= Q.getQueryText q
-             , "prepared_arguments" J..= map (txtEncodedPGVal . snd) prepArgs
+             , "prepared_arguments" J..= map (pgScalarValueToJson . snd) prepArgs
              ]
```

ecthiender:
Sorry, I missed @0x777's comment. We can do that also. But that changes it from a list of prepared arguments to an object with index being the keys.

Naveenaidu:
Yikes! The solution was that simple and I went the other way around. Thanks for all the help, fixed it now :)

ecthiender (approved):

PR #5659, opened 2020-08-24
server: Bump GHC version to 8.10.2

lexi-lambda:
This bumps GHC to version 8.10.2, pulling in a few patches that, among other things, should hopefully improve compiler performance. This is just internal housekeeping; we do not expect this to result in any observable difference in `graphql-engine` behavior or performance.


.circleci/config.yml
```diff
@@ -3,7 +3,7 @@ refs:
   constants:
   # TODO upload to hasura docker hub:
   # - &server_builder_image hasura/graphql-engine-server-builder:2020-01-14
-  - &server_builder_image jberryman/graphql-engine-server-builder-8.10:2020-04-29
+  - &server_builder_image lexilambda/graphql-engine-server-builder:2020-08-24
```
tirumaraiselvan, line 6: @lexi-lambda @jberryman BTW, both of you have push access to hasura repo in dockerhub :D

Screen Shot 2020-08-26 at 8 42 29 PM

lexi-lambda: Yes, I intend to push the image to the hasura org before merging.

tirumaraiselvan (approved):

PR #5675, opened 2020-08-26
rfc for remote schema permissions

tirumaraiselvan:
Preview here: https://github.com/tirumaraiselvan/graphql-engine/blob/rfc-remote-schema-perms/rfcs/remote-schema-permissions.md

Related issues/PRs:

https://github.com/hasura/graphql-engine/issues/2461
https://github.com/hasura/graphql-engine/pull/2690


tirumaraiselvan:
I was wondering instead of describing the permission in yaml metadata, if we can do it as a [graphql document](https://spec.graphql.org/June2018/#sec-Language.Document). For example:

```
type Todo {
  id: Int!
  name: String!
  description: String
  priority: Priority!
}

type Query {
  # using the default syntax
  todo(id: Int = "x-hasura-user-id"): Todo
}
```

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+**Scalars**
+
+```
+allowed_custom_scalars:
+- name: UUID
+- name: Geography
+```
+
+Only `UUID` and `Geography` custom scalars are allowed in the schema. Note that built-in scalars like `Int`, `String`, etc cannot be masked. 
+
```
jberryman, line 61: Does this mean that fields of this type will be masked implicitly? Why not just insist the user masks such fields themselves?

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+**Enums**
+
+```
+allowed_enums:
+- name: Direction
+  values: [North, South]
+```
+
+Enum types can specify a subset of values for the enum.
+
```
jberryman, line 92: Would these just replace any offending value with a `null` should they be returned? And/or will we filter nulls where we have a list, e.g. `[Direction]`, and so these would be like row-level permissions?

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+  user(id: "<x-hasura-user-id>", limit: 1) {
+    a
+    b
+  }
+}
+```
+
+## Few questions to tackle:
+
+1. This will be a breaking change as all remote schemas will be hidden for all role except `admin`.
```
jberryman, line 236: I guess a nice UI would allow editing schema for a new role by either pre-populating with a completely unfiltered schema, or cloning an existing role's schema

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+This was tackled in #2690 by introducing a feature flag for enabling remote schema permissions explicitly.
+2. UX of this feature in the console.
+3. The API proposed above is complete in defining the permission for a role and a remote schema. Should we have smaller additive permission API e.g. 
+one which takes permissions for a single type (of any type) and adds them together to yield the final schema?
```
jberryman, line 240: I'm also wondering what happens when the schema changes; seems like the whitelisting approach might be difficult to maintain? On the other hand it's a way of getting a more stable or simple schema even when the remote changes.

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+4. How will it work with remote joins?
+5. How will it work with multiple roles?
```
jberryman, line 242: The user assumes only one role at a time, right? Or are you referring to upcoming changes?

jberryman:
Summarizing briefly my understanding of how above relates to permissions in hasura (from PG)

- **type level permissions**: These are like table-level permissions (which are further broken down into query/mutate permissions)
- **argument presets**: this is unique, but certain use-cases are equivalent to row-level permissions
- **field masking for objects**: this is equivalent to column-level permissions on tables

jberryman:
> I was wondering instead of describing the permission in yaml metadata, if we can do it graphql SDL. For example:

I think the idea of using SDL syntax here makes some sense to me. But I'm imagining something that was valid SDL, but which has slightly different semantics, and which we use to merge with the source schema: so e.g. there's no way to express default arguments in the permissions SDL. In fact I suppose an argument list _without_ a default would be invalid in the permissions SDL

> We will probably need to remove the entire where argument from todoDeep. Making partial types does not seem like a good DX.

What do you mean? Sorry to be dense.

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+                value:
+                  from_session_variable: "x-hasura-user-id"
+          - name: "limit"
+            value:
+              static: 1
+    - name: User
+      fields:
+      - name: a
+      - name: b
+      - name: c
```
jberryman, line 182: This (just tackling Objects + fields, and argument presets) seems enough to me; is there a reason we need to deal with interfaces, etc?

rfcs/remote-schema-permissions.md
```diff
@@ -0,0 +1,244 @@
...
+
+1. This will be a breaking change as all remote schemas will be hidden for all role except `admin`.
+This was tackled in #2690 by introducing a feature flag for enabling remote schema permissions explicitly.
+2. UX of this feature in the console.
+3. The API proposed above is complete in defining the permission for a role and a remote schema. Should we have smaller additive permission API e.g. 
+one which takes permissions for a single type (of any type) and adds them together to yield the final schema?
+4. How will it work with remote joins?
+5. How will it work with multiple roles?
+6. Implementation: how do we generate input types for each field when there are presets (currently input types are shared across fields)?
+Possible solution: Suffix type name, field name and role name to the role-based type
```
jberryman, line 244: Yeah I can't think of any other solution

tirumaraiselvan:
> > We will probably need to remove the entire where argument from todoDeep. Making partial types does not seem like a good DX.
> 
> What do you mean? Sorry to be dense.

My fault, this definitely requires more explanation. So if we preset an argument which is an input object, then according to the yaml syntax proposed, we should be creating a new type after stripping _only_ the preset input field. For example in the SDL schema:

```
input BoolExp {
  id: IntCmpExp
  name: TxtCmpExp
}

todoDeep(where: BoolExp = "{id : {_eq: x-hasura-user-id}}"): [Todo]
```
would lead to creating a new type, say `input BoolExp_role_user { name: TxtCmpExp } ` (`id: IntCompExp` is stripped). 

This new type generation doesn't really make sense in SDL syntax because the SDL defines all the types and fields. Hence, I was proposing to not create any new "partial" input types and simply remove the `where` argument entirely from `todoDeep` field.

rfcs/remote-schema-permissions.md
tirumaraiselvan: Yes, this was for later.

rfcs/remote-schema-permissions.md
tirumaraiselvan: No, they are always visible. Since the spec dictates that each graphql schema have built-in scalard we only scope permissions to custom scalars.

rfcs/remote-schema-permissions.md
tirumaraiselvan: Hmm, not sure. Should we worry about response validation as well?

rfcs/remote-schema-permissions.md
tirumaraiselvan: Not really, it is just for completeness. We can scope it to just object types.

rfcs/remote-schema-permissions.md
tirumaraiselvan: Good question. Will add it to the list.

jberryman:
My thoughts so far:

- we should focus only on objects + input types + argument presets
- sticking with the original YAML format specification is best; it is fairly simple, and I think should make sense to users, and is close to the actual data model (the SDL proposal seems to be at a higher level, which makes the spec more difficult to talk about in my opinion)

What do you think?

The handling of input types does seem like the main sticking point. A couple alternatives:

- insist that presets fully saturate a field (so only a complete `BoolExp` would be allowed in `where`); then we don't need to mess with input types at all
- or: same as above, but also allow _defining a new argument field_ that will be referenced to saturate any nested input types, so e.g. the new rewritten schema would look like:
```
type Query {
  todoDeep(name_that_internally_gets_piped_to_BoolExp: TxtCmpExp): [Todo]
}
```
- or: do the above automatically by essentially "unboxing" input type fields into the argument list; we'd probably have to do a similar renaming/namespacing thing as in the proposal to create a new input type, but this would be better as it doesn't muddy the schema so much with new input types

tirumaraiselvan:
> My thoughts so far:
> 
> * we should focus only on objects + input types + argument presets
> * sticking with the original YAML format specification is best; it is fairly simple, and I think should make sense to users, and is close to the actual data model (the SDL proposal seems to be at a higher level, which makes the spec more difficult to talk about in my opinion)
> 
> What do you think?

Sure, I am fine with these too. I would still like a SDL like syntax in the UI. And then the SDL is converted to this YAML. Incidentally, this is what "create action" in console does -- takes input in SDL and converts it to yaml.

> The handling of input types does seem like the main sticking point. A couple alternatives:
> 
> * insist that presets fully saturate a field (so only a complete `BoolExp` would be allowed in `where`); then we don't need to mess with input types at all
> * or: same as above, but also allow _defining a new argument field_ that will be referenced to saturate any nested input types, so e.g. the new rewritten schema (that the client would see)  would look like:
> 
> ```
> type Query {
>   todoDeep(name_that_internally_gets_piped_to_BoolExp: TxtCmpExp): [Todo]
> }
> ```
> 
> * or: do the above automatically by essentially "unboxing" input type fields into the argument list; we'd probably have to do a similar renaming/namespacing thing as in the proposal to create a new input type, but this would be better as it doesn't muddy the schema so much with new input types

I like point 1 too, atleast for the first version.  I also like the idea of user supplying new arguments but I am not sure if I understand your proposal fully, do you mean if a field has an argument which is an input object with 10 keys and one of them is preset then we would end up creating 9 (new) arguments for the field? If yes, then we are creating new arguments instead of types now. Is that better?

But what if we just ask the user for a new input type name when they are presetting an input object? We can create a new type by removing the preset path (as proposed initially) but give it this new name.

jberryman:
> I also like the idea of user supplying new arguments but I am not sure if I understand your proposal fully, do you mean if a field has an argument which is an input object with 10 keys and one of them is preset then we would end up creating 9 (new) arguments for the field? 

Yeah that's right, I'm not sure what the YAML API would look like exactly, but maybe something like 

```
    allowed_objects:
    - name: Query
      fields:
      - name: foo
        proxy_field: true
        type: UUID
      - name: hello
      - name: user
        argument_presets:
        - name: "identifier"
          argument_presets:
          - name: user_id
            value:
               from_session_variable: "x-hasura-user-id"
          - name: group_id
            value:
               from_proxy_variable: "foo"
```

> If yes, then we are creating new arguments instead of types now. Is that better?

 I think it's a little better than having an input type with a weird name that is used just once, but probably not that much better...

> But what if we just ask the user for a new input type name when they are presetting an input object? We can create a new type by removing the preset path (as proposed initially) but give it this new name.

That seems okay to me too. But it also seems like we're halfway to a more general feature for arbitrarily reshaping and amending the schema...

jberryman:
## The promised Dhall-based schema-shaping idea

I hit a wall here since dhall (at least in the current form) doesn't quite work for us, and set this aside. But FWIW here is a rough sketch of what I was thinking...

### Schema shaping using a statically-typed embedded language

The idea is to give admins the ability to reshape the schema using all the power of a general purpose (-ish) programming language. Or maybe the feature here would be used at first only by frontend devs to implement and iterate on features without requiring backend changes. This might also result in shareable user "plugins".

#### generalizing the remote schema permissions feature

Imagine we have an embedded language (our console users can write little programs, and graphql-engine can interpret them) that is identical to haskell (i.e. I'll use haskell syntax here). 

Imagine [this is the schema we generate](https://github.com/hasura/graphql-engine/blob/master/server/tests-py/remote_schemas/nodejs/index.js) and which the user would like to massage or reshape in some way.

The main idea is to allow the user to define **a function from one field to another** while defining any new types they need. The second boring bit is a **configuration that maps these reshaping functions onto specific fields in the schema** via some path specification (this mapping will be type-checked).

**Note**: we can think of arguments to a field in an object type as a sort of punning; if we remove the arguments we get types that describe the _output_ format. It's these kinds of types that our embedded language will deal with.

**Example1**: say the user wants the top-level `message` field to construct an `id` by concatenating two int strings (maybe that's how they are constructed internally but it's not a great UX). We can think of `message` as a function:

    message :: Int -> Message

That's what we have to work with, so what the admin wants to define then is:

    ourMessage :: (Int -> Message) -> (Int -> Int -> Message)
    ourMessage message = \user_id group_id -> message (read (show user_id <> show group_id))

**Note**: Here `graphql-engine` has already brought into scope the `Message` type (see note above).

Then the user specifies (in some other configuration) that the `ourMessage` field transformation should map to `message` in the `Query` entrypoint type (there are lots of options/possibilities here, which I want to ignore, e.g. we can replace the existing field, make a sibling field, etc)

What happens here, and nice things about it:
- since `ourMessage` is well-typed we can be sure the user's program won't go too wrong
- `ourMessage` encodes both the change to the schema (from the type) as well as how to get there (the code `graphql-engine` will interpret at runtime)
- the user has the full power of a programming language at their disposal; they don't need to open an issue

**Example 2**: The admin wants `Query.messages` to only return messages with no error message, and provide an empty includes:

    nonErrorMessages :: (MessageWhereInpObj -> IncludeInpObj -> [Message]) -> (MessageWhereInpObj -> [NonErrorMessage]
    nonErrorMessages messages = \ _where -> map convert $ filter (null . errorMsg) $ messages _where (IncludeInpObj [] [])
        where convert Message{..} = NonErrorMessage{..} 

    -- Here we define the new type that we need. This will be added to the schema when we interpret
    -- TODO: but what if we need it to have field arguments...? What about `implementsCommunication`?
    data NonErrorMessage = NonErrorMessage  {
        id :: Int,
        name :: String,
        msg :: String
     }

#### What about roles?

actually we want the functions above to be _closed over role and session variable data for the user_. So something instead like:

    ourMessage :: RoleAndVariables -> (Int -> Message) -> (Int -> Int -> Message)

aaaand but actually we'd probably want the result _type_ (in this case `Int -> Int -> Message`) _to vary with the role_. There are a couple ways we could do this with our haskell-like DSL above, but it's getting a bit weird...

### Dhall

I was initially excited thinking dhall might be a really good fit since it has:
- strong safety guarentees
- super easy to embed in haskell

unfortunately the fact that recursive types can't be defined naturally would seem to be a deal-breaker for us (maybe not?):    https://github.com/dhall-lang/dhall-lang/issues/3

I'm not sure what alternatives there are. I guess there are several typed luas. Can idris be embedded?

Link dump:
https://hackage.haskell.org/package/egison
https://hackage.haskell.org/package/mappy
https://hackage.haskell.org/package/idris
https://hackage.haskell.org/package/hint (requires ghc on host machine)
https://github.com/amuletml/amulet 
https://github.com/roblox-ts/roblox-ts
https://github.com/TypeScriptToLua/TypeScriptToLua
https://github.com/teal-language/tl/blob/master/docs/tutorial.md

PR #5683, opened 2020-08-27
track graphql paths, tables, etc. responsible for trace messages in the metadata

evertedsphere:
currently this adds a `field_paths` metadata field to remote joins, which shows which gql paths triggered a request to a remote api

### Changelog

### Affected components

- [ ] Server

#### Catalog upgrade

- [ ] No

#### Metadata

- [ ] No

#### GraphQL

#### Breaking changes

- [ ] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/GraphQL/RemoteServer.hs
+++ b/server/src-lib/Hasura/GraphQL/RemoteServer.hs
@@ -10,6 +10,7 @@ import           Control.Monad.Unique
 import           Data.Aeson                             ((.:), (.:?))
 import           Hasura.HTTP
 import           Hasura.Prelude
+import qualified Data.Text as T
 
 import qualified Data.Aeson                             as J
 import qualified Data.ByteString.Lazy                   as BL
@@ -314,8 +315,11 @@ execRemoteGQ'
   -> GQLReqUnparsed
   -> RemoteSchemaInfo
   -> G.OperationType
+  -- TODO should we leave this as a list or use a nonempty?
+  -- perhaps it _has_ to be empty for a toplevel remote schema call
+  -> [FieldPath]
   -> m (DiffTime, [N.Header], BL.ByteString)
-execRemoteGQ' env manager userInfo reqHdrs q rsi opType =  do
+execRemoteGQ' env manager userInfo reqHdrs q rsi opType fieldPaths =  do
   when (opType == G.OperationTypeSubscription) $
     throw400 NotSupported "subscription to remote server is not supported"
   confHdrs <- makeHeadersFromConf env hdrConf
@@ -337,10 +341,13 @@ execRemoteGQ' env manager userInfo reqHdrs q rsi opType =  do
            , HTTP.responseTimeout = HTTP.responseTimeoutMicro (timeout * 1000000)
            }
   Tracing.tracedHttpRequest req \req' -> do
+    Tracing.attachMetadata [("field_paths", renderedFieldPaths)]
     (time, res)  <- withElapsedTime $ liftIO $ try $ HTTP.httpLbs req' manager
     resp <- either httpThrow return res
     pure (time, mkSetCookieHeaders resp, resp ^. Wreq.responseBody)
   where
+    renderFieldPath fp = T.intercalate "." $ map getFieldNameTxt (unFieldPath fp)
+    renderedFieldPaths = T.pack . show $ map renderFieldPath fieldPaths
     RemoteSchemaInfo url hdrConf fwdClientHdrs timeout = rsi
     httpThrow :: (MonadError QErr m) => HTTP.HttpException -> m a
     httpThrow = \case
--- a/server/src-lib/Hasura/RQL/DML/RemoteJoin.hs
+++ b/server/src-lib/Hasura/RQL/DML/RemoteJoin.hs
@@ -72,10 +72,6 @@ executeQueryWithRemoteJoins env manager reqHdrs userInfo q prepArgs rjs = do
   where
     rjMap = Map.fromList $ toList rjs
 
--- | Path to the remote join field in query response JSON from Postgres.
-newtype FieldPath = FieldPath {unFieldPath :: [FieldName]}
-  deriving (Show, Eq, Semigroup, Monoid, Hashable)
-
 type Alias = G.Name
 
 appendPath :: FieldName -> FieldPath -> FieldPath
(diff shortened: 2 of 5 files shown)


server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -310,3 +312,7 @@ resolveWebhook env (InputWebhook urlTemplate) = do
   let eitherRenderedTemplate = renderURLTemplate env urlTemplate
   either (throw400 Unexpected . T.pack)
     (pure . ResolvedWebhook) eitherRenderedTemplate
+
+-- | Path to a remote join field in e.g. query response JSON from Postgres.
```
paf31, line 316:
```suggestion
-- | A path to a substructure in a JSON document via a series of fields.
```

server/src-lib/Hasura/RQL/DML/RemoteJoin.hs
```diff
@@ -417,7 +416,11 @@ fetchRemoteJoinFields env manager reqHdrs userInfo remoteJoins = do
                                  (map _rjfField batchList)
         gqlReqUnparsed = (GQLQueryText . G.renderExecutableDoc . G.ExecutableDocument . unGQLExecDoc) <$> gqlReq
     -- NOTE: discard remote headers (for now):
-    (_, _, respBody) <- execRemoteGQ' env manager userInfo reqHdrs gqlReqUnparsed rsi G.OperationTypeQuery
+    (_, _, respBody) <- do
+      -- TODO(evertedsphere): move this into metadata, so that a trace further
+      -- down the callstack (i.e. while performing the http req) will print 
+      -- these field aliases.
+      execRemoteGQ' env manager userInfo reqHdrs gqlReqUnparsed rsi G.OperationTypeQuery (toList $ fmap _rjfPath batch)
```
paf31, line 423: I was going to say let's pass a `Maybe (NonEmpty FieldPath)` here, but what about adding something like

```haskell
data RemoteCallReason
  = TopLevelRemoteCall
  -- ^ pass the whole query to the remote schema as an optimization
  | RemoteJoinCall (NonEmpty FieldPath)
  -- ^ fetching a non-empty set of fields in the course of executing a remote join
```

to document the exact nature of the new argument?

server/src-lib/Hasura/GraphQL/RemoteServer.hs
```diff
@@ -337,10 +341,13 @@ execRemoteGQ' env manager userInfo reqHdrs q rsi opType =  do
            , HTTP.responseTimeout = HTTP.responseTimeoutMicro (timeout * 1000000)
            }
   Tracing.tracedHttpRequest req \req' -> do
+    Tracing.attachMetadata [("field_paths", renderedFieldPaths)]
     (time, res)  <- withElapsedTime $ liftIO $ try $ HTTP.httpLbs req' manager
     resp <- either httpThrow return res
     pure (time, mkSetCookieHeaders resp, resp ^. Wreq.responseBody)
   where
+    renderFieldPath fp = T.intercalate "." $ map getFieldNameTxt (unFieldPath fp)
+    renderedFieldPaths = T.pack . show $ map renderFieldPath fieldPaths
```
paf31, line 350:
```suggestion
    renderedFieldPaths = T.pack . show @[Text] $ map renderFieldPath fieldPaths
```

or 

```suggestion
    renderedFieldPaths = T.intercalate ", " (map renderFieldPath fieldPaths)
```

server/src-lib/Hasura/RQL/DML/RemoteJoin.hs
evertedsphere: this sounds like a much better idea

evertedsphere:
cc @paf31

server/src-lib/Hasura/GraphQL/RemoteServer.hs
```diff
@@ -314,8 +315,11 @@ execRemoteGQ'
   -> GQLReqUnparsed
   -> RemoteSchemaInfo
   -> G.OperationType
+  -- TODO should we leave this as a list or use a nonempty?
```
paf31, line 318: Remove this TODO now?

paf31 (changes requested):
Looks good, but we should remove the old comment please.

server/src-lib/Hasura/GraphQL/RemoteServer.hs
evertedsphere: Good catch.

paf31:
@evertedsphere Need to fix the build warnings before CI can pass.

paf31:
@evertedsphere Can you please resolve the conflicts?

server/src-lib/Hasura/GraphQL/RemoteServer.hs
```diff
@@ -337,10 +338,16 @@ execRemoteGQ' env manager userInfo reqHdrs q rsi opType =  do
...
            }
   Tracing.tracedHttpRequest req \req' -> do
+    Tracing.attachMetadata [("field_paths", renderedFieldPaths)]
     (time, res)  <- withElapsedTime $ liftIO $ try $ HTTP.httpLbs req' manager
     resp <- either httpThrow return res
     pure (time, mkSetCookieHeaders resp, resp ^. Wreq.responseBody)
   where
+    renderFieldPath fp = T.intercalate "." $ map getFieldNameTxt (unFieldPath fp)
+    renderedFieldPaths = case fieldPaths of
+      TopLevelRemoteCall -> "toplevel"
```
paf31, line 348: I don't think we should include this string in the metadata. Instead, I think it would be better to emit a list of root fields in the case of a top-level remote call. However, we can add that later, and for now, just not emit the `field_paths` metadata at all in this case.

server/src-lib/Hasura/RQL/DML/RemoteJoin.hs
```diff
@@ -420,7 +420,12 @@ fetchRemoteJoinFields env manager reqHdrs userInfo remoteJoins = do
                                  (map _rjfField batchList)
         gqlReqUnparsed = (GQLQueryText . G.renderExecutableDoc . G.ExecutableDocument . unGQLExecDoc) <$> gqlReq
     -- NOTE: discard remote headers (for now):
-    (_, _, respBody) <- execRemoteGQ' env manager userInfo reqHdrs gqlReqUnparsed rsi G.OperationTypeQuery
+    (_, _, respBody) <- do
+      -- TODO(evertedsphere): move this into metadata, so that a trace further
```
paf31, line 424: I'm not following this note. We can't put this in metadata here, or it'll appear at the wrong place in the trace.

evertedsphere: right, yeah, i put this note in a while ago so i can't say what i was thinking then...

PR #5694, opened 2020-08-31
Introduce `safeSelectionSet` to prevent conflicts in root fields

nicuveo:
### Description
We recently found an oversight of selection sets: we don't check for duplicates. In the unlikely (but, as evidenced, not impossible) case where we don't prevent a conflict from arising, said conflict should be detected when we build the schema. And in turns out that in some degenerate cases, we would **construct a schema without detecting the conflict** (specifically: when the conflicting root fields would yield types with the exact same fields, that therefore can't be differentiated).

(See discussion with @codingkarthik on Slack for more info, the test case, and the context.)

### Affected components

- [ ] Server

### Solution and Design
This is a very simple fix, that only fixes the problem for root fields. There are two other ways we could prevent this from happening:
  - make all combinators monadic and throw on invalid schema (bad, no)
  - make all combinators require a hashmap of field parsers, to delegate the responsibility of checking for duplicates to the callers (better, but perhaps overkill)

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Parser.hs
+++ b/server/src-lib/Hasura/GraphQL/Parser.hs
@@ -21,6 +21,7 @@ module Hasura.GraphQL.Parser
   , list
   , object
   , selectionSet
+  , safeSelectionSet
   , selectionSetObject
 
   , InputFieldsParser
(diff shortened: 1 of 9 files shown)


abooij (approved):
LGTM. Thanks @codingkarthik for adding tests!

tirumaraiselvan (approved):
no cl required

PR #5706, opened 2020-09-01
console: remove ONLY as default for ALTER TABLE in column alter operations (close #5512)

### LINKED ISSUE #5512: console: should add option to disable ONLY on ALTER TABLE
TimescaleDB hypertable doesn't support `ALTER TABLE ONLY` SQL statement. We should have an ENV option to disable `ONLY`

```json
{
    "statement": "ALTER TABLE ONLY \"public\".\"table_name\" ALTER COLUMN \"foo\" SET DEFAULT 0;",
    "prepared": false,
    "error": {
        "exec_status": "FatalError",
        "hint": null,
        "message": "ONLY option not supported on hypertable operations",
        "status_code": "0A000",
        "description": null
    },
    "arguments": []
}
```

Screenshot_20200803_121017

jasvithaMad:
### Description
This pr changes the default to not add ONLY for ALTER TABLE in column alter operations.

### Changelog

### Affected components

- [x] Console

### Related Issues
Closes the issue #5512


console/src/components/Common/utils/v1QueryUtils.ts
```diff
@@ -493,9 +493,7 @@ export const generateCreateEventTriggerQuery = (
         : null,
       update: state.operations.update
         ? {
-            columns: state.operationColumns
-              .filter(c => !!c.enabled)
-              .map(c => c.name),
+            columns: ['*'],
```
kolharsam, line 496: @jasvithaMad these changes are not supposed to be here right?

kolharsam (commented):
@jasvithaMad these changes look okay! 👌 
I need to try these out though. Will revert once I test these changes.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Next release
 
+- console: remove ONLY as default for ALTER TABLE in column alter operations (close #5512) #5706
+
```
beerose, line 6: It doesn't require changelog.
```suggestion
```

beerose (changes requested):
Why was `console/src/components/Common/utils/v1QueryUtils.ts` removed? Could you revert this change?

Apart from that, it looks good.

CHANGELOG.md
tirumaraiselvan: I think a changelog is good to have because this may affect users (those using inherited tables)

```
CREATE TABLE cities (
  name       text,
  population real,
  elevation  int     -- (in ft)
);

CREATE TABLE capitals (
  state      char(2)
) INHERITS (cities);

postgres=# alter table only cities drop column name;
ALTER TABLE
postgres=# select * from cities;
 population | elevation 
------------+-----------
(0 rows)

postgres=# select * from capitals;
 name | population | elevation | state 
------+------------+-----------+-------
(0 rows)
```

What has essentially happened above is that the set of inherited columns has reduced to the 2 columns: `population` and `elevation`. So the `capitals` table is still inherited from `cities` table;


The reverse is not possible though, you can't drop a column ONLY from the inherited table as it would not be compatible with it's parent table:

```
postgres=# alter table only capitals drop column population;
ERROR:  cannot drop inherited column "population"
```

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -1576,7 +1576,7 @@ const saveColumnChangesSql = (colName, column, onSuccess) => {
     if (colDefault !== '') {
       // ALTER TABLE ONLY <table> ALTER COLUMN <column> SET DEFAULT <default>;
```
beerose, line 1577:
```suggestion
      // ALTER TABLE <table> ALTER COLUMN <column> SET DEFAULT <default>;
```

tirumaraiselvan (approved):
changelog

beerose (approved):
Console

PR #5549, opened 2020-08-07
cli: fix error on seed create command

kolharsam:
fixes #5532 

### Description

Fixes an issue while using the seed migrations with tables that have capital letters in them.

### Affected components

- [x] CLI

PR #5710, opened 2020-09-01
cli: allow for schema specification when using hasura seeds create (close #5711)

### LINKED ISSUE #5711: [CLI] `hasura seeds create` doesn't allow for specifying a table schema
The following release [v1.3.1-beta.1](https://github.com/hasura/graphql-engine/releases/tag/v1.3.1-beta.1) fixed an issue with creating seed migrations for tables with capital letters, but introduced another one by removing the possibility for specifying tables from schemas other than `public`.

So, `v1.3.1` doesn't allow for something like:

`hasura seeds create seeds_from_auth --from-table auth.providers --endpoint endpoint --admin-secret secret`

https://github.com/hasura/graphql-engine/pull/5710 fixes the issue for both capitalized tables and schemas other than `public`.

nunopato:
closes #5711 

### Description

This PR fixes a regression from #5549. Now capitalized Schema names and table names are permitted in the seeds command. 

### Affected components

- [x] CLI


cli/migrate/migrate.go
```diff
@@ -1850,7 +1851,12 @@ func (m *Migrate) ExportDataDump(tableNames []string) ([]byte, error) {
 	// to support tables starting with capital letters
 	modifiedTableNames := make([]string, len(tableNames))
 	for idx, val := range tableNames {
-		modifiedTableNames[idx] = fmt.Sprintf(`"%s"`, val) 
+		split := strings.Split(val, ".")
+		if len(split) == 2 {
+			modifiedTableNames[idx] = fmt.Sprintf(`%s."%s"`, split[0], split[1])
```
kolharsam, line 1856:
```suggestion
			modifiedTableNames[idx] = fmt.Sprintf(`"%s"."%s"`, split[0], split[1])
```

kolharsam, line 1856: 2 points here : 

1. I would suggest that we wrap the schema part (`split[0]`) with `"` as well, for good measure :)
2. I was also thinking about the case when the user provides the input with the `"` in place? I'm thinking we should be handling that case too. Prior to the latest release of the CLI (v1.3) I think some users passed in table names like so `\"Schema_Name\".\"Table_Name\"`, so I guess by handling this we would still be supporting such inputs. What do you think?

cli/migrate/migrate.go
```diff
@@ -1850,7 +1851,12 @@ func (m *Migrate) ExportDataDump(tableNames []string) ([]byte, error) {
 	// to support tables starting with capital letters
 	modifiedTableNames := make([]string, len(tableNames))
 	for idx, val := range tableNames {
-		modifiedTableNames[idx] = fmt.Sprintf(`"%s"`, val) 
+		split := strings.Split(val, ".")
```
kolharsam, line 1854: I think we should be throwing an error if `len(split)` isn't either 2 or 1, as it's not a valid table/schema name.

kolharsam (commented):
Hey @nunopato! 👋 Thanks for your PR!
Just leaving some notes here :)

cli/migrate/migrate.go
nunopato: @kolharsam thanks for your comments! 

1. thanks for pointing that out, I didn't miss it on purpose.
2. although not explicitly here, that case is already "being dealt" because either Go or Postgres are removing the duplicates for `"`. I tried all those cases with this implementation and they all work. I might add a couple of tests for those special cases.

cli/migrate/migrate.go
nunopato: Sounds good! thank you.

cli/migrate/migrate.go
kolharsam: Re 2: Yup, it's go that does the trick. But as per my observations, you can escape the \" and it'd be part of the input.

nunopato: @kolharsam I meant with the escape. The escape is simply allowing for a duplicate `"` which is, eventually, considered a duplicate and removed.

cli/migrate/migrate.go
nunopato: @kolharsam let me know if this is ready to go

kolharsam (approved):
@nunopato I think these are good to go! 👍 

cc: @scriptonist @tirumaraiselvan

nunopato:
@kolharsam I just saw that you changed the check condition from an AND to an OR and I don't think that's what you want to do.

nunopato:
@kolharsam @scriptonist @tirumaraiselvan any chance we could get this merged soon? if there is anything missing here, just let me know (the seeds command, starting on `v1.3.1` is basically broken without this fix)

scriptonist (approved):

PR #5713, opened 2020-09-02
cli: add exclude-schema & full flags to migrate create command

### LINKED ISSUE #4301: Exporting a migration from server does not export extensions
When using `hasura migrate create "name" --from-server`, the migration does not contain the Postgres extensions. This will fail when used on a fresh database instance.

### Example
We're using a multi-database single-tenant setup, which means we have an Hasura instance for each database, and every database belongs to a single, specific tenant. All databases are structured the same way.

When creating the first migration from an existing database, the migration will not contain the `uuid-ossp`, and thus fails when applied to a fresh instance.

### Steps to reproduce
- **Setup**
  - Postgres
    - Create a new database
    - Add the `uuid-ossp` extension
    - Set up a table using that extension (e.g column `id uuid DEFAULT public.uuid_generate_v4() NOT 
    NULL`)
  - Hasura
    - Run a fresh instance and connect to the database
    - `hasura init --directory my-project --endpoint http://localhost:8080`
    - `cd my-project`
    - `hasura migrate create "init" --from-server`
- **Reproduce**
  - Drop the database and close the Hasura instance
  - Create a new database and open a new Hasura instance
  - Apply the migration: `hasura migrate apply --version "<version>"`

This will fail (you can examine the generated up.sql. It does not contain the extension).

kolharsam:
resolves #4301 

### Description

This PR adds 2 flags to the `migrate create` command.
 - `exclude-schema` - meant to be used while you'd like to exclude certain schema's while running pg_dump
 - `full` - meant to be used when you'd like to take to run pg_dump across all schemas along with extensions

### Changelog

### Affected components

- [x] CLI


tirumaraiselvan:
Even `sql-from-server` should support both these flags?

PR #5719, opened 2020-09-03
server: customize tracking tables with a custom name

### LINKED ISSUE #3811: Custom GraphQL types for tables not conforming to GraphQL naming spec
**Version: 1.1.0-beta.2**

I have a table from an underlying, existing database that is a legacy table with spaces in the name. Let's call it "User Accounts". This name does not conform to the GraphQL naming standard and thus cannot be exposed over the API.

In an attempt to solve this issue - I went to Modify -> Custom GraphQL Root Fields and , under the Query and Subscription section made the Select "UserAccounts" and hit save. The message said it was successful but I am not seeing "UserAccounts" anywhere under my available Schemas on the GraphIQL page. 

The [docs](https://docs.hasura.io/1.0/graphql/manual/schema/custom-field-names.html#) say this is available as of version v1.0.0-beta.8 and higher but I am using 1.1.0-beta.2 and it doesn't appear to be working.

codingkarthik:
This PR introduces the concept of custom table identifiers. An identifier is an alias for a table. When a table is tracked with an identifier, the nodes generated for the table will be according to the identifier. The identifier can be useful to use when 
tracking tables with non GraphQL compliant names or it can be useful if the nodes that are generated should not be dependent on the table name i.e when a table is tracked with an identifier and if the table has been renamed, the node names will not be changed due to the identifier.

### Changelog

- [X] `CHANGELOG.md` is updated with user-facing content relevant to this PR. If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [X] Server
- [ ] Console
- [ ] CLI
- [X] Docs
- [ ] Community Content
- [ ] Build System
- [X] Tests
- [ ] Other (list it)

### Solution and Design

fixes #3811 

#### Catalog upgrade

- [ ] No
- [x] Yes

#### Metadata

- [ ] No
- [x] Yes
    - [x] Yes
    - [ ] Not required
    - [x] Yes
    - [ ] Not required
    - [x] Yes
    - [ ] Not required

#### GraphQL
- [x] New GraphQL schema is being generated:
   - [x] New types and typenames are correlated
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     If a table has an identifier set, then the types and names are generated according to the identifier.

     Schema Resolve:-
     
  3. Logging

## LIMITATIONS:

To be able to track a table, the constraint names should be GraphQL compliant, otherwise the table cannot be tracked.


docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -304,6 +357,67 @@ Args syntax
      - :ref:`CustomColumnNames`
      - Customise the column fields
 
+.. _set_table_configuration:
+
+set_table_configuration
+-----------------------
+
+``set_table_configuration`` sets the configuration of a table, through which
+an identifier, custom root fields or custom column names of an already tracked
```
marionschleifer, line 366: Is it only one of those (identifier, custom root fields, custom column names) that can be set? Or actually all three?

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -304,6 +357,67 @@ Args syntax
      - :ref:`CustomColumnNames`
      - Customise the column fields
 
+.. _set_table_configuration:
+
+set_table_configuration
+-----------------------
+
+``set_table_configuration`` sets the configuration of a table, through which
```
marionschleifer, line 365:
```suggestion
``set_table_configuration`` sets the configuration of a table through which
```

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -304,6 +357,67 @@ Args syntax
...
+an identifier, custom root fields or custom column names of an already tracked
+table can be set. This will **replace** the already present configuration.
```
marionschleifer, line 367: Can we link to the deprecated configuration here?

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -242,13 +289,19 @@ Custom Root Fields
 
 .. _set_table_custom_fields:
 
-set_table_custom_fields
------------------------
+set_table_custom_fields (deprecated)
+-----------------------------------
```
marionschleifer, line 293:
```suggestion
------------------------------------
```

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -242,13 +289,19 @@ Custom Root Fields
 
 .. _set_table_custom_fields:
 
-set_table_custom_fields
------------------------
+set_table_custom_fields (deprecated)
```
marionschleifer, line 292: Should we also say in [this table](https://deploy-preview-5719--hasura-docs.netlify.app/graphql/core/api-reference/schema-metadata-api/index.html#request-types) that this is depracated?

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -242,13 +289,19 @@ Custom Root Fields
...
+-----------------------------------
+
+``set_table_custom_fields`` has been deprecated now, use the
```
marionschleifer, line 295:
```suggestion
``set_table_custom_fields`` has been deprecated. Use the
```

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -149,6 +149,48 @@ Add a table/view ``author``:
       }
    }
 
+An ``identifier`` can be used to track an table with the identifier. This can
```
marionschleifer, line 152:
```suggestion
An ``identifier`` can be used to track a table with the identifier. This can
```

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -149,6 +149,48 @@ Add a table/view ``author``:
...
+be useful when a table name is not GraphQL compliant, like ``"Users Address"``
```
marionschleifer, line 153:
```suggestion
be useful when a table name is not GraphQL compliant, like ``Users Address``.
```

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -149,6 +149,48 @@ Add a table/view ``author``:
...
+- ``users_address_aggregate``
+- ``insert_users_address``
+- ``insert_users_address_one``
+- ``update_users_address``
+- ``update_users_address_by_pk``
+- ``delete_users_address``
+- ``delete_users_address_by_pk``
+
+.. note::
+  graphql-engine expects the constraint names(if any) of a table to be GraphQL compliant_ to be
```
marionschleifer, line 189:
```suggestion
  graphql-engine requires the constraint names (if any) of a table to be GraphQL compliant_ in order to be
```

marionschleifer (dismissed):

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
codingkarthik: Oh, we can set all the three. I'll make the change.

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
codingkarthik: Sure, I'll do that

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
codingkarthik: I think yes, but would like to hear what @tirumaraiselvan also has to say

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
```diff
@@ -182,6 +224,11 @@ Table Config
      - Required
      - Schema
      - Description
+   * - identifier
```
lexi-lambda, line 227: The name “identifier” here strikes me as very vague and a little confusing. We already use the phrase “custom fields” for the field names, why not call this a “custom name”?

server/src-lib/Hasura/GraphQL/Parser/Class.hs
```diff
@@ -122,6 +123,23 @@ askTableInfo tableName = do
...
 
+-- | Helper function to get the table display name. A table may have an
+-- identifier configured with it. When the identifier exists, the GraphQL nodes
+-- that are generated according to the identifier. For example: Let's say,
+-- we have a table called `users address`, the name of the table is not GraphQL
+-- compliant so we configure the table with a GraphQL compliant name,
+-- say `users_address`
+-- The generated top-level nodes of this table will be like `users_address`,
+-- `insert_users_address` etc
+getTableDisplayName
```
lexi-lambda, line 134: “Display name” is a bit vague for this concept, I think—“display” usually means “user-facing,” which this is not really in the context of the console. How about `getTableGraphQLName`, or `getTableGQLName` if the former is too verbose?

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
```diff
@@ -196,11 +199,19 @@ runSetExistingTableIsEnumQ (SetTableIsEnum tableName isEnum) = do
...
+  , _stcConfiguration :: !TableConfig
+  } deriving (Show, Eq, Lift)
+$(deriveJSON (aesonDrop 4 snakeCase) ''SetTableConfiguration)
+
 data SetTableCustomFields
   = SetTableCustomFields
   { _stcfTable             :: !QualifiedTable
   , _stcfCustomRootFields  :: !TableCustomRootFields
   , _stcfCustomColumnNames :: !CustomColumnNames
+  , _stcfIdentifier        :: !(Maybe G.Name)
```
lexi-lambda, line 214: If `set_table_custom_fields` is now deprecated, should we just not allow a custom name to be given to this API? That way, people who want to use custom names will have a reason to migrate to the new API.

server/src-lib/Hasura/RQL/Types/Table.hs
```diff
@@ -421,7 +423,8 @@ type TableCoreInfo = TableCoreInfoG FieldInfo PGColumnInfo
 
 tciUniqueOrPrimaryKeyConstraints :: TableCoreInfoG a b -> Maybe (NonEmpty Constraint)
 tciUniqueOrPrimaryKeyConstraints info = NE.nonEmpty $
-  maybeToList (_pkConstraint <$> _tciPrimaryKey info) <> toList (_tciUniqueConstraints info)
+  maybeToList (_pkConstraint <$> _tciPrimaryKey info)
+  <> (toList (_tciUniqueConstraints info))
```
lexi-lambda, line 427: The added parentheses here are redundant:

```suggestion
  <> toList (_tciUniqueConstraints info)
```

docs/graphql/core/api-reference/schema-metadata-api/index.rst
```diff
@@ -104,6 +104,11 @@ The various types of queries are listed in the following table:
      - 2
      - Add a table/view with configuration
 
+   * - :ref:`set_table_configuration <set_table_configuration>`
+     - :ref:`set_table_configuration_args <set_table_configuration_syntax>`
+     - 1
+     - Set table configuration of an already tracked table
```
lexi-lambda, line 110: The name “configuration” here is also kind of vague. We have other options on tables, such as `is_enum`, and those things are really just as much “configuration” as the things this API supports. I think a better term would be *customization*, as all these things seem customization-related. So what about calling this `set_table_customization`, instead?

lexi-lambda (changes requested):
This generally LGTM, though I’ve left a few comments, largely around tightening up the names we choose for these concepts.

server/src-lib/Hasura/GraphQL/Parser/Class.hs
codingkarthik: I remember having lot of confusion while choosing the name for this function 😅 . `getTableGQLName` is much better. Thank you!

server/src-lib/Hasura/RQL/DDL/Schema/Table.hs
codingkarthik: Yes, it makes sense. I'll remove the custom name part from it.

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
codingkarthik: I have changed `identifier` to `custom name`

docs/graphql/core/api-reference/schema-metadata-api/index.rst
codingkarthik: I've refactored the API to `set_table_customization`.

docs/graphql/core/api-reference/schema-metadata-api/table-view.rst
tirumaraiselvan: Doesn't it say already? https://github.com/hasura/graphql-engine/pull/5719/files#diff-e8eabc9c3c339a6063c7b2c29b6fdb009017d9e0459be2978aee7617f31c9afeR115

Or do you mean add an extra column called deprecated to the table? I don't think that is necessary at this point. Maybe when we have few more deprecated APIs.

codingkarthik: I added the `(deprecated)` indicator yesterday. Earlier, there was nothing. So, Marion wanted some sort of indicator that the API is deprecated

0x777 (approved):

0x777 (approved):
LGTM

tirumaraiselvan (approved):

marionschleifer (approved):

PR #5730, opened 2020-09-05
console: select first operator by default (close #5729)

### LINKED ISSUE #5729: Select first operator by default on the browse rows screen
Most of the time people filter by searching for an exact value, so setting the default operator to the first operator might save people a little bit of time

image

bertyhell:
### Description
When filtering rows in the data tab in the hasura console, the first operator will be selected by default.

### Changelog

not yet, I'm not sure under which version I need to add this.

### Affected components
- Console

### Related Issues
#5729

### Solution and Design
image

### Steps to test and verify
* login to the console
* go to the data tab
* select a table
* see the the first filter has the first operator selected by default. This will probably be _eq in most setups. 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


rikinsk (changes requested):
Hey @bertyhell. Thanks a ton for raising this PR. We really appreciate it. 

Unfortunately the solution implemented here does not work as expected. The operator dropdown is a controlled element and the selected value comes from the redux state. hence just displaying the first element as selected will not actually set the operator in the state and when the Run Query button is hit the filter will not actually be applied. The actual solution might involve setting the appropriate state and shouldnt need any changes to the UI components.

beerose:
Closing in favour of #6032, as discussed in https://github.com/hasura/graphql-engine/issues/5729.

PR #5736, opened 2020-09-07
Server: fix `created_at` bug related to scheduled triggers

### LINKED ISSUE #5722: Scheduled event triggers `created_at` time shifted +04:00
I don't quite understand how nobody has reported this yet, but scheduled event triggers do not work. (Or I am somehow missing something major here)

Step 1: I create a scheduled event trigger in the Hasura UI at 17:35:04 to be fired at 17:36:25
Step 2 I create a scheduled event trigger from my server at 17:35:27 to be fired at 17:36:27

Here's a screen shot of what I see in "pending triggers"
Screen Shot 2020-09-03 at 5 35 54 PM

The first thing to notice: look at the created_at and the "scheduled_time" ... something is off already.

Then, at 17:37:01, my webhook endpoint is hit, first with the UI-created trigger:
Screen Shot 2020-09-03 at 5 37 49 PM

And immediately after, with the programmatically-created trigger:
Screen Shot 2020-09-03 at 5 38 20 PM

Notice how the `created_at` times are incorrect.

Further confirmation of the `created_at` times being incorrect on the "processed events" tab:
Screen Shot 2020-09-03 at 5 53 27 PM

At first I thought it might have been something I'm doing wrong with Javascript dates, which are often tricky... So  here, I set the `schedule_at` time to be 4 hours ahead. In this case, I set it to be: `schedule_at 2020-09-04T01:56:34.530Z`

And here's what the pending trigger looks like in the UI:
Screen Shot 2020-09-03 at 5 59 00 PM

I hope this is enough to help you sort this out. I replicated this on v1.3.0 and 1.3.1

codingkarthik:
Earlier, the time of the event invocation was sent as the `created_at` value in the webhook's payload. This PR fixes the issue by sending the correct value i.e in case of one-off scheduled events, it will be the time at which the scheduled event 
was created and in case of cron triggers, the `created_at` field will be omitted.

Also, the `created_at` was stored in a column with type `timestamp without time zone`, which ignores the timezone/offset while writing into the db, which then causes problems while parsing it in the console
. So, changed `created_at` type to `timestamp with time zone`.

### Description

fix #5722 

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [x] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

There was no `created_at` column in the `hdb_catalog.hdb_cron_triggers` table, added it in this PR for sending the value of `created_at` in the webhook payload. The default of this column is `now()`, so the existing cron triggers's `created_at` will get set to `now()`.


server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -192,26 +192,26 @@ data ScheduledEventFull
   , sefRetryConf     :: !STRetryConf
   , sefHeaders       :: ![EventHeaderInfo]
   , sefComment       :: !(Maybe Text)
+  , sefCreatedAt     :: !UTCTime
```
tirumaraiselvan, line 195: Let's not send `createdAt` for cron events as it is an implementation detail and can be confusing to consumers.

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ This release contains the [PDV refactor (#4111)](https://github.com/hasura/graph
 - server: miscellaneous description changes (#4111)
 - server: treat the absence of `backend_only` configuration and `backend_only: false` equally (closing #5059) (#4111)
 - server: allow remote relationships joining `type` column with `[type]` input argument as spec allows this coercion (fixes #5133)
+- server: fix some `created_at` issues related to scheduled triggers, the value of the `created_at` in the case of an one-off scheduled event will be the creation time of the one-off event and in case of cron triggers, the `created_at` will be the creation time of the cron trigger.
```
tirumaraiselvan, line 57: we can just say, change created_at column to timestamptz (prev timestamp) (fix #blah)

CHANGELOG.md
```diff
@@ -54,7 +54,7 @@ This release contains the [PDV refactor (#4111)](https://github.com/hasura/graph
 - server: miscellaneous description changes (#4111)
 - server: treat the absence of `backend_only` configuration and `backend_only: false` equally (closing #5059) (#4111)
 - server: allow remote relationships joining `type` column with `[type]` input argument as spec allows this coercion (fixes #5133)
-- server: fix some `created_at` issues related to scheduled triggers, the value of the `created_at` in the case of an one-off scheduled event will be the creation time of the one-off event and in case of cron triggers, the `created_at` will be the creation time of the cron trigger.
+- server: change `created_at` column type from `timestamp` to `timestamptz` (fix #5722)
```
tirumaraiselvan, line 57: for scheduled trigger tables?

codingkarthik: Oh no, forgot to add it. I'll add it

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -488,7 +497,7 @@ processScheduledEvent
           headers = addDefaultHeaders $ map encodeHeader sefHeaders
           extraLogCtx = ExtraLogContext (Just currentTime) sefId
           webhookReqPayload =
-            ScheduledEventWebhookPayload sefId sefName sefScheduledTime sefPayload sefComment currentTime
+            ScheduledEventWebhookPayload sefId sefName sefScheduledTime sefPayload sefComment sefCreatedAt
```
tirumaraiselvan, line 500: I wonder if we should introduce a new payload field `invoked_at` for `currentTime`

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -192,26 +191,32 @@ data ScheduledEventFull
   , sefRetryConf     :: !STRetryConf
   , sefHeaders       :: ![EventHeaderInfo]
   , sefComment       :: !(Maybe Text)
+  , sefCreatedAt     :: !(Maybe UTCTime)
```
tirumaraiselvan, line 194: This is arguable so I leave it up to you. But it feels the internal representation of `ScheduledEventFull` can have `UTCTime` for both cron and one-off events.

We can avoid exposing it in the payload for cron events because it is not user-facing.

This would mean we change the type to `sefCreatedAt     :: !UTCTime` and add the comment for `sewpCreatedAt`.

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
codingkarthik: When I added `created_at` initially (before this PR), I was thinking about the invocation time, but clearly `created_at` is a bad name for it. I think there's value to add `invoked_at` in the payload.

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
codingkarthik: > But it feels the internal representation of ScheduledEventFull can have UTCTime for both cron and one-off events.

Yes, we can have, but what's the point of having the `sefCreatedAt` for cron events, if it's not going to be used anywhere. 

But, now that I think of it, maybe it'll be useful to include it in the logs.

codingkarthik: I have included the `created_at` field in `CronEventPartial` as well and not sending the value in the payload.

tirumaraiselvan (approved):

lexi-lambda (approved):

PR #5748, opened 2020-09-08
console: trim role names when submitting the form (close #5553)

### LINKED ISSUE #5553: Role names without white spaces
Hey folks!

Whenever I create a new Role under the Permissions tab, a new role. I cannot create a role name with white spaces. For example, `Sales Manager`
Interestingly, I made that type of role name before. To fix that I usually copy `Sales Manager` to an input field and then change words separately. For example, `Sales Manager` becomes `Sales Moderator`.

Thank you

abhi40308:
Earlier, when creating a role in permissions, we could not give whitespaces, as whitespaces were getting trimmed. This pull request fixes that issue.

### Description

fix #5553 

### Changelog (TODO)

### Affected components

- [x] Console

### Steps to test and verify

Go to data section, select a table and select permissions, create a role with trailing whitespaces. For example ` admin role  `.


console/src/components/Services/Data/TablePermissions/Actions.js
```diff
@@ -723,7 +723,7 @@ const permChangePermissions = changeType => {
     const limitEnabled = permissionsState.limitEnabled;
 
     const table = permissionsState.table;
-    const role = permissionsState.role;
+    const role = permissionsState.role.replace(/ /g, '');
```
beerose, line 726:
```suggestion
    const role = permissionsState.role.trim();
```

beerose (changes requested):

console/src/components/Services/Data/TablePermissions/Actions.js
beerose: Doing replace whitespaces won't solve the original issue.

beerose (approved):
Console

tirumaraiselvan (approved):
A changelog might have been good since it's a user reported issue so it's definitely user-facing ;)

But, it's also very trivial so it's okay.

PR #5755, opened 2020-09-09
Console: Enter the Role name in permission tab with white space

Varun-Choudhary:
Closes @ #5553 

### Description
Removed `trim()` from RoleName. So, that we can write Role name with white space in permission tab.

### Affected components

- [X] Console


Varun-Choudhary:
Already solved by some else.

PR #5759, opened 2020-09-09
server: accept only non-negative integers for batch size and refetch interval (close #5653)

### LINKED ISSUE #5653: server accepts negative values for refetch delay and batch size
both the options to tune subscription worker accepts negative values

kolharsam:
closes #5653 

### Description

This PR fixes the bug that allowed users to pass negative values for batch size and the refetch interval. This PR also adds a spec file for the `NonNegativeInt` type that is introduced in this PR.

### Changelog

### Affected components

- [x] Server

#### Catalog upgrade
- [x] No

#### Metadata
- [x] No

#### GraphQL

#### Breaking changes
- [x] No Breaking changes


server/src-test/Data/NonNegativeIntSpec.hs
```diff
@@ -0,0 +1,24 @@
...
+nonNegIntSpec :: Spec
+nonNegIntSpec =
+    describe "non negative integer type" $ do
+        it "only validates non negative integers" $ do
+            NonNegativeInt 23 `shouldBe` 23
+            NonNegativeInt (-23) `shouldNotBe` (-23)
+
+        it "JSON accepts only non negative integers" $ do
+            toJSON (23 :: NonNegativeInt) `shouldBe` Number 23
+            toJSON (-23 :: NonNegativeInt) `shouldNotBe` Number (-23)
```
kolharsam, line 22: - I can think of why L18 should be failing, there's nothing stopping me from setting that value.
 - I don't understand full well why L22 should be failing.

Naveenaidu: IIUC L22 fails because you don't have a `ToJSON` instance which will reject from creating a JSON object if the value is negative. 

You have only created the `FromJSON` instance and that's why the `decode` tests below work.

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -277,12 +278,26 @@ isSystemDefined = unSystemDefined
 successMsg :: EncJSON
 successMsg = "{\"message\":\"success\"}"
 
+newtype NonNegativeInt = NonNegativeInt { getNonNegativeInt :: Int }
+  deriving (Show, Eq, ToJSON, Generic, NFData, Cacheable, Num, Read)
+
+maybeInt :: MonadFail m => Maybe Int -> m Int
+maybeInt x = case x of
+  Just v  -> return v
+  Nothing -> fail "integer passed is out of bounds"
```
codingkarthik, line 287: Can move this function to the `FromJSON` instance of `NonNegativeInt`, since this function is used only once.

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -277,12 +278,26 @@ isSystemDefined = unSystemDefined
 successMsg :: EncJSON
 successMsg = "{\"message\":\"success\"}"
 
+newtype NonNegativeInt = NonNegativeInt { getNonNegativeInt :: Int }
```
codingkarthik, line 281: There should be a function `mkNonNegativeInt` which takes an `Int` and then returns a `Maybe Int`. The function should return `Nothing` when a negative value is passed otherwise it should return `Just x`. Check out `mkNonEmptyText` for reference.

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -41,6 +41,7 @@ module Hasura.RQL.Types.Common
        , InputWebhook(..)
        , ResolvedWebhook(..)
        , resolveWebhook
+       , NonNegativeInt(..)
```
codingkarthik, line 44:
```suggestion
       , NonNegativeInt
```
Exposing the constructor of `NonNegativeInt` may not be a very good idea. Instead, you can expose the `mkNonNegativeInt` function that I have suggested.

server/src-lib/Hasura/RQL/Types/Common.hs
Naveenaidu: Just a small question, `mkNonNegativeInt` should return `Maybe NonNegativeInt` right? Instead of `Maybe Int`?

server/src-lib/Hasura/Server/Init/Config.hs
```diff
@@ -260,7 +261,7 @@ instance FromEnv LQ.BatchSize where
   fromEnv = fmap LQ.BatchSize . readEither
 
 instance FromEnv LQ.RefetchInterval where
-  fromEnv = fmap (LQ.RefetchInterval . milliseconds . fromInteger) . readEither
+  fromEnv = fmap (LQ.RefetchInterval . NonNegativeDiffTime . milliseconds . fromInteger) . readEither
```
codingkarthik, line 264: Instead of using the `NonNegativeDiffTime` constructor here, use the `mkNonNegativeDiffTime`. Currently, negative values will also be accepted.

server/src-test/Data/NonNegativeIntSpec.hs
codingkarthik: Line 18 and 22 fail because of the `NonNegativeInt` constructor. The `NonNegativeInt` is just a type, we need to control what value we construct, hence the check in the `FromJSON` instance i.e. if the number is negative we throw an error before constructing the `NonNegativeInt` value.

server/src-lib/Hasura/RQL/Types/Common.hs
codingkarthik: Ah sorry, Yes, you're right, it should return `Maybe NonNegativeInt`

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Options.hs
```diff
@@ -32,10 +35,20 @@ instance J.FromJSON LiveQueriesOptions where
...
                        <*> o J..: "refetch_delay"
 
-newtype BatchSize = BatchSize { unBatchSize :: Int }
+newtype BatchSize = BatchSize { unBatchSize :: NonNegativeInt }
   deriving (Show, Eq, J.ToJSON, J.FromJSON)
 
+mkBatchSize :: Int -> Maybe BatchSize
+mkBatchSize x = do
+  t <- mkNonNegativeInt x
+  return $ BatchSize t
```
codingkarthik, line 44:
```suggestion
mkBatchSize :: Int -> Maybe BatchSize
mkBatchSize =  BatchSize <$> mkNonNegativeInt
```

server/src-test/Data/NonNegativeIntSpec.hs
```diff
@@ -0,0 +1,23 @@
...
+spec = do
+  nonNegIntSpec
+
+nonNegIntSpec :: Spec
+nonNegIntSpec =
+    describe "non negative integer type" $ do
+        it "only validates non negative integers" $ do
+            (evalMaybeNonNegativeInt $ mkNonNegativeInt 23) `shouldBe` True
+            (evalMaybeNonNegativeInt $ mkNonNegativeInt (-23)) `shouldBe` False
+
```
codingkarthik, line 23: You can directly compare the value instead of converting it into a `Bool` and I think it will be more useful that way.
You can do `mkNonNegativeInt shouldBe (Just 23)`

server/src-lib/Hasura/GraphQL/Execute/LiveQuery/Options.hs
```diff
@@ -3,11 +3,14 @@ module Hasura.GraphQL.Execute.LiveQuery.Options
   , BatchSize(..)
```
codingkarthik, line 3: We should not be exposing the constructor part, when we have `mkBatchSize`

codingkarthik (approved):
LGTM

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@ This release contains the [PDV refactor (#4111)](https://github.com/hasura/graph
 - server: add action-like URL templating for event triggers and remote schemas (fixes #2483)
 - server: change `created_at` column type from `timestamp` to `timestamptz` for scheduled triggers tables (fix #5722)
 - server: allow configuring timeouts for actions (fixes #4966)
+- server: accept only non-negative integers for batch size and refetch interval (close #5653) (#5759)
```
tirumaraiselvan, line 60: I don't think we need a changelog atleast in the current form as "batch size" and "refetch interval" are pretty vague.

tirumaraiselvan (approved):
There is an optional changelog comment. but not important.

Issue #5297, opened 2020-07-04
Migration directory names documentation wrong


dan-lee:
> Docker engine: 19.03.12
> Docker image: hasura/graphql-engine:v1.3.0-beta.4.cli-migrations-v2

The [documentation for migrations suggests to use](https://hasura.io/docs/1.0/graphql/manual/hasura-cli/config-reference.html#environment-variables):

```
HASURA_GRAPHQL_METADATA_DIRECTORY: /path/to/metadata
HASURA_GRAPHQL_MIGRATIONS_DIRECTORY: /path/to/migrations
```

But when executing it with these env vars I get:

> graphql-engine_1  | time="2020-07-04T20:26:24Z" level=fatal msg="failed to apply metadata: cannot apply metadata on the database: [parse-failed] key \"tables\" not found ($[1])"

After a few hours of digging, I found out that the reason is that the env var names are actually prefixed with `_DIR` instead of `_DIRECTORY`:

https://github.com/hasura/graphql-engine/blob/1e5e7d93dd29a5323918090bff567f3b0afc1417/scripts/cli-migrations/v2/docker-entrypoint.sh#L71-L73

https://github.com/hasura/graphql-engine/blob/1e5e7d93dd29a5323918090bff567f3b0afc1417/scripts/cli-migrations/v2/docker-entrypoint.sh#L90-L99

Now my docker container boots up without a problem.
Just wanted to report this issue, I don't know if the problem is just outdated documentation or something else.

scriptonist:
@dan-lee The documentation for `cli-migrations-v2` image can be found here: https://hasura.io/docs/1.0/graphql/manual/migrations/advanced/auto-apply-migrations.html#applying-migrations, which mentions the Environment variables to use. 

But I understand this has created some confusion. Will be exploring how to improve it.

dan-lee:
@scriptonist Aah I see, thanks for the link. Didn't see these docs right away.

My suggestion would be to maybe either sync these environment variables or have a more descriptive error message. The former, of course, could have some side effects I am not aware of.

PR #5767, opened 2020-09-10
Fix migrations & metadata env vars in config reference

Shark:
### Description

In the [CLI config reference](https://hasura.io/docs/1.0/graphql/core/hasura-cli/config-reference.html#cli-config-reference), the environment variables for `metadata_directory` and `migrations_directory` are wrong.

Instead of `HASURA_GRAPHQL_*_DIRECTORY`, they should be `HASURA_GRAPHQL_*_DIR` as per the logs in version 1.3.2:

```
{"timestamp":"2020-09-10T16:06:40.000+0000","level":"info","type":"startup","detail":{"kind":"migrations-startup","info":"env var HASURA_GRAPHQL_MIGRATIONS_DIR is not set, defaulting to /hasura-migrations"}}
{"timestamp":"2020-09-10T16:06:40.000+0000","level":"info","type":"startup","detail":{"kind":"migrations-startup","info":"env var HASURA_GRAPHQL_METADATA_DIR is not set, defaulting to /hasura-metadata"}}
```

### Changelog

- [x] ~~`CHANGELOG.md` is updated with user-facing content relevant to this PR.~~ If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [x] Docs

### Related Issues

None

### Solution and Design

Just a simple fix.

### Steps to test and verify

Just a simple fix.

Just a simple fix.

### Server checklist

No action needed.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/docs/graphql/core/hasura-cli/config-reference.rst
+++ b/docs/graphql/core/hasura-cli/config-reference.rst
@@ -210,11 +210,11 @@ The configuration can also be set in the form of environment variables:
      - ``api_paths.version``
      - Version API endpoint. More details at :ref:`version_api`.
 
-   * - ``HASURA_GRAPHQL_METADATA_DIRECTORY``
+   * - ``HASURA_GRAPHQL_METADATA_DIR``
      - ``metadata_directory``
      - Defines the directory where the metadata files were stored.
 
-   * - ``HASURA_GRAPHQL_MIGRATIONS_DIRECTORY``
+   * - ``HASURA_GRAPHQL_MIGRATIONS_DIR``
      - ``migrations_directory``
      - Defines the directory where the migration files were stored.
 


scriptonist:
@Shark Thanks for contribution. Just wanted to clarify something, are you using the [hasura/graphql-engine:<version>.cli-migrations-v2](https://hasura.io/docs/1.0/graphql/core/migrations/advanced/auto-apply-migrations.html) image ?

Shark:
@scriptonist yes, I used `hasura/graphql-engine:v1.3.2.cli-migrations-v2`

scriptonist:
@Shark in that case, I think you are referring to a couple of env vars specific to cli-migrations image (https://hasura.io/docs/1.0/graphql/core/migrations/advanced/auto-apply-migrations.html#applying-migrations)

cllns:
Same issue as https://github.com/hasura/graphql-engine/issues/5297

This issue tripped me up too. Why are there separate ENV variables with such similar names? If they need to be different for some reason, a conspicuous note that points out the difference between the two would help others avoid this issue :)

scriptonist:
@cllns Understood, that's a very fair point to make. we'll look into it and see how this can be improved. Also, you might not need to set `HASURA_GRAPHQL_METADATA_DIRECTORY` and `HASURA_GRAPHQL_MIGRATIONS_DIRECTORY` explicitly, since they are the default values.

PR #5785, opened 2020-09-15
console: trim migration names exceeding 255 characters (close #4403)

### LINKED ISSUE #4403: Hasura console and migrations mishandles long file names; freezes hasura console UI feedback on long multi-column constraints submit
May be more specifically related to the automatic migration generation functionality.

Long multi-name composite column constraints generate huge migration names based on concatenating the name of every column included without character limit.

Although some functionality does truncate the name to a usable string (the constraint name itself appears to be truncated after I refresh or check the code), the migrations generation of exceedingly and unlimited long file names freeze both the Hasura console feedback (Message: creating constraint... never resolves in the web console) and results in errors blocking the hasura migrate CLI from performing `migrate status` `migrate apply` functionality as well as causing issues with `git` version control and pre-commit/linting hooks mishandling long paths (not using `git` `longpaths` flag due to keep consistent tool interoperability compatibility).
* The default version control support not handling long file names can result in a hard to track down and troubleshoot error due to e.g. `git` not tracking the newly created migration files (hidden files causing errors in the background and not in version history!).

# Reproduce:
* Set up hasura migrations https://hasura.io/docs/1.0/graphql/manual/migrations/index.html
* Create a table with 20 (or any combination of columns + long column names to generate a long concatenated string for composite constraint migrations automatically)
* Create a composite constraint (include those 20+ columns in a `UNIQUE` constraint) that results in a concatenated constraint and subsequent even longer unlimited concatenated migration version folder
* Hasura console pops up the "creating..." message on submit that does not resolve and leaves the submission form open for editing (and provides UX case to edit submit more conflicting constraints multiple times without feedback).
* Check version control to see(!) your migrations were not tracked due to `longpaths` 😨 😃 

# Workaround: 
Deleting large appended strings from each fun migration version folder name works without issue, subsequently allowed the hasura CLI and `git` to continue working (hasura migrate either identifies based on the first prefix version epoch timestamp or has cached the data in some place).

May have been conducting a large amount of weird and wonderful denormalised data scraping 😅 

In having verbose naming conventions within obtuse denormalised data scraping for a database migration :P,
Example breakdown : 

- 33 characters default prefix: `1586847853419_alter_table_public_`
  - including 13 characters unix epoch: `1586847853419`
  - 10 to 20 characters action text: `add_unique_` `alter_table_`, `add_relationship_`,etc
* 15 to 30 characters table name depending on convention, standards: `ProductRevision` (or 63 characters)
* 400 characters... 20 to 30 characters column name multiplied by ... twenty columns (Postgres default maximum 63 characters identifier column, table name; Generally 10-30 characters) 
* ~20 characters all the underscores in between the columns, i.e. 1 per column: `_`

Example:
30 characters repo directory path + 33 + 30 + (... 20 * n) + (n)
Easily 100 to 150 characters by default (including base repo directory) even without long concatenated column names.

Minimal 60 to 100 characters for the generated migration version folder when not taking into account any parent directories or any column names in composite:
* `1586813845928_add_relationship_CustomerProductRevision_table_public_undefined`
* `1586412310362_add_relationship__table_public_ProductRevision`

^(not sure what inconsistencies are present here e.g. `__` vs `_`

>  By default, NAMEDATALEN is 64 so the maximum identifier length is 63 bytes.  The system uses no more than NAMEDATALEN-1 bytes of an identifier; longer names can be written in commands, but they will be truncated. If this limit is problematic, it can be raised by changing the NAMEDATALEN constant in src/include/pg_config_manual.h. 
https://www.postgresql.org/docs/12/sql-syntax-lexical.html 
`max_identifier_length` https://www.postgresql.org/docs/12/runtime-config-preset.html

Default 255 total file path (code repository may not be a flattened monorepo) 
😆 Wondering how to make both informative and super compatible auto migration names (whether the scaffolded folder generation is the preferred standard for developer-customisability)

Naming things https://martinfowler.com/bliki/TwoHardThings.html

abhi40308:
Migration name longer than 255 characters lead to unpredictable behaviour due to filename character limits.

### Description

fix #4403

### Changelog

### Affected components

- [ ] Server
- [x] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Steps to test and verify

Open hasura console, create a table with long name and some columns with long column names, then try creating a constraint (say unique key) on a combination of columns.


soorajshankar (changes requested):
The console is still failing with the following steps.
1. Go to console > Create Table 
2. Try creating a table with name "test_test_test_test_test_test_test_test_test_test_test_test_test_123" (length greater than 63 chars)
3. console is getting crashed

Note: Adding a front-end char length validation might resolve this.

Also noticed the migration is getting created with the full table name, while pg is still truncating the name to 63 chars

--
>The system uses no more than NAMEDATALEN-1 bytes of an identifier; longer names can be written in commands, but they will be truncated. By default, NAMEDATALEN is 64 so the maximum identifier length is 63 bytes. If this limit is problematic, it can be raised by changing the NAMEDATALEN constant in src/include/pg_config_manual.h.

source : https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

soorajshankar (changes requested):
@abhi40308 I was testing the functionality, I found that the UI is trimming the migration length and sending the correct value to both migrations and the PG.

But when the user performs this action, the user is nowhere notified that the name is getting truncated. and the text input is having the full length but the moment when we submit, it gets truncated without letting the user know.

What I was expecting is to show a front end field validation, so that the user could only type characters with a maximum limit (text validation) 

It would be great if we can make inline validation or we can continue showing the notification (like if you put a space on the table name, it would show an error notification when you try to submit)

soorajshankar (approved):

abhi40308:
Closing in favour of https://github.com/hasura/graphql-engine-mono/pull/1360

Issue #3757, opened 2020-01-22
support custom SQL functions having composite type (rowtype) input arguments


rakeshkky:
via https://github.com/hasura/graphql-engine/issues/3232#issuecomment-576796667

Tracking an SQL function which returns `SETOF <table>` and type of any one of input arguments is a table row (already tracked) gives the exception described in above referred comment. User creates such functions for [computed fields](https://docs.hasura.io/1.0/graphql/manual/schema/computed-fields.html#computed-fields) and tracks them as [custom SQL functions](https://docs.hasura.io/1.0/graphql/manual/schema/custom-functions.html#).

Solution:- 

1. Allow table row type as scalars for function arguments in "args" input field. The name of such scalar types will conflict with the names of object types generated for the table. The possible solution I think as of now is to append `_input` for scalars generated for composite types. Input value for those row type scalars can be constructed as String literals (See [Postgres docs](https://www.postgresql.org/docs/12/rowtypes.html#id-1.5.7.24.6)).
~2. The "args" input for computed field should be named as `<table>_<computed-field-name>_input_args` and not `<function-name>_input_args`~ We can't change the generated type name as it breaks the schema. So, we can support solution `1` by restricting tracking an SQL function as either a custom SQL function (in queries) or a computed field of a table but not both.

- [ ] Update docs with appropriate example

nicuveo:
This also happens for SQL functions that return a scalar, FWIW.

charmpitz:
Any ideas when will this be solved? I'm stuck and it's really disappointing..

beerose:
Related issue: https://github.com/hasura/graphql-engine/issues/5514

When a function with an argument being a table row is tracked, then introspection query is successful. However, using its result with `graphql-js` library to [build a schema definition](https://github.com/graphql/graphql-js/blob/7b3241329e1ff49fb647b043b80568f0cf9e1a7c/src/utilities/buildClientSchema.js#L58) for graphiql explorer results in an error. 

Example function:
```
CREATE FUNCTION sample_function(arg text_result) RETURNS SETOF text_result LANGUAGE sql STABLE AS $ function $
SELECT
  q.*
FROM
  (...) q $ function $
```

Error from `graphql-js`:

<img width="528" alt="Screenshot 2020-08-19 at 13 31 26" src="https://user-images.githubusercontent.com/9019397/90629825-4a346500-e220-11ea-8a43-ffcac79093f6.png">

altitudems:
[...]

PR #5787, opened 2020-09-15
rfc: improvements to GraphQL API over PG functions

0x777:
[Rendered](https://github.com/0x777/graphql-engine/blob/pg-function-improvements/rfcs/pg-function-improvements.md)

### DIFF
--- a/rfcs/pg-function-improvements.md
+++ b/rfcs/pg-function-improvements.md
@@ -20,12 +20,13 @@ past the `stable`/`immutable` requirement.
 
 ### Proposal
 
-1. Every function is allowed to be exposed through GraphQL API. If a function
-   is marked as `immutable`/`stable`, it is exposed under `query_root`
+1. A function is allowed to be tracked if any of the following holds true:
+   - Returns a scalar
+   - Returns a table type or setof table type that has already been tracked.
+
+1. If a function is marked as `immutable`/`stable`, it is exposed under `query_root`
    otherwise under `mutation_root`.
 
-2. A function `f` is only exposed for a role `r` in the GraphQL API if a
-   permission is defined on `f` for the role `r`.
 
 ### Permissions
 
@@ -33,12 +34,9 @@ past the `stable`/`immutable` requirement.
 
 Either a static value or something that can be filled from a session variable.
 
-#### Permissions on the return types
-
-1. If the return type is a scalar, no further configuration is allowed.
-1. If the return type is a table type or setof <table> type that is already
-   tracked, we can maybe allow overriding the filter? Note that the allowed columns
-   cannot be changed without changing the name of the type.
+If a function returns a static type, the function is exposed without further
+checks.  If the function returns a table or set of table type then there have to be
+select permissions defined on the table for the role.
 
 ### Backwards compatibility
 
--- a/rfcs/pg-function-improvements.md
+++ b/rfcs/pg-function-improvements.md
@@ -1,0 +1,47 @@
+## Improvements to GraphQL API over SQL functions
+
+Currently we only support stable/immutable functions which return a list of rows.
+The rationale behind this decision was to avoid building a permission system
+for functions. The permissions are implicitly inferred as follows:
+
+> A function is allowed for a role only if the table it returns is allowed.
+
+This assumption is safe for stable/immutable functions as
+1. The functions cannot modify the database.
+2. The user will only see the data that is allowed through configured permissions
+   on the return tables.
+
+However the above system is insufficient to expose functions which are not
+`stable` or `immutable` - when the functions mutate the database, they cannot
+be exposed even if there is a select permission on the return type. This
+necessitates a permission system that is designed for functions to help us move
+past the `stable`/`immutable` requirement.
+
+
+### Proposal
+
+1. Every function is allowed to be exposed through GraphQL API. If a function
+   is marked as `immutable`/`stable`, it is exposed under `query_root`
+   otherwise under `mutation_root`.
+
+2. A function `f` is only exposed for a role `r` in the GraphQL API if a
+   permission is defined on `f` for the role `r`.
+
+### Permissions
+
+#### Argument presets
+
+Either a static value or something that can be filled from a session variable.
+
+#### Permissions on the return types
+
+1. If the return type is a scalar, no further configuration is allowed.
+1. If the return type is a table type or setof <table> type that is already
+   tracked, we can maybe allow overriding the filter? Note that the allowed columns
+   cannot be changed without changing the name of the type.
+
+### Backwards compatibility
+
+Since we currently automatically expose functions, we need a command line
+flag such as `ENABLE_FUNCTION_PERMISSIONS` (which defaults to false) to
+enable permissions for maintaining backwards compatibility.


tirumaraiselvan:
I think we should also allow for a `session_argument` which populates the entire session_argument in a json argument as currently available for stable functions. This is just for consistency with what we already have and giving more developer options without any loss.

rfcs/pg-function-improvements.md
```diff
@@ -0,0 +1,45 @@
+## Improvements to GraphQL API over SQL functions
+
+Currently we only support stable/immutable functions which return a list of rows.
+The rationale behind this decision was to avoid building a permission system
+for functions. The permissions are implicitly inferred as follows:
+
+> A function is allowed for a role only if the table it returns is allowed.
```
jberryman, line 7: I've just finished getting acquainted with the details of postgres functions, as well as hasura's docs on computed fields and function tracking; I haven't gotten into the code just yet, so this may be a dumb question...

" if the table it returns ... "

I think what you mean precisely here is: "if it returns a composite type implicitly created during `CREATE TABLE` on table..."

Not trying to be pedantic, this was actually a little confusing to me

rfcs/pg-function-improvements.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Permissions
+
+#### Argument presets
+
+Either a static value or something that can be filled from a session variable.
+
+If a function returns a static type, the function is exposed without further
+checks.  If the function returns a table or set of table type then there have to be
+select permissions defined on the table for the role.
```
jberryman, line 39: This seems pretty intuitive to me

jberryman:
I'm just starting actually playing more with functions and digging into the code, but one thought that came up for me while reading the docs:

**It's confusing that `track_function` and `add_computed_field` are separate API calls and features**

My tentative proposal is that `add_computed_field` should subsume all of this functionality (`track_function` also "adds a computed field", it's just at the top level") and work like this:

- the target function takes...
  - _at most 1 table-corresponding composite arguments_
    - if present then things work like `add_computed_field` today: the argument becomes a sub-field of the input type
    - otherwise this is a top-level field
  - zero or more additional base type arguments; these will be passed through into the resulting schema, or mapped from session vars
- it can return a scalar or `SETOF`/`TABLE` (as it currently works)
- If a function is marked as immutable/stable, it is exposed under query_root otherwise under mutation_root (as in the RFC here)

I'm also wondering why `computed_fields` create a single `args` field argument rather than just having them be individual regular arguments to the field... maybe that will become clear when I dig in.

rfcs/pg-function-improvements.md
jberryman: I think there's also an interesting question here:

should it be that we attach permissions, not to tables, _but to types_?

If so then we could allow, e.g. someone to `CREATE TYPE` and then define permissions on it and then create and track a function that returns that type (maybe this is already how things work and I just forget...)

jberryman:
@0x777 @tirumaraiselvan thoughs on above? ^

tirumaraiselvan:
> at most 1 table-corresponding composite arguments
>
> - if present then things work like add_computed_field today: the argument becomes a sub-field of the input type
otherwise this is a top-level field

Although, I like the idea of having one API for user-defined functions, I wonder if these constraints will always hold true: can't an table computed field ever have more than one composite argument. Similarly, can't a top-level computed field have composite arguments? We actually have people adding the same function both at top-level and inside a table which leads to a runtime bug currently (as of v1.3, maybe master too): https://github.com/hasura/graphql-engine/issues/3757 . It's not clear if they are doing it unintentionally or if they expect a certain behavior for their use-case (see: https://github.com/hasura/graphql-engine/issues/3757#issuecomment-692884522). Hence, it may be required to have an API which explicitly takes the "location" : query_root or table when adding functions to schema. 

Secondly, function permissions (as defined in the RFC) only apply to top-level fields? It's not clear if table computed fields will require a similar permission model; for one, they are only used in reads.

Unless these two issues have clear answers, I am inclined to doing the API unification outside the scope of the permissions work.

> I'm also wondering why computed_fields create a single args field argument rather than just having them be individual regular arguments to the field.

For a computed field which returns a `SETOF table`, there are default arguments like `where`, `limit`, `offset` etc also present. These are not part of a scalar computed field though but maybe just to have consistent arguments?

rfcs/pg-function-improvements.md
tirumaraiselvan: Yeah, I think the point here is that it should continue to work like that even for mutation functions.

rakeshkky:
@jberryman 
The notion of `computed field` is the field which is computed from existing table columns. My initially API for adding computed field for a table uses the same `track_function` which automatically detects whether it is computed field or a top level query field. Upon @0x777 suggestion, I went with new API for better discoverability and being explicit with functionality of an SQL function in the final schema. 
> These are not part of a scalar computed field though but maybe just to have consistent arguments?

Yes.

rfcs/pg-function-improvements.md
0x777: > Not trying to be pedantic, this was actually a little confusing to me. I think I could improve our docs a little here

:+1: 

> should it be that we attach permissions, not to tables, but to types?

This is something that has been on my mind for quite sometime now. It definitely makes sense to attach 'select' permissions to Postgres's composite types but we'll need to figure out some things:
1. `limit` in select permissions? Does it make sense to attach it to a type? Or should we introduce a notion of 'resolvers' and `limit` is attached to a resolver?
1. What about other permissions: insert/update/delete? Do the features that we have currently extend well for types?
1. What is the user experience gonna look like? i.e, currently the building blocks for user's GraphQL schema are mostly tracking tables, and defining relationships, both of which are built around the notion of 'tables/views'.

tirumaraiselvan: FWIW, I think we should see if there is an incremental path from table types (which is what our users have asked for mainly) to arbitrary postgres composite types for functions and scope this only to table types if possible.

jberryman:
> Although, I like the idea of having one API for user-defined functions, I wonder if these constraints will always hold true: can't an inner computed field ever have more than one composite argument. 

Currently we don't make an effort to support a good DX for working with _already written_ functions (e.g. by defining a mapping between function arguments and session variables); instead, we expect the user to be authoring functions for hasura (e.g. by taking a separate session variable json argument, etc). 

I think this is a good choice, and would allow us to insist that: if a function takes a composite type argument which was generated from a table `T`, the function becomes a subfield of `T`.

> Similarly, can't a top-level computed field have composite arguments?

I don't see what value it would bring to support composite type inputs to a function that would be passed through to the resulting schema. This seems to just make the feature (and resulting graphql schema) more complicated; instead, just tell users to write their function differently (with separate scalar arguments).

> Secondly, function permissions (as defined in the RFC) only apply to top-level fields? It's not clear if inner computed fields will require a similar permission model; for one, they are only used in reads.

My understanding is that computed fields [already work the way the RFC describes](https://hasura.io/docs/1.0/graphql/core/schema/computed-fields.html#computed-fields-permissions), but I haven't actually played with them yet. I'll write up some more tests to check this out... But that was my reasoning in proposing that `computed_fields` should be extended slightly to work for the top level.

Actually, I think it might help me if someone could clarify the RFC with regard to permissions. From the text:

> If a function returns a static type, the function is exposed without further checks. If the function returns a table or set of table type then there have to be select permissions defined on the table for the role.

Isn't this how `track_function` already works? Is this supposed to read "there have to be insert/update/delete permissions defined on the table for this role"?

@rakeshkky :

> My initially API for adding computed field for a table uses the same track_function which automatically detects whether it is computed field or a top level query field. Upon @0x777 suggestion, I went with new API for better discoverability and being explicit with functionality of an SQL function in the final schema.

That makes sense. Do you feel like that ended up being a good choice? As someone just coming to the docs, the fact that we have two different APIs is confusing to me personally: in both cases the user wants to add a new computed field to the schema.

>> For a computed field which returns a SETOF table, there are default arguments like where, limit, offset etc also present. These are not part of a scalar computed field though but maybe just to have consistent arguments?
> 
> Yes

Sorry, I'm still not getting it... is it an issue of namespacing? Like we don't know what to do if the user defines a function like:

    CREATE FUNCTION foo(baz baz_table, where text) ...  // oops "where" would clash!

...? 

Put another way: isn't the resulting graphql schema here pretty weird?: 

```
query {
  author {
   last_name
    # the user must wonder: why not just ` filtered_articles(search: "Hasura")` ...?
    filtered_articles(args: {search: "Hasura"}){
      title
      content
    }
  }
}
```

If this is a name clashing issue then I think one solution is, again, to simply disallow the user to add a function with name clashes as a computed field; they'd need to change the name. Making this change at this point might or might not be possible or worth it. In particular I'm not sure how we'd have forwards compatibility when adding new hasura fields...

But maybe doing the above would require a huge refactor and that's why a separate "args" field is used? I hope my question makes sense; just trying to understand the "why".

rakeshkky:
@jberryman One reason I could think of for the confusion among users regarding which API to use to create a computed field is, the console support for computed field is added much later and docs are incomplete since we can't embed the console UI screenshots for the same.

If a computed field returns `SETOF table`, we generate `where`, `limit`, `offset` input arguments just like for table select queries. In order to isolate them from SQL function input arguments we put the later in `args` field.

tirumaraiselvan:
> Currently we don't make an effort to support a good DX for working with already written functions (e.g. by defining a mapping between function arguments and session variables); instead, we expect the user to be authoring functions for hasura (e.g. by taking a separate session variable json argument, etc).

This RFC does propose individual mapping of each session variable to function arguments no? Then, `hasura_session json` would just be a "nice to have" argument (and also for backward compatibility).

> My understanding is that computed fields already work the way the RFC describes.

Yes, but my concern with this unification was about writes. What should be the behaviour in the unified API if a VOLATILE function with a table type argument is tracked? We do not have a model for computed fields in `insert_table` so would it throw an error? In other words, I was thinking computed fields do not even exist in the context of mutations and hence we should consider the permission RFC here only wrt top-level functions.

jberryman:
> This RFC does propose individual mapping of each session variable to function arguments no?

Right, I guess I misinterpreted that... so why don't we want to just do the json `session_argument` thing here? That seems like more complicated for users and a more complicated implementation.

This is in contrast to the proposed schema mapping for remote schemas where we have to assume the graphql-engine admin _has no control over the remote schema_. Does that make sense?

> I was thinking computed fields do not even exist in the context of mutations and hence we should consider the permission RFC here only wrt top-level functions.

Right, it only makes sense to map `VOLATILE` functions to top-level fields of the mutations root. In the context of the unified functions API I'm tentatively proposing, we'd raise a helpful error if a user tried to track a volatile function with a table/composite-type argument. Computed fields make perfect sense in the context of the _return_ type of mutations though, and presumably this already works right?

jberryman:
Sorry, from your older comment:

> We actually have people adding the same function both at top-level and inside a table which leads to a runtime bug currently (as of v1.3, maybe master too): #3757 . It's not clear if they are doing it unintentionally or if they expect a certain behavior for their use-case (see: #3757 (comment)). Hence, it may be required to have an API which explicitly takes the "location" : query_root or table when adding functions to schema.

This comment does seem to indicate they misunderstood the APIs and tracked function for no reason, which is understandable to me: https://github.com/hasura/graphql-engine/issues/3232#issuecomment-546387034

I think it supports the case for having a unified API. I think the two options would then be to:

- omitting `table` implies top-level field, or...
- if `table` is omitted insist on a `top-level: true` parameter or something

But now I wonder what happens when you `track_function` on a function with a composite type argument...

tirumaraiselvan:
> > This RFC does propose individual mapping of each session variable to function arguments no?
> 
> Right, I guess I misinterpreted that... so why don't we want to just do the json `session_argument` thing here? Mapping individual session variables to arguments seems just more complicated for users and a more complicated implementation.
> 

Apart from making the permission more declarative, it has the advantage of what you had said earlier: `Currently we don't make an effort to support a good DX for working with already written functions`. This makes the DX of already written functions and also to-be written functions much better because you don't have to litter your function body with local variables and expressions which extract individual session variables from the json blob (it also reduces the need to check for null values, etc). This makes me think the interface is actually _less_ complicated for function writers and the UI to specify this can be made very intuitive. We could eventually deprecate the `hasura_session json` style input argument. 

> This is in contrast to the proposed schema mapping for remote schemas where we have to assume the graphql-engine admin _has no control over the remote schema_. Does that make sense?
>
I am afraid I didn't understand this. Although, we should consider the case where we have no control over an already written custom function.
 
> > I was thinking computed fields do not even exist in the context of mutations and hence we should consider the permission RFC here only wrt top-level functions.
> 
> Right, it only makes sense to map `VOLATILE` functions to top-level fields of the mutations root. In the context of the unified functions API I'm tentatively proposing, we'd raise a helpful error if a user tried to track a volatile function with a table/composite-type argument. Computed (non-`VOLATILE`) fields make perfect sense in the context of the _return_ type of mutations though, and presumably this already works right?

Sure, this makes sense. I think this answers the second question I had originally.

> I think it supports the case for having a unified API.

I agree. Having a unified API will make the console also treat them in a unified manner and it's all win-win.

> Seems like one possible API is to take one of either input_table or e.g. top_level: true (or something)
>
> But now I wonder what happens when you track_function on a function with a composite type argument...

This is the first question I had. I am happy if we can come with a decent API here. I am feeling inferring the "location" of the computed field based on input arguments may cause random confusion in the future or constrain it to be inflexible (theoretically as of now) so it doesn't harm to be explicit here. We can have a radio button on the UI to mark if it's top-level or inside table.

jberryman:
>>This is in contrast to the proposed schema mapping for remote schemas where we have to assume the graphql-engine admin has no control over the remote schema. Does that make sense?
>
> I am afraid I didn't understand this. Although, we should consider the case where we have no control over an already written custom function.

Sorry, I was trying to say that in this proposal https://github.com/hasura/graphql-engine/pull/5675 having functionality to map arguments is a necessity, since we can't assume the hasura admin can make changes to the remote schema to accomadate a particular way of passing session variables.

In contrast with SQL functions this is trivial: you can write a small wrapper postgres function that just passes fields of the `session_argument` as arguments to the already-written function you have. It's easier IMO to do that (or learn to do that) than to read and understand a new yaml configuration spec. Consider also that the user would frequently need to do some casts, or other trivial massaging of the arguments anyway.

So that's why this doesn't seem terribly worthwhile to me.

tirumaraiselvan:
@jberryman Could you add the permission API as you are thinking as a comment here or in the RFC? This will help in understanding the permission API post-unification as proposed in https://github.com/hasura/graphql-engine-internal/issues/502

0x777:
Permissions part is superseded by https://github.com/hasura/graphql-engine/blob/master/rfcs/function-permissions.md.

PR #5794, opened 2020-09-16
[WIP] Heterogeneous execution

abooij:
### Description

The server could already execute queries that either fetch data from the database, or, through remotes, from other GraphQL servers.  This PR also enables mixing such data sources within one query.  So one can have queries such as

```
query {
  articles {
    title
  }
  weather {
    temperature
  }
}
```

where the articles are fetched from the database, and the weather is fetched from a remote server.

### Changelog

### Affected components

- [x] Server
- [x] Tests

### Solution and Design

For an incoming query we already used to generate an "execution plan" which specified how to execute that query. However, so far an execution plan has been a single execution step, where an execution step can be of database-type, of remote-type or of "raw"-type (used for introspection). This PR changes the type of execution plans to be an insertion-ordered hashmap of execution steps

```haskell
type ExecutionPlan db remote raw = InsOrdHashMap Text (ExecutionStep db remote raw)
```

where the key corresponds to a root field name.

### Steps to test and verify

A test suite is included.

Although this PR allows to mix root fields originating from different data sources, it does not allow joining between them arbitrarily beyond the existing "remote joins" that load relationships from Postgres to a remote. Such "generalized joins" are part of future work.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


abooij:
Closing in favor of #5869.

Issue #5376, opened 2020-07-14
[Console][Feature request] Schema for foreign key tables should be default to current table's schema


revskill10:
When adding foreign key to a table, currently Hasura only show first schema as default.

Suggestion:

Default schema should be default to current table's schema for ease of setting foreign key .

rikinsk:
@revskill10 the console already seems to be doing this. could you confirm if this is consistent behaviour for you? some more info might be needed to see why this might be happening

revskill10:
@rikinsk I can confirm that the issue is when you create new table.

PR #5795, opened 2020-09-16
Console: Setting Default schema to current table's schema for ease of setting Foreign Key

Varun-Choudhary:
Closes: #5376 

### Description
Earlier table's first schema is set as a default schema while setting foreign key. This PR solve that issue by setting default schema to current table's schema for the ease of setting foreign key.

### Affected components

- [X] Console

### DIFF
--- a/console/src/components/Services/Data/Add/ForeignKeyWrapper.js
+++ b/console/src/components/Services/Data/Add/ForeignKeyWrapper.js
@@ -57,6 +57,7 @@ const ForeignKeyWrapper = ({
         dispatch={dispatch}
         setForeignKeys={setForeignKeys}
         schemaList={orderedSchemaList}
+        currentSchema={currentSchema}
       />
     );
     // TODO handle ongoing request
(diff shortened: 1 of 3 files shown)


kolharsam (commented):
These changes look good to me for the `Reference Schema` part alone. One more thing you'll have to ensure is that the `Reference Tables` shown are from the `currentSchema`. And also of course make sure that you can add foreign keys and the whole process is working fine. :)

kolharsam (commented):
Apart from one small issue, which I will describe here. I think it is working fine. 👍 
So here's how I spotted the issue:
- From `public` schema, switch to a new schema.
- Click on `Create Table`
- Click to add Foreign Keys
- And you will see that under the `Reference Tables` you'd still be seeing tables from the `public` schema.

console/src/components/Services/Data/Add/ForeignKeyWrapper.js
```diff
@@ -105,6 +106,10 @@ const ForeignKeyWrapper = ({
 
     const expandCallback = () => {
       dispatch(toggleFk(i));
+      const newFks = JSON.parse(JSON.stringify(foreignKeys));
+      newFks[i].refSchemaName = event.target.value;
+      dispatch(setForeignKeys(newFks));
```
kolharsam, line 111: I feel we shouldn't mutate `newFks` directly. Here's a suggestion to change that, if required that is:

```suggestion
      const newFks = JSON.parse(JSON.stringify(foreignKeys));
       const updatedFks = {
          ...newFks,
          [i]: {
              ...newFks[i],
              refSchemaName: event.target.value,
          },
      };
      dispatch(setForeignKeys(updatedFks));
```

kolharsam (commented):
The code looks good overall. Just the one small suggestion here. 
Still haven't tested it out, will do that and the approve this PR.

beerose:
Closing in favor of https://github.com/hasura/graphql-engine-mono/pull/813

PR #5797, opened 2020-09-17
multiple postgres database & metadata storage separation

rakeshkky:
### TODO
- [x] Avoid hdb_catalog.hdb_catalog.check_violation 
- [x] Cron/Scheduled triggers
- [x] Console support - metadata storage data APIs
- [x] Backwards compatibilty of `DATABASE_URL`
- [x] v1/query API backwards compatibility
  - [x] Make existing tests work
- [x] pg_dump API to accept source
- [x] Heterogeneous execution (i.e, merge master)
- [ ] Action queries/mutations execution in appropriate source (when there are relationships defined)
- [ ] Tests for multiple sources**

- [ ] Add/modify column issue:
The server gets stuck with requests like the following one:
   ```
    {
         "type": "run_sql",
         "source": "default",
         "args": {
                "sql": "alter table \"public\".\"test\" add column \"test_col\" text null;",
                "cascade": false,
                "read_only": false
         }
     }
   ```

- [ ] `pg_invoke_event_trigger` results in the following error:
   ```
   {
    "internal": {
        "statement": "\n           SELECT hdb_catalog.insert_event_log($1, $2, $3, $4, $5)\n                ",
        "prepared": true,
        "error": {
            "exec_status": "FatalError",
            "hint": null,
            "message": "invalid input syntax for type json",
            "status_code": "22P02",
            "description": "The input string ended unexpectedly."
        },
        "arguments": [
            "(Oid 25,Just (\"public\",Binary))",
            "(Oid 25,Just (\"test\",Binary))",
            "(Oid 25,Just (\"lala\",Binary))",
            "(Oid 25,Just (\"MANUAL\",Binary))",
            "(Oid 114,Just (\"{\\\"row\\\":{\\\"id\\\":2}}\",Binary))"
        ]
    },
    "path": "$",
    "error": "database query error",
    "code": "unexpected"
   }
   ```
- [ ] can't set comments for computed fields. Request:
   ```
       {
            "type": "pg_add_computed_field",
            "args": {
                "table": {
                    "schema": "public",
                    "name": "users"
                },
                "name": "test",
                "definition": {
                    "function": {
                        "schema": "public",
                        "name": "author_full_name"
                    },
                    "table_argument": "author_row"
                },
                "comment": "some comment",
                "source": "my db 1"
            }
        }
   ```

`** - can be done in parallel`

### Description

**Metadata**:-
- This PR overhauls the way in which the metadata is persisted in Postgres database. Now, the metadata is stored as a single `json` blob in `hdb_catalog.hdb_metadata` table in a single row. This reduces the Hasura footprint in the `hdb_catalog` schema.
- Metadata can be stored in a separate database, use `--metadata-database-url` or `HASURA_GRAPHQL_METADATA_DATABASE_URL` option. By default it is stored in the database specified via `--database-url`. One can start the server with only metadata database specified and configure the postgres database as sources (see below). 

**Multiple Databases**
- Introduction of new notion in metadata - `Source`, a user Postgres database where all `tables/views` and `functions` can reside in. We can configure multiple sources via Metadata API or console UI.
- This change will pave the way for supporting sources of multiple backends other than Postgres. (MySQL, MSSQL etc.)

### Changelog

### Affected components

- [x] Server
- [x] Tests

**Actions**:-
1. Action query subscriptions
2. Action with relationships to postgres tables

### Metadata

The metadata version is bumped to `3`. The only difference from version `2` is that now the `tables` and `functions` will be present inside new `sources` field for each time. The other top level fields remain as it is.
Example:- 
version `2`:
```yaml
version: 2
tables:
- table:
    schema: public
    name: test
functions:
- function:
    schema: public
    name: search_test
```
version `3`:
```yaml
version: 3
sources:
- name: default
  tables:
  - table:
      schema: public
      name: test
  functions:
  - function:
      schema: public
      name: test
- name: chinook
  tables:
  - table:
      schema: public
      name: actors
  configuration:
    database_url: postgres://postgres@127.0.0.1:5432/chinook
    connection_pool_settings:
      retries: 1
      idle_timeout: 180
      max_connections: 50
```
Where `default` is the source configured via `--database-url` server option and `chinook` is newly added postgres source via metadata API.

### New APIs
It is recommended to use newly added `/v1/metadata` to manage metadata and `/v2/query` to query postgres sources via JSON DSL, instead of `/v1/query` API. 

#### /v1/metadata
It contains all query types in `/v1/query` except database DML operations like `insert`, `select` etc. 
Query types for managing postgres related metadata are appended with `pg_`. 
For example, `track_table` is changed to `pg_track_table` and similarly for all query types for managing tables, functions, relationships, remote relationships, event triggers and computed fields.
This also supports `bulk` query type for running multiple metadata queries in a single request.

**APIs for console data requirements from metadata storage**:
1. Get the console state:

request,
```json
{
  "type": "get_catalog_state",
  "args": {}
}
```
response,
```json
{
    "id": "fc0bc750-b343-44cc-a7f4-46637b0067f7",
    "cli_state": {},
    "console_state": {}
}
```
2. Set the console state:

request,
```json
{
  "type": "set_catalog_state",
  "args": {
    "type": "console",
    "state": {"key": "value"}
  }
}
```
`{ "message": "success" }` is the response.

3. Get invocation logs for cron/scheduled (one-off) events.

request,
```json
{
  "type": "get_event_invocations",
  "args": {
    "type": "one_off",
    "limit": 10,
    "offset": 10
  }
}
```
```json
{
  "type": "get_event_invocations",
  "args": {
    "type": "cron",
    "trigger_name": "some_trigger_name",
    "limit": 10,
    "offset": 10
  }
}
```

By event id,
```json
{
  "type": "get_event_invocations",
  "args": {
    "type": "cron",
    "event_id": "some_event_id"
  }
}
```

response,
```json
{
    "invocations": <list of invocation objects>,
    "count": <count of total invocations"
}
```

By Event Id:

4. Get all scheduled events.

One off:-
**pending**
```json
{
    "type": "get_scheduled_events",
    "args": {
        "type": "one_off",
        "limit": 10,
        "offset": 10,
        "status": ["scheduled"] 
    }
}
```
**processed**
```json
{
    "type": "get_scheduled_events",
    "args": {
        "type": "one_off",
        "limit": 10,
        "offset": 10,
        "status": ["delivered", "error", "dead"] 
    }
}
```
Cron:-
```json
{
    "type": "get_scheduled_events",
    "args": {
        "type": "cron",
        "trigger_name": "test"
    }
}
```
`trigger_name` is required if `type` is `cron`.

Response:-
```json
{
    "events": <list of events, the object structure is different for cron and one_off>,
    "count": <total-count>
}
```

5. Delete cron/scheduled (one-off) event by event id

request:-
```json
{
  "type": "delete_scheduled_event",
  "args": {
    "type": "one_off",
    "event_id": "00a8d34c-de0d-417a-ac9f-22df310af989"
  }
}
```
`{"message": "success"}` is the response.

#### /v2/query
All database DML operations like `insert`, `select`, `delete`, `update`, `count` and `run_sql` from `/v1/query`. An optional top level field `source` is accepted in payload and its absence treated as `default` source. These queries are run in the Postgres source specified.
Example:-
```yaml
---
type: select
source: chinook
args:
  table: actor
  columns:
  - id
  - name
```

---

#### Console issues:

- ~get_invocation_logs: filter by **cron trigger name**, for **one-off** return all invocations~
- ~get_invocation_logs: pagination~
- ~get_invocation_logs: count all~
```
// example req
    {
      "type": "get_event_invocations",
      "args": {
        "type": "one_off",
        "offset": 10,
        "limit": 10,
      }
    }

// response
{
  invocations: [...], count: 122,
}
```

- ~get_scheduled_events: filter by status - pending/processed~
- ~get_scheduled_events: pagination~
- ~get_scheduled_events: count~


stevepro-lab:
This is great work 👍
Looming forward to try out this feature

kolharsam:
These are some of the issues that I ran into (Will keep updating this comment as and when I encounter any issue)
cc: @rakeshkky 

1. ```JSON {
    "type": "select",
    "source": "default",
    "args": {
        "table": {
            "name": "event_log",
            "schema": "hdb_catalog"
        },
        "columns": ["*"]
    }}

So, any `v2/query` on `hdb_catalog` of the `default` source results in the following error:
```JSON
{
    "path": "$",
    "error": "table \"hdb_catalog.event_log\" does not exist",
    "code": "not-exists"
}
```

....

rakeshkky:
> These are some of the issues that I ran into (Will keep updating this comment as and when I encounter any issue)
> cc: @rakeshkky @beerose
> 
>     1. ```json
>         "type": "select",
>         "source": "default",
>         "args": {
>             "table": {
>                 "name": "hdb_metadata",
>                 "schema": "hdb_catalog"
>             },
>             "columns": ["*"]
>         }}
>        ```
> 
> 
> So, any `v2/query` on `hdb_catalog` of the `default` source results in the following error:
> 
> ```json
> {
>     "path": "$",
>     "error": "table \"hdb_catalog.event_log\" does not exist",
>     "code": "not-exists"
> }
> ```
> 
> ....

Those tables aren't added to metadata. Hence you can't use any graphql or RQL query to CRUD the data. You can use `run_sql`.

kolharsam:
@rakeshkky 
I was facing an issue with creating custom types. I was trying this request on `v1/metadata`

```{
    "type": "set_custom_types",
    "args": {
        "scalars": [],
        "input_objects": [
            {
                "name": "SampleInput",
                "description": null,
                "fields": [
                    {
                        "name": "username",
                        "type": "String!",
                        "description": null
                    },
                    {
                        "name": "password",
                        "type": "String!",
                        "description": null
                    }
                ]
            }
        ],
        "objects": [
            {
                "name": "SampleOutput",
                "description": null,
                "fields": [
                    {
                        "name": "accessToken",
                        "type": "String!",
                        "description": null
                    }
                ]
            }
        ],
        "enums": []
    }
}
```




So this was the error that was being reported: 
```
{
    "path": "$.objects[0]",
    "error": "key \"relationships\" not found",
    "code": "parse-failed"
}
```

But if instead, I tried to add a `relationships` key to the above request within objects as shown [here](https://hasura.io/docs/1.0/graphql/core/api-reference/schema-metadata-api/custom-types.html#objecttype). I got the following error:
```
{
    "path": "$.objects[0].relationships",
    "error": "parsing NonEmpty failed, unexpected empty list",
    "code": "parse-failed"
}
```

0x777:
Handling `metadata-database-url` and `database-url` in 1.4.

```haskell
case (db-url-option, md-db-url-option) of

   -- db-url is given, the backwards compatibility case
  (Just db-url, _) -> do

    -- metadata db url is the db-url if it isn't explicitly specified
    let md-url = fromMaybe db-url md-db-url-option

    try reading `hdb_catalog.hdb_version` from md-db-url
    if the catalog hasn't been initialized:
      initialize the catalog
      add a source with name 'default' and 'db-url' as the source's url
    catalog is initialized but pre-1.4:
      get the old metadata
      add a source with name 'default' and 'db-url' as the source's url with the old metadata as the source's metadata
    catalog is initialized and is the latest:
      nothing to do

  -- metadata url is present but db-url isn't
  (Nothing, Just md-url) -> do
    try reading `hdb_catalog.hdb_version` from md-db-url
    if the catalog hasn't been initialized:
      initialize the catalog
    catalog is initialized but pre-1.4:
      get the old metadata
      if the old metadata is empty:
        migrate the catalog
      else:
        fail (there is some metadata but we don't know what source it belongs to)
    catalog is initialized and is the latest:
      nothing to do

   (Nothing, Nothing) -> error
```

The idea is that during the upgrade from the older catalog to 1.4 catalog, we add the `db-url` as a `default` source.

lexi-lambda:
What is the purpose of the explicit “if the old metadata is empty, migrate the catalog” clause in the case where a metadata URL is given but no DB URL is given? It seems simpler (and more uniform) to just fail in that situation if we see pre-1.4 metadata, since that scenario has no reason to ever happen.

Aside from that, I think this sounds reasonable to me.

0x777:
> It seems simpler (and more uniform) to just fail in that situation if we see pre-1.4 metadata, since that scenario has no reason to ever happen.

I agree, we can just throw an error if we see pre-1.4 catalog.

rakeshkky:
@kolharsam The 'relationships' field in custom types if fixed in https://github.com/hasura/graphql-engine/pull/5797/commits/74302add619f9af4e92ef784a6d6f6e5f9b78ba4.

PR #5805, opened 2020-09-18
console: automatically wrap default column values in quotes (close #418)

### LINKED ISSUE #418: Suggest adding quotes while setting default values for JSON types
For a JSON column called `roles`, say, setting a default type to `["user"]`  doesn't work but `'["user"]'` works.

This escaping/quoting is probably required for a bunch of other use-cases?

Varun-Choudhary:
closes: #418 

### Description
This PR automatically wraps the default value of json type column in quotes `' '`. So that it will throw no error. 

### Affected components

- [X] CONSOLE


console/src/components/Services/Data/utils.js
```diff
@@ -791,6 +791,18 @@ WHERE
...
   ['text', 'varchar', 'char', 'bpchar', 'name'].includes(colType);
 
+export const wrapQuotes = (colType, colDefault) => {
+  const checkIfFunctionFormat = isPostgresFunction(colDefault);
+  if (
+    (isColTypeString(colType) && !checkIfFunctionFormat) ||
+    colType === 'json' ||
+    colType === 'jsonb'
+  ) {
+    return "'" + colDefault + "'";
```
kolharsam, line 801: I would suggest that you use template strings here.

console/src/components/Services/Data/TableModify/ModifyActions.js
```diff
@@ -1561,15 +1556,8 @@ const saveColumnChangesSql = (colName, column, onSuccess) => {
...
     }
 
-    const colDefaultWithQuotes =
-      isColTypeString(colType) && !isPostgresFunction(colDefault)
-        ? `'${colDefault}'`
-        : colDefault;
-    const originalColDefaultWithQuotes =
-      isColTypeString(colType) && !isPostgresFunction(originalColDefault)
-        ? `'${originalColDefault}'`
-        : originalColDefault;
```
kolharsam, line 1571: Please compare the arguments/values that are used here against the latest changes made by you.

kolharsam (commented):
Minor changes suggested. Otherwise, it looks good! 👍

console/src/components/Services/Data/utils.js
```diff
@@ -791,6 +791,18 @@ WHERE
...
 
+export const wrapQuotes = (colType, colDefault) => {
+  const checkIfFunctionFormat = isPostgresFunction(colDefault);
+  if (
+    (isColTypeString(colType) && !checkIfFunctionFormat) ||
+    colType === 'json' ||
+    colType === 'jsonb'
+  ) {
+    return `'${colDefault}'`;
+  }
```
beerose, line 802: This would be needed for other data types as well. E.g. all the date/time types. Could you go over all possible data types and handle them as well? It may turn out that we could wrap everything in quotes.

beerose (changes requested):

beerose (approved):
Console

PR #5832, opened 2020-09-24
Better concurrency mechanic for shutdown-latch

sordina:
### Description

There is a shutdown-latch datatype in use and exported by the graphql-engine haskell library. This has an interface of `newShutdownLatch`, `waitForShutdown`, and `shutdownGracefully`.

There is an issue with the current implementation that doesn't allow for reuse of the latch:

* Calling shutdown twice will block the second call - This is unlikely to be expected.
* Waiting for shutdown more than once will block - This prevents using the same latch to trigger multiple shutdown procedures

### Changelog

### Affected components

- [x] Server

### Related Issues

No known related issues.

### Solution and Design

This patch addresses this by changing the implementation of

```haskell
waitForShutdown = C.takeMVar . unShutdownLatch
```

to

```haskell
waitForShutdown = C.readMVar . unShutdownLatch
```

and

```haskell
shutdownGracefully' = flip C.putMVar () . unShutdownLatch
```

to

```haskell
shutdownGracefully' = void . flip C.tryPutMVar () . unShutdownLatch
```

Thus allowing multiple non-blocking calls to `shutdownGracefully`, and multiple instances of `waitForShutdown` that all unblock immediately after `shutdownGracefully` has been called once.

### Steps to test and verify

In Haskell library-usage of graphql-server you may now pass the same latch to multiple running services and terminate them all simultaneously with a single shutdown call.

No known limitations.

### Server checklist

N/A

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/App.hs
+++ b/server/src-lib/Hasura/App.hs
@@ -274,11 +274,11 @@ waitForShutdown = C.readMVar . unShutdownLatch
 
 -- | Initiate a graceful shutdown of the server associated with the provided
 -- latch.
-shutdownGracefully :: InitCtx -> IO ()
+shutdownGracefully :: InitCtx -> IO Bool
 shutdownGracefully = shutdownGracefully' . _icShutdownLatch
 
-shutdownGracefully' :: ShutdownLatch -> IO ()
-shutdownGracefully' = void . flip C.tryPutMVar () . unShutdownLatch
+shutdownGracefully' :: ShutdownLatch -> IO Bool
+shutdownGracefully' = flip C.tryPutMVar () . unShutdownLatch
 
 -- | If an exception is encountered , flush the log buffer and
 -- rethrow If we do not flush the log buffer on exception, then log lines
--- a/server/src-lib/Hasura/App.hs
+++ b/server/src-lib/Hasura/App.hs
@@ -270,7 +270,7 @@ newShutdownLatch = fmap ShutdownLatch C.newEmptyMVar
 
 -- | Block the current thread, waiting on the latch.
 waitForShutdown :: ShutdownLatch -> IO ()
-waitForShutdown = C.takeMVar . unShutdownLatch
+waitForShutdown = C.readMVar . unShutdownLatch
 
 -- | Initiate a graceful shutdown of the server associated with the provided
 -- latch.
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/App.hs
```diff
@@ -270,15 +270,15 @@ newShutdownLatch = fmap ShutdownLatch C.newEmptyMVar
...
+waitForShutdown = C.readMVar . unShutdownLatch
 
 -- | Initiate a graceful shutdown of the server associated with the provided
 -- latch.
 shutdownGracefully :: InitCtx -> IO ()
 shutdownGracefully = shutdownGracefully' . _icShutdownLatch
 
 shutdownGracefully' :: ShutdownLatch -> IO ()
-shutdownGracefully' = flip C.putMVar () . unShutdownLatch
+shutdownGracefully' = void . flip C.tryPutMVar () . unShutdownLatch
```
paf31, line 281: I think it would be good to return the result instead of discarding it. Trying to shut down twice should result in an error in some cases IMO.

sordina: ok

sordina: I don't call void here any more.

server/src-lib/Hasura/App.hs
```diff
@@ -270,15 +270,15 @@ newShutdownLatch = fmap ShutdownLatch C.newEmptyMVar
 
 -- | Block the current thread, waiting on the latch.
 waitForShutdown :: ShutdownLatch -> IO ()
-waitForShutdown = C.takeMVar . unShutdownLatch
+waitForShutdown = C.readMVar . unShutdownLatch
 
 -- | Initiate a graceful shutdown of the server associated with the provided
 -- latch.
```
paf31, line 276:
```suggestion
-- | Initiate a graceful shutdown of the server associated with the provided
-- latch. Returns 'True' if the shutdown was initiated successfully - that is,
-- the latch is not blocked by another existing shutdown attempt.
```

paf31 (approved):
LGTM 👍

server/src-lib/Hasura/App.hs
paf31: I'll just commit this and save you having to wait for another CI cycle.

Issue #4751, opened 2020-05-12
console: reset permissions editor state after creating a new role


beerose:
![permissions-editor](https://user-images.githubusercontent.com/9019397/81736573-14504980-9497-11ea-9b82-c08f110b602d.gif)

### Steps to reproduce

1. Go to the permissions tab.
0. Create a new role. 
0. Open editor for arbitrary action for this role.
0. Start creating a new role — start typing in the `Enter new role` input.

**Current behavior:** the editor is prefilled with the state from the previous role.
**Expected behavior:** the state should be reset.

PR #5833, opened 2020-09-24
Console: Reset the Permission editor after creating new role

Varun-Choudhary:
Closes: #4751 
### Description

Closes the Editor every-time when we enter new role name to create permission

### Affected components

- [X] Console


beerose (approved):
Console

PR #5837, opened 2020-09-25
console: catch api errors, resolves console getting stuck on loading screen (close #5774)

### LINKED ISSUE #5774: /apis/migrate/settings HTTP 500 error isn't displayed in the UI
Hi! :wave:

It would be nice to read this error message in console UI below "Something went wrong..." heading.

image

abhi40308:
Earlier, console was stuck in loading state in case of some API errors, this PR fixes that issue.

**Edit** : Few observations: 
- The error in the reported issue only happens in cli mode.
- It was getting stuck due to error in catching error response if it comes as text, and handling the error object.
- We don't need an api fallback page as console does not get stuck to the `loading...` screen now, and we can reach the console, where we show the error notification like so
Screenshot 2021-03-16 at 2 27 08 PM

### Description
fix #5774 

### Changelog

### Affected components

- [x] Console


console/src/components/Error/ErrorBoundary.tsx
```diff
@@ -19,16 +20,19 @@ export interface Metadata {
 export interface ErrorBoundaryProps {
   metadata: Metadata;
   dispatch: (arg: unknown) => Promise<unknown>; // TODO update when Redux is migrated to TS;
+  errorValue: any;
+  requestError: any;
+  requestURL: any;
```
kolharsam, line 25: We can  try using the correct types for these: 
- errorValue: Error;
- requestError: Error;
- requestURL: string;

console/src/components/Error/ErrorBoundary.tsx
```diff
@@ -79,21 +83,25 @@ class ErrorBoundary extends React.Component<
...
   render() {
-    const { metadata } = this.props;
-    const { hasError, type, error } = this.state;
+    const { metadata, errorValue, requestError, requestURL } = this.props;
+    const { hasReactError, type, error } = this.state;
 
-    if (hasError && metadata.ongoingRequest) {
+    if (hasReactError && metadata.ongoingRequest) {
       return <Spinner />;
     }
```
kolharsam, line 91: Question: When does this case occur?

console/src/components/Error/ErrorBoundary.tsx
```diff
@@ -79,21 +83,25 @@ class ErrorBoundary extends React.Component<
...
         <PageNotFound resetCallback={this.resetState} />
       ) : (
         <RuntimeError resetCallback={this.resetState} error={error} />
       );
     }
 
+    if (requestError===null && errorValue) {
+        return <ApiError error={errorValue} requestURL={requestURL} />
+    }
+
```
kolharsam, line 104: I think in most cases, it should work well. But yes, we should try out as many different scenarios and make sure all requests are being made using `requestAction` and these errors are caught appropriately.

kolharsam (commented):
Just leaving a list of things that need to be done for this PR:
 - [ ] Understand how a failed request is handled atm. (on the `Loading...` screen)
 - [ ] `RequestAction` is used in mulitple places when the console is loaded, so yeah, if those requests fail, we should not be showing this API error page, but only on the `Loading...` page.
(We can perhaps use `isLoading` along with the current conditions that you have set)

console/src/utils/requestAction.ts
```diff
@@ -46,9 +46,27 @@ const requestAction = (
...
+              isResponseJSON = false;
+            }
+
+            if (!isResponseJSON) {
+              return response.text().then(errorMsg => {
+                dispatch({ type: FAILED_REQUEST });
+                if (ERROR) {
+                  dispatch({ type: ERROR, response, data: errorMsg });
+                }
+                // a bool to check whether loading screen is shown, and based on this call error request
```
kolharsam, line 63: TODO

console/src/routes.js
```diff
@@ -54,7 +54,8 @@ const routes = store => {
           cb();
         },
         r => {
-          if (r.code === 'data_api_error') {
+          // TODO: let's add one more case here when 'r' is not defined.
```
kolharsam, line 57: TODO

console/src/components/Error/ErrorBoundary.tsx
```diff
@@ -79,21 +83,26 @@ class ErrorBoundary extends React.Component<
...
-    if (hasError) {
+    if (hasReactError) {
       return type === '404' ? (
         <PageNotFound resetCallback={this.resetState} />
       ) : (
         <RuntimeError resetCallback={this.resetState} error={error} />
       );
     }
 
+    // TODO: Check the condition here, it causes the console to keep making repeated requests
```
kolharsam, line 101: TODO

kolharsam (commented):
Just pointing these TODOs out to you @abhi40308

console/src/routes.js
```diff
@@ -54,17 +54,18 @@ const routes = store => {
...
         r => {
-          if (r.code === 'data_api_error') {
+          if (r && r.code === 'data_api_error') {
             dispatch(showErrorNotification('Error', null, r));
           } else {
             dispatch(
               showErrorNotification(
                 'Connection error',
-                'Hasura console is not able to reach your Hasura GraphQL engine instance. Please ensure that your ' +
+                'Hasura console is not able to reach your Hasura CLI instance. Please ensure that your ' +
```
beerose, line 63: Can you elaborate on this change? What if I run console in server mode and there's no CLI involved?

beerose (changes requested):

console/src/routes.js
abhi40308: Thanks for review. We are only dealing with the cli mode here (its in a `if` condition for CLI), and I guess this error message was mistaken, check for reference: https://github.com/hasura/graphql-engine-mono/blob/main/console/src/routes.js#L57

abhi40308: we don't need an api error fallback page in my opinion, I've updated the pr and pr description accordingly

beerose:
Closing in favor of https://github.com/hasura/graphql-engine-mono/pull/856

PR #5845, opened 2020-09-26
Generalize combinators from Monad to Applicative

tomjaguarpaw:
### Description

There were some combinators whose type specifies `Monad` when `Applicative` is sufficient. This PR generalizes the signatures to `Applicative`.

### Affected components

- [x] Server

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


tomjaguarpaw:
This doesn't require a changelog entry. I don't think I have the power to set that label.

abooij (approved):
LGTM. While we're at it, perhaps we can remove the redundant brackets around the `Applicative m` constraint.

tomjaguarpaw:
> perhaps we can remove the redundant brackets

Sure, done.

tirumaraiselvan (approved):

PR #5856, opened 2020-09-28
console: fix radio behaviour in table insert/edit item (close #3391)

### LINKED ISSUE #3391: Unexpected radio behaviour in table edit item

Varun-Choudhary:
Closes: #3391 

### Description

Earlier when we press tab to switch to the next textbox while editing or inserting rows, radio button does not activate. This PR resolve that issue. Now when we write some in textbox, radio button activates.

### Affected components

- [X] Console


console/src/components/Services/Data/TableBrowseRows/EditItem.js
```diff
@@ -77,6 +77,31 @@ class EditItem extends Component {
...
+        if (
+          textValue === undefined ||
+          textValue === null ||
+          textValue.length === 0
+        ) {
+          const radioToSelectWhenEmpty = prevValue
+            ? refs[colName].defaultNode
+            : refs[colName].nullNode;
+
+          refs[colName].insertRadioNode.checked = false;
```
soorajshankar, line 101: I think it would be nice if we can plan for removing direct DOM controls and using local state instead.
@beerose thoughts?

soorajshankar (changes requested):
Left a minor comment

console/src/components/Services/Data/TableBrowseRows/EditItem.js
beerose: Yeah, totally agree, it would be. However, this is a pretty complex and big portion of code and refactoring it may take some time. So I'd leave it for later.

beerose:
Close in favor of https://github.com/hasura/graphql-engine-mono/pull/420

PR #5857, opened 2020-09-28
docs: clarify JWT type `RS512`

afitzek:
### Description

This is just clarifying a sentence in the docs, which might lead to confusion.

The jwt type `RS512` does not mean RSA keys with 512-bit
keysize, but that the hash algorithmn used, that creates
the hash, which is in turn signed using the RSA key is
SHA-512.

### Affected components

- [x] Docs


afitzek:
I tried to add the `no-changelog-required` label, but it seems like I don't have the permissions to do this. :)

GavinRay97 (approved):

marionschleifer (approved):

rikinsk (approved):

PR #5865, opened 2020-09-29
server: redesign MonadExecuteQuery in preparation for heterogeneous execution

nicuveo:
## Motivation

This PR updates `MonadExecuteQuery` to simplify / clarify its semantics in the context of heterogeneous execution / remote joins. In its previous state, `MonadExecuteQuery` had several concerns: perform the postgres call, optionally do some caching, compute which headers should be added to the result... all combined in one function: `executeQuery`.

Furthermore, it was used under the assumption that a query would translate to one DB call: the cache key was the generated SQL query, and `executeQuery` was a wrapper around one call to postgres. As we are working on heterogeneous execution, support for other databases, and hopefully at some point on generalized joins, we are moving towards a state in which one graphql query corresponds to several database calls, potentially to several different databases.

## Description

The proposed change solves the following issues:
  * it introduces a new cache key that does not depend on the generated SQL, which allows for top-level query caching (at the cost of preventing cache hits for users with different roles but similar permissions), regardless of how the query gets executed
  * it removes the responsibility of calling postgres from `executeQuery`: both instances were calling the same function (`runQueryTx`), and it isn't desirable to duplicate the logic that translates a query plan into several db calls in the typeclass itself
  * it separates the responsibilities in several functions rather than attempting to be in one all-encompassing function call.

At time of writing, rather than wrapping the call to Postgres, the typeclass is now split in two functions:
  * `cacheLookup`: the preprocessing step, responsible for attempting to retrieve the query from the cache, and computing the headers
  * `cacheStore`: the post-processing step, takes the result of the computation and the cache key, and is responsible for updating the cache.

## Limitations

- the actual call to postgres is the same for both implementations of `MonadExecuteQuery`; it might be required to introduce a new `runAgainstPG` class function to allow for specific specialization of the call, if additional tracing capabilities are required
- this change will require the existing implementation of `executeQuery` to be split into several functions, the monad won't have full control over the execution

## Alternative design

A simpler approach would have been to have the following:
```
class Monad m => MonadExecuteQuery m where
  executeQuery :: .... -> m a -> m a
```
where `executeQuery` takes as an argument the monadic action to be performed, and has therefore freedom to decide how to execute it. However, this makes writing monad transformer instances for `MonadExecuteQuery` painful, as it means having to unstack / restack, à la `MonadBaseControl`.

It is, however, a viable alternative, should we discover that this new approach is too limited.


server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -404,22 +411,26 @@ onStart env serverEnv wsConn (StartMsg opId q) = catchAndIgnore $ do
...
       let telemLocality = Telem.Local
       logOpEv ODStarted (Just requestId)
       -- log the generated SQL and the graphql query
       logQueryLog logger q genSql requestId
-      withElapsedTime (runExceptT action) >>= \case
-        (_,      Left err) -> postExecErr requestId err
+      result <- withElapsedTime (runExceptT action) >>= \case
+        (_,      Left err) -> do
+          postExecErr requestId err
+          throwError ()
```
0x777, line 423: The `completed` message has to be sent even when there is an error. So `throwError` shouldn't be called, Instead the type of this function can be `ExceptT () m (Maybe EncJSON)`? i.e, only the successful response is cached.

0x777: I'm not very sure why this function has the signature `ExceptT () ..`.

nicuveo: I fixed it using your suggestion; not sure why that function has ExceptT either, I'll have a look at whether it can be simplified.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -191,26 +220,25 @@ runQueryDB
...
   -- spent in the PG query; for telemetry.
-runQueryDB reqId (query, queryParsed) asts _userInfo (tx, genSql) =  do
+runQueryDB reqId query _userInfo (tx, genSql) =  do
   -- log the generated SQL and the graphql query
   E.ExecutionCtx logger _ pgExecCtx _ _ _ _ <- ask
   logQueryLog logger query (Just genSql) reqId
-  (telemTimeIO, respE) <- withElapsedTime $ runExceptT $ trace "Query" $
-    Tracing.interpTraceT id $ executeQuery queryParsed asts (Just genSql) pgExecCtx tx
-  (respHdrs,resp) <- liftEither respE
+  (telemTimeIO, respE) <- withElapsedTime $ runExceptT $ trace "Postgres Query" $
```
0x777, line 235: Looks like we are changing the paths in the traces? `Query -> Postgres Query`, is that okay?

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -42,24 +43,45 @@ import qualified Language.GraphQL.Draft.Syntax          as G
...
-    -> Maybe EQ.GeneratedSqlMap
-    -> PGExecCtx
-    -> TraceT (LazyTx QErr) EncJSON
-    -> TraceT (ExceptT QErr m) (HTTP.ResponseHeaders, EncJSON)
+  cacheLookup
+    :: [QueryRootField UnpreparedValue]
+    -- ^ Used to check that the query is cacheable
+    -> QueryCacheKey
+    -- ^ Key that uniquely identifies the result of a query execution
+    -> TraceT m (HTTP.ResponseHeaders, Maybe EncJSON)
```
0x777, line 62: What's the case where response headers are returned from the cache but not the response?

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
nicuveo: This was discussed with @paf31; since this trace used to cover all that was done in `MonadExecuteQuery`, the top-level "Query" trace call has been moved to the calling function; this is now just the span where we call Postgres.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
nicuveo: We've kept `cacheLookup` as a name, but it is a bit of a misnomer: it's more of a preprocessing steps; it *both* computes the response headers and does the cache lookup; we always use the response headers that it returns, regardless of whether there was a cache hit or not.

0x777 (approved):
LGTM

nicuveo:
Do we still merge this one, or do we decide to simply include it as part of #5869?

tirumaraiselvan (approved):
no cl required

PR #5869, opened 2020-09-30
server: heterogeneous execution of GraphQL queries

abooij:
### Description

The server could already execute queries that either fetch data from the database, or, through remotes, from other GraphQL servers.  This PR also enables mixing such data sources within one query.  So one can have queries such as

```
query {
  articles {
    title
  }
  weather {
    temperature
  }
}
```

where the articles are fetched from the database, and the weather is fetched from a remote server.

### Changelog

### Affected components

- [x] Server
- [x] Tests

### Related Issues

This PR depends on #5865.

### Solution and Design

For an incoming query we already used to generate an "execution plan" which specified how to execute that query. However, so far an execution plan has been a single execution step, where an execution step can be of database-type, of remote-type or of "raw"-type (used for introspection). This PR changes the type of execution plans to be an insertion-ordered hashmap of execution steps

```haskell
type ExecutionPlan db remote raw = InsOrdHashMap Text (ExecutionStep db remote raw)
```

where the key corresponds to a root field name.

### Steps to test and verify

A test suite is included.

Although this PR allows to mix root fields originating from different data sources, it does not allow joining between them arbitrarily beyond the existing "remote joins" that load relationships from Postgres to a remote. Such "generalized joins" are part of future work.

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/GraphQL/Execute/Prepare.hs
```diff
@@ -46,7 +44,7 @@ type PrepArgMap = IntMap.IntMap (Q.PrepArg, PGScalarValue)
 -- | Full execution plan to process one GraphQL query.  Once we work on
 -- heterogeneous execution this will contain a mixture of things to run on the
 -- database and things to run on remote schemas.
-type ExecutionPlan db remote raw = ExecutionStep db remote raw
+type ExecutionPlan db remote raw = InsOrdHashMap Text (ExecutionStep db remote raw)
```
lexi-lambda, line 47: Could `ExecutionStep` be made a little less polymorphic to avoid the need for so many type parameters?

1. Could `remote` be hardcoded to `RemoteCall`, or is it sometimes something else?

2. Could `raw` be hardcoded to `Aeson.Value`, or is it sometimes something else?

server/src-lib/Hasura/GraphQL/Execute/Mutation.hs
```diff
@@ -161,51 +170,25 @@ convertMutationSelectionSet env logger gqlContext sqlGenCtx userInfo manager req
...
-            varDefs
-            (map (G.SelectionField . snd . snd) remotes)
-            varValsM
-      if all (\remote' -> fst (snd firstRemote) == fst (snd remote')) remotes
-        then return $ ExecStepRemote (fst (snd firstRemote), remoteOperation, varValsM')
-        else throw400 NotSupported "Mixed remote schemas are not supported"
-    _ -> throw400 NotSupported "Heterogeneous execution of database and remote schemas not supported"
-  -- Build and return an executable action from the generated SQL
+  let userSession = _uiSession userInfo
+      rjCtx = (manager, reqHeaders, userInfo)
```
lexi-lambda, line 174: I think `rjCtx` is pushing the limit of our [style guide’s section on naming](https://github.com/hasura/graphql-engine/blob/4062cdcbeabd10541f9e51e8b76b97f079f19b00/server/STYLE.md#naming). Can we expand it to `remoteJoinCtx`?

server/src-lib/Hasura/GraphQL/Execute/Query.hs
```diff
@@ -40,55 +36,56 @@ import           Hasura.EncJSON
...
+-- | Required to log in `query-log`
+instance J.ToJSON PreparedSql where
+  toJSON (PreparedSql q prepArgs _) =
+    J.object [ "query" J..= Q.getQueryText q
+             , "prepared_arguments" J..= fmap (pgScalarValueToJson . snd) prepArgs
              ]
 
+-- | The computed SQL with alias which can be logged. Nothing here represents no
+-- SQL for cases like introspection responses. Tuple of alias to a (maybe)
+-- prepared statement
```
lexi-lambda, line 68: This comment says this is a “tuple,” but this is a map. I’ve revised the comment to what I think is correct, but feel free to change my wording if I’m wrong or if you have something better.

```suggestion
-- | Maps root fields to the SQL used to execute them. For fields that do not
-- generate any SQL, the value is 'Nothing'. This information is retained in
-- this form only for logging; it is not consulted during actual execution.
```

server/src-lib/Hasura/GraphQL/Execute/Query.hs
```diff
@@ -135,56 +130,49 @@ mkCurPlanTx
...
-      RFPPostgres (PGPlan q _ prepMap remoteJoins) -> do
-        let args = withUserVars (_uiSession userInfo) $ IntMap.elems prepMap
-        return $ RRSql $ PreparedSql q args remoteJoins
-      RFPActionQuery tx -> pure $ RRActionQuery tx
-    return (alias, fldResp)
-
-  (,) <$> mkLazyRespTx env manager reqHdrs userInfo directives resolved <*> pure (mkGeneratedSqlMap resolved)
+    RFPPostgres ps@(PreparedSql q prepMap remoteJoinsM) ->
+      let args = withUserVars (_uiSession userInfo) prepMap
+          -- TODO this quietly assumes the intmap keys are contiguous
```
lexi-lambda, line 141: Does this TODO need to be fixed?

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -84,64 +120,112 @@ runGQ
...
-  (telemTimeTot_DT, (telemCacheHit, telemLocality, (telemTimeIO_DT, telemQueryType, !resp))) <- withElapsedTime $ do
+  (telemTimeTot_DT, (telemQueryType, telemTimeIO_DT, telemLocality, resp)) <- withElapsedTime $ do
     E.ExecutionCtx _ sqlGenCtx pgExecCtx {- planCache -} sc scVer httpManager enableAL <- ask
 
     -- run system authorization on the GraphQL API
     reqParsed <- E.checkGQLExecution userInfo (reqHeaders, ipAddress) enableAL sc reqUnparsed
                  >>= flip onLeft throwError
 
-    (telemCacheHit, execPlan) <- E.getResolvedExecPlan env logger pgExecCtx {- planCache -}
+    (_telemCacheHit, execPlan) <- E.getResolvedExecPlan env logger pgExecCtx {- planCache -}
```
lexi-lambda, line 130: Is it okay that we’re throwing this information away? And if so, can it be eliminated from the result altogether?

server/src-lib/Hasura/GraphQL/Execute/Remote.hs
```diff
@@ -0,0 +1,46 @@
...
+
+collectVariables
+  :: forall fragments var
+   . (Foldable fragments, Hashable var, Eq var)
+  => G.SelectionSet fragments var
+  -> Set.HashSet var
+collectVariables =
+  Set.unions . fmap (foldMap Set.singleton)
+
+buildTypedOperation
```
lexi-lambda, line 32: This is subjective, but I found the name `buildTypedOperation` a little bit mystifying, though I suppose in some sense it is ultimately what the function does. I think I would find a name like `buildExecStepRemote` more helpful.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -191,26 +275,21 @@ runQueryDB
...
-  -> [QueryRootField UnpreparedValue]
+  -> GQLReqUnparsed
   -> UserInfo
-  -> (Tracing.TraceT (LazyTx QErr) EncJSON, EQ.GeneratedSqlMap)
-  -> m (DiffTime, Telem.QueryType, HTTP.ResponseHeaders, EncJSON)
-  -- ^ Also return 'Mutation' when the operation was a mutation, and the time
-  -- spent in the PG query; for telemetry.
-runQueryDB reqId (query, queryParsed) asts _userInfo (tx, genSql) =  do
+  -> Text
+  -> (Tracing.TraceT (LazyTx QErr) EncJSON, Maybe EQ.GeneratedSqlMap)
```
lexi-lambda, line 283: It seems like the `GeneratedSqlMap` here only ever contains a single element, and the `fieldName` is given elsewhere anyway. Could this just accept a chunk of SQL? The map wrapper seems unnecessary.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
```diff
@@ -84,64 +120,112 @@ runGQ
...
+        (responseHeaders, cachedValue) <- Tracing.interpTraceT id $ cacheLookup asts cacheKey
+        case cachedValue of
+          Just cachedResponseData ->
+            pure (Telem.Query, 0, Telem.Local, HttpResponse cachedResponseData responseHeaders)
+          Nothing -> do
+            conclusion <- runExceptT $ forWithKey queryPlans $ \fieldName -> \case
+              E.ExecStepDB txGenSql -> doQErr $ do
+                (telemTimeIO_DT, resp) <-
+                  runQueryDB reqId reqUnparsed userInfo fieldName
+                  (Just . Map.singleton (G.unsafeMkName fieldName) <$> txGenSql)
```
lexi-lambda, line 145: Could `QueryExecutionPlan` be changed to include a `Name` already so that we don’t need to use `unsafeMkName`?

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
```diff
@@ -346,31 +352,53 @@ onStart env serverEnv wsConn (StartMsg opId q) = catchAndIgnore $ do
...
+              runRemoteGQ fieldName execCtx requestId userInfo reqHdrs opDef rsi varValsM
+            E.ExecStepRaw json ->
+              buildRaw json
+          buildResult Telem.Query timerTot requestId conclusion
+          case conclusion of
+            Left _ -> pure ()
+            Right results ->
+              Tracing.interpTraceT id $ cacheStore cacheKey (encJFromInsOrdHashMap (fmap rfResponse results))
+      sendCompleted (Just requestId)
+
```
lexi-lambda, line 384: This chunk of code seems *incredibly* similar to the corresponding piece of logic in the HTTP module. Could we get some sharing here, or is the duplication incidental?

lexi-lambda (changes requested):

server/src-lib/Hasura/GraphQL/Execute/Prepare.hs
abooij: Done. The `db` type variable is used in a rather minor way in `Execute/Query.hs`.

server/src-lib/Hasura/GraphQL/Execute/Query.hs
abooij: No, it's rather a warning for the next person to touch this code. I've made it into a `WARNING` instead.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
abooij: Well, since we're not doing query plan caching at the moment, its value will always be `Telem.Miss`. But I've now pushed two commits that plumb this constant value through the code :)

server/src-lib/Hasura/GraphQL/Execute/Remote.hs
abooij: You're absolutely right, this method should've been named after its role, not after its implementation.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
abooij: Okay, I've pushed the responsibility of constructing the hashmap down to `runQueryDB` so that it can take a `PreparedSql` argument instead. We can't get rid of the hashmap entirely without further discussion as this is used in the type class `MonadQueryLog` which is also used by other codebases.

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
abooij: Done.

server/src-lib/Hasura/GraphQL/Execute/Mutation.hs
abooij: This isn't new code (it just got moved around), and `rjCtx` is not a new name. But I do agree that it deserves a better name, so I'll do a search/replace for this as part of this PR.

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
abooij: No, you're absolutely right. For what it's worth, there should have already been code sharing before this PR. And I did try to deduplicate some code here. But I got stuck juggling monad stacks. So I didn't manage to do this in a reasonable time.

(Note that there are some mild operational differences: e.g. while the WebSocket code writes the results to the socket, the HTTP code returns it as a Haskell value.)

abooij:
@lexi-lambda Thank you very much for the thorough review!

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
lexi-lambda: How difficult do you think it would be to do that bit of refactoring as part of this PR? If we’re going to do it at some point, now seems as good a time as any. Of course, if it’s really hard, that’s one thing, but it’s not clear to me why it would be.

server/src-lib/Hasura/GraphQL/Transport/HTTP/Protocol.hs
```diff
@@ -99,48 +98,18 @@ encodeGQErr :: Bool -> QErr -> J.Value
 encodeGQErr includeInternal qErr =
   J.object [ "errors" J..= [encodeGQLErr includeInternal qErr]]
 
-data GQResult a
-  = GQSuccess !a
-  | GQPreExecError ![J.Value]
-  | GQExecError ![J.Value]
-  deriving (Show, Eq, Functor, Foldable, Traversable)
+type GQResult a = ExceptT GQExecError Identity a
```
lexi-lambda, line 101: `ExceptT` over `Identity` is basically just an awkward version of `Either`. Is there a reason we need `ExceptT` here specifically?

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
lexi-lambda: Do we ever actually call `logQueryLog` with a map containing more than one entry in it? If not, I’d be in favor of just changing the signature—I don’t think the tiny bit of extra downstream effort to update to the new interface is worth the misleading API. (Of course, if we *do* call it with more than one value somewhere, please point me to it!)

server/src-lib/Hasura/GraphQL/Execute/Query.hs
```diff
@@ -135,56 +130,49 @@ mkCurPlanTx
...
-    return (alias, fldResp)
-
-  (,) <$> mkLazyRespTx env manager reqHdrs userInfo directives resolved <*> pure (mkGeneratedSqlMap resolved)
+    RFPPostgres ps@(PreparedSql q prepMap remoteJoinsM) ->
+      let args = withUserVars (_uiSession userInfo) prepMap
+          -- WARNING: this quietly assumes the intmap keys are contiguous
+          prepArgs = fst <$> IntMap.elems args
+      in (, Just ps) $ case remoteJoinsM of
+           Nothing -> do
+             Tracing.trace "Postgres" . (runExtractProfile ep =<<) . liftTx $ asSingleRowJsonResp (instrument q) prepArgs
```
lexi-lambda, line 145: This is another instance of an operator section involving `=<<`, which I missed on the first pass, but I personally find confusing. There’s a lot going on all on one line here. I recognize this kind of thing is subjective, but I happen to prefer a formulation like this:

```suggestion
             Tracing.trace "Postgres" $ runExtractProfile ep =<< liftTx do
               asSingleRowJsonResp (instrument q) prepArgs
```

This puts all the “wrapping bits” on a separate line from the actual interesting functionality being executed. Another approach I use sometimes when the “wrapping bits” get complicated is something like this:

```suggestion
             asSingleRowJsonResp (instrument q) prepArgs
               &   liftTx
               >>= runExtractProfile ep
               &   Tracing.trace "Postgres"
```

Which has sort of a “pipeline” flavor to it. This works out nicely because `&` and `>>=` are both `infixl 1`. But in this particular case, I’m not sure that it’s actually better; most of the parts of the “pipeline” are pretty boring.

lexi-lambda (changes requested):
Many thanks for the prompt changes; I think this PR generally looks good to me now. I think it would be nice to find some way to deduplicate all the tricky query execution logic and parameterize it over the transport mechanism, but I’m happy to have a call to discuss that, since it seems possibly subtle.

server/src-lib/Hasura/GraphQL/Transport/HTTP/Protocol.hs
nicuveo: Done!

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
nicuveo: We wanted to avoid doing too many changes for pro; since there was no direct need to change `logQueryLog` for the purpose of this PR, we chose not to change anything.
However, while @paf31 agreed with the changes as first done in #5865, we're thinking of doing a second PR to do some cleanup, after this one is submitted and further work on dataloader is unblocked. We could definitely include that change as part of the second PR!

server/src-lib/Hasura/GraphQL/Transport/WebSocket.hs
nicuveo: We're thinking of doing it while working on data-loader, since it will mean changing this logic significantly.

server/src-lib/Hasura/GraphQL/Execute/Query.hs
nicuveo: Done!

server/src-lib/Hasura/GraphQL/Transport/HTTP.hs
nicuveo: After discussion: pushed a change to `MonadQueryLog`.

lexi-lambda (approved):

tirumaraiselvan (approved):
changelog

PR #5928, opened 2020-10-05
server: implement websocket compression setting (fixes #3292)

### LINKED ISSUE #3292: [Question] How activate compression on subscriptions
Hi

I'm starting with Hasura and I found compression for queries over http

https://docs.hasura.io/1.0/graphql/manual/deployment/compression.html

but I don't know if hasura engine has compression for subscriptions enabled over web sockets

If not, how can I enable it?

gahag:
### Description

As discussed, the websocket package has a compression setting built-in.
This commit implements the wiring to allow such configuration through
the `connection-compression` command line flag, or the
`HASURA_GRAPHQL_CONNECTION_COMPRESSION` environment variable.

### Changelog

### Affected components

- [x] Server

### Related Issues

#3292

### Solution and Design

As discussed, the websocket package has a compression setting built-in.
This commit implements the wiring to allow such configuration through
the `connection-compression` command line flag, or the
`HASURA_GRAPHQL_CONNECTION_COMPRESSION` environment variable.

### Steps to test and verify

Run graphql with the compression flag, and send a big but low entropy string through a graphql client.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


abooij (approved):
This looks good to me, thanks for this PR, @gahag!

@0x777, do we want to have some kind of unit test for this? I suppose it might be difficult to integrate a test for this into our testing suite...

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -1010,3 +1023,15 @@ downgradeOptionsParser =
...
+connectionCompressionEnv =
+  ( "HASURA_GRAPHQL_CONNECTION_COMPRESSION"
+  , "Enable Websocket Compression (default: false)"
+  )
+
+parseConnectionCompression :: Parser Bool
+parseConnectionCompression =
+  switch ( long "connection-compression" <>
+           help (snd connectionCompressionEnv)
+         )
```
lexi-lambda, line 1037: I think this PR generally looks great. My only quibble is that I think “connection compression” is pretty vague. I think it would be helpful to be a little more specific:

```suggestion
webSocketCompressionEnv :: (String, String)
webSocketCompressionEnv =
  ( "HASURA_GRAPHQL_WEBSOCKET_COMPRESSION"
  , "Enable WebSocket permessage-deflate compression (default: false)"
  )

parseWebSocketCompression :: Parser Bool
parseWebSocketCompression =
  switch ( long "websocket-compression" <>
           help (snd webSocketCompressionEnv)
         )
```

This way, we’re at least a little more clear about what the compression scheme is, and we’ll have hedged our bets in case some other compression scheme becomes popular in the future.

(Of course, the use sites of these bindings would have to be updated, too, but that would be awkward to do as a GH PR suggestion.)

lexi-lambda (changes requested):

server/src-lib/Hasura/Server/Init.hs
gahag: Fair point! I'll adjust it.

gahag:
@lexi-lambda I've made a new commit with the suggested changes. Thanks for the review!

abooij:
> @lexi-lambda I've made a new commit with the suggested changes. Thanks for the review! Should I squash the two commits?

You can leave the commits as-is, we usually squash the commits in a PR at merge time. Let's give Alexis some more time to have another look.

lexi-lambda (approved):

tirumaraiselvan (approved):
changelog

mehtavishwa30:
Hello @gahag!

Thanks so much for your contribution to Hasura OSS during the Hacktoberfest 2020! We'd love to send you some Hasura swag as a token of appreciation and you will be getting an email with the swag form shortly. ✨

kokroo:
Compression on websockets doesn't seem to actually work anymore. The headers indicate that permessage-deflate is on, but the packets are still not compressed. After checking with wireshark, even with that header present, the frame size is the same as the length of the payload. Manually using the deflate tool shows that the text is 1000% compressible.

PR #5938, opened 2020-10-06
Server: Validate remote schema queries

codingkarthik:
Earlier, the graphql-engine just proxied the remote schema requests to the remote server without validation, the only validation that was done at the root level. This PR changes that behaviour by deeply parsing a query i.e validate every field and argument provided in a remote schema query

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -201,8 +196,52 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+       -- We need to do this because the `SelectionSet`(s) that are
+       -- inputted to this function have the fragments (if any) flattened.
+       -- (Check `flattenSelectionSet` in 'Hasura.GraphQL.Parser.Collect' module)
+       -- This function makes a valid interface query by:
+       -- 1. Getting the common interface fields in all the selection sets
+       -- 2. Remove the common fields obtained in #1 from the selection sets
+       -- 3. Construct a selection field for every common interface field
+       -- 4. Construct inline fragments for non-common interface fields
+       --    using the result of #2 for every object
+       -- 5. Contruct the final selection set by combining #3 and #4
```
abooij, line 212: I think this explanation of the algorithm deserves to be a `Note` as it is fairly involved. It could also benefit from a running example.

Can we organize the code so that its relation with the English algorithm is easier to spot? For instance, since you've numbered the steps in the algorithm, you can refer to those steps in the code.

abooij, line 212:
```suggestion
       --    using the result of #2 for every object
       -- 5. Construct the final selection set by combining #3 and #4
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -201,8 +196,52 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+       let selSetFields' = flip (fmap . fmap . fmap) objNameAndSelSets $
+                             \case
+                               G.SelectionField fld -> Just fld
+                               _                    -> Nothing
+           selSetFields = (fmap . fmap) catMaybes selSetFields'
```
abooij, line 217: Can we rename these two variables to the role that they play in the algorithm, rather than essentially spelling out their type?

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -201,8 +196,52 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+       -- 1. Getting the common interface fields in all the selection sets
+       -- 2. Remove the common fields obtained in #1 from the selection sets
+       -- 3. Construct a selection field for every common interface field
+       -- 4. Construct inline fragments for non-common interface fields
+       --    using the result of #2 for every object
+       -- 5. Contruct the final selection set by combining #3 and #4
+       let selSetFields' = flip (fmap . fmap . fmap) objNameAndSelSets $
+                             \case
+                               G.SelectionField fld -> Just fld
+                               _                    -> Nothing
```
abooij, line 216: This needs some explanation. Why can we ignore all other entries of the selection set?

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -201,8 +196,52 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+             toList $
+             Map.mapMaybe allTheSame $
+             Map.groupOn G._fName $
+             concat $ flip map selSetFields $ \(_, flds) ->
+               filter (flip elem interfaceFieldNames . G._fName) flds
+
+           interfaceFieldNames = map G._fldName fields
+
+           allTheSame [] = Nothing
+           allTheSame (x:xs) = bool Nothing (Just x) $ and $ map (== x) xs
```
abooij, line 231:
```suggestion
           -- common interface fields that exist in every
           -- selection set provided
           commonInterfaceFields =
             Map.elems $
             Map.mapMaybe allTheSame $
             Map.groupOn G._fName $
             concatMap (\(_, flds) ->
               filter ((`elem` interfaceFieldNames) . G._fName) flds) selSetFields

           interfaceFieldNames = map G._fldName fields

           allTheSame [] = Nothing
           allTheSame (x:xs) = bool Nothing (Just x) $ all (== x) xs
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -219,14 +258,24 @@ remoteSchemaUnion
...
 remoteSchemaUnion schemaDoc defn@(G.UnionTypeDefinition description name _directives objectNames) =
   P.memoizeOn 'remoteSchemaObject defn do
-  objDefs <- traverse getObject objectNames
-  objs :: [Parser 'Output n ()] <- traverse (remoteSchemaObject schemaDoc) objDefs
+  objs :: [Parser 'Output n (Name, (SelectionSet NoFragments G.Name))] <-
+    traverse (\objName -> do
+                 obj <- getObject objName
+                 rsObj <- remoteSchemaObject schemaDoc obj
+                 return $ (objName,) <$> rsObj)
+             objectNames
```
abooij, line 269: This code looks very similar to the code that's part of `remoteSchemaInterface`. Can we deduplicate this?

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -219,14 +258,24 @@ remoteSchemaUnion
...
   when (null objs) $
     throw400 RemoteSchemaError $ "List of member types cannot be empty for union type " <> squote name
-  pure $ void $ P.selectionSetUnion name description objs
+  pure $ P.selectionSetUnion name description objs <&>
+    (\objNameAndSelSets ->
+       catMaybes $ flip map objNameAndSelSets $ \(objName, selSet) ->
+        case selSet of
+          [] -> Nothing
+          _  -> Just (G.SelectionInlineFragment $ G.InlineFragment (Just objName) mempty selSet)
+    )
```
abooij, line 278: Could you please add a comment explaining what this code does, and why?

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -362,8 +411,8 @@ argumentsParser
   => G.ArgumentsDefinition
   -> G.SchemaIntrospection
   -> m (InputFieldsParser n ())
-argumentsParser args schemaDoc =
-  sequenceA_ <$> traverse (inputValueDefinitionParser schemaDoc) args
+argumentsParser args schemaDoc = do
+  sequenceA_ <$> flip traverse args (inputValueDefinitionParser schemaDoc)
```
abooij, line 415: `flip traverse = for`
```suggestion
  sequenceA_ <$> for args (inputValueDefinitionParser schemaDoc)
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +425,43 @@ remoteField
...
-      pure $ void $ P.subselection fieldName description argsParser remoteSchemaObj
+      pure $ P.rawSubselection fieldName description argsParser remoteSchemaObj <&>
+        (\(alias, args, _, selSet) ->
+           (G.Field alias fieldName (fmap getName <$> args) mempty selSet))
     G.TypeDefinitionUnion unionTypeDefn -> do
       remoteSchemaObj <- remoteSchemaUnion sdoc unionTypeDefn
-      pure $ void $ P.subselection fieldName description argsParser remoteSchemaObj
+      pure $ P.rawSubselection fieldName description argsParser remoteSchemaObj <&>
+        (\(alias, args, _, selSet) ->
+           (G.Field alias fieldName (fmap getName <$> args) mempty selSet))
```
abooij, line 457: Visually there's some duplication going on. I understand that the details are different between the various cases. But at least some of them use equal anonymous functions, which we can make into local named definitions, for instance. And we either construct the final `FieldParser` using `rawSelection` or `rawSubselection`. So we can have two locally defined combinators instead of calling `rawSelection` and `rawSubselection` directly, which should clean up the code a little bit.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +425,43 @@ remoteField
...
     _ -> throw400 RemoteSchemaError "expected output type, but got input type"
 
 remoteFieldScalarParser
   :: MonadParse n
-  => G.Name
-  -> Maybe G.Description
+  => G.ScalarTypeDefinition
   -> Parser 'Both n ()
-remoteFieldScalarParser name description =
+remoteFieldScalarParser (G.ScalarTypeDefinition description name _) =
```
abooij, line 464: Make explicit what we're ignoring:
```suggestion
remoteFieldScalarParser (G.ScalarTypeDefinition description name _directives) =
```

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
```diff
@@ -178,9 +178,9 @@ data IntrospectionResult
 
 data ParsedIntrospection
   = ParsedIntrospection
-  { piQuery        :: [P.FieldParser (P.ParseT Identity) (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
-  , piMutation     :: Maybe [P.FieldParser (P.ParseT Identity) (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
-  , piSubscription :: Maybe [P.FieldParser (P.ParseT Identity) (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  { piQuery        :: [P.FieldParser (P.ParseT Identity) (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
+  , piMutation     :: Maybe [P.FieldParser (P.ParseT Identity) (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
+  , piSubscription :: Maybe [P.FieldParser (P.ParseT Identity) (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
abooij, line 183: Is this the same? (I might be missing some imports.)
```suggestion
  { piQuery        :: [P.FieldParser (P.ParseT Identity) RemoteField]
  , piMutation     :: Maybe [P.FieldParser (P.ParseT Identity) RemoteField]
  , piSubscription :: Maybe [P.FieldParser (P.ParseT Identity) RemoteField]
```

server/tests-py/graphql_server.py
```diff
@@ -53,6 +55,8 @@ def post(self, request):
 class User(graphene.ObjectType):
     id = graphene.Int()
     username = graphene.String()
+    blah = graphene.String()
```
abooij, line 58: Perhaps the role of this field `blah` can be made more clear by renaming it into something like `generate_error`, so that future readers know why we'd explicitly add a field that cannot be queried.

abooij (changes requested):

server/tests-py/graphql_server.py
codingkarthik: Yes, I'll rename the field to `generateError`.

server/src-lib/Hasura/RQL/Types/SchemaCache.hs
codingkarthik: Thanks for the suggestion, I have applied it.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Yes, you're right. I have applied the suggested change.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Done!

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Sure, I'll add a comment

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -206,7 +206,7 @@ query'
      )
   => HashSet QualifiedTable
   -> [FunctionInfo]
-  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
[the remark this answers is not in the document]
codingkarthik: Yes, it can. I've made the change

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -11,77 +11,63 @@ import           Hasura.Prelude
...
-    makeFieldParser :: G.FieldDefinition -> m (P.FieldParser n (RemoteSchemaInfo, Field NoFragments Variable))
+    makeFieldParser :: G.FieldDefinition -> m (P.FieldParser n (RemoteSchemaInfo, RemoteField))
     makeFieldParser fieldDef = do
-      fp <- remoteField' sdoc fieldDef
-      return $ do
-        raw <- P.unsafeRawField (P.fDefinition fp)
-        return (info, raw)
-    makeParsers :: G.Name -> m [P.FieldParser n (RemoteSchemaInfo, Field NoFragments Variable)]
+      fldParser <- remoteField' sdoc fieldDef
+      pure $ fldParser <&> (\fp -> (info, fp))
```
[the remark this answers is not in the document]
codingkarthik: Applied the suggestion!

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -187,12 +178,16 @@ remoteSchemaInterface
...
   P.memoizeOn 'remoteSchemaObject defn do
   subFieldParsers <- traverse (remoteField' schemaDoc) fields
-  objs :: [Parser 'Output n ()] <-
-    traverse (getObject >=> remoteSchemaObject schemaDoc) possibleTypes
+  objs :: [Parser 'Output n (Name, (SelectionSet NoFragments G.Name))] <-
+    traverse (\objName -> do
+                 obj <- getObject objName
+                 rsObj <- remoteSchemaObject schemaDoc obj
+                 return $ (objName,) <$> rsObj)
+             possibleTypes
```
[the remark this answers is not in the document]
codingkarthik: Sure, I'll do that

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -201,8 +196,52 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+               let nonCommonFields = filter (not . flip elem commonInterfaceFields) flds
+               in mkObjInlineFragment (objName, map G.SelectionField nonCommonFields)
+
+           mkObjInlineFragment (_, []) = Nothing
+           mkObjInlineFragment (objName, selSet) =
+             Just $ G.SelectionInlineFragment $
+               G.InlineFragment (Just objName) mempty selSet
+
+       in (map G.SelectionField commonInterfaceFields) <> nonCommonInterfaceFields
+    )
```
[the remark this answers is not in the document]
codingkarthik: Yes, I should have done that in the first place. I'll make the change

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Thanks for the suggestion, I've applied this.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Done!

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: I've added a comment explaining the same

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Done

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Done!

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -181,18 +169,87 @@ remoteSchemaObject schemaDoc defn@(G.ObjectTypeDefinition description name inter
...
+After parsing, the above query, the selection set is modified to:
+
+{
+   hero {
+     id
+     name
+     primaryFunction
+     homePlanet
+   }
+}
```
abooij, line 230: This isn't quite right, and has to be phrased more carefully, I think.

1. I don't think any code in the current PR would ever see a selection set with 4 fields as in the example above.
2. Selection sets aren't mutable, so we can't say it's "modified". You can say that some value is returned by some parser, but then it'd be valuable to specify which one.

The point is that when parsing the interface selection set for the `hero` field, we parse the selection set twice: once for the `Droid` object type, which would be passed a selection set containing the fields `id`, `name`, and `primaryFunction`, and once for the `Human` object type, which would be passed a selection set containing the fields `id`, `name` and `homePlanet`. The result of the interface selection set parsing would then be the results of the parsing of the object types when passed their corresponding _restricted_ (and flattened) selection sets. So `Droid` might wish to report a selection set of `{id name primaryFunction}`, and `Human` might report a selection set of `{id name homePlanet}`.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,80 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
-  objDefs <- traverse getObject objectNames
-  objs :: [Parser 'Output n ()] <- traverse (remoteSchemaObject schemaDoc) objDefs
+  objs :: [Parser 'Output n (Name, (SelectionSet NoFragments G.Name))] <-
+    traverse (getObjectParser schemaDoc getObject) objectNames
   when (null objs) $
     throw400 RemoteSchemaError $ "List of member types cannot be empty for union type " <> squote name
-  pure $ void $ P.selectionSetUnion name description objs
+  pure $ P.selectionSetUnion name description objs <&>
+    (\objNameAndSelSets ->
+       catMaybes $ flip map objNameAndSelSets $ \(objName, selSet) ->
```
abooij, line 338:
```suggestion
       catMaybes $ objNameAndSelSets <&> \(objName, selSet) ->
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,80 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+        case selSet of
+          -- In the case of an incoming query containing an union type, which has
+          -- 3 objects in it's object member types, but an incoming query only
```
abooij, line 341:
```suggestion
          -- 3 objects in its object member types, but an incoming query only
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,80 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+          -- queries 2 out of the 3 objects using fragments, then we get the
+          -- selectionSet for the object that was not queried as `[]`. So,
+          -- we omit it while constructing the query
```
abooij, line 344: I had to read this a few times before I was convinced I got the point. Let me try writing this as well.

> The return value obtained from the parsing of a union selection set specifies, for each object in the union type, a fragment-free selection set for that object type. In particular, if, for a given object type, the selection set passed to the union type did not specify any fields for that object type (i.e. if no inline fragment applied to that object), the selection set resulting from the parsing through that object type would be empty, i.e. `[]`. We exclude such object types from the reconstructed selection set for the union type, as selection sets cannot be empty.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +493,63 @@ remoteField
...
+      -> (Parser 'Both n ())
+      -> m (FieldParser n (Field NoFragments G.Name))
+    mkFieldParserWithoutSelectionSet fldName desc argsParser outputParser =
+      let selectionFldParser = P.rawSelection fldName desc argsParser outputParser
+      in
+        pure $
+          (selectionFldParser
+           <&>
+           (\(alias, args, _)
+              -> (G.Field alias fieldName (fmap getName <$> args) mempty [])))
```
abooij, line 531: Since this is always `pure`, rather make the function `mkFieldParserWithoutSelectionSet` itself pure and use `<&>` rather than `>>=` above in the 2 cases.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +493,63 @@ remoteField
...
+      -> (Parser 'Output n (SelectionSet NoFragments G.Name))
+      -> m (FieldParser n (Field NoFragments G.Name))
+    mkFieldParserWithSelectionSet fldName desc argsParser outputParser =
+      let selectionFldParser = P.rawSubselection fldName desc argsParser outputParser
+      in
+        pure $
+          (selectionFldParser
+           <&>
+           (\(alias, args, _, selSet)
+              -> (G.Field alias fieldName (fmap getName <$> args) mempty selSet)))
```
abooij, line 546: Ditto, let's make `mkFieldParserWithSelectionSet` pure and use `<&>` rather than `>>=` in the above 3 cases.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
abooij: The `let ... in ...` syntax doesn't add much to the readability for me. If you want to clarify why you're using `P.rawSelection`, you can write inline comments, even in the middle of an expression.

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -302,7 +302,7 @@ query
   => G.Name
   -> HashSet QualifiedTable
   -> [FunctionInfo]
-  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
abooij, line 305:
```suggestion
  -> [P.FieldParser n RemoteField]
```

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -402,8 +402,8 @@ queryWithIntrospection
      )
   => HashSet QualifiedTable
   -> [FunctionInfo]
-  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
-  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
+  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
abooij, line 406:
```suggestion
  -> [P.FieldParser n RemoteField]
  -> [P.FieldParser n RemoteField]
```

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -464,8 +464,8 @@ unauthenticatedQueryWithIntrospection
    . ( MonadSchema n m
      , MonadError QErr m
      )
-  => [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
-  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  => [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
+  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
abooij, line 468:
```suggestion
  => [P.FieldParser n RemoteField]
  -> [P.FieldParser n RemoteField]
```

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -477,7 +477,7 @@ mutation
   :: forall m n r
    . (MonadSchema n m, MonadTableInfo r m, MonadRole r m, Has QueryContext r, Has Scenario r)
   => HashSet QualifiedTable
-  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  -> [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
abooij, line 480:
```suggestion
  -> [P.FieldParser n RemoteField]
```

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -556,7 +556,7 @@ mutation allTables allRemotes allActions nonObjectCustomTypes = do
 unauthenticatedMutation
   :: forall n m
    . (MonadError QErr m, MonadParse n)
-  => [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments P.Variable)]
+  => [P.FieldParser n (RemoteSchemaInfo, G.Field G.NoFragments G.Name)]
```
abooij, line 559:
```suggestion
  => [P.FieldParser n RemoteField]
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -11,77 +10,61 @@ import           Hasura.Prelude
...
+       , Maybe [P.FieldParser n RemoteField]
+       , Maybe [P.FieldParser n RemoteField])
 buildRemoteParser (IntrospectionResult sdoc query_root mutation_root subscription_root) info = do
   queryT <- makeParsers query_root
   mutationT <- traverse makeParsers mutation_root
   subscriptionT <- traverse makeParsers subscription_root
   return (queryT, mutationT, subscriptionT)
   where
-    makeFieldParser :: G.FieldDefinition -> m (P.FieldParser n (RemoteSchemaInfo, Field NoFragments Variable))
+    makeFieldParser :: G.FieldDefinition -> m (P.FieldParser n (RemoteSchemaInfo, (Field NoFragments G.Name)))
```
abooij, line 39:
```suggestion
    makeFieldParser :: G.FieldDefinition -> m (P.FieldParser n RemoteField))
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -11,77 +10,61 @@ import           Hasura.Prelude
...
     makeFieldParser fieldDef = do
-      fp <- remoteField' sdoc fieldDef
-      return $ do
-        raw <- P.unsafeRawField (P.fDefinition fp)
-        return (info, raw)
-    makeParsers :: G.Name -> m [P.FieldParser n (RemoteSchemaInfo, Field NoFragments Variable)]
+      fldParser <- remoteField' sdoc fieldDef
+      pure $ (info, ) <$> fldParser
+    makeParsers :: G.Name -> m [P.FieldParser n (RemoteSchemaInfo, (Field NoFragments G.Name))]
```
abooij, line 43:
```suggestion
    makeParsers :: G.Name -> m [P.FieldParser n RemoteField]
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -181,18 +169,87 @@ remoteSchemaObject schemaDoc defn@(G.ObjectTypeDefinition description name inter
...
+--   'remoteSchemaObject' functions. Both of these have a slightly
+--   different implementation of 'getObject', which is the
+--   reason 'getObject' is an argument to this function
+getObjectParser
+  :: forall n m
+  . (MonadSchema n m, MonadError QErr m)
+  => SchemaIntrospection
+  -> (G.Name -> m G.ObjectTypeDefinition)
+  -> G.Name
+  -> m (Parser 'Output n (Name, (SelectionSet NoFragments G.Name)))
```
abooij, line 183:
```suggestion
  -> m (Parser 'Output n (Name, SelectionSet NoFragments G.Name))
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,80 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
           Just _  -> throw400 RemoteSchemaError $ "Interface type " <> squote name <>
             " can only include object types. It cannot include " <> squote objectName
 
+    -- 'constructInterfaceQuery' constructs a remote interface query.
+    constructInterfaceSelectionSet
+      :: [(G.Name, SelectionSet NoFragments G.Name)]
+      -> (SelectionSet NoFragments G.Name)
```
abooij, line 276:
```suggestion
      -> SelectionSet NoFragments G.Name
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,80 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
   . (MonadSchema n m, MonadError QErr m)
   => SchemaIntrospection
   -> G.UnionTypeDefinition
-  -> m (Parser 'Output n ())
+  -> m (Parser 'Output n (SelectionSet NoFragments G.Name))
 remoteSchemaUnion schemaDoc defn@(G.UnionTypeDefinition description name _directives objectNames) =
   P.memoizeOn 'remoteSchemaObject defn do
-  objDefs <- traverse getObject objectNames
-  objs :: [Parser 'Output n ()] <- traverse (remoteSchemaObject schemaDoc) objDefs
+  objs :: [Parser 'Output n (Name, (SelectionSet NoFragments G.Name))] <-
```
abooij, line 332:
```suggestion
  objs :: [Parser 'Output n (Name, SelectionSet NoFragments G.Name)] <-
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +493,63 @@ remoteField
...
+      remoteSchemaUnion sdoc unionTypeDefn >>=
+        mkFieldParserWithSelectionSet fieldName description argsParser
     _ -> throw400 RemoteSchemaError "expected output type, but got input type"
+  where
+    mkFieldParserWithoutSelectionSet
+      :: G.Name
+      -> (Maybe Description)
+      -> (InputFieldsParser n ())
+      -> (Parser 'Both n ())
+      -> m (FieldParser n (Field NoFragments G.Name))
```
abooij, line 523:
```suggestion
      :: G.Name
      -> Maybe Description
      -> InputFieldsParser n ()
      -> Parser 'Both n ()
      -> m (FieldParser n (Field NoFragments G.Name))
```

abooij (changes requested):
Hi @codingkarthik, thanks for your improvements. The PR looks better. We're getting close, I think.

I am adding a few more items to discuss. It seems to me that this code will be playing an increasingly important role, and so I think it would be valuable if we go the extra mile with this PR.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +219,86 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+          selSetFieldsWithoutFragments = (fmap . fmap) catMaybes selSetFieldsWithoutFragments'
+
+          -- common interface fields that exist in every
+          -- selection set provided
+          commonInterfaceFields =
+            Map.elems $
+            Map.mapMaybe allTheSame $
+            Map.groupOn G._fName $
+            concatMap (\(_, flds) ->
+                         filter ((`elem` interfaceFieldNames) . G._fName) flds)
```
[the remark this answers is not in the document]
codingkarthik: Done

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +219,86 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+                         $ selSetFieldsWithoutFragments
+
+          interfaceFieldNames = map G._fldName fields
+
+          allTheSame [] = Nothing
+          allTheSame (x:xs) = bool Nothing (Just x) $ all (== x) xs
```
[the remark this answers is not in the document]
codingkarthik: Done

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Yes, that's right. Also, Your explanation is much better. I'll use your explanation in the code.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Applied suggestion

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Removing the type annotation from here, because moved the logic to `getObject` function

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +219,86 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+                -- We're omitting a selection field with fragments
+                -- here because the `SelectionSet` we recieve is
+                -- a "flattened" selection set without any fragments
+                -- (atleast at the top level of the selection set).
+                -- This is needed to extract the common interface
+                -- fields from the various objects that have been
+                -- queried.
+                G.SelectionField fld -> Just fld
+                _                    -> Nothing
+          selSetFieldsWithoutFragments = (fmap . fmap) catMaybes selSetFieldsWithoutFragments'
```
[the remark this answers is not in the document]
codingkarthik: Thank you for this suggestion, indeed it avoids the partiality and the code is cleaner now!

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,71 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+
+          interfaceFieldNames = map G._fldName fields
+
+          allTheSame (x:xs) | all (== x) xs = Just x
+          allTheSame _ = Nothing
+
+          -- #2 of Note [Querying remote schema interface fields]
+          nonCommonInterfaceFields =
+            catMaybes $ flip map objNameAndFields $ \(objName, flds) ->
+              let nonCommonFields = filter (not . flip elem commonInterfaceFields) flds
```
abooij, line 297:
```suggestion
            catMaybes $ flip map objNameAndFields $ \(objName, fields) ->
              let nonCommonFields = filter (not . flip elem commonInterfaceFields) fields
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -213,20 +270,71 @@ remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name
...
+          -- object type, the selection set passed to the union type did not
+          -- specify any fields for that object type (i.e. if no inline fragment
+          -- applied to that object), the selection set resulting from the parsing
+          -- through that object type would be empty, i.e. []. We exclude such
+          -- object types from the reconstructed selection set for the union
+          -- type, as selection sets cannot be empty.
+          [] -> Nothing
+          _  ->
+            Just (G.SelectionInlineFragment
+                    $ G.InlineFragment (Just objName) mempty $ fmap G.SelectionField flds))
```
abooij, line 337:
```suggestion
       catMaybes $ objNameAndSelSets <&> \(objName, fields) ->
        case fields of
          -- The return value obtained from the parsing of a union selection set
          -- specifies, for each object in the union type, a fragment-free
          -- selection set for that object type. In particular, if, for a given
          -- object type, the selection set passed to the union type did not
          -- specify any fields for that object type (i.e. if no inline fragment
          -- applied to that object), the selection set resulting from the parsing
          -- through that object type would be empty, i.e. []. We exclude such
          -- object types from the reconstructed selection set for the union
          -- type, as selection sets cannot be empty.
          [] -> Nothing
          _  ->
            Just (G.SelectionInlineFragment
                    $ G.InlineFragment (Just objName) mempty $ fmap G.SelectionField fields))
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Thank you for the explanation, I was a bit confused because I did a `Debug.trace` on the output of `P.selectionSetInterface` and thought it returns a selection set with the 4 fields as I mentioned earlier. I've modified the note based on your explanation.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +484,57 @@ remoteField
...
+        mkFieldParserWithSelectionSet fieldName description argsParser
     _ -> throw400 RemoteSchemaError "expected output type, but got input type"
+  where
+    mkFieldParserWithoutSelectionSet
+      :: G.Name
+      -> Maybe Description
+      -> InputFieldsParser n ()
+      -> Parser 'Both n ()
+      -> FieldParser n (Field NoFragments G.Name)
+    mkFieldParserWithoutSelectionSet fldName desc argsParser outputParser =
```
abooij, line 518: Since `fldName` and `desc` are just `fieldName` and `description`, which are in scope, we can use `fieldName` and `description` instead, instead of passing them as arguments.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
```diff
@@ -376,32 +484,57 @@ remoteField
...
+      P.rawSelection fldName desc argsParser outputParser
+      <&> (\(alias, args, _) -> (G.Field alias fieldName (fmap getName <$> args) mempty []))
+
+    mkFieldParserWithSelectionSet
+      :: G.Name
+      -> Maybe Description
+      -> InputFieldsParser n ()
+      -> Parser 'Output n (SelectionSet NoFragments G.Name)
+      -> FieldParser n (Field NoFragments G.Name)
+    mkFieldParserWithSelectionSet fldName desc argsParser outputParser =
```
abooij, line 528: ditto

abooij (commented):
Some minor code cleanups.

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: I was about to use `fields` too but it clashes with the `fields` defined in the arguments defined in 

```haskell

remoteSchemaInterface
  :: forall n m
  . (MonadSchema n m, MonadError QErr m)
  => SchemaIntrospection
  -> G.InterfaceTypeDefinition [G.Name]
  -> m (Parser 'Output n (G.SelectionSet NoFragments G.Name))
remoteSchemaInterface schemaDoc defn@(G.InterfaceTypeDefinition description name _directives fields possibleTypes) =
```

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Yes, I'll make the change

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: done

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Changing `flds` to `objFields`

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
codingkarthik: Done

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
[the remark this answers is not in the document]
codingkarthik: Done

server/src-lib/Hasura/GraphQL/Schema/Remote.hs
abooij: Ah, I see :)

abooij (approved):
Awesome. Thanks for all the improvements, @codingkarthik!

codingkarthik:
Thank you so much @abooij for the review and for bearing with me :)

tirumaraiselvan (approved):
changelog

PR #5945, opened 2020-10-07
server: fix issue with tracking custom functions that return `SETOF` materialized view (close #5294)

### LINKED ISSUE #5294: Bug: cannot track a custom function which returns materialized view
I have a materialized view called `course_info_table`. This view is tracked by Hasura. I've created a custom function as follows:

```sql
CREATE FUNCTION search_course_info(query text)
RETURNS SETOF course_info_table AS $$
    SELECT *
    FROM course_info_table
    WHERE info @@ websearch_to_tsquery('english', query)
    ORDER BY ts_rank(info, websearch_to_tsquery('english', query)) DESC
$$ language sql stable;
```

When I attempt to track this function in Hasura, I get the following error:
```json
[
    {
        "definition": {
            "schema": "public",
            "name": "search_course_info"
        },
        "reason": "in function \"search_course_info\": the function \"search_course_info\" cannot be tracked because the function does not return a SETOF table",
        "type": "function"
    }
]
```

I'm running Hasura v1.2.2 via Docker.

kolharsam:
fixes #5294 

### Description
This PR adds a fix to help track custom functions that return a `setof` elements from a materialized view.

### Affected components

- [x] Server
- [x] Tests 

### Steps to test and verify

- Create a Materialized View
- Create a custom function that returns `setof` materialized view
- Click on track function on `Data` page to see if this custom function is tracked.
- Run Graphql queries from GraphiQL editor to verify,.

#### Catalog upgrade

- [x] Yes

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes
- [x] No Breaking changes


sordina:
Could the two migrations be combined?

server/src-lib/Hasura/SQL/Types.hs
```diff
@@ -195,30 +195,34 @@ instance ToTxt TableName where
...
 tableTyToTxt :: TableType -> T.Text
-tableTyToTxt TTBaseTable      = "BASE TABLE"
-tableTyToTxt TTView           = "VIEW"
-tableTyToTxt TTForeignTable   = "FOREIGN TABLE"
-tableTyToTxt TTLocalTemporary = "LOCAL TEMPORARY"
+tableTyToTxt TTBaseTable        = "BASE TABLE"
+tableTyToTxt TTView             = "VIEW"
+tableTyToTxt TTMaterializedView = "MATERIALIZED VIEW"
+tableTyToTxt TTForeignTable     = "FOREIGN TABLE"
+tableTyToTxt TTLocalTemporary   = "LOCAL TEMPORARY"
```
sordina, line 208: Generally good to try to avoid fixing alignment in a PR so that it's obvious what the semantic changes are. No big deal, but just harder to skim the changes.

sordina (commented):
Is the change actually restricted to setof return types, or materialised views in general?

kolharsam:
> Is the change actually restricted to setof return types, or materialised views in general?

setof return types. So for custom functions, the return type has to be table type. So yeah, that would include Tables, Views and materialized views

server/src-lib/Hasura/SQL/Types.hs
kolharsam: noted!

server/tests-py/queries/graphql_query/functions/teardown.yaml
```diff
@@ -9,4 +9,6 @@ args:
       DROP TABLE integer_column cascade;
       DROP TABLE "user" cascade;
       DROP TABLE text_result cascade;
+      DROP TABLE author cascade;
+      DROP MATERIALIZED VIEW IF EXISTS author_mat_view cascade;
```
codingkarthik, line 13: I think it would be better if we don't have `IF EXISTS` here

codingkarthik: A minor change suggested, it's up to you to include it.

tirumaraiselvan (approved):
changelog approved

rakeshkky (approved):
server LGTM

reinoldus:
Is there a docker container available with this fix? Or would I have to build it myself, if I need this fix?

kolharsam:
@reinoldus there is a built image available [here](https://193077-137724480-gh.circle-artifacts.com/0/server/image.tar) if you'd like to use it. You could also build the latest from the `master` branch to use this fix.

PR #5957, opened 2020-10-08
server: hlint the Haskell code as part of the CI

abooij:
### Description

We've previously attempted to improve the coding style of the Haskell backend. #2175 introduced a code style guide, which we are partially starting to follow. #4372 proposed to add an hlint configuration file and apply a bunch of the hints obtained through hlint (although, being a large PR, this never got merged).

This PR is another attempt at making hlint a regular part of our toolbox. It creates an hlint configuration file, and switches off the vast majority of hints, so that only very few hints are enforced. In fact, I have simply disabled hints until none were left, without changing any Haskell code. It also adds a Github Action workflow that runs hlint whenever a PR is created. Finally, it adds an hlint pass to `scripts/dev.sh`, so that this can be run easily without memorizing how to invoke hlint.

The reason that the vast majority of hints is switched off is that it means we can make hlint a part of our CI without having to discuss exactly which hints we want to enforce. Additionally, we don't have to go through the codebase and make any changes. The only changes to Haskell code in this PR disable hlint entirely in certain places where its Haskell code parser disagrees with the one from GHC, causing unavoidable warnings. (In particular this happens when a multi-line GHC pragma is used, with `#-}` on the final line, which hlint erroneously interprets as a CPP pragma.)

The Github Action workflow is set up so that hlint is only run against files that were actually changed by the PR. This should avoid a PR from being blocked because code linting fails in an unrelated Haskell file.

Additionally, the workflow is set up so that when any hints are found, they are shown in the "Files changed" tab, inline with the code, as follows:

Impression of an inline hlint warning

As you can see, the hlint output is made visible inline, giving immediate feedback on the code, and a proposal on how to fix it.

My vision would be that we can tighten up the hlint rules after merging this PR. So the idea is that we first merge this PR, and then have various discussions on which hints we want to enable. As part of enabling hints, we'd need to (eventually) fix the warnings given by hlint.

Another future possibility is to write our own hints, which hlint supports as well: so for specific modules, such as `Hasura.GraphQL.Schema`, which follow certain patterns, we can make sure the logic is written a certain way.

This PR does not check code formatting, which is normally fixed by stylish-haskell. One future possibility would be to add a CI phase that either runs stylish-haskell on the codebase directly, or that checks that running stylish-haskell on the codebase doesn't change anything.

### Changelog

### Affected components

- [x] Server
- [x] Build System
- [x] Tests

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


0x777 (approved):
LGTM

tirumaraiselvan (approved):

PR #5998, opened 2020-10-13
console: fix dropping Postgres dependencies (close #5993)

### LINKED ISSUE #5993: dropping Postgres dependencies results in invalid SQL error
When trying to drop Postgres dependant objects, `CASCADE` is being added to all the _up queries_. In some cases, it results in invalid SQL. 

### Steps to reproduce

1. Create table `users` with columns `id` (PK) and e.g. `id2`.
2. Create table `posts` with columns `id` (PK), `user_id`.
3. Add foreign key relation: `posts.user_id -> users.id`.
4. Go to `users` table, try to modify primary key by, for example, setting `id2` as a new primary key.
5. There will be an error notification with an option to drop Postgres dependant objects:
Screenshot 2020-10-12 at 20 05 53

6. Clicking `Continue` will result in SQL error:

Screenshot 2020-10-12 at 20 06 05

### The issue

In this case, it happens because when we try to modify a primary key we create two SQL queries:
1. `drop constraint <old_PK ...>`,
2. `add constraint <new_PK ...>`,

and current _cascade_ logic adds `CASCADE` to all the queries. In this case `add constraint <new_PK ...> CASCADE` results in an error.

---

This is one place where I noticed it, but the same bug may occur in other places as well. Solving the issue would mean checking all the possible places where the bug happens and fixing them.

soorajshankar:
### Description
when `bulk_queries` were called and one of the query was returning postgres error `2BP01`, upon user retry the console was trying to add `cascade` on all the up SQL queries 

### Affected components
- [x] Console

### Related Issues
close #5993 

### Solution and Design
added `is cascadable` check before adding `cascade`  to SQL query.

### Steps to reproduce

1. Create table `users` with columns `id` (PK) and e.g. `id2`.
2. Create table `posts` with columns `id` (PK), `user_id`.
3. Add foreign key relation: `posts.user_id -> users.id`.
4. Go to `users` table, try to modify primary key by, for example, setting `id2` as a new primary key.
5. There will be an error notification with an option to drop Postgres dependant objects:
Screenshot 2020-10-12 at 20 05 53

6. Clicking `Continue` should change the primary key

The solution based on the assumption that cascade is added only with queries that contain ` drop ` command (from SQL that has ` drop ` with spaces on both sides). Any other scenarios might require `cascade` will be ignored with this code.


beerose (approved):
Console

tirumaraiselvan (approved):
changelog not required

PR #6017, opened 2020-10-14
clean SQL text functions

nicuveo:
### Description

This PR cleans text manipulation functions out of the SQL directory, and merges them with the ones defined in `Data.Text.Extended`. It deletes the `DQuote` typeclass, in favour of the already redundant `ToTxt`. It also cleans some existing code, by making use of already exising functions such as `dquoteList`, and by introducing `commaSeparated`.


0x777:
@nicuveo Why not move those functions to `Data.Text.Extended` instead of  `SQL.Text`? Are they in any way tied to `SQL`?

nicuveo:
They conflict with `Data.Text.Extended`: same names, different types... They are used exclusively for SQL generation, hence their presence in `SQL`. I would definitely be in favour of replacing the ones in Data.Text.Extended by those ones and get rid of `SQL.Text` entirely. WYT?

0x777:
> They conflict with `Data.Text.Extended`: same names, different types... They are used exclusively for SQL generation, hence their presence in `SQL`. I would definitely be in favour of replacing the ones in Data.Text.Extended by those ones and get rid of `SQL.Text` entirely. WYT?

Get rid of `SQL.Text`!

nicuveo:
The only problem is the instances for GraphQL types; they don't really belong in `Data.Extended`... I could keep them as orphan instances in `SQL.Types`, but... that feels wrong. :/

0x777:
> they don't really belong in `Data.Extended`

Why not? Since `Data.Text.Extended` is introducing a typeclass `ToTxt`, it seems reasonable for the instances to go in that module?

nicuveo:
Ah ok! My understanding was that stuff in `Data` was to belong as if it could be a standalone library. But if not, yeah, of course, that works, will do!

0x777:
We can even use `TextShow` from [this](https://hackage.haskell.org/package/text-show) package and avoid defining our own `ToTxt` typeclass? That way the instances for graphql types for `TextShow` type will go into graphql-parser library. I guess it depends on what the semantics of `ToTxt` typeclass are?

nicuveo:
That seems like a good idea! I'll just need to patch the parser first. But sounds good to me!

nicuveo:
Done! It's a bit verbose, but at least that's out of the way now. :)

nicuveo:
Pushed a few more things: introduced `commaSeparated`, and replaced several instances of `map dquote . toList` by the appropriate text function.

nicuveo:
Fixed the kerfuffle re: `TextShow`. This should be better!

0x777 (approved):

tirumaraiselvan (approved):
no cl required

PR #6021, opened 2020-10-14
Create remote schema permissions

abhi40308:
### Description

Permissions UI for Remote Schemas

### CHANGELOG (TODO)


kolharsam (commented):
@abhi40308 these were some of the points that I had, which I thought could make this much better! :)

PR #6003, opened 2020-10-13
server: SQL that grows

abooij:
### Description

We are starting to work towards implementing solid support for different database backends beyond Postgres. To this end, we are transforming the Haskell module `Hasura.RQL` to play the role of an intermediate representation, as follows:

```
GraphQL -> RQL -> SQL
```

Different databases support a different flavor of SQL. This means that some RQL representations cannot easily be executed on MySQL, for instance. To this end, this PR adds a type parameter to some RQL types, so that we can specialize RQL to the various backends. This type parameter is referred to as an _extension description parameter_ in the Trees That Grow (Najd&Jones 2016) paper.

This PR does not add the parameter throughout the RQL modules, and there may be various places where code is left too general. This PR should be seen as an initial pass at implementing the ideas of "Trees That Grow" in our RQL modules.

The PR adds type parameters (of kind `Backend`, defined in `Hasura.SQL.Backend`) to some types. Some existing code is already backend-agnostic in the sense that it can accept any value for this type parameter. Other code works only for the few things that have been specialized to Postgres, such as the `PGScalarType` being replaced by a type family `ScalarType`, with `Scalartype 'Postgres = PGScalarType`. In the latter case, this PR specifies `'Postgres` as a type parameter value.

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR. If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [x] Server

#### Catalog upgrade

Does this PR change Hasura Catalog version?
- [x] No

#### Metadata

Does this PR add a new Metadata feature?
- [x] No

#### GraphQL
- [x] No new GraphQL schema is generated

#### Breaking changes

- [x] No Breaking changes

PR #6059, opened 2020-10-22
server: make more use of hlint

abooij:
### Description

#5957 added an hlint code quality checking phase to the CI. However, as described in that PR, we left many hints disabled, as they may be considered controversial. This PR re-enables a number of hints, hopefully in a non-controversial way. Future PRs will be checked against them.

Using the hlint "refactor" tool, I have also applied many of the changes suggested by hlint. However, this tool needs to be babysit somewhat: it can generate code that does not compile, such as the following instance of the "Redundant bracket" hint:
```haskell
-buildPostgresSpecs :: (HasVersion) => RawConnInfo -> IO Spec
+buildPostgresSpecs :: (HasVersion => RawConnInfo -> IO Spec
 buildPostgresSpecs pgConnOptions = do
   env <- getEnvironment
```
and it can remove comments, such as the following instance of the "Functor law" hint:
```haskell
         runExceptT
         . withExceptT qeCode -- just look at Code for purposes of tests
-        . fmap _uiRole -- just look at RoleName for purposes of tests
-        . fmap fst -- disregard Nothing expiration
-        . getUserInfoWithExpTime_ userInfoFromAuthHook processJwt () () rawHeaders
+        . fmap (_uiRole . fst) . getUserInfoWithExpTime_ userInfoFromAuthHook processJwt () () rawHeaders
         where
           -- mock authorization callbacks:
```
(lines starting with `-` are removed by hlint; lines starting with `+` are added)

I have gone through the code after hlint's refactoring was applied, to make sure not too strange code is generated. However, please do feel free to scrutinize the changes introduced by this PR. (Actually, the first commit in this PR, e09db0c, which only enables hlint hints, does *not* pass CI. This is due to a mistake in the way the hlint script is run, and I have implemented a fix for this as part of #6003 where I already encountered this.)

As a reminder, the fact that this PR both enables hints and implements them is a _choice_: #5957 only runs hlint on files that were changed by a PR.

### Changelog

### Affected components

- [x] Server
- [x] Tests

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


nicuveo:
This PR should not touch `bench-wrk`, which is not code that we own.

server/src-lib/Hasura/RQL/DML/Returning.hs
```diff
@@ -135,7 +135,7 @@ mkMutationOutputExp qt allCols preCalAffRows cte mutOutput strfyNum =
     mutationResultAlias = Iden $ snakeCaseQualObject qt <> "__mutation_result_alias"
     allColumnsAlias = Iden $ snakeCaseQualObject qt <> "__all_columns_alias"
     allColumnsSelect = S.CTESelect $ S.mkSelect
-                       { S.selExtr = map S.mkExtr $ map pgiColumn $ sortCols allCols
+                       { S.selExtr = map (S.mkExtr . pgiColumn) (sortCols allCols)
```
0x777, line 138: Why not `map (S.mkExtr . pgiColumn) $ sortCols allCols`?

server/src-lib/Hasura/Server/Init.hs
```diff
@@ -47,7 +47,7 @@ newtype PGVersion = PGVersion { unPGVersion :: Int } deriving (Show, Eq, J.ToJSO
 
 getDbId :: Q.TxE QErr Text
 getDbId =
-  (runIdentity . Q.getRow) <$>
+  runIdentity . Q.getRow <$>
```
0x777, line 50: I prefer the original in terms of readability but I don't feel particularly strongly about it.

server/src-lib/Hasura/RQL/DML/Returning.hs
abooij: Ah, well hlint doesn't have a clear preference for for `$` or `()`. From our point of view, I agree that indeed perhaps that reads more nicely. So I'll implement this suggestion.

server/src-lib/Hasura/Server/Init.hs
abooij: I think hlint's reasoning is that if you shouldn't use parentheses if you don't need them. This happens when the operators automatically bind as desired.

abooij: As discussed in our meeting today, I reinstated these particular parentheses and disabled the hint for now until we make up our mind about a more precise plan.

0x777 (approved):

tirumaraiselvan (approved):

PR #5070, opened 2020-06-12
console: notifications feature

kolharsam:
### Description
Add notifications to console

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR. If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [x] Console

### Related Issues

builds on #4583. should resolve https://github.com/hasura/graphql-engine-internal/issues/302 eventually.

PR #6067, opened 2020-10-23
console: notifications bug fixes

kolharsam:
### Description

This PR fixes some issues with #5070 for cloud and pro usage. There are also some additional types added related to the ReduxState which are kinda related to these fixes.

### Affected components

- [x] Console

### Related issues

https://github.com/hasura/graphql-engine-pro/pull/600


console/src/components/Main/Main.js
```diff
@@ -35,10 +35,45 @@ import {
...
+  getUserType,
 } from './utils';
 import { getSchemaBaseRoute } from '../Common/utils/routesUtils';
 import LoveSection from './LoveSection';
 import { Help, ProPopup } from './components/';
+import { UPDATE_DATA_HEADERS } from '../Services/Data/DataActions';
+import { HASURA_COLLABORATOR_TOKEN } from '../../constants';
+import { UPDATE_CONSOLE_NOTIFICATIONS } from '../../telemetry/Actions';
+
+const updateRequestHeaders = props => {
```
karthikvt26, line 47: Should notification be refreshed whenever any header is changed in the API explorer tab? The state at `state.apiexplorer.displayedApi.request.headers` would be updated on any change in the `GraphiQL` header tab. The notification section should be dependent on the headers (stored in `~table.dataHeaders`) IMO.

kolharsam: Yeah, fair point. Since these changes are ending up on Cloud and/or Pro and since `state.table.dataHeaders` has the collaborator token already. I can make this change.

kolharsam: @karthikvt26, I have pushed the changes related to this. You can take another look.

console/src/components/Main/Main.js
```diff
@@ -35,10 +35,40 @@ import {
...
+  const { requestHeaders, dispatch } = props;
+
+  const collabTokenPresent = Object.keys(requestHeaders).find(
+    hdr => hdr.toLowerCase() === HASURA_COLLABORATOR_TOKEN
+  );
+
+  if (collabTokenPresent) {
+    const userID = getUserType(collabTokenPresent.value);
```
karthikvt26, line 54: This line is causing a crash probably due to `collabTokenPresent` being a string object and `.value` doesn't exist

console/src/components/Main/Main.js
```diff
@@ -35,10 +35,40 @@ import {
...
+    if (props.console_opts.console_notifications) {
```
karthikvt26, line 55: initially `props.console_opts` seems to be `null` and this is also crashing.

console/src/telemetry/Actions.ts
```diff
@@ -159,10 +159,14 @@ const updateConsoleNotificationsState = (updatedState: NotificationsState) => {
...
-    if (headers?.[HASURA_COLLABORATOR_TOKEN]) {
-      const collabToken = headers[HASURA_COLLABORATOR_TOKEN];
+    const headerHasAdminToken = Object.keys(headers).find(
+      header => header.toLowerCase() === HASURA_COLLABORATOR_TOKEN
+    );
+    if (headerHasAdminToken) {
+      const collabToken = headers[headerHasAdminToken];
       userType = getUserType(collabToken);
     }
+
```
karthikvt26, line 169:
```
if (userType !== 'admin') {
      const currentState = restState?.console_notifications;
      const currentState = restState?.console_notifications;
      if (Object.keys(currentState ?? {}).length > 1) {
      if (Object.keys(currentState ?? {}).length > 1) {
        composedUpdatedState = {
        composedUpdatedState = {
          ...restState,
          ...restState,
          console_notifications: {
          console_notifications: {
            ...currentState,
            ...currentState,
            [userType]: updatedState,
            [userType]: updatedState,
          },
          },
        };
        };
      }
      }
    }
    }
```

When an admin tries to click on the notification bar and an update to the database is triggered, the payload only has `admin` related information and other collaborators details are removed. I am sensing we should add the data of other collaborators when the current user is `admin`

karthikvt26 (changes requested):
Before we save the notification state to the database, we should refetch and save. The problem here is lets say

1. Admin has opened the console and the `console_notification` would have `admin` notification usage info
2. Collaborator has opened the console and the `console_notification` would have `admin` related usage info and `collaborator` notification usage info. 

Now if the admin doesn't refresh the page and does any operation such as `read` and this operation would remove the collaborator's activity and if collaborator refreshes the page they would see the badge and all the notification being unread

console/src/telemetry/Actions.ts
kolharsam: @karthikvt26 I think this case would be resolved, once I make the changes for the cases that you've mentioned here: https://github.com/hasura/graphql-engine/pull/6067#pullrequestreview-519327097

karthikvt26 (approved):
Verified the functionality

beerose (approved):
Console

PR #6068, opened 2020-10-23
server: SQL Server backend (isolated changes)

chrisdone:
As discussed with @nicuveo, this is a rebase of the SQL Server backend--which consists of a "from IR to SQL" module, no schema support.

This consists of only new modules, so should be an easy merge.

I'm not sure whether something needs to be added to the cabal lock file business as I use stack, but I think the CI will tell me what to do upon opening the PR.


chrisdone:
Screenshot from 2020-10-23 14-54-42

@nicuveo Which file should I modify or what command should I run to satisfy cabal?

nicuveo:
Ah, yes, sorry, you need to run a `cabal freeze` to fix the dependencies. Make sure you're using 8.10.2!

chrisdone:
Looks like this is blocked on getting the right Microsoft odbc package installed. I'll take a look at the circle ci script and come back with an update.

.circleci/config.yml
```diff
@@ -172,7 +172,7 @@ jobs:
         command: |
           apt-get -y update
           apt-get -y install postgresql-client-12
-          apt-get -y install unixodbc freetds-dev
+          apt-get -y install unixodbc-dev freetds-dev
```
chrisdone, line 175: I assume this'll let it build. When it comes to testing runtime behavior, we'll need the MS driver, which is a separate "dynamically loaded" component given by the connection string to unixodbc.

nicuveo: Gotcha. Can it be trivially installed via `apt`?

chrisdone: There's a self-contained debian-slim Dockerfile that installs it here: https://github.com/fpco/odbc/blob/master/Dockerfile#L8-L16

I just did a `docker image build` with that Dockerfile and it works, so it's still good.

nicuveo:
Hmmmm.... it looks like the proper way of making the CI happy would be to patch the `graphql-engine-server-builder` docker image. I... have no idea how that image was created, I'll have a look. @jberryman you're the one who made it IIRC?

jberryman:
Right... I can't recall if it's enough to just modify the `server-builder.dockerfile` or if we need to manually push the image somewhere.

abooij:
Here's how `.circleci/server-builder.dockerfile` needs to be changed:
```
 # Don't update this without updating the
 # packager imager of graphql-engine
 FROM haskell:8.10.2-stretch
 
 ARG docker_ver="17.09.0-ce"
 ARG postgres_ver="12"
 ARG node_ver="12.x"
 
 # Install GNU make, curl, git and docker client. Required to build the server
 RUN apt-get -y update \
     && apt-get -y install curl gnupg2 \
     && echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" > /etc/apt/sources.list.d/pgdg.list \
+    && curl -s https://packages.microsoft.com/config/debian/9/prod.list > /etc/apt/sources.list.d/mssql-release.list \
     && curl -s https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - \
+    && curl -s https://packages.microsoft.com/keys/microsoft.asc | apt-key add - \
     && apt-get -y update \
     && apt-get install -y \
          g++ gcc git gnupg libc6-dev libffi-dev libgmp-dev libkrb5-dev \
          libpq-dev libssl-dev make netcat postgresql-client-${postgres_ver} \
          postgresql-client-common python3 python3-pip upx xz-utils zlib1g-dev \
+    && ACCEPT_EULA=Y apt-get -y install msodbcsql17 \
     && curl -sL https://deb.nodesource.com/setup_${node_ver} | bash - \
     && apt-get install -y nodejs \
     && curl -Lo /tmp/docker-${docker_ver}.tgz https://download.docker.com/linux/static/stable/x86_64/docker-${docker_ver}.tgz \
     && tar -xz -C /tmp -f /tmp/docker-${docker_ver}.tgz \
     && mv /tmp/docker/* /usr/bin \
     && apt-get -y purge curl \
     && apt-get -y auto-remove \
     && apt-get -y clean \
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf /usr/share/doc/ \
     && rm -rf /usr/share/man/ \
     && rm -rf /usr/share/locale/
 
 ENV LANG=C.UTF-8 LC_ALL=C.UTF-8
```
After changing the dockerfile, the image needs to actually be built and published, as the CI does not build any docker images. I've asked @0x777 if he could prepare an image for us.

chrisdone:
@abooij @nicuveo Here's a sketch start at generalizing `UnpreparedValue` and such: https://github.com/chrisdone/graphql-engine/commit/48101b446a0aceda61c47a6fdbaac01d669765c0

This compiles. If you swap/delete the `'Postgres` in `Column.hs` then you'll see a bunch of modules light up with PG-specificness.

chrisdone:
I won't pursue that UnpreparedValue generalization for now.

server/src-lib/Hasura/Backends/MSSQL/Types/Instances.hs
```diff
@@ -0,0 +1,123 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+-- | Instances that're slow to compile.
```
0x777, line 3: How is having instances in a different module help with the compile speed?

server/src-lib/Hasura/Backends/MSSQL/FromIr.hs
```diff
@@ -0,0 +1,969 @@
...
+import           Hasura.Backends.MSSQL.Types        as TSQL
+import           Hasura.SQL.Backend
+
+
+--------------------------------------------------------------------------------
+-- Types
+
+-- | Most of these errors should be checked for legitimacy.
+data Error
+  = FromTypeUnsupported (IR.SelectFromG 'MSSQL Expression)
```
0x777, line 44: @abooij A lot of these errors can be avoided by correctly parameterizing the IR as per the 'trees that grow' idea. Note that the equivalent function of `mkSQLSelect` for Postgres does not throw *any* error.

server/src-lib/Hasura/Backends/MSSQL/Plan.hs
```diff
@@ -0,0 +1,260 @@
...
+
+-- TODO: real env object.
+envObjectExpression :: TSQL.Expression
+envObjectExpression =
+  ValueExpression (ODBC.TextValue "[{\"result_id\":1,\"result_vars\":{\"synthetic\":[10]}}]")
+
+--------------------------------------------------------------------------------
+-- Resolving values
+
+data PrepareError
```
0x777, line 124: `PrepareError` too shouldn't exist.

0x777 (changes requested):

server/src-lib/Hasura/Backends/MSSQL/Types/Instances.hs
chrisdone: When editing the Types.hs module your editor/ghci can show type errors immediately instead of the 5s~ it takes to compile all the derived instances code.

nicuveo:
We can close this now I guess. ^^

PR #6069, opened 2020-10-23
add check fk_is_also_primary_key

### LINKED ISSUE #6009: FK that is unique should become object relationship
Would be nice if a foreign key that also is unique would be recognized as object-relationship both ways. At the moment the FK is recognized as array-relationship, but because of the uniqueness its impossible to have multiple referring rows. This makes it easier to share primary keys on tables and join the data as objects...

Example schema:
```sql
create table base_table (
  id serial primary key,
  base_field varchar
);

create table sub_table (
  id integer primary key references base_table on delete cascade,
  sub_field varchar
);

with base_table as (
  insert into base_table (base_field) values ('base value') returning id
)
insert into sub_table(id, sub_field)
select b.id, 'sub value'
from base_table b;

select *
from base_table
natural join sub_table;
```

Could allow as graphql query:
```graphql
query {
  base_table {
    base_field
    sub_table {
      sub_field
    }
  }
  
  sub_table {
    sub_field
    base_table {
      base_field
    }
  }
}```

Workaround is removing the recognized array relationship and manually add an object relationship.

sanderhahn:
### Description
Fix for #6009 (fk that is pk is also unique)

### Affected components
- [x] Console

### Related Issues
close #6009

### Solution and Design
Added function `fk_is_also_primary_key` that lookups the table and checks if the fk is also the primary key and should be considered unique.

~~Extended the is_unique sql query to include primary key constraint. Was removed because it will be refactored in the near future...~~

### Steps to test and verify
Import sql schema from the issue and see that the relationship is now an object instead of an array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,7 +122,7 @@ This release contains the [PDV refactor (#4111)](https://github.com/hasura/graph
 - server: fix bug which arised when renaming a table which had a manual relationship defined (close #4158)
 - server: limit the length of event trigger names (close #5786)
 - server: Configurable websocket keep-alive interval. Add `--websocket-keepalive` command-line flag
-          and handle `HASURA_GRAPHQL_WEBSOCKET_KEEPALIVE` env variable (fix #3539) 
+          and handle `HASURA_GRAPHQL_WEBSOCKET_KEEPALIVE` env variable (fix #3539)
 **NOTE:** If you have event triggers with names greater than 42 chars, then you should update their names to avoid running into Postgres identifier limit bug (#5786)
 - server: validate remote schema queries (fixes #4143)
 - server: fix issue with tracking custom functions that return `SETOF` materialized view (close #5294) (#5945)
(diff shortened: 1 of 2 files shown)


mehtavishwa30:
Hello @sanderhahn!

Thanks so much for your contribution to Hasura OSS during the Hacktoberfest 2020! We'd love to send you some Hasura swag as a token of appreciation. If you can share your email address, we will send you an email with the swag form shortly. ✨

sanderhahn:
@mehtavishwa30 Thanks, no problem great to be able to contribute back to a great product! Already received some very cool Hasura swag earlier, thank you for that again! :)

sanderhahn:
Squashed the commits...

beerose (changes requested):
Thanks for your PR! We really appreciate it and your solution works fine 🙂

However, we're currently working on some bigger changes to the console code (https://github.com/hasura/graphql-engine/pull/6114) — among other things, we're changing the way we extract data and we're getting rid of many SQLs including the one that you edited. We plan to merge said PR pretty soon, thus if I merged this PR now, your work would be erased or the fix would need to be implemented from scratch. 

So after a bit of thinking, I see two options:
* You could go back to your previous solution (look up the table schema to see whether a column is also PK). It will work because this part of the code won't change. We'll still have the same information. We'll just be getting them in a completely different way. 
* You could wait for https://github.com/hasura/graphql-engine/pull/6114 to be merged and try some other solution with the new code (it may take a few weeks, tho).

sanderhahn:
Hello @beerose, no problem have covered the first proposed solution. There is no reason for hurry because i don't think users complained before about thing :) Would be nice if it can get merged, but can also redo a solution once the console refactor gets merged.

PR #6072, opened 2020-10-25
server: split RQL.DML into four different parts

nicuveo:
## Description

Currently, the code in `RQL.DML` has several different purposes:
- it implements the RQL API
- it contains an intermediary representation for queries (used by both RQL and GraphQL)
- it contains the code to transform from that IR to postgres' SQL
- it contains a subset of the Postgres execution code (namely mutations)

This was fine in a Postgres-only world, but complicates things as we move towards a multi-backends codebase.

This PR attempts to solve this problem by splitting `RQL.DML` in four different parts:
- `RQL.DML` only contains the code for the RQL API: execute a query and return its result
- `RQL.IR` contains the **backend agnostic** IR (formerly `RQL/DML/*/Types.hs`)
- `Backends.Postgres.Translate` contains all the code responsible for translating from the IR to a SQL expression
- `Backends.Postgres.Execute` contains the code responsible for executing a query

After this, not only will the import graph be simplified, but it'll also make it easier to continue splitting the code between abstract IR and Postgres-specific implementation.

## Limitations

Most files did not map to a single purpose and had to be split. This PR is consequently a bit messy. A couple known limitations:
- this PR doesn't address all remaining `PG*` types in the IR
- this PR doesn't attempt to split / clean / simplify `RQL.DML.Internal`
- ~~the distinction between Translate and Execute for RemoteJoin feels arbitrary~~; furthermore, that code probably doesn't need to be Postgres-specific and should probably be moved to `GraphQL.Execute`
- this PR attempts to minimize code changes (for ease of submission), and therefore keeps some now nonsensical qualified imports, such as `import Hasura.Backends.Postgres.Translate.Select as RQL`; this will be modified last, to avoid unpleasant rebases during the review process
- this PR does not consolidate all the IR into `RQL.IR`


nicuveo:
(I opened this for review, but it's quite unreadable while the previous one isn't merged. I'll also go over the code another time to see if there are other quick wins in the meantime. The reason I opened it early is that I'd rather proceed step by step rather than do a months long PR that does everything right on the first try.)

nicuveo:
Ok! I've done some cleanup now that all dependencies have been merged. Let me know what you think!

abooij (approved):
Too many files are being touched to really dive into the code changes being made here, so I trust that you've made no substantial changes apart from moving code around, perhaps disambiguating some names here and there.

I like the approach you have taken with this PR. I was always confused by the precise role of `Hasura.RQL.DML`, and this helps to clarify that. The work towards support for other database backends will require more code reorganization, and this RP takes a pragmatic approach towards that. Let's merge this now, and then we'll see which item in your list has the highest priority.

What is the status of `Hasura.RQL.DDL`? Can it be merged into the new `Hasura.RQL.DML`, perhaps creating `Hasura.RQL.API`? As I feel these are doing similar things, although perhaps `Hasura.RQL.DML` is thought of as the Postgres-oriented part?

nicuveo:
I think DDL has a different purpose, and is seldom (if at all) used in the context of GraphQL queries, which is why I chose not to touch it at all for now. And yes, this PR does not change any code, just shuffles it around. I'll do a quick round of renaming the imports when we get closer to it getting submitted, for cleanliness.
And yeah, it is quite verbose, due to touching the imports of so many files... I think that's unavoidable, at least for now.

0x777 (approved):

nicuveo:
Kodiak doesn't seem to be picking it up, merging manually.

PR #6091, opened 2020-10-28
server: generalize SQLExp in RQL.IR

nicuveo:
### Description
This PR is in the continuation of #6003, and continues to generalize the IR, this time by abstracting over `SQLExp`. Additionally, this PR *reverts some generalization work*, as the changes in #6072 have made explicit which parts of the code need to be generic and which don't.

This is a purely internal PR, with no feature change, and it doesn't require a Changelog entry.


server/src-lib/Hasura/Backends/Postgres/Translate/BoolExp.hs
```diff
@@ -377,13 +377,13 @@ foldBoolExp f = \case
...
 mkFieldCompExp
-  :: S.Qual -> FieldName -> OpExpG 'Postgres S.SQLExp -> S.BoolExp
+  :: S.Qual -> FieldName -> OpExpG 'Postgres (SQLExp 'Postgres) -> S.BoolExp
 mkFieldCompExp qual lhsField = mkCompExp (mkQField lhsField)
   where
     mkQCol = S.SEQIdentifier . S.QIdentifier qual . toIdentifier
     mkQField = S.SEQIdentifier . S.QIdentifier qual . Identifier . getFieldNameTxt
 
-    mkCompExp :: S.SQLExp -> OpExpG 'Postgres S.SQLExp -> S.BoolExp
+    mkCompExp :: (SQLExp 'Postgres) -> OpExpG 'Postgres (SQLExp 'Postgres) -> S.BoolExp
```
abooij, line 386:
```suggestion
    mkCompExp :: SQLExp 'Postgres -> OpExpG 'Postgres (SQLExp 'Postgres) -> S.BoolExp
```

server/src-lib/Hasura/Backends/Postgres/Translate/Select.hs
```diff
@@ -340,22 +340,22 @@ mkSimilarArrayFields annFields maybeOrderBys =
...
   AFArrayRelation (ASConnection _) -> Nothing
   AFArrayRelation ar               -> Just (f, ar)
   _                                -> Nothing
 
 
 withWriteJoinTree
-  :: (MonadWriter (JoinTree backend) m)
-  => (JoinTree backend -> b -> JoinTree backend)
+  :: (MonadWriter (JoinTree 'Postgres) m)
+  => (JoinTree 'Postgres -> b -> JoinTree 'Postgres)
```
abooij, line 358:
```suggestion
  => JoinTree 'Postgres -> b -> JoinTree 'Postgres
```

abooij (approved):
Indeed this PR follows the style of #6003, and replaces all instances of the Postgres-specific `SQLExp` in `src-lib/Hasura/RQL/IR/` with a `SQLExp` type family.

Perhaps it will make sense to import the postgres-specific `SQLExp` as `PG.SQLExp` rather than `S.SQLExp` (i.e. change the import qualifier from `S` to `PG`), such as in `src-lib/Hasura/GraphQL/Execute/Insert.hs` and `src-lib/Hasura/RQL/DML/Update.hs`. But it's not necessary to do that now.

0x777 (approved):
LGTM

server/src-lib/Hasura/Backends/Postgres/Translate/BoolExp.hs
nicuveo: Done.

server/src-lib/Hasura/Backends/Postgres/Translate/Select.hs
nicuveo: Hmmmm, nope: that first argument is a function, and requires the parentheses!

abooij: Ah, thanks for paying more attention than me! :-)

PR #6082, opened 2020-10-28
server: generalize BoolExp

nicuveo:
### Description
This small PR is one more step towards a multi-backends codebase: it continues the work started in #6003 and generalizes `BoolExp`. Additionally, it moves RQL's BoolExp code according to the logic of #6072:
- `RQL.Types.BoolExp` is moved to `RQL.IR.BoolExp`
- `RQL.GBoolExp` is moved to `Backends.Postgres.Translate.BoolExp`

### Limitations
This PR does not _fully_ generalize BoolExp, due to the questions surrounding `PGCol` (see Slack discussion).

### Notes

I recommend toggling on `Hide whitespace changes` when reviewing this PR.

PR #6101, opened 2020-10-29
server: split RQL.Types.DML

nicuveo:
_(Note: this PR is on top of #6082 and #6091.)_

### Description
This PR is in the continuation of our IR work. It splits `RQL.Types.DML` in several distinct parts:
- `RQL.DML.Types`, which contains types specific to the RQL query language, and is isolated from the rest of the code base
- `RQL.IR.OrderBy`
- `RQL.IR.BoolExp` (which was already introduced in #6082)

Additionally, it fixes a few mistakes with the previous PRs and starts renaming some imports.

### DIFF
--- a/server/src-lib/Hasura/GraphQL/Schema/OrderBy.hs
+++ b/server/src-lib/Hasura/GraphQL/Schema/OrderBy.hs
@@ -7,18 +7,20 @@ import           Hasura.Prelude
 import qualified Data.List.NonEmpty                 as NE
 import qualified Language.GraphQL.Draft.Syntax      as G
 
-import           Hasura.Backends.Postgres.SQL.DML   as SQL
-import qualified Hasura.GraphQL.Parser              as P
-import qualified Hasura.RQL.IR.Select               as RQL
-import           Hasura.RQL.Types                   as RQL
-
 import           Data.Text.Extended
+
+import qualified Hasura.Backends.Postgres.SQL.DML   as PG
+import qualified Hasura.GraphQL.Parser              as P
+import qualified Hasura.RQL.IR.OrderBy              as IR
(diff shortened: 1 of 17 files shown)


server/src-lib/Hasura/GraphQL/Schema/OrderBy.hs
```diff
@@ -64,28 +66,32 @@ orderByExp table selectPermissions = memoizeOn 'orderByExp table $ do
...
 
 -- local definitions
 
-type OrderInfo = (SQL.OrderType, SQL.NullsOrder)
+type OrderInfo = (PG.OrderType, PG.NullsOrder)
+
 
+-- FIXME!
+-- those parsers are directly using Postgres' SQL representation of
+-- order, rather than using a general intermediary representation
```
0x777, line 88: Why is this bad?

0x777: as we are building parsers that would parse a graphql query into an IR specific to Postgres? Are we planning on reusing the parser generation logic too across backends?

nicuveo: Yeah, definitely, we will want to reuse as much as possible of the existing parsers. How much of that will be possible, I honestly don't know. We will probably want to use the same "trees that grow" logic, where parsers carry around some backend type information, so that they can delegate to backend-specific parsers when required... I can remove this FIXME if you prefer; we'll have to address this at some point anyway. ^^

0x777 (approved):

0x777 (approved):

tirumaraiselvan (approved):

PR #6103, opened 2020-10-29
server: metadata separation: reorganize metadata types

abooij:
### Description

This is an incremental PR towards #5797. It refactors metadata types that will be used to incrementally modify the metadata.

~It also adds a big type class in `Hasura.Class` that will be used as a general interface for providing a metadata API. We may want to discuss this type class in some more detail, to decide whether it follows a design pattern we'd like to include in our code base, or if it needs further thought.~ This change will be made in subsequent incremental PR.

Although this PR adds types that can hold new metadata, the intention is that it does not change what metadata actually gets stored. Any new style metadata code included should not be used. However, we should check this carefully.

### Changelog

### Affected components

- [x] Server

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/RQL/DDL/Metadata.hs
```diff
@@ -72,86 +72,12 @@ runClearMetadata _ = do
   buildSchemaCacheStrict
   return successMsg
 
-applyQP1
```
rakeshkky, line 75: The logic this function does has been moved to `parseJSON` of the `Metadata` type. See `parseListAsMap` [function](https://github.com/hasura/graphql-engine/pull/6103/files#diff-7b8382ab20e441fa214586be491eb6f7ca904d8b20ed97894c16a339be45219aR154).

rakeshkky (approved):
LGTM

tirumaraiselvan (approved):

server/src-lib/Hasura/RQL/DDL/Metadata/Generator.hs
```diff
@@ -48,17 +43,26 @@ genReplaceMetadata = do
...
   where
-    genFunctionsMetadata :: MetadataVersion -> Gen FunctionsMetadata
+    genFunctionsMetadata :: MetadataVersion -> Gen Functions
     genFunctionsMetadata = \case
-      MVVersion1 -> FMVersion1 <$> arbitrary
-      MVVersion2 -> FMVersion2 <$> arbitrary
+      MVVersion1 -> OM.fromList . map (\qf -> (qf, FunctionMetadata qf emptyFunctionConfig)) <$> arbitrary
+      MVVersion2 -> arbitrary
+
+instance (Arbitrary k, Eq k, Hashable k, Arbitrary v) => Arbitrary (InsOrdHashMap k v) where
```
nicuveo, line 51: Having those instances here seem.... arbitrary :drum: 
Jokes aside: I think we could move those to a dedicated module... maybe `Hasura.Generators` or something? But that could be done in a separate PR.

server/src-lib/Hasura/RQL/Types/Permission.hs
```diff
@@ -36,7 +38,7 @@ instance Q.FromCol PermType where
     "delete" -> Just PTDelete
     _        -> Nothing
 
-permTypeToCode :: PermType -> Text
+permTypeToCode :: PermType -> T.Text
```
nicuveo, line 41: nit: that's not required, we have an export of `Text` in the Prelude.

nicuveo (approved):
LGTM, only a few nits.

server/src-lib/Hasura/RQL/DDL/Metadata/Generator.hs
abooij: Let me at least add a docstring/comment to this module suggesting this.

server/src-lib/Hasura/RQL/Types/Permission.hs
abooij: Agreed, will fix this.

PR #6114, opened 2020-11-02
console: data sources support and metadata separation

beerose:
### Description
TBD

### Changelog

### Affected components

- [x] Console

### Test

* UX when building an entirely new app using postgres
* UX when building using an existing Postgres - successful case, error cases (eg: permissions)
* UX for people upgrading from 1.3 to 1.4 using DATABASE_URL as it is
* UX for runtime failures and error reporting in case Postgres source DB is down


soorajshankar (approved):
Functionalities are working as expected.

Issue #6041, opened 2020-10-19
change isolation level of event fetch to ReadCommitted


tirumaraiselvan:
Since the [query that is run](https://github.com/hasura/graphql-engine/blob/4c3890b3eadde0f7c04b03885c182c85a098516f/server/src-lib/Hasura/Eventing/EventTrigger.hs#L417) is of the form:

```
SELECT FOR UPDATE .. SKIP LOCKED
UPDATE ..
```

Even with concurrent queries running in ReadCommitted isolation level, this should be safe from re-deliveries. The first transaction to lock the event will "win" and others will ignore that event.

tirumaraiselvan:
This should fix errors of the following type which are seen in fairly active event trigger usage:

```
{"type":"event-trigger","timestamp":"2020-08-31T17:08:47.158+0000","level":"error","detail":{"internal":{"statement":"\n UPDATE hdb_catalog.event_log\n SET locked = 't'\n WHERE id IN ( SELECT l.id\n FROM hdb_catalog.event_log l\n WHERE l.delivered = 'f' and l.error = 'f' and l.locked = 'f'\n and (l.next_retry_at is NULL or l.next_retry_at <= now())\n and l.archived = 'f'\n ORDER BY created_at\n LIMIT $1\n FOR UPDATE SKIP LOCKED )\n RETURNING id, schema_name, table_name, trigger_name, payload::json, tries, created_at\n ","prepared":true,"error":{"exec_status":"FatalError","hint":null,"message":"could not serialize access due to concurrent update","status_code":"40001","description":null},"arguments":["(Oid 20,Just (\"\\NUL\\NUL\\NUL\\NUL\\NUL\\NUL\\NULd\",Binary))"]},"path":"$","error":"database query error","code":"unexpected"}​} 
```

tirumaraiselvan:
closed by #6120

PR #6120, opened 2020-11-03
use readcommitted isolation level for event fetch

tirumaraiselvan:
### Description

#6041 

### Changelog

### Affected components

- [x] Server

### Steps to test and verify

`FOR UPDATE .. SKIP LOCKED` will enforce non-overlapping rows on concurrent event fetches and hence it is sufficient to run this in ReadCommitted. Read more here: https://www.postgresql.org/docs/9.5/sql-select.html#SQL-FOR-UPDATE-SHARE

Further, we do not expect the UPDATE to have any bad outcome in ReadCommitted with any other writes as it would either be a new INSERT (non-conflicting with fetched rows) or an UPDATE on a `locked=t` row (again non-conflicting with fetched rows). The only conflict that can happen is when the event trigger is deleted, in which case `archived=t` will be set and the event may or may not be delivered depending on which tx wins the lock first -- this is acceptable.


server/src-lib/Hasura/Eventing/EventTrigger.hs
```diff
@@ -187,7 +187,7 @@ processEventQueue logger logenv httpMgr pool getSchemaCache eeCtx@EventEngineCtx
   where
     fetchBatchSize = 100
     popEventsBatch = do
-      let run = liftIO . runExceptT . Q.runTx pool (Q.RepeatableRead, Just Q.ReadWrite)
+      let run = liftIO . runExceptT . Q.runTx pool (Q.ReadCommitted, Just Q.ReadWrite)
```
codingkarthik, line 190: I think we should justify why `Q.ReadCommitted` is sufficient here, the description of the PR should suffice .

codingkarthik (approved):
LGTM, with a request for the comment.

0x777 (approved):

Issue #6112, opened 2020-11-02
cli: scripts update-config-v2 should remove metadata.yaml from migrations folder


tirumaraiselvan:
In config-v1, if there is a `metadata.yaml` file in `migrations/` folder, then `hasura scripts update-config-v2` should remove that file  because that file is no longer used for anything and is confusing (anyway `migrations_backup` has it).

PR #6121, opened 2020-11-03
CLI: Remove metadata.yaml file from update-config-v2 in migrations folder (#fixes 6112)

purush7:
### Changelog

### Affected components

- [x] CLI

### Related Issues

https://github.com/hasura/graphql-engine/issues/6112

### Aim
To remove metadata.yaml file while updating the project from version 1 to version 2.


cli/commands/scripts_update_config_v2.go
```diff
@@ -343,6 +343,14 @@ func newScriptsUpdateConfigV2Cmd(ec *cli.ExecutionContext) *cobra.Command {
 			}
 			ec.Spinner.Stop()
 			ec.Logger.Infoln("Updated config to version 2")
+
+			if _, err = os.Stat(filepath.Join(ec.MigrationDir, "metadata.yaml")); err == nil || os.IsExist(err) {
+				err = os.Remove(filepath.Join(ec.MigrationDir, "metadata.yaml"))
+				if err != nil {
+					return errors.Wrap(err, "cannot remove metadata.yaml file")
```
scriptonist, line 350: Maybe don't  "return" with a failure here, since we have already completed the migration. I think It'll be better If we do a "Warning" log here instead of returning with an error.

scriptonist (changes requested):

cli/commands/scripts_update_config_v2.go
purush7: okay

cli/commands/scripts_update_config_v2.go
```diff
@@ -343,6 +343,14 @@ func newScriptsUpdateConfigV2Cmd(ec *cli.ExecutionContext) *cobra.Command {
 			}
 			ec.Spinner.Stop()
 			ec.Logger.Infoln("Updated config to version 2")
+
+			if _, err = os.Stat(filepath.Join(ec.MigrationDir, "metadata.yaml")); err == nil || os.IsExist(err) {
```
scriptonist, line 347: There will be a file pointer only when we can successfully "stat" a file right? which means we are not interested in the error here, so maybe something like this also would work?
```
if f, _ = os.Stat(filepath.Join(ec.MigrationDir, "metadata.yaml")); f {
    // do stuff
}
```

purush7: yeah

scriptonist (approved):

scriptnull (approved):

PR #6123, opened 2020-11-03
server: avoid raising mutation check constraint violation in pg procedure

rakeshkky:
### Description

This is incremental PR towards #5797. It removes dependency on `hdb_catalog.check_violation` pg procedure to raise mutation (insert & update) check constraint violation exception. Instead, the check constraint violation error is thrown in Haskell code itself.

### Changelog

### Affected components

- [x] Server

### Related Issues

Refer the PR #5797 

### Solution and Design

SQL is generated such a way that it returns a nullable error message column `check__constraint` which is result of a check constraint evaluation. 

```
....
....
RETURNING *,
<check-constraint> AS "check__constraint"
```
In Haskell code, the error is raised when `check__error` column is not null. 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


abooij (commented):
I don't think I understand RQL well enough to be able to judge if this has any odd side-effects. But the Haskell code itself looks okay to me. @0x777, what do you think?

server/src-lib/Hasura/Backends/Postgres/Execute/RemoteJoin.hs
```diff
@@ -64,8 +66,26 @@ executeQueryWithRemoteJoins
...
+  => Env.Environment
+  -> HTTP.Manager
+  -> [N.Header]
+  -> UserInfo
+  -> BL.ByteString
+  -> RemoteJoins 'Postgres
+  -> m EncJSON
+processRemoteJoins env manager reqHdrs userInfo pgRes rjs = do
+  -- Step 1: Decode the given bytestring as a JSON value
   jsonRes <- either (throw500 . T.pack) pure $ AO.eitherDecode pgRes
```
abooij, line 89:
```suggestion
  jsonRes <- onLeft (AO.eitherDecode pgRes) $ throw500 . T.pack
```
(I'll add an hlint hint for this momentarily.)

server/src-lib/Hasura/RQL/IR/Returning.hs
```diff
@@ -79,3 +79,18 @@ hasNestedFld = \case
       AFObjectRelation _ -> True
       AFArrayRelation _  -> True
       _                  -> False
+
+data MutationCTE
```
0x777, line 83: These need to be documented. Maybe a note describing how we validate the check constraint and reference it?

0x777 (changes requested):

marionschleifer (approved):
docs

server/src-lib/Hasura/Backends/Postgres/Execute/Mutation.hs
```diff
@@ -186,51 +183,80 @@ mutateAndSel
...
   -- Perform select query and fetch returning fields
-  executeMutationOutputQuery env (Q.fromBuilder $ toSQL selWith) [] remoteJoins
+  executeMutationOutputQuery env qt allCols Nothing
+    (MCSelectValues select) mutationOutput strfyNum [] remoteJoins
+
+withCheckPermission :: (MonadError QErr m) => m (a, Bool) -> m a
+withCheckPermission sqlTx = do
+  (rawResponse, checkConstraint) <- sqlTx
+  unless checkConstraint $ throw400 PermissionError $
+    "Check constraint violation. insert/update check constraint failed"
```
tirumaraiselvan, line 195: why do we call it "check constraint" and not "permission check"?

rakeshkky: Because we used to capture the exception from the Postgres as check constraint violation. Now it is not. I'll make the necessary change.

0x777 (changes requested):

rakeshkky:
@0x777 Please have a final look at the PR.

0x777 (approved):

nicuveo:
Ah, this introduced some Postgres-specific code in `RQL.IR`. :(
I'll open a new PR to fix that.

PR #6126, opened 2020-11-03
server: improve hlint hints for Either and Maybe

abooij:
### Description

#6059 enabled a number of hlint hints, including a few new hints that ensured that we made proper use of some of our combinators for `Either` and `Maybe`, such as `onLeft` and `onNothing`. During the review of #6123 I noticed that these were not firing due to the use of `pure` rather than `return`. This PR moves some of the hints to the right place and duplicates them to work for `pure` as well as `return`. The duplication is unfortunately necessary: if one only has the hints in terms of `return`, then hlint doesn't see the ones in terms of `pure`, and vice versa.

This PR does *not* apply these hints throughout the codebase, as there are a number of PRs lined up touching some of these files. If this PR gets merged before those PRs, then that means that those PRs will get explicit hlint hints rather than a merge conflict. Arguably, this may be more helpful. The downside is that *if* those conflicts could be resolved automatically by git, then we get a hlint warning rather than an automatic merge conflict resolution. So, dear reviewer, please judge whether this is the right choice.

### Changelog

### Affected components

- [x] Server
- [x] Tests

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


nicuveo (approved):
LGTM!
Out of curiosity, if you run it on the entire repository, how many of those does it find?

tirumaraiselvan (approved):

PR #6127, opened 2020-11-03
server: generalize table name IR

nicuveo:
### Description
This PR is in the continuation of our IR work. It generalizes table names throughout the IR code. It uses a slightly different representation than the one we've been using so far, to avoid having an explosion of instance declarations for every type that uses it.

Additionally, this starts fixing a few imports.

### DIFF
--- a/server/src-lib/Hasura/RQL/Types/Common.hs
+++ b/server/src-lib/Hasura/RQL/Types/Common.hs
@@ -310,12 +310,8 @@ unsafeNonNegativeInt = NonNegativeInt
 instance FromJSON NonNegativeInt where
   parseJSON = withScientific "NonNegativeInt" $ \t -> do
     case t >= 0 of
-      True  -> NonNegativeInt <$> maybeInt (toBoundedInteger t)
+      True  -> maybe (fail "integer passed is out of bounds") (pure . NonNegativeInt) $ toBoundedInteger t
       False -> fail "negative value not allowed"
-    where
-      maybeInt x = case x of
-        Just v  -> return v
-        Nothing -> fail "integer passed is out of bounds"
 
 newtype NonNegativeDiffTime = NonNegativeDiffTime { unNonNegativeDiffTime :: DiffTime }
   deriving (Show, Eq,ToJSON,Generic, NFData, Cacheable, Num)
--- a/server/src-lib/Hasura/RQL/Types/Common.hs
+++ b/server/src-lib/Hasura/RQL/Types/Common.hs
@@ -9,6 +9,7 @@ module Hasura.RQL.Types.Common
        , ScalarType
        , Column
        , SQLExp
+       , Representation (..)
 
        , FieldName(..)
        , fromPGCol
@@ -71,16 +72,18 @@ import           Data.Aeson
 import           Data.Aeson.Casing
 import           Data.Aeson.TH
 import           Data.Bifunctor                     (bimap)
+import           Data.Kind                          (Type)
 import           Data.Scientific                    (toBoundedInteger)
 import           Data.Text.Extended
 import           Data.Text.NonEmpty
+import           Data.Typeable
 import           Data.URL.Template
 import           Instances.TH.Lift                  ()
 import           Language.Haskell.TH.Syntax         (Lift)
 
-import qualified Hasura.Backends.Postgres.SQL.DML   as S
+import qualified Hasura.Backends.Postgres.SQL.DML   as PG
+import qualified Hasura.Backends.Postgres.SQL.Types as PG
 
-import           Hasura.Backends.Postgres.SQL.Types
 import           Hasura.EncJSON
 import           Hasura.Incremental                 (Cacheable)
 import           Hasura.RQL.DDL.Headers             ()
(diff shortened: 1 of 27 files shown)


abooij (approved):
When I was working on #6003, I also encountered problems with type class instances: since our types are now defined in terms of type families, in order to derive instances, we need some control over which instances for those type families are in scope. At the time, I "solved" this by only writing instances for `'Postgres`, for which the required instances exist and are in scope. But that's not scalable, of course. It was clear we'd have to solve this problem in the future.

Well, it seems that "today is tomorrow". Whereas plain type families don't allow control over type class instances that are in scope, but type families associated to type classes do. This looks like an elegant and robust solution. We'll probably end up using this trick throughout our IR generalization work.

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -89,16 +92,31 @@ import           Hasura.SQL.Backend
...
+  , Lift (TableName b)
+  , NFData (TableName b)
+  , Cacheable (TableName b)
+  , Hashable (TableName b)
+  , Data (TableName b)
+  , Typeable b
+  ) => Representation (b :: Backend) where
+  type TableName b :: Type
+
+instance Representation 'Postgres where
```
0x777, line 118: Why do we need the `Representation` typeclass? It only seems to define the `TableName` type family?

0x777: Looks like Auke's comment above explains it:
> Whereas plain type families don't allow control over type class instances that are in scope, but type families associated to type classes do. This looks like an elegant and robust solution. We'll probably end up using this trick throughout our IR generalization work

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -292,12 +310,8 @@ unsafeNonNegativeInt = NonNegativeInt
 instance FromJSON NonNegativeInt where
   parseJSON = withScientific "NonNegativeInt" $ \t -> do
     case t >= 0 of
-      True  -> NonNegativeInt <$> maybeInt (toBoundedInteger t)
+      True  -> maybe (fail "integer passed is out of bounds") (pure . NonNegativeInt) $ toBoundedInteger t
```
abooij, line 313:
```suggestion
      True  -> NonNegativeInt <$> onNothing (toBoundedInteger t) $ fail "integer passed is out of bounds"
```
(not sure about the binding strength of `<$>` vs `$`)

server/src-lib/Hasura/RQL/Types/Common.hs
nicuveo: Every single type of the IR that wants to derive a class instance needs to explicitly list all its requirements, and **transitively the requirements of all its dependencies**: if a type has a field whose type has a field of type `TableName`, it will have to add a `Show (TableName b) =>` or `Eq (TableName b) =>` for each of its corresponding instances. The more types we generalize, the more unwieldy each new type declaration becomes... Using this new approach allows to list all the requirements on those types once, and then use a single `Representation b =>` at instance declaration time.

This PR introduces `Representation`, and subsequent PRs add to it. In the last commit of the pipeline it looks like this:
```
class
  ( Representable (Identifier b)
  , Representable (TableName b)
  , Representable (FunctionName b)
  , Representable (ConstraintName b)
  , Representable (BasicOrderType b)
  , Representable (NullsOrderType b)
  , Representable (Column b)
  , Representable (ScalarType b)
  , Representable (SQLExpression b)
  , Representable (SQLOperator b)
  , Ord (TableName b)
  , Ord (ScalarType b)
  , Lift (TableName b)
  , Lift (BasicOrderType b)
  , Lift (NullsOrderType b)
  , Data (TableName b)
  , Data (ScalarType b)
  , Data (SQLExpression b)
  , FromJSON (TableName b)
  , FromJSON (ScalarType b)
  , FromJSON (BasicOrderType b)
  , FromJSON (NullsOrderType b)
  , ToJSON (TableName b)
  , ToJSON (ScalarType b)
  , ToJSON (BasicOrderType b)
  , ToJSON (NullsOrderType b)
  , ToTxt (TableName b)
  , ToTxt (ScalarType b)
  , Typeable b
  ) => Representation (b :: Backend) where
  type Identifier     b :: Type
  type Alias          b :: Type
  type TableName      b :: Type
  type FunctionName   b :: Type
  type ConstraintName b :: Type
  type BasicOrderType b :: Type
  type NullsOrderType b :: Type
  type CountType      b :: Type
  type Column         b :: Type
  type ScalarType     b :: Type
  type SQLExpression  b :: Type
  type SQLOperator    b :: Type
```

server/src-lib/Hasura/RQL/Types/Common.hs
nicuveo: Yeah, I used `maybe` on purpose here, since the positive case is not just a `pure`; I find it more readable in this case?

abooij: Okay, I don't feel strongly about this.

server/src-lib/Hasura/RQL/Types/Common.hs
```diff
@@ -89,16 +92,31 @@ import           Hasura.SQL.Backend
...
+class
+  ( Show (TableName b)
+  , Eq (TableName b)
+  , Lift (TableName b)
+  , NFData (TableName b)
+  , Cacheable (TableName b)
+  , Hashable (TableName b)
+  , Data (TableName b)
+  , Typeable b
+  ) => Representation (b :: Backend) where
```
0x777, line 115: We rename `Backend` to `BackendType` and `Representation` to `Backend`?

nicuveo: Ah, yeah, that works! I've pushed a comment, I'll do the renaming.

abooij: This is an elegant solution.

nicuveo: Done!

0x777 (approved):

PR #6131, opened 2020-11-04
metadata storage abstraction for scheduled triggers

rakeshkky:
### Description

This is an incremental PR towards #5797. It introduces an abstraction for metadata storage `MonadMetadataStorage` so that the caller of the function `runHGEServer` can customize the implementations via type class instance. It is a step towards separation of metadata from users database. This PR ONLY abstracted the operations related to scheduled triggers eventing sub-system. Subsequent PRs will contain operations related to metadata management and async actions sub system.  

This also includes moving of types related to scheduled triggers.

**To Reviewers:** Feel free to post your thoughts on type class design and suggest improvements :sweat_smile: 

### Changelog

### Affected components

- [x] Server

### Solution and Design

Each function in the new type class operates in `MetadataStorageT m a` monad transformer where `m` is base App monad in which the function `runHGEServer` operates in. 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/Class.hs
```diff
@@ -0,0 +1,84 @@
+module Hasura.Class
```
abooij, line 1: Maybe this could use a more specific name. Something like `Hasura.Metadata.Class`?

server/src-lib/Hasura/Class.hs
```diff
@@ -0,0 +1,84 @@
...
+  getDeprivedCronTriggerStats :: MetadataStorageT m [CronTriggerStats]
+  getScheduledEventsForDelivery :: MetadataStorageT m ([CronEvent], [OneOffScheduledEvent])
+  insertScheduledEvent :: ScheduledEventSeed -> MetadataStorageT m ()
+  insertScheduledEventInvocation
+    :: Invocation 'ScheduledType -> ScheduledEventType -> MetadataStorageT m ()
+  setScheduledEventOp
+    :: ScheduledEventId -> ScheduledEventOp -> ScheduledEventType -> MetadataStorageT m ()
+  unlockScheduledEvents
+    :: ScheduledEventType -> [ScheduledEventId] -> MetadataStorageT m Int
+  unlockAllLockedScheduledEvents :: MetadataStorageT m ()
```
abooij, line 44: @0x777 and myself discussed a potential future design here that avoids spelling out all of these methods explicitly. TODO: expand this sentence.

server/src-lib/Hasura/Class.hs
```diff
@@ -0,0 +1,84 @@
...
+runMetadataStorageT =
+  runExceptT . unMetadataStorageT
+
+
+class (Monad m) => MonadMetadataStorage m where
+
+  -- Scheduled triggers
+  getDeprivedCronTriggerStats :: MetadataStorageT m [CronTriggerStats]
+  getScheduledEventsForDelivery :: MetadataStorageT m ([CronEvent], [OneOffScheduledEvent])
+  insertScheduledEvent :: ScheduledEventSeed -> MetadataStorageT m ()
```
abooij, line 37: Does this also make sense?
```suggestion
class (MonadError QErr m) => MonadMetadataStorage m where

  -- Scheduled triggers
  getDeprivedCronTriggerStats :: m [CronTriggerStats]
  getScheduledEventsForDelivery :: m ([CronEvent], [OneOffScheduledEvent])
  insertScheduledEvent :: ScheduledEventSeed -> m ()
```

So rather than writing an instance for `ServerAppM`:

```haskell
instance MonadMetadataStorage ServerAppM where
```

you would write an instance for `MonadStorageT ServerAppM`:

```haskell
instance MonadMetadataStorage (MonadStorageT ServerAppM) where
```

server/src-lib/Hasura/App.hs
```diff
@@ -632,42 +643,58 @@ instance HttpLog AppM where
...
+
+instance MonadMetadataStorage ServerAppM where
+
+  getDeprivedCronTriggerStats        = runTxInMetadataStorage getDeprivedCronTriggerStatsTx
+  getScheduledEventsForDelivery      = runTxInMetadataStorage getScheduledEventsForDeliveryTx
+  insertScheduledEvent               = runTxInMetadataStorage . insertScheduledEventTx
+  insertScheduledEventInvocation a b = runTxInMetadataStorage $ insertInvocationTx a b
+  setScheduledEventOp a b c          = runTxInMetadataStorage $ setScheduledEventOpTx a b c
+  unlockScheduledEvents a b          = runTxInMetadataStorage $ unlockScheduledEventsTx a b
+  unlockAllLockedScheduledEvents     = runTxInMetadataStorage unlockAllLockedScheduledEventsTx
```
abooij, line 696: Rather than using the `Tx` suffix everywhere, why don't we import `ScheduleTriggers.hs` qualified, to avoid the name clash that way?

server/src-lib/Hasura/Class.hs
abooij: Please see my experiment [here](https://github.com/rakeshkky/graphql-engine/pull/8).

rakeshkky: Implemented!

server/src-lib/Hasura/App.hs
rakeshkky: Type class functions are being used in the same module (`Hasura.Eventing.ScheduledTrigger`). So I need to suffix `Tx` to identify those functions as database transactions.

server/src-lib/Hasura/App.hs
```diff
@@ -632,42 +643,58 @@ instance HttpLog AppM where
...
+instance WS.MonadWSLog ServerAppM where
   logWSLog = unLogger
 
+runTxInMetadataStorage :: Q.TxE QErr a -> MetadataStorageT ServerAppM a
+runTxInMetadataStorage tx = do
+  pool <- lift ask
+  liftEitherM $ liftIO $ runExceptT $
+    Q.runTx pool (Q.RepeatableRead, Just Q.ReadWrite) tx
+
+instance MonadMetadataStorage (MetadataStorageT ServerAppM) where
```
lexi-lambda, line 688: The way these pieces fit together seems a little unusual to me:

  * It’s odd to see an instance like `instance MonadFoo (BarT BazM)` rather than `instance MonadFoo BazM` or `instance MonadBaz m => MonadFoo (BarT m)`. That is, you normally define instances on a concrete monad or on a transformer with a polymorphic monad underneath, not on a transformer with a concrete monad underneath.

  * `MetadataStorageT` sounds like it’s *the* transformer for metadata storage, but it looks like this instance really implements “Postgres metadata storage.”

  * The use of the `runTxInMetadataStorage` wrapper—rather than just using `liftTx`—is a bit opaque to me. (This also seems like it will run each of these methods in a totally separate transaction—is that desirable in general?)

Reasoning through the surrounding context, it seems like perhaps a couple concerns have been muddled a bit:

  1. We need a way to acquire a Postgres connection.

  2. We need a way to delimit the scope of a Postgres transaction.

  3. We need a way to handle errors that result from executing queries against Postgres.

  4. We need a way to select which implementation of metadata storage we want to use.

`MetadataStorageT` is a bit odd as-written, because it appears to be doing points 2 through 4, which in my mind are not really usefully coupled concerns. One could imagine having things that aren’t related to metadata storage that still want to raise errors; those errors should probably go to the same place as the metadata storage errors, but it seems odd for `MetadataStorageT` to be that place. Similarly, it’s not clear to me that the scope of a Postgres transaction should be limited to a single method of `MonadMetadataStorage`.

It’s worth thinking about what exactly `MetadataStorageT`’s purpose is. If it’s just a compile-time constant of sorts that’s used to guide typeclass instance selection, maybe it should be the identity transformer:

```haskell
newtype PGMetadataStorageT m a
  = PGMetadataStorageT { runPGMetadataStorageT :: m a }
```

Then you can hang a typeclass instance on it that looks like this:

```haskell
instance MonadTx m => MonadMetadataStorage (PGMetadataStorageT m) where
  getDeprivedCronTriggerStats = getDeprivedCronTriggerStatsTx
  getScheduledEventsForDelivery = getScheduledEventsForDeliveryTx
  insertScheduledEvent = insertScheduledEventTx
  ...
```

This seems like a division of responsibilities that would make more sense to me. Concerns 2 and 3 can be handled by some other transformer in the stack, which makes a bit more sense, since they’re all very closely related.

rakeshkky: @lexi-lambda 
> This also seems like it will run each of these methods in a totally separate transaction?

Yes.

> Is that desirable in general?

The Metadata storage abstraction need NOT to be tied to Postgres. That said, it is the implementer choice to use any storage. In OSS, we choose that to be Postgres. 

In a nutshell the `MonadMetadataStorage` have functions in which each of them should operate in `MonadError QErr m`. Maybe the name `MetadataStorageT` is mis-leading but it just a wrapper over `ExceptT` so as to fulfill the `MonadError QErr` constraint here.

> not clear to me that the scope of a Postgres transaction should be limited to a single method of  `MonadMetadataStorage`

We're implementing the eventing sub-system for scheduled triggers in metadata storage. Each method is an isolated interaction to the storage and it is not necessary to guarantee a transaction.

ecthiender: Probably it is also worth mentioning that the existing `AppM` type has been renamed to `ServerAppM`. I am not sure of the reasoning (maybe @rakeshkky can shed some light), but maybe something to keep in mind that this `ServerAppM` is not a newly introduced type but the existing app-monad type we had.

rakeshkky: @ecthiender The `ServerAppM` is specifically where `runHGEServer` function operates in. Also, now it carries the reader context of metadata storage pool (`Q.PGPool`).

server/src-lib/Hasura/Class.hs
abooij: I'm resolving this particular discussion, although spiritually @lexi-lambda has struck similar points below and I think that discussion should be continued.

server/src-lib/Hasura/App.hs
abooij: Would it be possible to add a comment to `ScheduledTrigger.hs` explaining the relationship between the `Tx`-variant and the non-`Tx`-variant of these methods?

server/src-lib/Hasura/App.hs
abooij: Would something along the following lines work and/or make sense, @rakeshkky?

```suggestion
instance (MonadReader Q.PGPool m, MonadTx m) => MonadMetadataStorage (MetadataStorageT m) where
```

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -0,0 +1,98 @@
...
+{- Note [Todo: Common interface for eventing sub-system]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Scheduled and Table event triggers have their own implementation
+and tables in Postgres. The core logic for both is same. So it
+is necessary to have a unified interface (maybe via a Postgres extension)
+where all event triggers implementation can extend from. By this we can
+fairly reduce the Postgres schema foot print and interactions to the database.
+
+TODO: Reference to open issue or rfc?
+-}
```
abooij, line 42: Could you please expand this note a little bit? I think an important point is being made here, but I find the language a bit too abstract to grasp.

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -0,0 +1,98 @@
...
+-- (TODO):
+-- This also includes functions to fetch events and their invocations so that the
+-- console can show them in the UI.
+-- For more details, refer description in 'Hasura.Eventing.ScheduledTrigger' module.
+--
+-- 3. Async Actions (TODO)
+-- ----------------
+-- Operations to implement async actions sub-system. This includes recording an
+-- async action event and retreiving the details of action delivery to the webhook.
+-- For more details see Note [Async action architecture] in 'Hasura.GraphQL.Execute.Action' module.
```
abooij, line 66: I think this also deserves to be expanded. The fact that the type class specifies functions related to these three topics is clear. The question is *why* they have been captured in a type class, and why these three areas are deemed to belong together in one type class.

server/src-lib/Hasura/RQL/Types/ScheduledTrigger.hs
```diff
@@ -166,3 +187,173 @@ instance FromJSON CreateScheduledEvent where
...
+  = SEOneOff
+  | SECron !TriggerName
+  deriving (Show, Eq)
+
+parseScheduledEvent :: Object -> Parser ScheduledEvent
+parseScheduledEvent o = do
+  ty <- o .: "type"
+  case ty of
+    Cron   -> SECron <$> o .: "trigger_name"
+    OneOff -> pure SEOneOff
```
abooij, line 230: Perhaps from a code organization point of view it makes sense to make this into a local definition as part of the `FromJSON GetEventInvocations` instance, and ditto for `parseScheduledEventPagination`, just like `parseEventId` already is?

server/src-lib/Hasura/RQL/Types/ScheduledTrigger.hs
```diff
@@ -166,3 +187,173 @@ instance FromJSON CreateScheduledEvent where
...
+    flip Q.fromColHelper bs $ PD.enum textToScheduledEventStatus
+
+instance ToJSON ScheduledEventStatus where
+  toJSON = String . scheduledEventStatusToText
+
+instance FromJSON ScheduledEventStatus where
+  parseJSON = withText "String" $ \s ->
+    case textToScheduledEventStatus s of
+      Nothing     -> fail $ T.unpack $ "unexpected status: " <> s
+      Just status -> pure status
```
abooij, line 329:
```suggestion
  parseJSON = withText "String" $ \s -> onNothing (textToScheduledEventStatus s) $
    fail $ T.unpack $ "unexpected status: " <> s
```
(I didn't check how lambdas and `$` associate, so the above might need some parentheses before it compiles.)

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -351,221 +179,199 @@ generateScheduleTimes from n cron = take n $ go from
...
+        let webhookUrl = unResolvedWebhook ctiWebhookInfo
+            payload = ScheduledEventWebhookPayload id' (Just name) st
+                      (fromMaybe J.Null ctiPayload) ctiComment
+                      Nothing
+            retryCtx = RetryContext tries ctiRetryConf
+        finally <- runMetadataStorageT $ flip runReaderT (logger, httpMgr) $
+                   processScheduledEvent logEnv id' ctiHeaders retryCtx
+                                         payload webhookUrl Cron
+        removeEventFromLockedEvents id' lockedCronEvents
+        onLeft finally logInternalError
```
abooij, line 215:
```suggestion
    CronTriggerInfo{..} <- onNothing (Map.lookup name cronTriggersInfo) $ 
      logInternalError $ err500 Unexpected "could not find cron trigger in cache"
    let webhookUrl = unResolvedWebhook ctiWebhookInfo
        payload = ScheduledEventWebhookPayload id' (Just name) st
                  (fromMaybe J.Null ctiPayload) ctiComment
                  Nothing
        retryCtx = RetryContext tries ctiRetryConf
    finally <- runMetadataStorageT $ flip runReaderT (logger, httpMgr) $
               processScheduledEvent logEnv id' ctiHeaders retryCtx
                                     payload webhookUrl Cron
    removeEventFromLockedEvents id' lockedCronEvents
    onLeft finally logInternalError
```

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
```diff
@@ -61,283 +61,111 @@ During the startup, two threads are started:
...
+      -- join stats with cron triggers and produce @[(CronTriggerInfo, CronTriggerStats)]@
+      cronTriggersForHydrationWithStats <-
+        catMaybes <$>
+        mapM (withCronTrigger cronTriggersCache) deprivedCronTriggerStats
+      insertCronEventsFor cronTriggersForHydrationWithStats
+
+    case eitherRes of
+      Right _ -> pure ()
       Left err -> L.unLogger logger $
         ScheduledTriggerInternalErr $ err500 Unexpected (T.pack $ show err)
```
abooij, line 145:
```suggestion
    void $ onLeft eitherRes $
      L.unLogger logger . ScheduledTriggerInternalErr . err500 Unexpected . T.pack . show
```
(I'm not 100% sure how `.` and `$` associate - please check.)

server/src-lib/Hasura/Class.hs
0x777: Ideally we would've liked to avoid having functions that are specific to scheduled/cron triggers and instead have functions that provide a generic 'event storage and retrieval' interface but we'll have to change a lot of existing code for scheduled and cron triggers. We can get to this after the multi-source work is done. (@rakeshkky maybe you can put this as a comment under TODO in `MonadMetadataStorage` class's description.

server/src-lib/Hasura/App.hs
rakeshkky: @abooij I think we shouldn't limit MetadataStorage implementation to Postgres. Hence we don't want to put `MonadReader Q.PGPool` constraint.

server/src-lib/Hasura/Eventing/ScheduledTrigger.hs
abooij: (This suggestion didn't actually work out, and so I have reverted it on the branch.)

server/src-lib/Hasura/RQL/Types/ScheduledTrigger.hs
rakeshkky: These types aren't necessary now. However, I'll keep these notes in mind while adding them in subsequent PRs.

server/src-lib/Hasura/App.hs
```diff
@@ -643,57 +644,58 @@ instance HttpLog ServerAppM where
...
-  insertScheduledEvent               = runTxInMetadataStorage . insertScheduledEventTx
-  insertScheduledEventInvocation a b = runTxInMetadataStorage $ insertInvocationTx a b
-  setScheduledEventOp a b c          = runTxInMetadataStorage $ setScheduledEventOpTx a b c
-  unlockScheduledEvents a b          = runTxInMetadataStorage $ unlockScheduledEventsTx a b
-  unlockAllLockedScheduledEvents     = runTxInMetadataStorage unlockAllLockedScheduledEventsTx
+instance MonadTx (MetadataStorageT PGMetadataStorageApp) where
+  liftTx tx = do
+    pool <- lift ask
+    -- Each operation in metadata storage is executed in an isolated transaction.
+    -- No two operations are not necessarily executed together.
```
abooij, line 687: I think there's a negation too many in this sentence. Or I don't understand what it is saying.

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -15,6 +16,7 @@ import           Hasura.RQL.Types
 
 import qualified Hasura.Tracing                         as Tracing
 
+-- | Metadata storage transformer that enables raising @'QErr' exceptions.
```
abooij, line 19: It's clear from the definition 2 lines below this comment that it enables raising `QErr` exceptions. The question to answer is _why_ we want to raise `QErr`ors, and if this monad transformer represents anything else in addition to that.

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -32,38 +34,49 @@ runMetadataStorageT =
...
-Scheduled and Table event triggers have their own implementation
-and tables in Postgres. The core logic for both is same. So it
-is necessary to have a unified interface (maybe via a Postgres extension)
-where all event triggers implementation can extend from. By this we can
-fairly reduce the Postgres schema foot print and interactions to the database.
+Postgres tables' event triggers and scheduled event triggers are similar in the
+core logic. But currently, their implementation is completely isolated and do not share
+a common schema in Postgres. We're having a plan to simplify them via a common interface
+(maybe via a Postgres extension). This will potentially reduce number of interactions made
+to database and schema foot print.
```
abooij, line 41: Thanks for adding this note. Could you specify in this note what code it is referring to, e.g. which methods would need to get changed in this plan? How would it reduce the number of interactions made to the database, and how would it reduce the schema footprint?

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -32,38 +34,49 @@ runMetadataStorageT =
...
 
 TODO: Reference to open issue or rfc?
 -}
 
 -- | Metadata storage abstraction via a type class.
 --
+-- This type class enables storing and managing Hasura metadata in an isolated
+-- database which will not interfere with user's database where tables/functions
+-- are defined. Hence, it'll enable support for databases of multiple backends
+-- like MySQL, MSSQL etc.
```
abooij, line 51: Why have these methods been abstracted in a type class, rather than just being implemented once?

Why do these three areas belong together in one type class? I think an element of the answer to this second question could be (if I understand correctly) that the "global state" of graphql-engine consists exactly of these three things: metadata, scheduled triggers, and async actions.

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -32,38 +34,49 @@ runMetadataStorageT =
...
+-- 3. Async Actions
 -- ----------------
+-- TODO
 -- Operations to implement async actions sub-system. This includes recording an
 -- async action event and retreiving the details of action delivery to the webhook.
 -- For more details see Note [Async action architecture] in 'Hasura.GraphQL.Execute.Action' module.
+--
+-- It is believed that all the above three are implemented in a single storage
+-- system (ex: a Postgres database). We can split the functions into appropriate and
+-- specific type classes in future iterations if required.
```
abooij: What do these `TODO`s refer to? What is the thing that remains to be done?

server/src-lib/Hasura/Metadata/Class.hs
rakeshkky: All these questions will be answered in an RFC for the same. I'm not sure about specifically what methods will be changed/dropped. One thing we're sure that the common interface would simplify things related to eventing sub-system. 

cc @0x777

server/src-lib/Hasura/Metadata/Class.hs
rakeshkky: > Why have these methods been abstracted in a type class, rather than just being implemented once?

Implementation will be different for OSS and Pro/Cloud graphql engines. Pro/Cloud graphql-engine code is using OSS code as library with there own implementation. 

> Why do these three areas belong together in one type class?

https://github.com/rakeshkky/graphql-engine/blob/664c43b1526d63cc12886089bcbc4897c353e91b/server/src-lib/Hasura/Metadata/Class.hs#L77

server/src-lib/Hasura/App.hs
rakeshkky: @abooij Deriving `MonadMetadataStorage` instance for any generalized `MetadataStorageT m` raises overlapping instance if Pro/Cloud wants to have their own implementation of the same.

server/src-lib/Hasura/App.hs
```diff
@@ -632,42 +644,58 @@ instance HttpLog AppM where
...
+-- | Each of the function in the type class is executed in a totally separate transaction.
+instance MonadMetadataStorage (MetadataStorageT PGMetadataStorageApp) where
+
+  getDeprivedCronTriggerStats        = runInSeparateTx getDeprivedCronTriggerStatsTx
+  getScheduledEventsForDelivery      = runInSeparateTx getScheduledEventsForDeliveryTx
+  insertScheduledEvent               = runInSeparateTx . insertScheduledEventTx
+  insertScheduledEventInvocation a b = runInSeparateTx $ insertInvocationTx a b
+  setScheduledEventOp a b c          = runInSeparateTx $ setScheduledEventOpTx a b c
+  unlockScheduledEvents a b          = runInSeparateTx $ unlockScheduledEventsTx a b
+  unlockAllLockedScheduledEvents     = runInSeparateTx unlockAllLockedScheduledEventsTx
```
abooij, line 697:
```suggestion
-- | Each of the function in the type class is executed in a totally separate transaction.
--
-- TODO: We are not quite sure about the role of `MetadataStorageT` and `PGMetadataStorageApp`.
-- During the technical development, the following concerns got muddled a bit:
-- 1. We need a way to acquire a Postgres connection.
-- 2. We need a way to delimit the scope of a Postgres transaction.
-- 3. We need a way to handle errors that result from executing queries against Postgres.
-- 4. We need a way to select which implementation of metadata storage we want to use.
-- In summary, it is not clear if this instance has a desirable shape, and it may undergo
-- redesign at a later stage.
instance MonadMetadataStorage (MetadataStorageT PGMetadataStorageApp) where

  getDeprivedCronTriggerStats        = runInSeparateTx getDeprivedCronTriggerStatsTx
  getScheduledEventsForDelivery      = runInSeparateTx getScheduledEventsForDeliveryTx
  insertScheduledEvent               = runInSeparateTx . insertScheduledEventTx
  insertScheduledEventInvocation a b = runInSeparateTx $ insertInvocationTx a b
  setScheduledEventOp a b c          = runInSeparateTx $ setScheduledEventOpTx a b c
  unlockScheduledEvents a b          = runInSeparateTx $ unlockScheduledEventsTx a b
  unlockAllLockedScheduledEvents     = runInSeparateTx unlockAllLockedScheduledEventsTx
```

abooij (approved):
Approving this PR to unblock further work, despite [ongoing discussions](https://github.com/hasura/graphql-engine/pull/6131#discussion_r518453115) that we may wish to return to at a later stage.

server/src-lib/Hasura/App.hs
lexi-lambda: I think the idea that we have a shared `MetadataStorageT` for which different implementations provide instances like `MonadMetadataStorage (MetadataStorageT SomeBaseMonad)` is suspect. Generally, the idea would be that each implementation of `MonadMetadataStorage` would use a *different* monad transformer. This makes sense, because the transformer might need to have a different representation—currently `MetadataStorageT` is just a wrapper around `ExceptT`, and some other `MonadMetadataStorage` instance might need to track some other information in their metadata storage transformer.

What you say about each method running in its own transaction makes sense; that is all reasonable. But I still think the other points from my comment are relevant. The idea of “bracketing a Postgres transaction,” “implementing the metadata operations,” and “having a place to raise errors to” are all independent concerns. For example, suppose in the Haskell integration test suite you wanted to use Postgres for metadata storage, but you (a) only have a single Postgres connection and (b) your base monad already has a `MonadError` instance. It would be nice to be able to reuse `PGMetadataStorageT` to select the Postgres-backed metadata implementation and just have that use the “acquire a Postgres connection and run a transaction” and “throw errors” capabilities of the underlying monad, whatever they happen to be.

If you require a `MonadError` constraint on the `MonadMetadataStorage` instance, then call sites can easily satisfy it by adding an extra `runExceptT` call around `runPGMetadataStorageT`. The `ExceptT` part doesn’t have to be embedded into the base monad for that to work.

The current approach in this PR seems very *direct*: the reasoning seems to be “we needed a monad transformer to hang aa `MonadMetadataStorage` instance on, and that instance needed error handling, so we might as well put the error handling in that monad transformer.” In isolation, that reasoning isn’t necessarily bad, but it’s totally at odds with the idea that `MetadataStorageT` is supposed to be a *general-purpose* transformer for hanging many different `MonadMetadataStorage` instances on. Even ignoring the `ExceptT` detail, the idiomatic `mtl`-style approach is to have one typeclass (which is the interface) and a transformer per implementation. This PR does something unidiomatic, which seems ill-advised to me unless you have a really good reason for it.

rakeshkky: @lexi-lambda We do have `MonadMetadataStorage (MetadataStorageT m)` as type class constraint for the functions in which the type class methods being called (See https://github.com/hasura/graphql-engine/blob/74c04362322a15fc569b3b9d826b894f813bb878/server/src-lib/Hasura/Eventing/ScheduledTrigger.hs). The idiomatic `mtl`-style is possible if we some how declare those functions as `MonadMetadataStorage m`.

rakeshkky: @lexi-lambda Addressed this conversation in a source code note added via https://github.com/hasura/graphql-engine/pull/6131/commits/b9b97eff670cf2708dc95743c921d2f369be79ca.

server/src-lib/Hasura/Metadata/Class.hs
```diff
@@ -0,0 +1,135 @@
...
+1. Functions which implements @'MonadMetadataStorage' have
+   `MonadMetadataStorage (MetadataStorageT m)` as type class constraint.
+   Hence, we derive instance for any monad, let's say 'MyMonad', as
+
+   `instance MonadMetadataStorage (MetadataStorageT MyMonad) where`
+
+2. We use @'runMetadataStorageT' to catch exception thrown by the class methods
+   at call site.
+
+-}
```
lexi-lambda, line 117: Here’s my suggested rewording of this comment, as promised. I’ve changed the name of the `Note` to be a little more descriptive as to what’s actually discussed here, but I haven’t renamed the various references (since this is just a GH PR suggestion). If you accept this suggestion, please rename the references, too!

```suggestion
{- Note [Generic MetadataStorageT transformer]
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
All methods of the MonadMetadataStorage class may fail, which we represent in
the usual way using a MonadError superclass:

    class MonadError QErr m => MonadMetadataStorage m

However, unusually, the location where we pick a concrete MonadMetadataStorage
instance is not a context where we can handle errors, and as such the monad at
that point has no MonadError instance! Instead, clients of MonadMetadataStorage
are expected to handle errors /locally/, even though the code is parameterized
over an arbitrary metadata storage mechanism.

To encode this, we take a slightly unorthodox approach involving the auxiliary
MetadataStorageT transformer, which is really just a wrapper around ExceptT:

    newtype MetadataStorageT m a
      = MetadataStorageT { unMetadataStorageT :: ExceptT QErr m a }

We then define MonadMetadataStorage instances on a transformer stack comprising
both MetadataStorageT and a concrete base monad:

    instance MonadMetadataStorage (MetadataStorageT PGMetadataStorageApp)

This looks unconventional, but it allows polymorphic code to be parameterized
over the metadata storage implementation while still handling errors locally.
Such functions include a constraint of the form

    MonadMetadataStorage (MetadataStorageT m) => ...

and use runMetadataStorageT at the location where errors should be handled, e.g.:

    result <- runMetadataStorageT do
      {- ... some metadata operations ... -}
    case result of
      Left err -> ...
      Right value -> ...

In other words, runMetadataStorageT serves as a marker that says “I’m going to
handle exceptions raised by metadata operations right here,” which allows them
to be handled more locally than the point at which the concrete
MonadMetadataStorage instance (and thus the particular metadata storage
implementation) is actually chosen. -}
```

lexi-lambda (commented):
After my conversation last night with @rakeshkky, I understand the issues at play here much better. I’ve left a suggestion to reword the comment to be more explicit about exactly what’s going on, but otherwise I’m happy with this PR as it stands

PR #6133, opened 2020-11-04
adds haskell-language-server/ghcide configuration

0x777:
Configuration file for [ghcide](https://github.com/haskell/ghcide)/[hls](https://github.com/haskell/haskell-language-server).


abooij (approved):
We might wish to improve on this configuration later on, especially to avoid apparent race conditions between HLS and terminal builds, and to avoid rebuilds when manually building with `cabal build` after HLS loads. However, this looks like a good start that might help some people get started with HLS.

Is it possible to configure this to *not* watch all files, as most of our PCs do not seem prepared to deal with this?

tirumaraiselvan (approved):

PR #6137, opened 2020-11-05
Fix the way replace_metadata drops event triggers (fix #5461)

### LINKED ISSUE #5461: "table or event-trigger not found in schema cache" error in logs
Just upgraded from 1.2.2 to 1.3.0, noticed this following error message popping up more than once:

```
graphql-engine {"type":"event-trigger","timestamp":"2020-07-24T23:37:43.617+0000","level":"error","detail":{"path":"$","error":"table or event-trigger not found in schema cache","code":"unexpected"}}
```

How do we go about debugging this error / what would be the cause? Hasura still functions normally, but the log message spams almost every second or minute.

lexi-lambda:
### Description

This PR fixes #5461 by changing the way `replace_metadata` updates event triggers. Instead of dropping and recreating them, HGE now performs a diff-and-patch against the existing metadata, which goes through the appropriate code paths to update the associated SQL functions and archive events.

### Affected components

- Server

### Related Issues
#5461

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes
- [x] No Breaking changes


server/src-rsr/initialise.sql
```diff
@@ -288,7 +288,8 @@ CREATE TABLE hdb_catalog.event_triggers
   configuration JSON,
   comment TEXT,
   FOREIGN KEY (schema_name, table_name)
-  REFERENCES hdb_catalog.hdb_table(table_schema, table_name) ON UPDATE CASCADE
+    REFERENCES hdb_catalog.hdb_table(table_schema, table_name)
+    ON UPDATE CASCADE DEFERRABLE INITIALLY DEFERRED
```
tirumaraiselvan, line 292: shouldn't it be `INITIALLY IMMEDIATE` as per `server/src-rsr/migrations/40_to_41.sql`

tirumaraiselvan: Would it be less noisy to just say `DEFERRABLE` ? After all, we are adding this migration just so that the `SET CONSTRAINTS` can only work for `DEFERRABLE` constraints. The default behaviour for DEFERRABLE feels sufficiently obvious.

server/src-rsr/migrations/41_to_42.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE hdb_catalog.event_triggers
+  DROP CONSTRAINT event_triggers_schema_name_table_name_fkey;
```
tirumaraiselvan, line 2: This unfortunately breaks cascading effects when renaming tables. If you rename a table with event triggers, it throws inconsistent metadata and does not allow renaming.

codingkarthik (approved):
LGTM

rakeshkky:
@lexi-lambda 
With the metadata storage separation (#5797) the migration present in the PR becomes obsolete and event triggers are no longer stored in `hdb_catalog.event_triggers table`. I'll take this up. Assigning myself.

tirumaraiselvan:
@rakeshkky This PR should be merged regardless because we need to cherry-pick it for v1.3.3

The new event trigger changes should happen on top of this.

tirumaraiselvan:
`test_server_upgrade` is failing because circleci/postgres:11-alpine-postgis has:

```
postgres=# \d hdb_catalog.event_triggers;
          Table "hdb_catalog.event_triggers"
    Column     | Type | Collation | Nullable | Default 
---------------+------+-----------+----------+---------
 name          | text |           | not null | 
 type          | text |           | not null | 
 schema_name   | text |           | not null | 
 table_name    | text |           | not null | 
 configuration | json |           |          | 
 comment       | text |           |          | 
Indexes:
    "event_triggers_pkey" PRIMARY KEY, btree (name)
Foreign-key constraints:
    "event_triggers_schema_name_fkey" FOREIGN KEY (schema_name, table_name) REFERENCES hdb_catalog.hdb_table(table_schema, table_name) ON UPDATE CASCADE
```

rakeshkky (approved):
Server LGTM. However, these changes will become obsolete with metadata storage separation (see #5797).

tirumaraiselvan (approved):

PR #6145, opened 2020-11-06
server: add hlint hints to replace case analysis with combinators

abooij:
### Description

This PR implements 8 new hlint hints that enforce usage of our `onNothing` and `onLeft` combinators. Thus it further improves on #6126. Example hints provided by these new hlint hints:

```
src-lib/Hasura/RQL/Types/Common.hs:(292,20)-(294,57): Warning: Use onNothing
Found:
  case x of
    Just v -> return v
    Nothing -> fail "integer passed is out of bounds"
Perhaps:
  onNothing x (fail "integer passed is out of bounds")
```
```
src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs:(434,7)-(436,25): Warning: Use onLeft
Found:
  case eitherScalar of
    Left _ -> throwError $ InvalidGraphQLName $ toSQLTxt scalarType
    Right s -> pure s
Perhaps:
  onLeft
    eitherScalar
    (\ _c -> throwError $ InvalidGraphQLName $ toSQLTxt scalarType)
```

These new hints are triggered 23 times on the current master codebase.

This PR also removes some dead code, namely `coerceSet`, from `Hasura.Prelude`.

### Affected components

- [x] Server
- [x] Tests

### DIFF
--- a/server/src-lib/Hasura/Prelude.hs
+++ b/server/src-lib/Hasura/Prelude.hs
@@ -17,7 +17,6 @@ module Hasura.Prelude
   , liftEitherM
   -- * Efficient coercions
   , coerce
-  , coerceSet
   , findWithIndex
   , mapFromL
   , oMapFromL
@@ -81,13 +80,11 @@ import qualified Data.ByteString.Lazy              as BL
 import           Data.Coerce
 import qualified Data.HashMap.Strict               as Map
 import qualified Data.HashMap.Strict.InsOrd        as OMap
-import qualified Data.Set                          as Set
 import qualified Data.Text                         as T
 import qualified Data.Text.Encoding                as TE
 import qualified Data.Text.Encoding.Error          as TE
 import qualified GHC.Clock                         as Clock
 import qualified Test.QuickCheck                   as QC
-import           Unsafe.Coerce
 
 alphabet :: String
 alphabet = ['a'..'z'] ++ ['A'..'Z']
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/Prelude.hs
```diff
@@ -142,16 +139,6 @@ spanMaybeM f = go . toList
...
--- | Efficiently coerce a set from one type to another.
---
--- This has the same safety properties as 'Set.mapMonotonic', and is equivalent
--- to @Set.mapMonotonic coerce@ but is more efficient. This is safe to use when
--- both @a@ and @b@ have automatically derived @Ord@ instances.
---
--- https://stackoverflow.com/q/57963881/176841
-coerceSet :: Coercible a b=> Set.Set a -> Set.Set b
-coerceSet = unsafeCoerce
-
```
codingkarthik, line 154: Not sure why this is removed from here. Is this change relevant to this PR?

abooij: No, it's not relevant. Would you rather this change be merged separately?

codingkarthik: I guess it's just dead code removal. It's fine this way

codingkarthik (approved):
LGTM

server/src-lib/Hasura/Prelude.hs
nicuveo: We should check that Pro doesn't use it.

abooij: @nicuveo According to some quick `grep`s, it's not used by pro.

nicuveo: Perfect, then. LGTM! :)

nicuveo (approved):

tirumaraiselvan (approved):
no cl required

PR #6147, opened 2020-11-06
server: make LazyTxT strict

abooij:
### Description

The `LazyTxT` type was introduced to avoid connecting to Postgres when a given request did not require this. However, through the introduction of `ExecutionStep`s, this has now _mostly_ been taken care of in a different way, since no `LazyTxT` action is generated for GraphQL requests that do not fetch data from postgres.

This PR removes the laziness of `LazyTxT` by simply making it a newtype wrapper around `Q.TxET`. This simplifies a lot of code in `Hasura.Backends.Postgres.Connection`.

If I understood correctly, there are still some remaining RQL-style requests that would require a database connection through this change where they would have been able to avoid this through laziness before this PR. However, these requests are admin-only, so that this should have no user-visible impact.

### Changelog

### Affected components

- [x] Server

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes


abooij:
This has now been done in 438477fbc75bae20154728a30466cb5fac352ebf.

PR #6148, opened 2020-11-07
server: move MutationCTE code to Backends.Postgres.Translate

nicuveo:
### Description
This PR fixes an oversight of #6123: namely, it introduced postgres-specific code in `RQL.IR`, which is where our (soon to be) backend-agnostic query representation lives. This PR moves the code back to `Backends.Postgres`, and also fixes the imports of the impacted file.

I am not sure about adding this in `Postgres.Translate.Returning`, though? Maybe a new `Postgres.Translate.Types`?


abooij (approved):
The aim is for RQL to be an intermediate step between GraphQL and SQL, and should therefore be fairly backend-agnostic. This small change makes that happen.

We may wish to enforce RQL IR code to not depend on anything backend specific, and this discussion is taking place in #6151.

0x777 (approved):

PR #6149, opened 2020-11-07
server: remove postgres-specific code in Insert IR

nicuveo:
_(Note: this PR is on top of #6127.)_

### Description
This PR is one further step in our work to make the IR backend-agnostic; this time, tacking the representation of inserts. This PR does not change any behaviour, and therefore does not require a changelog.

### DIFF
--- a/server/src-lib/Hasura/RQL/Types/SchemaCacheTypes.hs
+++ b/server/src-lib/Hasura/RQL/Types/SchemaCacheTypes.hs
@@ -11,7 +11,7 @@ import           Data.Aeson.Types
 import           Data.Text.NonEmpty
 
 import           Hasura.Backends.Postgres.SQL.Types
-import           Hasura.RQL.Types.Common
+import           Hasura.RQL.Types.Common             hiding (ConstraintName)
 import           Hasura.RQL.Types.ComputedField
 import           Hasura.RQL.Types.EventTrigger
 import           Hasura.RQL.Types.Permission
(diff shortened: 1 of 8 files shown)


abooij (approved):
This PR adds some `hiding` import statements. Do these correspond to names that were duplicated before, but now their duplication is getting more visible? Can we somehow avoid these `hiding` qualifiers?

nicuveo:
Yeah, most of the hiding statements in this PR and the next ones are due to the generalized type in `Representation` having the same name as the specific type:

    instance Representation 'Postgres where
      ...
      ConstraintName 'Postgres = PG.ConstraintName

I'm hoping that those hiding statements are temporary (except, you know, famous last words...) because of two things:
  1. as we move forward, fewer parts of the code will end up having both in scope (PG code being isolated, RQL.Types being broken down)
  2. files that DO need PG-specific code will more consistently use qualified imports

server/src-lib/Hasura/RQL/Types/SchemaCacheTypes.hs
```diff
@@ -11,7 +11,7 @@ import           Data.Aeson.Types
 import           Data.Text.NonEmpty
 
 import           Hasura.Backends.Postgres.SQL.Types
-import           Hasura.RQL.Types.Common
+import           Hasura.RQL.Types.Common             hiding (ConstraintName)
```
0x777, line 14: I guess this is needed because Postgres's `ConstraintName` is used in this module? It looks like this is because `TableObjId` etc uses `ConstraintName`. I guess we'll have to move these types too to the Postgres's specific module. (Just a note)

0x777 (approved):

server/src-lib/Hasura/RQL/Types/SchemaCacheTypes.hs
0x777: My bad, already mentioned in an earlier comment.

nicuveo: Correct, yeah. There's a lot to clean in `RQL.Types`, and we will do that in due time. :)

PR #6150, opened 2020-11-07
server: remove postgres-specific code from OrderBy IR

nicuveo:
_(Note: this PR is on top of #6149.)_

### Description
This PR is one further step in our work to make the IR backend-agnostic; this time, tacking the representation of `OrderBy`. 

This PR does a few big changes:
- it deprecates the IR types for `OrderType` and `NullsOrder`, which were only thing `newtype` wrappers around specific types, and which were therefore not adding a lot of type safety
- it moves the types that are only used in the RQL DML in the according folder, including the ad-hoc FromJSON instance.


nicuveo:
Interestingly, it seems like all those JSON instances on IR types are only used in the case of RQL requests, not anywhere else?

server/src-lib/Hasura/RQL/IR/OrderBy.hs
```diff
@@ -111,52 +62,21 @@ orderByColFromTxt =
...
+deriving instance (Representation b, Lift a) => Lift (OrderByItemG b a)
+instance (Representation b, Hashable a) => Hashable (OrderByItemG b a)
+
+type OrderByItem b = OrderByItemG b OrderByCol
+
+instance (Representation b, FromJSON a) => FromJSON (OrderByItemG b a) where
+  parseJSON = genericParseJSON (aesonDrop 3 snakeCase){omitNothingFields=True}
+
+instance ToJSON a => ToJSON (OrderByItemG 'Postgres a) where
+  toJSON = genericToJSON (aesonDrop 3 snakeCase){omitNothingFields=True}
```
abooij: I like the pattern that the files in `Hasura.RQL` are assuming with these changes.

abooij (approved):
LGTM

0x777 (approved):

PR #6151, opened 2020-11-08
ci: add a github action that detects forbidden imports

nicuveo:
### Description
As we clean and contribute to the codebase, we will want to maintain some invariants that cannot be expressed at the language level. One such invariant is the fact that different subdirectories are meant to be dependencies of one another, with clear separation of purposes; and as such, some imports are "unlawful": they introduce an unwanted edge in the import graph, and are a code smell.

This PR introduces a new linter step on all PRs, that flags some uncontroversial edges as errors. **THIS SHOULD NOT BE SUBMITTED YET** as, at time of writing, this would break all open PRs: some ongoing work on the IR needs to be merged first to remove some of the unlawful edges.

### Limitations
The grepping mechanism is quite straightforward, and as such cannot be used yet to express something a bit more refined than "A isn't allowed in B". For instance, it's not possible to express "no file in Data is allowed to import a file in Hasura _except for Hasura.Prelude_" yet. Arguably, the Prelude could be considered an exception and grepped out of all matches?

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -38,3 +38,32 @@ jobs:
           then
             echo "$CHANGED_HS_FILES" | xargs ${{env.working-directory}}/hlint --json --hint=${{env.working-directory}}/.hlint.yaml | jq -r "$JQ_SCRIPT"
           fi
+
+  import-checks:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Check imports
+        shell: bash
+        run: |
+          FOUND_ERRORS=0
+          function check_imports() {
+              rootfolder="$1"
+              forbidden="$2"
+              message="$3"
+              for match in $(find "server/src-lib/$rootfolder" -type f -iname "*.hs" | xargs grep -Hn -E "^import.*$forbidden" | cut -d ':' -f 1,2); do
+                  filename=$(echo $match | cut -d ':' -f 1)
+                  line=$(echo $match | cut -d ':' -f 2)
+                  echo "::error file=$filename,line=$line::Forbiden import: $message"
+                  FOUND_ERRORS=1
+              done
+          }
+          check_imports "Hasura/RQL/IR" "Hasura.RQL.DML"  "the IR is backend-agnostic and cannot use RQL's DML."
+          check_imports "Hasura/RQL/IR" "Hasura.Backends" "the IR is backend-agnostic."
+          check_imports "Hasura/SQL"    "Hasura.RQL"      "SQL is a dependency of RQL."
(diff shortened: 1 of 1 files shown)


.github/workflows/hlint.yml
```diff
@@ -37,3 +37,32 @@ jobs:
...
+                  filename=$(echo $match | cut -d ':' -f 1)
+                  line=$(echo $match | cut -d ':' -f 2)
+                  echo "::error file=$filename,line=$line::Forbiden import: $message"
+                  FOUND_ERRORS=1
+              done
+          }
+          check_imports "Hasura/RQL/IR" "Hasura.RQL.DML"  "the IR is backend-agnostic and cannot use RQL's DML."
+          check_imports "Hasura/RQL/IR" "Hasura.Backends" "the IR is backend-agnostic."
+          check_imports "Hasura/SQL"    "Hasura.RQL"      "SQL is a dependency of RQL."
+          check_imports "Hasura/SQL"    "Hasura.GraphQL"  "SQL is a dependency of GraphQL."
```
abooij, line 67: These error messages can be improved, I think. Why is this an error, i.e. what did the CI step expect instead of what it found? Can it suggest an alternative? Here's an attempt:

```
Forbidden import: Hasura.SQL.[module] is not allowed to depend on Hasura.RQL.[module]: the SQL code should be usable by any intermediate representation. Perhaps the modules can be rearranged?
```

.github/workflows/hlint.yml
```diff
@@ -37,3 +37,32 @@ jobs:
...
+        run: |
+          FOUND_ERRORS=0
+          function check_imports() {
+              rootfolder="$1"
+              forbidden="$2"
+              message="$3"
+              for match in $(find "server/src-lib/$rootfolder" -type f -iname "*.hs" | xargs grep -Hn -E "^import.*$forbidden" | cut -d ':' -f 1,2); do
+                  filename=$(echo $match | cut -d ':' -f 1)
+                  line=$(echo $match | cut -d ':' -f 2)
+                  echo "::error file=$filename,line=$line::Forbiden import: $message"
```
abooij, line 60:
```suggestion
                  echo "::error file=$filename,line=$line::Forbidden import: $message"
```

.github/workflows/hlint.yml
```diff
@@ -37,3 +37,32 @@ jobs:
...
+
+      - name: Check imports
+        shell: bash
+        run: |
+          FOUND_ERRORS=0
+          function check_imports() {
+              rootfolder="$1"
+              forbidden="$2"
+              message="$3"
+              for match in $(find "server/src-lib/$rootfolder" -type f -iname "*.hs" | xargs grep -Hn -E "^import.*$forbidden" | cut -d ':' -f 1,2); do
```
abooij, line 57: To be on the safe side, it would be great if this would only test files that are touched by the PR.

Stefmore02:
Closing this PR because it has been open for more than 90 days with no activity.

PR #6152, opened 2020-11-08
server: simplify JSON instances

nicuveo:
### Description
A trivial cleanup PR: we have a lot of instances of `toJSON` and `fromJSON` that explicitly specify how many characters should be dropped from the field names, using `aesonDrop x`. However, there's a `aesonPrefix` function that does it automatically, by dropping all characters before the first uppercase one. This PR simply replaces all instances of `aesonDrop x` by `aesonPrefix`, except for the only instance in which the provided number yielded a different result.


nicuveo:
This will also avoid potential errors such as [this one](https://github.com/hasura/graphql-engine/commit/b2f88ff28a943e6140e03d974003cdaaabb7c674#diff-9a24ad5eb5e29e7a053a44ad24ae83855babb8a10274131affe2b317f4343809).

abooij:
A few days ago @jberryman suggested we may want to derive parsers that fail when they see unknown fields. Since this code touches such a similar set of lines in the codebase, do we maybe want to combine that proposal with this change?

nicuveo:
Hmmmm, I have conflicting opinions on this: on one hand, this is a no-op; there's no single behaviour change here; on the other, yes, might as well do both at once...

I was also thinking of introducing an alias in the Prelude, to unify all of those instances, wyt?

abooij:
@nicuveo Yes, the idea of defining our own default also came up in the discussion last week.

nicuveo:
Ok, I thought about it, I think it makes sense to do both changes at the same time; @jberryman how did you plan to do this? I'm not aware of a "reject objects with unknown fields" option in Aeson?

jberryman:
Cool, I didn't know about that! I see no reason not to merge this (after review) and tackle `rejectUnknownFields` later, if the work here is already done.

Also maybe we could use that refactoring tool thingy from facebook to make this easier?

nicuveo:
Our codebase is small enough that `sed` is my refactoring tool of choice. :P

But jokes aside: I was looking at the wrong version of the documentation, which is why I couldn't find `rejectUnknownFields`... I'll update this PR: it needs to be rebased anyway, so might as well do both at once!

jberryman:
@tirumaraiselvan do you want to weigh in on `rejectUnknownFields` as a potential breaking change?
My thought is it only breaks client code where:
- there's maybe a typo in a field (i.e. it's probably not doing what they think it is)
- we deprecated and removed a field but kept the same semantics

jberryman:
An aside: there's a lot of lint-y / format-y changes here that are unrelated to JSON instances. If automated tooling is doing this I think we should run them on the entire codebase at once and fix everything. I think I also wish we didn't care about formatting of imports; all I see in these import list changes is merge conflicts and git history noise...

server/src-lib/Hasura/Server/CheckUpdates.hs
```diff
@@ -18,16 +18,18 @@ import qualified Network.Wreq                as Wreq
...
+import           Hasura.Server.Version       (HasVersion, Version, currentVersion)
 
 
 newtype UpdateInfo
   = UpdateInfo
   { _uiLatest :: Version
   } deriving (Show)
 
+-- note that this is erroneous and should drop three characters or use
+-- aesonPrefix, but needs to remain like this for backwards compatibility
```
jberryman, line 32: good catch

server/src-lib/Hasura/Server/CheckUpdates.hs
```diff
@@ -53,12 +55,10 @@ checkForUpdates (LoggerCtx loggerSet _ _ _) manager = do
...
       return . buildUrl $ case ciM of
         Nothing -> "server"
-        Just ci -> "server-" <> (T.toLower . T.pack $ show ci)
+        Just ci -> "server-" <> T.toLower (tshow ci)
 
     -- ignoring if there is any error in response and returning the current version
-    decodeResp bs = case A.eitherDecode bs of
-      Left _  -> return $ UpdateInfo currentVersion
-      Right a -> return a
+    decodeResp bs = pure $ either (const $ UpdateInfo currentVersion) id $ A.eitherDecode bs
```
jberryman, line 61:
```suggestion
    decodeResp = pure . fromRight (UpdateInfo currentVersion) . A.eitherDecode
```

jberryman (approved):
Thanks for doing this!

nicuveo:
Yeah, we had decided against doing hlint on the entire repo, and doing it incrementally, because that kind of PRs is very difficult to merge, and very quickly outdated. The only downside of that approach is that large, sweeping PRs like this one get penalized... But it should converge over time. On that note: I have a PR to auto-format all PR so that no one has to do it manually anymore: https://github.com/hasura/graphql-engine-mono/pull/320 ^^

Would you mind approving it on the monorepo-side? https://github.com/hasura/graphql-engine-mono/pull/367.

PR #6153, opened 2020-11-09
server: split and clean IR.Select

nicuveo:
_(Note: this PR is on top of #6150.)_

### Description
This PR is one further step in our work to make the IR backend-agnostic; this time, tackling the representation of `Select`. 

This PR does a few big changes:
- it introduces `Backends.Postgres.Translate.Types` for types that are only used as intermediary representation steps during the translation to Postgres
- it adds more type to the `Representation` class and cleans up constraints
- it generalizes several instance declarations for which we still had an explicit `'Postgres`
- it makes a few names more explicit

### DIFF
--- a/server/src-lib/Hasura/RQL/IR/BoolExp.hs
+++ b/server/src-lib/Hasura/RQL/IR/BoolExp.hs
@@ -238,8 +238,8 @@ data OpExpG (b :: BackendType) a
   | ALIKE !a -- LIKE
   | ANLIKE !a -- NOT LIKE
 
-  | AILIKE (XAILIKE b) !a -- ILIKE, case insensitive
-  | ANILIKE (XANILIKE b) !a-- NOT ILIKE, case insensitive
+  | AILIKE () !a -- ILIKE, case insensitive
+  | ANILIKE () !a-- NOT ILIKE, case insensitive
 
   | ASIMILAR !a -- similar, regex
   | ANSIMILAR !a-- not similar, regex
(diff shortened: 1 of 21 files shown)


abooij (approved):
Commit bc1eea350999c904265554399c5f4d9f1157d307 (the main commit in this PR) LGTM.

server/src-lib/Hasura/RQL/IR/BoolExp.hs
```diff
@@ -238,8 +238,8 @@ data OpExpG (b :: BackendType) a
   | ALIKE !a -- LIKE
   | ANLIKE !a -- NOT LIKE
 
-  | AILIKE (XAILIKE b) !a -- ILIKE, case insensitive
-  | ANILIKE (XANILIKE b) !a-- NOT ILIKE, case insensitive
+  | AILIKE () !a -- ILIKE, case insensitive
```
0x777, line 241: I missed this in my previous review. What about having all the common operators across all databases like `_eq`, `_gt` etc. under `OpExpG` and then let databases extend this by providing an escape hatch with a type family? 

```
data OpExpG (b :: BackendType) a
  = ACast !(CastExp b a)
  | AEQ !Bool !a
  | ANE !Bool !a
  | AIN  !a
  | ANIN !a
  | ABackend !(BackendOp b a)
```

nicuveo: That's a good idea. Alternatively (or in addition?), we could also choose to use the same IR for both (simply LIKE and NLIKE), regardless of actual implementation?
I have no strong feeling on this, and not enough knowledge of the subtle differences between SQL flavours, so I'll let y'all decide what's best.

abooij: Hold up, I also missed this in the first review. The point of `XAILIKE` is exactly to make the `AILIKE` constructor accessible exactly for those databases that support it. This is along the lines of the "Trees that grow" approach: give every constructor an extra configurable field, which, if set to `Void`, makes the constructor unusable. Additionally, provide an additional constructor, in this case it'd be named `XOpExpG`, with a type family `XXOpExpG`, that allows any backend to specify its own additional "constructors".

Before this PR, we had the former. Vamshi is proposing something along the lines of the latter, but with a different naming convention. I think there is some value in following the naming convention of "Trees that grow" for now.

But I did not notice that this PR got rid of this `XANILIKE`. That doesn't seem like the right thing to do. And if we want to get rid of it, then let's remove the constructor field altogether rather than replacing it with `()`.

abooij: See the data type `HsExpr` [here](https://haskell-code-explorer.mfix.io/package/ghc-8.6.5/show/hsSyn/HsExpr.hs) for an example. A given constructor, such as `HsTick`, is given a field of type `XTick p`, where `XTick` is a type family that allows that particular constructor to be switched off for a certain GHC phase. Additionally, there's an `XExpr` constructor with an `XXExpr p` field, where the `XXExpr` type family specifies additional `HsExpr` values beyond the listed constructors in that data type.

nicuveo: I'll move this to be part of the `Backend` class then, I'd rather have everything together.

0x777: @nicuveo I'm okay with just removing the `()` from `AILIKE` and `ANILIKE`. We can introduce a new constructor with MSSQL server related code.

nicuveo: @0x777 I did what I should have done in the first place: keep the code the way it was, but simply move the type family to the Backend class.

0x777: @nicuveo Sorry I wasn't clear in my earlier comment, of the two approaches that @abooij  outlined, I prefer the latter, i.e, additional constructor rather than an extra field per constructor as this would quickly explode - the number of type families in backend class for operators would the union of all the operators supported across all databases. That being said, the approach that we take is not that important right now, we'll know better once we have another backend, so I don't mind deferring this decision.

0x777 (approved):

server/src-lib/Hasura/RQL/IR/BoolExp.hs
abooij: @0x777 100% agreed! I initially added `A(N)ILIKE` as a "hotfix", and you are right that we may wish to primarily "grow" the `OpExpG` data type in the latter way. It's simply that "Trees that grow" provides two options here, which overlap to some extent.

PR #6155, opened 2020-11-09
server: generalize PGColumnType

nicuveo:
_(Note: this PR is on top of #6153.)_

### Description
This PR is one further step in our work to make the IR backend-agnostic; this time, changing `PGColumnType` into a more generic `ColumnType`. There's a lot more work to be done regarding postgres types, and to further decouple postgres from RQL, but this is a first step.


abooij (approved):
Commit b0ad75582f457bd9907af35241685b68c6794771 LGTM.

PR #5858, opened 2020-09-29
Support tracking VOLATILE SQL functions as mutations. Closes #1514

jberryman:
### Description

Support tracking `VOLATILE` functions as mutations (top-level fields under `mutation` root field). Support is added to the v2 API while remaining backwards compatible. See comments for API design decisions.

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR. If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

https://github.com/hasura/graphql-engine/issues/1514

### Solution and Design

I re-used the machinery from support for tracking functions as queries and tried to do as little as possible.

### Steps to test and verify

The tests written here represent the extent that I've tested/tried this out, so you can run those.

### Limitations, known bugs & workarounds

The tests aren't particularly thorough; since so much of the queries code is re-used this seems all right to me.

Aggregations on the result set isn't implemented; it might be as simple as uncommenting the relevant line (I just noticed this as an unresolved question).

#### Catalog upgrade

Does this PR change Hasura Catalog version?
- [x] No
- [ ] Yes
  - [ ] Updated docs with SQL for downgrading the catalog 

#### Metadata

Does this PR add a new Metadata feature?
- [x] No
- [ ] Yes
  - Does `run_sql` auto manages the new metadata through schema diffing?
    - [ ] Yes
    - [ ] Not required
  - Does `run_sql` auto manages the definitions of metadata on renaming?
    - [ ] Yes
    - [ ] Not required
  - Does `export_metadata`/`replace_metadata` supports the new metadata added?
    - [ ] Yes
    - [ ] Not required

#### GraphQL
- [x] No new GraphQL schema is generated
- [ ] New GraphQL schema is being generated:
   - [ ] New types and typenames are correlated
   
   

#### Breaking changes

- [x] No Breaking changes
- [ ] There are breaking changes:

  1. Metadata API

     Existing `query` types:
     - [ ] Modify `args` payload which is not backward compatible
     - [ ] Behavioural change of the API
     - [ ] Change in response `JSON` schema
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     - [ ] Change in any `NamedType`
     - [ ] Change in table field names
     

     Schema Resolve:-
     
[...]

PR #6160, opened 2020-11-10
Support tracking SQL functions as mutations. Closes #1514

### LINKED ISSUE #1514: extend support for custom SQL functions (mutations, other functions types)
1. Support for SQL functions in mutations
2. Support other function types

jberryman:
### Description

Support tracking functions as mutations (top-level fields under `mutation` root field). Support is added to the v2 API while remaining backwards compatible. See comments for API design decisions.

This is the second version of this feature implemented. The first is at: https://github.com/hasura/graphql-engine/pull/5858

- new field (replacing `as_mutation`): `expose_as: query` or `expose_as: mutation`
    - if included we'll place the function where requested, regardless of its volatility
    - ...with the exception that _non-volatile functions requested exposed as mutations will raise an error_, since this probably indicates a mistake either in choice of function, or when defining the function
    - ~we'll raise a warning in some way whenever the volatility and `expose_as` are "mismatched"~ **UPDATE:** I punted on this
- if that field is omitted, we'll infer `expose_as` from the volatility (`VOLATILE` implies mutation, otherwise query)

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [x] Tests
- [ ] Other (list it)

### Related Issues

https://github.com/hasura/graphql-engine/issues/1514

### Solution and Design

I re-used the machinery from support for tracking functions as queries and tried to do as little as possible.

### Steps to test and verify

The tests written here represent the extent that I've tested/tried this out, so you can run those.

The tests aren't particularly thorough; since so much of the queries code is re-used this seems all right to me.

Aggregations on the result set isn't implemented; it might be as simple as uncommenting the relevant line (I just noticed this as an unresolved question).

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/docs/graphql/core/api-reference/schema-metadata-api/custom-functions.rst
+++ b/docs/graphql/core/api-reference/schema-metadata-api/custom-functions.rst
@@ -24,7 +24,7 @@ Only tracked custom functions are available for querying/mutating/subscribing da
 track_function
 --------------
 
-``track_function`` is used to add a custom SQL function to the GraphQL schema.
+``track_function`` is used to add a custom SQL function to the ``query`` root field of the GraphQL schema.
 Also refer a note :ref:`here <note>`.
 
 Add an SQL function ``search_articles``:
@@ -48,10 +48,12 @@ Add an SQL function ``search_articles``:
 track_function v2
 -----------------
 
-Version 2 of ``track_function`` is used to add a custom SQL function to the GraphQL schema with configuration.
+Version 2 of ``track_function`` is used to add a custom SQL function to the GraphQL schema.
+It supports more configuration options than v1, and also supports tracking
+functions as mutations.
 Also refer a note :ref:`here <note>`.
 
-Add an SQL function called ``search_articles`` with a Hasura session argument.
+Track an SQL function called ``search_articles`` with a Hasura session argument:
 
 .. code-block:: http
 
@@ -72,6 +74,48 @@ Add an SQL function called ``search_articles`` with a Hasura session argument.
            }
        }
    }
+
+Track ``VOLATILE`` SQL function ``reset_widget`` as a mutation, so it appears
+as a top-level field under the ``mutation`` root field:
+
+.. code-block:: http
+
+   POST /v1/query HTTP/1.1
+   Content-Type: application/json
+   X-Hasura-Role: admin
+
+   {
+       "type": "track_function",
+       "version": 2,
+       "args": {
+           "function": {
+               "schema": "public",
+               "name": "reset_widget"
+           },
+           "configuration": {
+               "exposed_as": "mutation"
+           }
+       }
+   }
+
+If ``exposed_as`` is omitted, the location in the schema to expose the function
+will be inferred from the function's volatility, with ``VOLATILE`` functions
+appearing under the ``mutation`` root, and others ending up under
+``query/subscription``.
+
+In most cases you will want ``VOLATILE`` functions to only be exposed as
+mutations, and only ``STABLE`` and ``IMMUTABLE`` functions to be queries.
+When tracking ``VOLATILE`` functions under the ``query`` root, the user needs
+to guarantee that the field is idempotent and side-effect free, in the context
+of the resulting graphql API. 
+
+One such use case might be a function that wraps a simple query and performs
+some logging visible only to administrators.
+
+.. note::
+
+   It's easy to accidentally give an SQL function the wrong volatility (or for a
+   function to end up with ``VOLATILE`` mistakenly, since it's the default).
 
 .. _track_function_args_syntax_v2:
 
@@ -110,6 +154,10 @@ Function Configuration
      - false
      - `String`
      - Function argument which accepts session info JSON
+   * - exposed_as
+     - false
+     - `String`
+     - In which part of the schema should we expose this function? Either "mutation" or "query".
 
 .. _note:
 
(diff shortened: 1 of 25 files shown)


server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -171,11 +179,16 @@ takeValidTables = Map.filterWithKey graphQLTableFilter . Map.filter tableFilter
...
       || (isJust . _tcCustomName . _tciCustomConfig . _tiCoreInfo $ tableInfo)
 
+-- TODO and what about graphql-compliant function names here too?
 takeValidFunctions :: FunctionCache -> [FunctionInfo]
 takeValidFunctions = Map.elems . Map.filter functionFilter
   where
     functionFilter = not . isSystemDefined . fiSystemDefined
 
+takeExposedAs :: FunctionExposedAs -> [FunctionInfo] -> [FunctionInfo]
+takeExposedAs x = filter ((== x) . fiExposedAs)
```
abooij, line 189: This is a nice combinator.

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -171,11 +179,16 @@ takeValidTables = Map.filterWithKey graphQLTableFilter . Map.filter tableFilter
...
+
+
```
abooij, line 191:
```suggestion

```

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -347,12 +362,14 @@ buildPostgresQueryFields allTables allFunctions = do
...
-    requiredFieldParser f = fmap $ Just . fmap f
-
     mapMaybeFieldParser :: (a -> b) -> m (Maybe (P.FieldParser n a)) -> m (Maybe (P.FieldParser n b))
     mapMaybeFieldParser f = fmap $ fmap $ fmap f
 
+requiredFieldParser 
+  :: (Functor n, Functor m)=> (a -> b) -> m (P.FieldParser n a) -> m (Maybe (P.FieldParser n b))
+requiredFieldParser f = fmap $ Just . fmap f
+
+
```
abooij, line 372:
```suggestion
```

server/src-lib/Hasura/GraphQL/Schema.hs
```diff
@@ -626,17 +643,40 @@ buildMutationParser
   => [P.FieldParser n RemoteField]
   -> [ActionInfo 'Postgres]
   -> NonObjectTypeMap
+  -> [FunctionInfo]
+  -- ^ all "valid" functions
   -> [P.FieldParser n (MutationRootField UnpreparedValue)]
   -> m (Maybe (Parser 'Output n (OMap.InsOrdHashMap G.Name (MutationRootField UnpreparedValue))))
-buildMutationParser allRemotes allActions nonObjectCustomTypes pgMutationFields = do
+buildMutationParser allRemotes allActions nonObjectCustomTypes allFunctions pgMutationFields = do
+  let mutationFunctions = takeExposedAs FEAMutation allFunctions
```
abooij, line 651: For consistency:

```suggestion
buildMutationParser allRemotes allActions nonObjectCustomTypes (takeExposedAs FEAMutation -> mutationFunctions) pgMutationFields = do
```

server/src-lib/Hasura/RQL/DDL/Schema/Function.hs
```diff
@@ -89,7 +89,22 @@ mkFunctionInfo qf systemDefined config rawFuncInfo =
...
+      -- We mostly take the user at their word here and will, e.g. expose a
+      -- function as a query if it is marked VOLATILE (since perhaps the user
+      -- is using the function to do some logging, say). But this is also a
+      -- footgun we'll need to try to document (since `VOLATILE` is default
+      -- when volatility is omitted). See the original approach here:
+      -- https://github.com/hasura/graphql-engine/pull/5858 
+      --
+      -- This is the one exception where we do some validation. We're not
+      -- commited to this check, and it would be backwards compatible to remove
+      -- it, but this seemed like an obvious case:
```
abooij, line 101: Do you think this documentation has been implemented adequately?

Perhaps, in order to avoid footgunning, can we output some kind of warning message when users expose a `VOLATILE` function as a `query` field?

docs/graphql/core/api-reference/schema-metadata-api/custom-functions.rst
```diff
@@ -73,6 +75,39 @@ Add an SQL function called ``search_articles`` with a Hasura session argument.
...
+
+If ``exposed_as`` is omitted, the location in the schema to expose the function
+will be inferred from the function's volatility, with ``VOLATILE`` functions
+appearing under the ``mutation`` root, and others ending up under
+``query/subscription``.
+
+.. note::
+
+   It's easy to accidentally give an SQL function the wrong volatility (or for a
+   function to end up with ``VOLATILE`` mistakenly, since it's the default).
```
abooij, line 109: I think this deserves one or two more sentences. Could you add here what would be the scenarios that would make sense (i.e. non-volatile as query, and volatile as mutation), and indicate how the scenario that's odd (volatile as query would be odd, since queries are thought of as non-mutating) and why they are supported regardless (e.g. logging)?

abooij (approved):
Thanks, Brandon, this is a really nice PR, and I hope you manage to merge this soon. I have lefta few notes for you to use as you wish - from my point of view they're not blockers.

Are you sure you want to tell kodiak to "merge" this PR, rather than "squash" it (which is the default if you only apply the `auto-update-auto-merge` label)?

server/src-lib/Hasura/RQL/DDL/Schema/Function.hs
jberryman: Yeah I decided to punt on; I was on the fence anyway, and there wasn't a `logger` available in nearby callers

docs/graphql/core/api-reference/schema-metadata-api/custom-functions.rst
jberryman: Thanks, added a bit of elaboration here.

jberryman:
> Are you sure you want to tell kodiak to "merge" this PR, rather than "squash" it (which is the default if you only apply the auto-update-auto-merge label)?

yep

PR #6164, opened 2020-11-10
ci: add hlint escape hatch

nicuveo:
### Description
This PR changes our check configuration:
- the changelog check will be re-run whenever a label changes
- the hlint check will be skipped if a label named `ignore-server-lint-checks` is present
- (it also adds one new hlint check)

This is hard to test, given that `label` events are only fired using the configuration as it is on main, according to the documentation. At the very least this runs, which means that even if it does not work as expected it won't break our workflows.

If someone has any experience with actions and can check this, that'd be great!

### DIFF
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -1,9 +1,9 @@
 name: server lint
 on:
-  label
-  pull_request:
-    paths:
-      - 'server/**'
+  - label
+  - pull_request:
+      paths:
+        - 'server/**'
 
 jobs:
   hlint:
--- a/.github/workflows/hlint.yml
+++ b/.github/workflows/hlint.yml
@@ -1,5 +1,6 @@
 name: server lint
 on:
+  label:
   pull_request:
     paths:
       - 'server/**'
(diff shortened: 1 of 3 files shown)


.github/workflows/hlint.yml
```diff
@@ -1,20 +1,21 @@
...
 on:
+  label:
   pull_request:
     paths:
       - 'server/**'
 
 jobs:
   hlint:
     runs-on: ubuntu-20.04
+    if: !contains(github.event.pull_request.labels.*.name, 'ignore-lint-checks')
```
tirumaraiselvan, line 11: calling it `hlint` will disambiguate from other lints?

nicuveo: SG! Sorry, this is still a draft, I'm testing that this works as expected, the documentation for actions is a bit lackluster. ^^

nicuveo: I might settle for `ignore-server-lint-checks`, since I'm working on other lint steps that are not hlint I'd like to be able to disable if necessary.

abooij (commented):
I think I agree with @tirumaraiselvan's proposal for somehow putting `hlint` in the name of the label. A PR might touch both the console and the server, for instance, and we wouldn't want to get mixed up in which code linter needs to be switched off. I think it'd be preferable to have many different labels for many different linters than to have one big on/off switch, that may end up only switching off part of the linters.

PR #6173, opened 2020-11-11
schema cache syncing without relying on a Postgres table

rakeshkky:
### Description

This is an incremental PR towards #5797. It avoids using `hdb_cataog.hdb_schema_update` table and associated trigger to notify schema syncing events to other server instances. Instead it is achieved in a much simpler way by directly using `pg_notify` procedure to publish event payload.

Other things this PR does:-
- Now `initialiseCtx` function is only used in `serve` command
- Using new `GlobalCtx` type for essentials required for all graphql-engine commands

### Changelog

### Affected components

- [x] Server

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### Breaking changes

- [x] No Breaking changes


server/src-lib/Hasura/App.hs
```diff
@@ -156,6 +155,23 @@ mkPGLogger :: Logger Hasura -> Q.PGLogger
 mkPGLogger (Logger logger) (Q.PLERetryMsg msg) =
   logger $ PGLog LevelWarn msg
 
+-- | Context required for all graphql-engine commands
```
abooij, line 158: What does "all graphql-engine commands" mean? Does this refer to CLI commands? graphql-engine supports a variety of commands through a variety of interfaces and protocols :-)

server/src-lib/Hasura/App.hs
```diff
@@ -156,6 +155,23 @@ mkPGLogger :: Logger Hasura -> Q.PGLogger
...
+data GlobalCtx
+  = GlobalCtx
+  { _gcHttpManager :: !HTTP.Manager
+  , _gcInstanceId  :: !InstanceId
```
abooij, line 162: `_gcInstanceId` is only used by `serve`, right? And `InitCtx` is also only used by `serve`, right? If these two are right, then perhaps we can move `_gcInstanceId` to `InitCtx`, and rename `InitCtx` to something more fitting, such as `ServeCtx`.

server/src-lib/Hasura/App.hs
```diff
@@ -181,71 +198,64 @@ data Loggers
 newtype AppM a = AppM { unAppM :: IO a }
   deriving (Functor, Applicative, Monad, MonadIO, MonadBase IO, MonadBaseControl IO, MonadCatch, MonadThrow, MonadMask)
 
--- | this function initializes the catalog and returns an @InitCtx@, based on the command given
--- - for serve command it creates a proper PG connection pool
--- - for other commands, it creates a minimal pool
--- this exists as a separate function because the context (logger, http manager, pg pool) can be
--- used by other functions as well
+-- | Initializes the catalog and returns an @InitCtx@ for given @'ServeOptions'
```
abooij, line 201: As a rule of thumb: a comment that does little more than spell out the Haskell type of the thing being documented is not documentation. If the comments contain information that can be extracted from the Haskell code itself, they're not adding information, just duplicating it.

[Here](https://carlanderson.xyz/how-to-write-useful-comments/)'s one of many guides you might find useful to read about writing comments.

abooij (commented):
I can't reliably judge the syncing logic itself. It seems to me that the core of tho changes are in `fetchLastUpdate`, `recordSchemaUpdate`, and `notifySchemaCacheSync`. With the old `fetchLastUpdate` and `recordSchemaUpdate`, I think I understood the logic: `record` sends a message to all other graphql-engine instances, and `fetch` received messages. Is `notify` the new `record` or the new `fetch`? I believe it's the former. If so, where's the new `fetch`? Where's the code that receives these PG notifications?

Edit: Sorry, perhaps the failing PR was just a flake. [This used to say: @rakeshkky, please invite reviewers *after* you got the PR to pass CI. Otherwise, you ensure that the reviewer does *not* review the code that will end up getting merged.]

server/src-lib/Hasura/App.hs
rakeshkky: > Does this refer to CLI commands? 

Yes.

abooij: Could you please clarify this in the comment?

rakeshkky:
@0x777 Please take a look at this PR. You have the better context of the schema syncing process than other reviewers.

PR #6179, opened 2020-11-12
archive invalid events and drop existing invalid postgres triggers

tirumaraiselvan:
Although #6137 fixes the problem of dropping event triggers, we need to run additional steps in the migration for ppl already having orphaned triggers (and events) in their db.


codingkarthik (approved):
LGTM

PR #6190, opened 2020-11-12
Enable HASURA_GRAPHQL_PG_CONN_LIFETIME by default with a default of 1…

jberryman:
…20 sec

See: https://github.com/hasura/pg-client-hs/pull/26

We're still experiencing segfaults in certain configurations from the
libpq shrinking approach. This should be very safe, but is also
inadequate in several ways:

- negotiating a new connection is quite heavyweight compared to
  allocating a 1mb buffer
- thundering herd issues (TODO maybe add some jitter here)
- memory is not reclaimed promptly; there is no good default here.

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/Server/Init.hs
```diff
@@ -201,7 +201,10 @@ mkServeOptions rso = do
       -- hard throughput cap at 1000RPS when db queries take 50ms on average:
       conns <- fromMaybe 50 <$> withEnv c (fst pgConnsEnv)
       iTime <- fromMaybe 180 <$> withEnv i (fst pgTimeoutEnv)
-      connLifetime <- withEnv cl (fst pgConnLifetimeEnv)
+      connLifetime <- withEnv cl (fst pgConnLifetimeEnv) <&> \case
+        Nothing -> Just 120 -- Not set by user; use the default timeout
```
tirumaraiselvan, line 205: The server_lifetime is really high for pgbouncer: https://www.pgbouncer.org/config.html. 3600 seconds i.e. 1 hour.

Considering the rough heuristic that a high memory connection will also take longer time to execute, should we set this to something like 600?

tirumaraiselvan: At the least, it should be higher than the idle timeout (180s) ?

jberryman: I'm not sure there's any good solution: if it's higher than, say, 10 seconds then the user won't observe memory going down after a big result set. But 10 seconds would probably have a huge performance impact and be unacceptable. Anything over a very small number would also make an OOM situation much more likely it seems.

So I guess I don't know?... I'm fine changing it to 600 if you have an intuition that's better

tirumaraiselvan: @jberryman I do think having a higher number to start with is better than starting with a low number. We can then fine tune this based on user reports later (afaict only [one report](https://github.com/hasura/graphql-engine/issues/5087#issuecomment-654570203) was directly related to huge data fetch).  10 minutes just seems like a good compromise between pgbouncer's config and OOM probabilities. Coupled with the fact that not many users are not fetching GBs of data in one go.

0x777: 10minutes sounds good to me.

PR #6227, opened 2020-11-19
server: support remote relationship with ID scalar types

codingkarthik:
This PR will allow users to create remote relationship with `ID` scalar types. As, Hasura doesn't generate the `ID` scalar type, we're allowing `String` and `Int` (and related types like `bigint`, `smallint`) to join with the `ID` scalar type

### Changelog

### Affected components

- [x] Server
- [x] Tests

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -409,7 +412,16 @@ isTypeCoercible actualType expectedType =
...
   in
-  if | actualBaseType /= expectedBaseType -> raiseValidationError
+  if | expectedBaseType == $$(G.litName "ID") ->
+         bool (throwError $ IDTypeJoin actualBaseType)
+              (pure ())
+              -- Check under `Input Coercion` https://spec.graphql.org/June2018/#sec-ID
+              -- We can also include the `ID` type in the below list but it will be
+              -- extraneous because at the time of writing this, we don't generate
+              -- the `ID` type in the DB schema
+              (G.unName actualBaseType `elem`
```
tirumaraiselvan, line 422: We are being arbitrarily lax in our type system. So why not all scalars (in any case, we need `uuid`)? Should we make this a configuration element "coerce into ID type"?

tirumaraiselvan: We should think what happens when hasura does support `ID` type. I guess the schema customisation roadmap will allow something like "make this column ID type" (and it will validate that the column has uniqueness constraint)

codingkarthik: I guess we can extend the coercible types here to custom scalar types, `Int` and `String` types? I'm consciously not supporting the `Boolean` and `Float` types because the spec says that when these values are received as input values, the remote schema should throw an error.

> When expected as an input type, any string (such as "4") or integer (such as 4) input value should be coerced to ID as appropriate for the ID formats a given GraphQL server expects. Any other input value, including float input values (such as 4.0), must raise a query error indicating an incorrect type.

codingkarthik: > Should we make this a configuration element "coerce into ID type"
Yes, I agree. That will be better, but will that only be for remote relationships or will the type of that column be changed in the schema also?

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -409,7 +412,16 @@ isTypeCoercible actualType expectedType =
   let (actualBaseType, actualNestingLevel) = getBaseTyWithNestedLevelsCount actualType
       (expectedBaseType, expectedNestingLevel) = getBaseTyWithNestedLevelsCount expectedType
   in
-  if | actualBaseType /= expectedBaseType -> raiseValidationError
+  if | expectedBaseType == $$(G.litName "ID") ->
+         bool (throwError $ IDTypeJoin actualBaseType)
+              (pure ())
+              -- Check under `Input Coercion` https://spec.graphql.org/June2018/#sec-ID
```
jberryman, line 418: Thanks for this comment :+1:

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
```diff
@@ -409,7 +412,16 @@ isTypeCoercible actualType expectedType =
...
+              -- We can also include the `ID` type in the below list but it will be
+              -- extraneous because at the time of writing this, we don't generate
+              -- the `ID` type in the DB schema
+              (G.unName actualBaseType `elem`
+               [ "Int", "String", "bigint", "smallint" ])
```
jberryman, line 423:
```suggestion
          -- Check under `Input Coercion` https://spec.graphql.org/June2018/#sec-ID
          -- We can also include the `ID` type in the below list but it will be
          -- extraneous because at the time of writing this, we don't generate
          -- the `ID` type in the DB schema
         unless (actualBaseType `elem` [ "Int", "String", "bigint", "smallint" ]) $
           throwError $ IDTypeJoin actualBaseType
```
since `Name` is instance `IsString` here you can omit the `unName` (I think!)

jberryman (approved):
LGTM, just the one suggestion if you agree.

server/src-lib/Hasura/RQL/DDL/RemoteRelationship/Validate.hs
tirumaraiselvan: I guess coercion is expected as per the spec. Lets just add `uuid` type to this list?

tirumaraiselvan (approved):

PR #6286, opened 2020-12-02
server: do not block catalog migration on inconsistent metadata

codingkarthik:
Earlier, we used to terminate the graphql-engine process during the startup if there were any inconsistent metadata, this only happened when there were some catalog migration(s). In the other case, when there are no catalog migrations there to be applied, we don't terminate the process, this PR replicates the same behaviour for when there are catalog migration(s) to be applied.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Steps to test and verify

To manually test this,

1. Setup a local GraphQL server and add it as a remote schema in the graphql-engine.
2. Shut down the graphql-engine create a dummy migration and increment the `catalog_version.txt`.
3. Shutdown the local GraphQL server and then start the graphql-engine.
4. The graphql-engine should have been started up successfully.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/Server/Migrate.hs
+++ b/server/src-lib/Hasura/Server/Migrate.hs
@@ -164,9 +164,7 @@ migrateCatalog env migrationTime = do
     buildCacheAndRecreateSystemMetadata :: m (RebuildableSchemaCache m)
     buildCacheAndRecreateSystemMetadata = do
       schemaCache <- buildRebuildableSchemaCache env
-      case scInconsistentObjs (lastBuiltSchemaCache schemaCache) of
-        [] -> view _2 <$> runCacheRWT schemaCache recreateSystemMetadata
-        _  -> pure schemaCache
+      view _2 <$> runCacheRWT schemaCache recreateSystemMetadata
 
 downgradeCatalog :: forall m. (MonadIO m, MonadTx m) => DowngradeOptions -> UTCTime -> m MigrationResult
 downgradeCatalog opts time = do
@@ -333,7 +331,7 @@ recreateSystemMetadata = do
     for_ tableRels \case
       Left relDef  -> insertRelationshipToCatalog tableName ObjRel relDef
       Right relDef -> insertRelationshipToCatalog tableName ArrRel relDef
-  buildSchemaCacheStrict
+  buildSchemaCache
   where
     systemMetadata :: [(QualifiedTable, [Either ObjRelDef ArrRelDef])]
     systemMetadata =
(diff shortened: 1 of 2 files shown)


server/src-lib/Hasura/Server/Migrate.hs
```diff
@@ -164,7 +164,9 @@ migrateCatalog env migrationTime = do
     buildCacheAndRecreateSystemMetadata :: m (RebuildableSchemaCache m)
     buildCacheAndRecreateSystemMetadata = do
       schemaCache <- buildRebuildableSchemaCache env
-      view _2 <$> runCacheRWT schemaCache recreateSystemMetadata
+      case scInconsistentObjs (lastBuiltSchemaCache schemaCache) of
+        [] -> view _2 <$> runCacheRWT schemaCache recreateSystemMetadata
+        _  -> pure schemaCache
```
codingkarthik, line 169: I'm not sure if this is the right solution. I tried to replicate the behaviour of when `previousVersion == latestCatalogVersionString` in the `migrateFrom` function in the `migrateCatalog` function. 

Calling the `runCacheRWT` function with a `schemaCache` containing inconsistent metadata throws an error, I don't understand the `runCacheRWT` function well enough to know why that happens.

@lexi-lambda Please do let me know if there is a better way to solve this issue 😄

lexi-lambda: This is definitely *not* the right solution: we can’t just skip updating the system metadata if there are inconsistent objects! Then the system metadata might be wrong/out of date, and it wouldn’t be fixed until the next migration.

The reason an exception is raised is that `recreateSystemMetadata` calls `buildSchemaCacheStrict`, which raises an error on inconsistent metadata. Fortunately, this has a simple solution: call `buildSchemaCache` instead, which does not raise an error on inconsistent metadata.

lexi-lambda (changes requested):

server/src-lib/Hasura/Server/Migrate.hs
codingkarthik: Oh, I didn't know that's what `buildSchemaCache` is used for 😅 . I have fixed this. Thanks!

0x777 (approved):

server/src-lib/Hasura/Server/Migrate.hs
lexi-lambda: No worries! No harm no foul, and the updated patch looks good to me.

lexi-lambda (approved):

PR #5783, opened 2020-09-14
RFC: Add virtual/contextual roles

webdeb:
This is a proposal for adding `virtual roles`. It's aim is to split rules logic into explicit definitions for "what can a user be?".
An `author`, an `owner`, an `admin` or `commenter` etc.

Issue #3442, opened 2019-11-28
allow defining multiple (row + column) permission tuples for a given role + table + action


rikinsk:
To restrict access to sensitive fields, the solution proposed in https://docs.hasura.io/1.0/graphql/manual/queries/control-access.html  requires one to modify the db schema (creating a view and setting up new relationships/permissions) 

Instead if it were possible to define multilple row + column permission rules this could be avoided. 

e.g. For table `user(id, name, email, phone)`, `select` permissions for role `user` can be

- Rule 1:
  - Row permission: `{}` i.e. without any checks
  - Col permission: `name`

- Rule 2:
  - Row permission: `{id: {_eq: 'x-hasura-user-id'}}` i.e. if user is the owner
  - Col permission: `name, email, phone`

yasinarik:
This is the exact need I had 2 days ago. I would solve this by moving sensitive & private data to a new table at the moment. However, if multiple rules for the same role in a table can be implemented, it will become easier to control & develop.

**Current Solution** 
Also, even though I did a research about it, I couldn't reach the above documentation page which suggests creating a view by using: 
```
Data -> SQL


CREATE VIEW user_public AS
  SELECT id, name, email
    FROM user_profile;
```

where the original fields of a user profile are:
```
user_profile (id, name, email, phone, address)
```

tirumaraiselvan:
Related to #877

bertyhell:
My usecase is similar to this, but a bit more complex.

eg:
* Rule 1:
** Row permission: {} i.e. without any checks
** Col permission: name

Rule 2:
** Row permission: {id: {_eq: 'x-hasura-user-id'}} i.e. if user is the owner
** Col permission: email

Rule 3:
** Row permission: {id: {_eq: 'x-hasura-user-id'}, {permissions: {name: {_eq: "CAN_VIEW_PHONE"}}}} i.e. if user has certain permissions linked to its user account
** Col permission: phone

Ideally, all checks would be run and the union of all columns would be used as the final result.

eg: if the user id matches but the user does not have the CAN_VIEW_PHONE permission, then he would be allowed to access the columns: ["name", "email"]

webdeb:
Guys let me get some attention to the small rfc I've created here https://github.com/hasura/graphql-engine/pull/5783 where I am touching exactly this problem, with a slightly other mindset. Instead of multiple rules, the idea is to define virtual roles
role + check + columns.

[...]

PR #6396, opened 2020-12-28
server: inherited roles for postgres read queries [preview]

### LINKED ISSUE #68: fix console tests
Tests are now failing for console, refer attachment.
[console-tests.log](https://github.com/hasura/graphql-engine/files/2180410/console-tests.log)

codingkarthik:
fixes #3868

docker image - `hasura/graphql-engine:inherited-roles-preview-48b73a2de`

Note:

To be able to use the inherited roles feature, the graphql-engine should be started with the env variable `HASURA_GRAPHQL_EXPERIMENTAL_FEATURES` set to `inherited_roles`.

Introduction
------------

This PR implements the idea of multiple roles as presented in this [paper](https://www.microsoft.com/en-us/research/wp-content/uploads/2016/02/FGALanguageICDE07.pdf). The multiple roles feature in this PR can be used via inherited roles. An inherited role is a role which can be created by combining multiple singular roles. For example, if there are two roles `author` and `editor` configured in the graphql-engine, then we can create a inherited role with the name of `combined_author_editor` role which will combine the select permissions of the `author` and `editor` roles and then make GraphQL queries using the `combined_author_editor`.

How are select permissions of different roles are combined?
------------------------------------------------------------

A select permission includes 5 things:

1. Columns accessible to the role
2. Row selection filter
3. Limit
4. Allow aggregation
5. Scalar computed fields accessible to the role

 Suppose there are two roles, `role1` gives access to the `address` column with row filter `P1` and `role2` gives access to both the `address` and the `phone` column with row filter `P2` and we create a new role `combined_roles` which combines `role1` and `role2`.

Let's say the following GraphQL query is queried with the `combined_roles` role.

```graphql
query {
   employees {
     address
     phone
   }
}
```

This will translate to the following SQL query:

```sql

 select 
    (case when (P1 or P2) then address else null end) as address,
    (case when P2 then phone else null end) as phone
 from employee
 where (P1 or P2)
```

The other parameters of the select permission will be combined in the following manner:

1. Limit - Minimum of the limits will be the limit of the inherited role
2. Allow aggregations - If any of the role allows aggregation, then the inherited role will allow aggregation
3. Scalar computed fields - same as table column fields, as in the above example

APIs for inherited roles:
----------------------

1. `add_inherited_role`

`add_inherited_role` is the [metadata API](https://hasura.io/docs/1.0/graphql/core/api-reference/index.html#schema-metadata-api) to create a new inherited role. It accepts two arguments

`role_name`: the name of the inherited role to be added (String)
`role_set`: list of roles that need to be combined (Array of Strings)

Example:

```json
{
  "type": "add_inherited_role",
  "args": {
      "role_name":"combined_user",
      "role_set":[
          "user",
          "user1"
      ]
  }
}
```

After adding the inherited role, the inherited role can be used like single roles like earlier

Note: 

An inherited role can only be created with non-inherited/singular roles.

2. `drop_inherited_role`

The `drop_inherited_role` API accepts the name of the inherited role and drops it from the metadata. It accepts a single argument:

`role_name`: name of the inherited role to be dropped

Example:

```json

{
  "type": "drop_inherited_role",
  "args": {
      "role_name":"combined_user"
  }
}
```

Metadata
---------

The derived roles metadata will be included under the `experimental_features` key while exporting the metadata. 

```json
{
  "experimental_features": {
    "derived_roles": [
      {
        "role_name": "manager_is_employee_too",
        "role_set": [
          "employee",
          "manager"
        ]
      }
    ]
  }
}
```

Scope
------

Only postgres queries and subscriptions are supported in this PR.

Important points:
-----------------

1. All columns exposed to an inherited role will be marked as `nullable`, this is done so that cell value nullification can be done.


webdeb:
Can we actually have multiple rules for a single table with this feature?

Table: spectacle

Roles
user:guest
user:director
user:artist

Now lets imagine that internally the role `user` would represent this three rules(guest, director, artist) and deliver based on the bool expressions of them? 
So from the usage we would only use role: user

@codingkarthik sry, if thats not clear, I would like to explain it better

codingkarthik:
@webdeb I'm afraid I do not understand your question completely. Can you maybe give an example along with it?

webdeb:
My question is related to this problem #3442 

It's about defining multiple rules for a table+role combination:
The rule defines a bool_exp + column selection for this particular table.

But defining roles for each rule will not scale, because roles are global. So my idea was to have local table rules which look like roles but are kind of subroles: user:owner, user:editor, user:admin. They all refer to the role user, but they can have different rules for the columns and bool_exp. 

So I thought that it might be possible to create table specific roles this way and use the global role user as the inherited role for them all.

Is it better to understand or should I provide a better example?

codingkarthik:
@web-devel Thank you for providing a more verbose explanation and IIUC, the answer to your question is yes! 
The column selection rule will depend upon the row bool expression i.e a column can be accessible in some of the rows and not in other rows depending upon the what the boolean expression evaluates to. I have written more docs about this [here](https://hasura.io/docs/latest/graphql/core/auth/authorization/inherited-roles.html). Please feel free to comment here if you have more questions about this 😄

webdeb:
@codingkarthik sorry, no, I think you did not understand correctly. Right now, when you create a role on a table, this role is created globally for all tables. This does not scale if you have a lot of tables with different requirements. 

Making a role/rule local to a table makes sense. For example the posts table could have the role "user:author", or the "projects" table could have a role/rule "user:manager" / "user:member" / "user:reporter". Basically -> "A:R1", "A:R2", "A:R3" (R for Rule)

The difference of this roles would be only the bool_exp and column selection. Now with the ability of combining this rules, we can use only the role A as an inherited role of these three?

codingkarthik:
@webdeb Ah, now I understand. Yeah, it's not possible currently. Can you open a new issue with the above?

webdeb:
@codingkarthik I am happy, that I could explain it properly. There is an RFC for this already, I'll rewrite it a bit so it aligns better with current capabilities of inherited roles. #5783

PR #6551, opened 2021-02-15
docs: update command seed -> seeds

gunar:
The `seed` command seems to have been renamed but in some parts of the codebase the old token is still referenced. This PR attempts to fix this without changing the entries in the CHANGELOG obviously.

---

```
$ hasura seed apply

FATA[0000] unknown command "seed" for "hasura"

Did you mean this?
        seeds 
```


rikinsk:
Hey. Thanks for raising this PR. We really appreciate it! 

Though these changes are now not required as the latest CLI now supports both `seed` and `seeds` as aliases. Hence closing this PR.

PR #6786, opened 2021-04-14
install manifests: persist caddy certificates

WNpoScocIdEShmAZymNILEgEonOustRAC:
Caddy v2 uses the `/data` directory. 

I was restarting my containers a few times while testing and I noticed that https randomly stopped working. Caddy did not persist the certificate and I was being rate limited by let's encrypt.

### DIFF
--- a/install-manifests/docker-compose-https/docker-compose.yaml
+++ b/install-manifests/docker-compose-https/docker-compose.yaml
@@ -34,7 +34,7 @@ services:
     - "443:443"
     volumes:
     - ./Caddyfile:/etc/caddy/Caddyfile
-    - caddy_certs:/root/.caddy
+    - caddy_certs:/data
 volumes:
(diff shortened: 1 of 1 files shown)


subwaymatch:
Thank you @san-serif! I was having the same issue after running `docker-compose` commands multiple times. I manually remapped the volume using your `- caddy_certs:/data` and it works well (I had to wait a full week since I hit the LetsEncrypt's single domain limit of 5 certificates per week.

I hope this PR gets merged since many others can suffer as well :(

san-serif:
I totally forgot about this. Maybe this wasn't merged because they couldn't verify it? In any case here's the link where I got the information from: 

https://registry.hub.docker.com/_/caddy

petrogad:
Lovely, I just hit this too. welp next week it'll be there.

PR #6796, opened 2021-04-19
Auth0 fixed the JWKS bug, which allows using the jwks url

afitzek:
### Description

Auth0 integration works now with jwks urls. Auth0 fixed the bug with
the incorrect kid calculation. One still has to revoke a key
that was created before the key was generated.

This just updates the docs, maybe also `https://hasura.io/jwt-config/` could be updated.

### Affected components

- [X] Docs


JamiesWhiteShirt:
Yes, please include this in the docs! The docs do not even mention the maintenance issues and security risks with the workaround for Auth0, so the docs could at least primarily suggest using the JWK URL and fixing the signing keys if necessary.

We figured this out on our own, and it turned out our signing keys were OK.

thetimbecker:
Happy birthday to this PR! I was about to type up and submit my own PR, luckily I searched first.

We also are having no issues with the JWK URL, the docs were confusing.

TheodorRene:
We uneccesaryily hardcoded the key in our project for some time. As @JamiesWhiteShirt says this has security implications and should be prioritized to be merged.

PR #6860, opened 2021-05-03
Update Auth0 Docs if using SPA's and regular web app

psilospore:
### Description
There are separate instructions on creating a rule for the SPA flow and the Regular Web Application flow. 
This adds documentation if a user has both types of applications.

### Changelog

^ Would this be required for this PR?

### Affected components

- [x] Docs

### Solution and Design

I switch on the client type. It would be better if I could switch on the app type (e.g. if SPA) but I didn't see a way to do that.

I've include auth0-react as an example instructing the user to replace with the library client name.

Alternatively, we could instruct the user to switch on the client name `context.clientName`.

### Steps to test and verify

Follow the documentation and create both types of applications.

### DIFF
--- a/docs/graphql/core/guides/integrations/auth0-jwt.rst
+++ b/docs/graphql/core/guides/integrations/auth0-jwt.rst
@@ -75,6 +75,28 @@ For auth0-spa-js:
       callback(null, user, context);
     }
 
+If you use both SDKs:
+
+.. code-block:: javascript
+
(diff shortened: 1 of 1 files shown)


psilospore:
This approach actually didn't end up working for some reason. Not sure why I can see that property in the context object in the Auth0 logs. I ended up adding metadata to the App and checking for that instead. Going to close this for now.

PR #6956, opened 2021-05-20
Handle commenting of multi-line SQL in getDownQueryComments

etrepum:
### Description

I noticed that the down migration placeholder generated by `getDownQueryComments` does not handle commenting of multi-line SQL, which can result in a down.sql that looks like this:

```sql
-- Could not auto-generate a down migration.
-- Please write an appropriate down migration for the SQL below:
-- alter table "public"."locations" add column "google_maps_url" text
null;
```

### Changelog

### Affected components

- [ ] Server
- [X] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [X] No
- [ ] Yes

#### Metadata

- [X] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [X] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/console/src/utils/migration/utils.ts
+++ b/console/src/utils/migration/utils.ts
@@ -355,13 +355,16 @@ export const getDownQueryComments = (
   source: string
 ) => {
   if (Array.isArray(upqueries) && upqueries.length >= 0) {
-    let comment = `-- Could not auto-generate a down migration.
--- Please write an appropriate down migration for the SQL below:`;
-    comment = upqueries.reduce(
-      (acc, i) => `${acc}
--- ${i.args.sql}`,
-      comment
-    );
+    const comment = [
+      "Could not auto-generate a down migration.",
+      "Please write an appropriate down migration for the SQL below:",
+      ...upqueries.map((i) => i.args.sql),
+      ""
+    ].join("\n")
+     // Normalize \r\n to \n and add comments before every line
+     .replace(/\r?(^|\n)(?!$)/g, "$1-- ")
+     // Eliminate trailing spaces
+     .replace(/ +\n/g, "\n");
     return [getRunSqlQuery(comment, source)];
   }
   // all other errors


console/src/utils/migration/utils.ts
```diff
@@ -349,13 +349,11 @@ export const getDownQueryComments = (
...
-    comment = upqueries.reduce(
-      (acc, i) => `${acc}
--- ${i.args.sql}`,
-      comment
-    );
+    const comment = [
+      "Could not auto-generate a down migration.",
+      "Please write an appropriate down migration for the SQL below:",
+      ...upqueries.map((i) => i.args.sql)
+    ].join("\n").replace(/\r?\n/g, "$1-- ");
```
soorajshankar, line 356: could you try this one?
small changes on the regex to make sure all carrier returns and line breaks are captured + commented out the first line directly.
```suggestion
    const comment = [
      '-- Could not auto-generate a down migration.',
      'Please write an appropriate down migration for the SQL below:',
      ...upqueries.map(i => i.args.sql),
    ]
      .join('\n')
      .replace(/(\r\n)+|\r+|\n+/g, '\n-- ');
```

soorajshankar: with the above suggestion, I was able to generate the following

```sql
-- Could not auto-generate a down migration.
-- Please write an appropriate down migration for the SQL below:
-- CREATE OR REPLACE VIEW author_publics_2 AS
--   SELECT id, name,city, email
--     FROM author;
```

soorajshankar (changes requested):
Hello @etrepum,

Thank you for raising this PR.

I was trying to make a multiline SQL migration, but it seem to be generating the following string.

```
Could not auto-generate a down migration.$1-- Please write an appropriate down migration for the SQL below:$1-- CREATE OR REPLACE VIEW author_public AS$1--   SELECT id, name,city, email$1--     FROM author;
```

left a small suggestion, could you try that and update?

console/src/utils/migration/utils.ts
etrepum: Bare carriage returns aren't used in the modern world, it's either `\n` or `\r\n`. I've revised and rebased the PR to ensure the first line is commented, newlines are normalized, lines do not have trailing spaces, and the file ends with a newline.

soorajshankar (approved):
code and functionality LGTM

PR #6990, opened 2021-05-27
Update hasura_completion.rst

msimon:
.oh-my-zsh completions folder does not seems to work for autocompleting on the latest version.
Adding it into the plugin, then adding the plugin to zsh `plugins` does the trick.

I got inspired to try that from from this project: https://github.com/g-plane/zsh-yarn-autocompletions.


m-Bilal:
Hi @msimon , thank you for your PR ❤️ 
This PR has been open for so long now that the file you intend to make changes to no longer exists in the codebase. I think we should close this PR for now. Thanks again!

Issue #6419, opened 2021-01-05
RFC: Mapping template for request body transformation for events


tirumaraiselvan:
Suppose we could transform the request body for events (event triggers, scheduled triggers) then lot webhook integrations might be easier (without needing to plumb a middleware).

Will solve use-cases like https://github.com/hasura/graphql-engine/issues/5050

tirumaraiselvan:
We need to figure out the templating mechanism. Something in Haskell which resembles the Go template language: https://golang.org/pkg/text/template/ is one possibility (example implementation: https://tyk.io/docs/transform-traffic/request-body/). 

There is also https://jsonnet.org/ which seems even more programmable (might be overkill for this use-case though).

PR #7063, opened 2021-06-15
first draft

tirumaraiselvan:
For https://github.com/hasura/graphql-engine/issues/6419

### DIFF
--- a/rfcs/transforms.md
+++ b/rfcs/transforms.md
@@ -1,0 +1,110 @@
+## Motivation
+
+Hasura integrates with external HTTP APIs via a variety of features: actions, event triggers, scheduled triggers. Since the client request that these external APIs expect may be different from what Hasura generates, we want to be able to transform these requests before invoking the external API.
+
+## Spec
+
+We will introduce a key called `transforms` in the configuration for actions, event triggers and scheduled triggers.
+
+The schema for `transforms` is described here:
+
+``` 
+data RequestMethod = GET | POST | PUT | PATCH | DELETE
+
+data Transforms 
+  = Transforms
+  { requestMethod ::  Maybe RequestMethod
+ -- ^ Change the request method to one provided here. Nothing means POST.
+  , requestURL    ::  Maybe TemplateText
+ -- ^ Change the request URL to one provided here. Nothing means original URL.
+  , request_body  ::  Maybe TemplateText                  # Raw templating
+ -- ^ Change the request_body to one provided here. Nothing means original payload. Only applicable for POST, PUT, PATCH methods
+  , query_params  ::  Maybe (HashMap Text TemplateText)   # Key-value based templating
+ -- ^ Add query params provided here to URI. Only applicable for GET and DELETE methods
+  , request_headers :: Maybe TransformHeaders
+ -- ^ Tranform headers as defined here.
+  }
+
+data TransformHeaders
+  = TransformHeaders { replaceHeaders :: [(HeaderName, HeaderValue}] }
+
+data TemplateText # TO BE DECIDED (see next section)
+
+```
+
+## Templating Language
+
+This is largely undecided at the moment.
+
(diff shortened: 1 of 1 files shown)


rfcs/transforms.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Simple string templating is also required (for example for requestURI transform):
+
+```
+# $url has the original URL
+$url/$.event.author_id
+```
+
+TODO: Explore Haskell templating packages which resemble this.
```
paf31, line 65: Something like [this](https://github.com/moleike/haskell-jsonnet) could be a very good option, with some small modifications.

rfcs/transforms.md
```diff
@@ -0,0 +1,110 @@
...
+     param1: "$.value1"
+     param2: "$.value2"
+```
+
+We should also be able to transform the request headers. Hasura already has a way to add headers hence what will be useful is a way to replace few of the system added headers. The following example changes the body type to `x-www-form-urlencoded` and allows a custom `User-Agent`.
+
+```
+transforms:
+   request_body: |
+     key1=$.value1&key2=$.value2&key3=$.session['x-hasura-user-id']
```
paf31, line 105: There's some subtlety here, because how is escaping going to be done? If this language is JSON-specific, presumably it uses JSON escaping, but URL-encoded bodies and query strings don't use JSON encoding.

tirumaraiselvan: Right...how do we solve this?

rfcs/transforms.md
```diff
@@ -0,0 +1,110 @@
...
+
+data Transforms 
+  = Transforms
+  { requestMethod ::  Maybe RequestMethod
+ -- ^ Change the request method to one provided here. Nothing means POST.
+  , requestURL    ::  Maybe TemplateText
+ -- ^ Change the request URL to one provided here. Nothing means original URL.
+  , request_body  ::  Maybe TemplateText                  # Raw templating
+ -- ^ Change the request_body to one provided here. Nothing means original payload. Only applicable for POST, PUT, PATCH methods
+  , query_params  ::  Maybe (HashMap Text TemplateText)   # Key-value based templating
```
tirumaraiselvan, line 22: I was thinking we can skip `query_params` and just transform `requestURL` to add query params also.

rfcs/transforms.md
tirumaraiselvan: @paf31 How do we use jsonpath with haskell-jsonnet or can we load them into variables first?

tirumaraiselvan: @paf31 This looks cool.

rfcs/transforms.md
tirumaraiselvan: Can we make this users' concern? Can jsonnet urlencode values like below?

```
key1=std.urlencode($.value1)
```

paf31: Probably the easiest thing to do for the query string is to modify the interpreter to output url-encoded text instead of JSON, but then the user would write `{ key1: $.value1 }` instead.

paf31: But again, that only works if we know the content type, and we'd have to have a mapping from content types to encoders, unless we make everything into plain text, which sort of defeats the purpose of using something like jsonnet, I think.

rfcs/transforms.md
```diff
@@ -0,0 +1,135 @@
...
+data Transforms 
+  = Transforms
+  { requestMethod ::  Maybe RequestMethod
+ -- ^ Change the request method to one provided here. Nothing means POST.
+  , requestURL    ::  Maybe TemplateText
+ -- ^ Change the request URL to one provided here. Nothing means original URL.
+  , request_body  ::  Maybe TemplateText                  # Raw templating
+ -- ^ Change the request_body to one provided here. Nothing means original payload. Only applicable for POST, PUT, PATCH methods
+ -- The encoding of the content is defined by the 'request_content_type' field defined below.
+  , request_content_type :: AllowedContentTypes
```
tirumaraiselvan, line 23: @paf31 What if we accept a field called `request_content_type` ? I have added examples of this in examples section.

PR #7313, opened 2021-07-28
Delete hasura_logo_horizontal_blue.svg

m23koushik:
Hey, @praveenweb wanted to know if deleting or renaming brand asset files will break the assets if they are used anywhere. We are not going to use this format of the logo anymore. We are renaming other logo assets as well along with slight modifications to the design. So please let me know what might be the best way to handle this.


Stefmore02:
Closing this PR because it has been open for more than 90 days with no activity.

PR #7529, opened 2021-09-08
Do not check name of unique key constraint when suggesting autorelations (closes #7525)

### LINKED ISSUE #7525: One to one relationships not suggested when using custom names for unique indexes
It has come to my attention that, while this was working in earlier versions of Hasura (at least until 1.3.3), one-to-one relations are not tracked in the expected way in current version(s) of Hasura. I'm running Hasura 2.0.8 for the moment.

At the owning side of the relation, it gets taken up in the right way, i.e. suggesting an object relationship from the owning side to the other.

From the perspective of the other side however, an array relationship is suggested. Clearly, this doesn't make a lot of sense and causes the automated GraphQL queries and mutations to have a rather weird syntax.

A concrete example in my case:

One-to-one relationship between table "users" and table "app_settings".
Users table has a foreign key constraint to app_settings, named "app_settings_id". This is made unique.

After applying all suggested trackings, a new relationship with name "users" will show up in table "app_settings". This used to be "user" in earlier versions, as expected.

Did I miss anything or is this genuine unexpected behavior?

BenoitRanque:
There is a mechanism for suggesting one-to-one relationships for auto-tracking in the console.
Specifically, a one to-one relationship is suggested if `tableA` has a unique field with a foreign key to `tableB`.

This can be seen with a schema like this one:
```sql
create table favorite_color(
  id INT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
  name TEXT
);
create table person (
  id INT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
  name TEXT,
  favorite_color_id INT NOT NULL UNIQUE REFERENCES favorite_color(id)
);
```
Once these tables are tracked, one to one relationships will be suggested for both tables.

This breaks if a custom name is used for the unique constraint:
```sql
create table favorite_color(
  id INT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
  name TEXT
);
create table person (
  id INT PRIMARY KEY GENERATED BY DEFAULT AS IDENTITY,
  name TEXT,
  favorite_color_id INT NOT NULL CONSTRAINT my_unique_favorite_color UNIQUE REFERENCES favorite_color(id)
);
```
With the custom name, the one to one relationship is not suggested, and instead we get a one to many.

This seems to happen because when choosing whether to suggest an object relationship or array relationship, we check if the constraint's name includes the string `_key`. I do not know why this was originally added.

This change removes that check.

### Affected components
- [x] Console

### Related Issues
#7525


iykekings:
@BenoitRanque can you run `npm run format` and push?

BenoitRanque:
@iykekings Done. This changed a lot of files, I don't know if that's good. If you'd like, I can also redo this change on the private repo instead, and close this PR?

PR #7607, opened 2021-09-29
changing “ (unicode) for " (UTF-8)

thiagodamas:
The “ looks like ", but I'm getting the following error:
                "message": "character with byte sequence 0xe2 0x80 0x9c in encoding \"UTF8\" has no equivalent in encoding \"LATIN1\"",


jberryman (approved):

thiagodamas:
Hi, can somebody review this pr? Thanks

PR #7736, opened 2021-10-28
Avoid Translation Errors

0xanon:
You may add Google translator to the the possible places to enhance people understanding for every single language.


rikinsk:
@0xanon hi there. the intent of this PR is not clear. was this intentional?

PR #7946, opened 2021-12-16
Server: Streaming subscriptions v1

codingkarthik:
This PR adds support for streaming subscriptions. Streaming subscriptions are subscriptions that can be configured to only stream the data which has been changed in the response. Also, streaming subscriptions can be used to stream large amounts of data.

#### How to use it?

When you track a table, you should be able to see the `<tablename>_stream` fields under the subscription root.

For example: If you've tracked a table called `messages`,

```graphql
subscription {
   messages_stream(cursor: {value: {created_at: "2021-01-01"}, ordering: ASC}, batch_size: 10) {
       from
       to
   }
}
```

The stream field accepts the following arguments:

1. `cursor` (required) - column on which the streaming should be done, good choice of a column for a cursor would be a unique and sortable column
2. `batch_size` (required) - maximum number of rows to be returned in one batch
3. `where` (optional) - filter expression

docker image: `hasurabuild/graphql-engine:mono-pr-2948-cd5ef5d96`

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


psyko255:
@codingkarthik , was wondering what the plan is for this PR.  This will significantly help some of our use cases where we struggle with the current non-streaming subscriptions.  We are trying to understand if we need to go a different route or if this is expected to make it into a release relatively soon.  

Also had a question.  How are rows which no longer match a where clause handled?  Are they sent with sent in the stream with a flag to indicate  it is removed from the subscription? We need to be able to take the updates the from the stream and keep the state updated on the client.

ecthiender:
@codingkarthik we can close this PR right?

codingkarthik:
Closing this PR as it is already merged in main

PR #7958, opened 2021-12-18
improve contributing.md

marionschleifer:
### Description

As discussed in the call with KP, I tried out the workflow to set up the dev environment. It largely worked, and I only made one comment with a section we should add.

### Changelog

### Affected components

- [x] CLI


marionschleifer:
Sorry, wrong repo :) will open the PR in `graphql-engine-mono`.

Issue #6958, opened 2021-05-20
Customise Health endpoint


jarleriksen:
Hi,

We are currently self-hosting our Hasura instance on Google Cloud App Engine.
We are trying to use the `healthz` endpoint exposed by the GraphQL engine to monitor for inconsistent metadata.

However, `healthz` is a reserved path for App Engine and as far as I can tell it's not possible to customise the path in Hasura.

If it's not something that's going to be implemented or you have alternative solutions, I think it would be awesome to have it added to the docs. I spent some time figuring this one out :)

I would be more than happy to provide a PR for the docs.

Thanks!

sordina:
Hi @jarleriksen,

Quick question, which of these would work for you?

* An additional health endpoint, such as `/healthz2` or `/healthz-hasura`?
* The health endpoint to be customizable via command-line-options, with `/healthz` returning a 404.

jarleriksen:
Hi @sordina,
Honestly, any of those would work fine. A common solution is to prefix it with something to avoid the clashing e.g. `/hasura/healthz` etc.

Google uses `/_ah` for their prefix.

kolharsam:
Hi @jarleriksen,
We've added `/hasura/healthz` which mirrors `/healthz`. 
This should be out in one of the upcoming releases.

bkniffler:
Its cool to have an alternative endpoint, but I can't do migrations now because the cli will attempt to check the endpoint health via /healthz, but since it is reserved in app engine / cloud run, it may just return 404. Could the CLI default to /hasura/healthz?

romainbriand:
> Its cool to have an alternative endpoint, but I can't do migrations now because the cli will attempt to check the endpoint health via /healthz, but since it is reserved in app engine / cloud run, it may just return 404. Could the CLI default to /hasura/healthz?

I'm having the same issue with the CLI to apply metadata, migrations, etc. to our Google Cloud Run hosted endpoint. Is there any possible workaround?

romainbriand:
I've submitted a PR here: https://github.com/hasura/graphql-engine/pull/7994

PR #7994, opened 2022-01-04
Use healthz mirror endpoint in CLI

romainbriand:
### Description
This PR could solve the problem here: https://github.com/hasura/graphql-engine/issues/6958#issuecomment-1003208434
CLI uses `/healthz` endpoint but it is a reserved URL on Google Cloud products for example. In this case route will always resolve to an http 404 error code while the mirror endpoint (`hasura/healthz`) will work just fine.

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

#6958

### Solution and Design

I'm offering here to use the `/hasura/healthz` mirror endpoint instead of `/healthz`. This could solve a blocking issue to use the CLI on infrastructures having `/healthz` as a reserved URL.

### Steps to test and verify

Deploy Hasura on Google Cloud Run and check that you can apply metadata for example.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/cli/util/server.go
+++ b/cli/util/server.go
@@ -104,7 +104,7 @@ func GetServerStatus(endpoint string) (err error) {
 	if err != nil {
 		return fmt.Errorf("error while parsing the endpoint :%w", err)
 	}
-	uri.Path = path.Join(uri.Path, "healthz")
+	uri.Path = path.Join(uri.Path, "hasura/healthz")
 	resp, err := http.Get(uri.String())
 	if err != nil {
 		return fmt.Errorf("making http request failed: %w", err)


mccare:
Same problem here, `/healthz` is filtered by google app engine flex but `hasura migrate status`  checks the healthz first and aborts. Other option is to skip health check by introducing a command line flag? I'm not sure if /hasura/healthz has always been present in graphql-engine.

nathanameye:
Thank you, same issue here with Cloud Run hosted Hasura. This fix is excellent.

DaddyMonster:
I'm using cli version 2.1.1 with hasura docker on Google cloud run
it still check for /healthz when trying to open console or migration via cli

Is there any workaround for now?

PR #8039, opened 2022-01-17
cli: fix --address option (close #8005)

### LINKED ISSUE #8005: cli: The --address option of the hasura console command no longer works with v2.1.0
### Version Information

- Server Version: v2.1.0
- CLI Version (for CLI related issue): v2.1.0

### Environment

- Server: Docker container launched on Amazon EC2 Instance(docker image: hasura/graphql-engine:v2.1.0)
  - Docker containers expose port 8080.
- Database: Docker container launched on Amazon EC2 Instance(docker image: postgres:12)
- CLI: CLI installed on Amazon EC2 Instance

The `--address` option should start the hasura console.

### Keywords

hasura console

```
hasura console --address XXX.com --admin-secret XXX
```

When opening the web console screen, the error message is displayed.

```
Hasura console is not able to reach your Hasura CLI instance. Please ensure that your instance is running and the endpoint is configured correctly.
```

Also, you will no longer be able to apply to migrate on the web console.

1. Run `hasura console` command with the `--address` option set to a value other than localhost.
2. Access the domain specified by `--address`option and open the web console screen.
3. An error message(`Hasura console is not able to reach your Hasura CLI instance...`) will appear and you will not be able to apply to migrate from the web console.

### Screenshots or Screencast
Screenshot 2022-01-06 14 25 49

### Any possible solutions?

As a temporary measure, you can work around this by setting the values of both the `--api-port` and `--address` options.

e.g.

```
hasura console --api-host http://XXX.com --address XXX.com
```

### Can you identify the location in the source code where the problem exists?

[cli/commands/console.go#L129](https://github.com/hasura/graphql-engine/blob/master/cli/commands/console.go#L129)

With the [v2.0.10 -> v.2.1.0](https://github.com/hasura/graphql-engine/compare/v2.0.10...v2.1.0#diff-e79cadc3cdbde4a7ff6a82c161f008fb204d28cb060217f03ecf69900a255ab2L128) change, `APIHost` no longer references the value of the address option.

This causes the default value to be used for the API Host to connect to the server, even if the `--address` option is given.

The default value of `api-host` is `localhost`, so if the port number of localhost `9693` or the port specified by --api-port is not public, the connection will fail.

### If the bug is confirmed, would you be willing to submit a PR?

Yes, I want to work for this :)

However, to submit a PR, you need to decide which value to use to launch the server when both the `--api-host` and `--address` options are specified.

sho-hata:
### Description
Closes #8005
Resolved the problem where the `--address` option would not work if the `--api-host` option was the default value.

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [x] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues
#8005

### Solution and Design
If the value of the `--api-host` option is the default value, the value specified by the `--address` option is now used for the console and migrate API.
If both `--apihost` and `--address` options are specified, `--address` has priority.

### Steps to test and verify
Make sure it passes all the e2e tests to show that the changes are OK.

#### Catalog upgrade
- [x] No
- [ ] Yes

#### Metadata
- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/cli/commands/console.go
+++ b/cli/commands/console.go
@@ -126,7 +126,7 @@ func (o *ConsoleOptions) Run() error {
 	}
 
 	consoleRouter, err := console.BuildConsoleRouter(templateProvider, consoleTemplateVersion, o.StaticDir, gin.H{
-		"apiHost":              o.addressOrAPIHost(),
+		"apiHost":              o.handleAPIHost(),
 		"apiPort":              o.APIPort,
 		"cliVersion":           o.EC.Version.GetCLIVersion(),
 		"serverVersion":        o.EC.Version.GetServerVersion(),
@@ -176,10 +176,10 @@ func (o *ConsoleOptions) Run() error {
 	return console.Serve(serveOpts)
 }
 
-func (o *ConsoleOptions) addressOrAPIHost() string {
-	if o.Address != "localhost" {
-		return "http://" + o.Address
+func (o *ConsoleOptions) handleAPIHost() string {
+	if o.APIHost != "http://localhost" {
+		return o.APIHost
 	}
 
-	return o.APIHost
+	return "http://" + o.Address
 }
--- a/cli/commands/console.go
+++ b/cli/commands/console.go
@@ -126,7 +126,7 @@ func (o *ConsoleOptions) Run() error {
 	}
 
 	consoleRouter, err := console.BuildConsoleRouter(templateProvider, consoleTemplateVersion, o.StaticDir, gin.H{
-		"apiHost":              o.APIHost,
+		"apiHost":              o.addressOrAPIHost(),
 		"apiPort":              o.APIPort,
 		"cliVersion":           o.EC.Version.GetCLIVersion(),
 		"serverVersion":        o.EC.Version.GetServerVersion(),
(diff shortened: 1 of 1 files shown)


cli/commands/console.go
```diff
@@ -175,3 +175,11 @@ func (o *ConsoleOptions) Run() error {
 
 	return console.Serve(serveOpts)
 }
+
+func (o *ConsoleOptions) addressOrAPIHost() string {
```
scriptonist, line 179: @sho-hata On a second thought, We can give `--api-host` the priority when both flags are set. When someone set the flags they probably want to set the value of "--api-host" different from `--address`.

sho-hata: @scriptonist OK, We will modify the code to give priority to the value of `--api-host` when both options are specified.

sho-hata: @scriptonist I fixed the behavior, please check it out.
https://github.com/hasura/graphql-engine/pull/8039/commits/7ea85f09434b63bb14bdcae4b3ef5c6e6116e4d9

scriptonist: Thanks @sho-hata will check it out and get back.

PR #8153, opened 2022-02-08
fix wrong env in docker-compose

pfuhrmann:
### Description
Fixing wrong env


rikinsk:
Hi. The env var name is actually intentionally `PG_DATABASE_URL`. Could you explain a bit why you would like this change?

pfuhrmann:
OK. It did not work for me as expected. Only after I changed it to `HASURA_GRAPHQL_DATABASE_URL` I could actually see my database as a source in the console. Closing.

rikinsk:
@pfuhrmann Thanks for the explanation. This is an intentional change we have made because with v2.0, users can connect multiple databases and can already have an existing db they want to connect to. With the current docker-compose setup users can explicitly choose which database they want to connect to post starting the server. This bit is documented [here](https://hasura.io/docs/latest/graphql/core/getting-started/docker-simple.html#step-3-connect-a-database) if you are interested. We understand how this is a little less ideal for users just starting off from scratch but its conscious decision  at least for now.
Thanks a ton for taking the effort to raise this PR. Your feedback is definitely useful.

PR #8238, opened 2022-02-24
don't drop nested typed null fields in actions (fix #8237)

### LINKED ISSUE #8237: Actions: A nullable field with nested type gets omitted from result if the field is null
### Version Information

Server Version: 2.2.0

### Environment

OSS

Given the following Action and type definitions:
```graphql
type Query {
  myAction: Item!
}

type Item {
  id: ID!
  str: String
  child: Item
}
```

And then performing the query
```graphql
query myQuery {
  myAction {
   id
   str
   child: { id }
  }
}
```

When the backend implementation produces the following response:
```json
{ "id": 1, "str": null, "child": null }
```

Hasura should produce the following response:
```json
{ "id": 1, "str": null, "child": null }
```

### Keywords

Action, nested, type, null, field, omitted

Hasura does not pass through the null valued nested child fields, but instead returns the following object as response:
```json
{ "id": 1, "str": null }
```

Define action + handler as defined in the "What is expected behaviour" section

### Can you identify the location in the source code where the problem exists?

Possibly line https://github.com/hasura/graphql-engine/blob/f43704fb4a071e392b5346972dbef9b6ed18dd4b/server/src-lib/Hasura/GraphQL/Execute/Action.hs#L216

That seems to be in a function that deals with transforming the REST API Action response to the GraphQL response, and it would seem that the handling of nested objects is missing handling for null values and unless the response is an object or array, it dropped completely

Without prior experience with Haskell or Hasuras codebase, the following might be a "the" solution:
```haskell
J.Null -> Just AO.Null
_ -> Nothing
```

### If the bug is confirmed, would you be willing to submit a PR?

PR: #8238

Huulivoide:
### Description
Given Action types:

```
type Query { myAction: Item! }

type Item {
  id: ID!
  child: Item
}
```

And executing the query:
```
query myQuery {
  myAction {
    id
    child: { id }
  }
}
```

With backend response of `{ "id": 1, "child": null }`
Hasura would drop the child field and response with
`{ "id": 1 }`, this change makes sure that the nulls are preserved on
Action type definitions on typed nested fields.

### Changelog

- [X] `CHANGELOG.md` is updated with user-facing content relevant to this PR. If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [X] Server
- [X] Tests

### Related Issues
Fix #8237 

### Solution and Design

Added handling for the null case.

### Steps to test and verify

A test case has been added

#### Catalog upgrade

- [X] No

#### Metadata

- [X] No

#### GraphQL

#### Breaking changes

- [X] No Breaking changes


sordina:
Fantastic!

Issue #8392, opened 2022-04-12
[console] bad layout css causes jankyness


haf:
It's very easy to get annoyed when using the console, because the console is very janky, i.e. when scrolling up and down, you sometimes get stuck in smaller blocks that scroll.

This is something that's not supposed to happen if you write good css; so I had look at the css and sorted out the start screen and data screen (to start with)

All in all, this case-base is *really really* in need of spring cleaning, including #8390 and #8387.

PR #8393, opened 2022-04-12
[no changelog required] Started removing cruft from CSS; much more work needed! #8392 

haf:
Work towards #8392 

**I will not complete this PR; but here's what you should do:**

- Go through the codebase and convert it properly from scss to Tailwind (since I see you've chosen that path)
- Remove all classes with underscores in them; they've never been changed and have outlived theire life
- Start looking at code-splitting and getting change-to-see-effect down to less than one second. Waiting 18 seconds for every code save is not an acceptable development experience; you can quickly code-split each module (SQL, Data Manager, ...) and get a much smaller bundle size to start with
- Whenever there's a `calc( ... )` statement in the code, *seriously* consider whether this *is a must*; from working 8+ hours on this in this PR, I can tell you that it is NOT a must in 80% of the places and is used instead of properly understanding how the CSS box model works
- Fix all the useEffects; not by deactivating the linter warning you about the bugs, but by actually fixing the refs that cause rerenders
- Remove *all* usages of `!important` throughout the code base. The !important rash started to spread in the tree view component and then spread into buttons, and it'll spread further if not watched
- Remove the old bootstrap theme; just delete the files and remove the dead classes and style it from there
- Remove all extraneous divs littering every react component; render with fragments instead
- Fix the hundreds of deprecated packages and vulnerabilities

### Affected components

- [ ] Server
- [x] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


console/src/components/Services/Data/DataSubSidebar.js
```diff
@@ -236,66 +236,41 @@ const DataSubSidebar = props => {
         setTreeViewItems(newItems);
       }
     );
-  }, [sources.length, tables, functions, enums, schemaList, currentTable]);
+  }, [sources.length, tables, functions, enums, schemaList, currentTable, allSourcesSchemas, sources, dispatch, getItems, isFetching, preLoadState]);
```
haf, line 239: This probably will cause the problems with rerenders; but fix it properly by making these objects memoised instead of not adding them to the dep list.

rikinsk:
@haf Thanks for taking the time for getting deep into the weeds of this, we really appreciate it. We are aware of the issues you've brought up and we'll be doing a refactoring exercise to address them. That being said we will be picking up that exercise at a later point in time as we are occupied with a few other priorities at the moment. Hence we will not be able to review this PR immediately and will be leaving it open till the time we get around to doing that refactor. Again, thanks a ton for raising this!

PR #8408, opened 2022-04-17
Update COMPILING-ON-M1.md

nicoburns:
### Description
 On my machine, it was necessary to:
- Install python2 in order to get the node install for the console to work
- Set the `CPPFLAGS` environment variable in order to get the python install to work

Given the nature of these tweaks, I imagine they are necessary on all machines but just happened to already by in place on the machine of whoever wrote these instructions.

### Affected components

- [x] Docs

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/COMPILING-ON-M1.md
+++ b/server/COMPILING-ON-M1.md
@@ -31,6 +31,8 @@
 
 3. Install console assets
 
+   This step may require you to have python2 installed and available in your $PATH.
+
    ```sh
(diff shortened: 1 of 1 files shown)


soupi:
Thanks @nicoburns. I did try to follow these commands on a clean VM at the time and it worked for me, but it is possible something has change. I will approve this PR shortly. Thank you for your contribution!

PR #8461, opened 2022-05-03
Clarify computed fields documentation

hangtwenty:
Computed fields docs - intro paragraph: the existing phrasing combined two ideas into one sentence, in this case resulting in some ambiguity. 

### Description

The existing phrasing said "Computed fields are computed when requested for" (that's the first idea) "via custom SQL functions ..." (that's the second idea). When put together, "Computed fields are computed when requested for via custom SQL functions," it can be interpreted as "Computed fields are computed when [requested-by] custom SQL functions," for example. There should be no ambiguity about when computed fields actually get run/computed, i.e. upon each request. So hopefully this small change would make that clearer, without having to reread the sentence a couple of times to make sure.

### Changelog

- [ ] ~`CHANGELOG.md` is updated with user-facing content relevant to this PR.~ If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [x] Docs

### DIFF
--- a/docs/docs/graphql/core/databases/postgres/schema/computed-fields.mdx
+++ b/docs/docs/graphql/core/databases/postgres/schema/computed-fields.mdx
@@ -21,7 +21,8 @@ import TabItem from '@theme/TabItem';
 
 Computed fields are virtual values or objects that are dynamically
 computed and can be queried along with a table/view's columns. Computed
-fields are computed when requested for via [custom SQL functions](https://www.postgresql.org/docs/current/sql-createfunction.html)
+fields are computed upon request. They are computed by 
+[custom SQL functions](https://www.postgresql.org/docs/current/sql-createfunction.html)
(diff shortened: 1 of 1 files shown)


docs/docs/graphql/core/databases/postgres/schema/computed-fields.mdx
```diff
@@ -21,7 +21,8 @@ import TabItem from '@theme/TabItem';
 
 Computed fields are virtual values or objects that are dynamically
 computed and can be queried along with a table/view's columns. Computed
-fields are computed when requested for via [custom SQL functions](https://www.postgresql.org/docs/current/sql-createfunction.html)
+fields are computed upon request. They are computed by
```
hangtwenty, line 24: It might be even better to say "upon API requests" but that feels like more of a judgement call, so I'll defer to reviewers about that. I just wanted to suggesting "upon request" rather than "when requested for", and splitting into two sentences.

PR #8472, opened 2022-05-06
Improve COMPILING-ON-MACOS.md

nicoburns:
### Description

- Add note about the user's homebrew prefix potentially being different to the one used in the instructions
- Add extra clarifications about cabal project configuration files for those who are unfamiliar with cabal.
- Add step to set version number in `server/CURRENT_VERSION` file
- Disambiguate compile target by adding `exe:` prefix (cabal build now fails otherwise)

### Affected components

- [x] Docs

#### Breaking changes

- [x] No Breaking changes


soupi:
Thanks for the PR @nicoburns! I've made a few minor formatting changes and will merge this PR shortly.

PR #8531, opened 2022-05-27
Fix branch path and filter out index file.

nicuveo:
### Description
This PR does two small fixes to the haddock indexing pages:
- it makes the link relative to the index; the absolute link was wrong, as it was missing the `graphql-engine` root; making it relative is probably a most robust fix than making it absolute?
- it filters out the "index.html" page out of the list of branches ^^


i-am-tom (approved):
Thank you! This is what I get for being "clever"...

scriptnull (approved):

PR #8575, opened 2022-06-10
Add default privileges to pg user permissions example (close #3671)

### LINKED ISSUE #3671: docs: improve docs for "metadata reload" for databases migrations and a separate hasurauser
Is this the right place to report this?  It relates to docs for "reload" which I think was added here?
https://github.com/hasura/graphql-engine/pull/3264

A Very Important Thing is missing in the docs
-------------
specifically at 
https://docs.hasura.io/1.0/graphql/manual/migrations/manage-metadata.html

For someone who is managing their database schema with Rails migrations...

After doing the migration (especially one that adds a table) be sure to also run:
```
GRANT ALL ON ALL TABLES IN SCHEMA public TO hasurauser; 
GRANT ALL ON ALL SEQUENCES IN SCHEMA public TO hasurauser;
```
(per https://docs.hasura.io/1.0/graphql/manual/deployment/postgres-permissions.html )

otherwise the new table you just added in Rails will not show in hasura after doing a 'reload metadata' in the dashboard.  And if you do not do the "ON ALL SEQUENCES" one, it will fail/error when inserting.

(That was me a minute ago!)

BenoitRanque:
### Description
Adds an example of `ALTER DEFAULT PRIVILEGES` to grant permissions to newly created database objects by default.

### Affected components

- [x] Docs

### Related Issues

#3671

### Solution and Design

Reference: postgres documentation on [GRANT](https://www.postgresql.org/docs/current/sql-grant.html), [REVOKE](https://www.postgresql.org/docs/current/sql-revoke.html), and [ALTER DEFAULT PRIVILEGES](https://www.postgresql.org/docs/current/sql-alterdefaultprivileges.html).

### Steps to test and verify

Follow edited example, check if hasura user defaults to having permissions to tables newly created by other users

### DIFF
--- a/docs/docs/graphql/core/deployment/postgres-requirements.mdx
+++ b/docs/docs/graphql/core/deployment/postgres-requirements.mdx
@@ -107,6 +107,17 @@ GRANT ALL ON ALL FUNCTIONS IN SCHEMA public TO hasurauser;
 -- GRANT ALL ON ALL TABLES IN SCHEMA <schema-name> TO hasurauser;
 -- GRANT ALL ON ALL SEQUENCES IN SCHEMA <schema-name> TO hasurauser;
 -- GRANT ALL ON ALL FUNCTIONS IN SCHEMA <schema-name> TO hasurauser;
+
+-- By defaults users won't have access to tables they have not created (and thus do not own).
+-- You can change these default prvileges to grant access to any object created in the future.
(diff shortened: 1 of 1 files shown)


BenoitRanque:
Apparently I can't add tags to this, need to add the /no-changelog-required tag

Issue #7914, opened 2021-12-12
Apply seeds in cli-migrations v3


preetjdp:
### Is your proposal related to a problem?

The CLI Migrations docker image currently only applies the migrations and the metadata, the proposal is for it to also apply the `seeds` when the directory is present

### Describe the solution you'd like

- If the `hasura-seeds` directory is mounted to the container, apply the seeds

### If the feature is approved, would you be willing to submit a PR?

Yes, already on it

scriptonist:
The reason why this might be a bad idea is, that seeds are currently stateless, as opposed to migrations. ie migrations are in a way idempotent, so a cli-migrations container can be restarted multiple times in its lifetime since the migrations "state" is persisted.

But on the other hand, seeds are not handled in a similar fashion, so unless the user writes seeds as idempotent SQL a similar behaviour cannot be achieved.

For example, if I've some seeds to apply from my cli-migrations image, I'll start my container and the image will apply my seeds. Now let's say I spin up another replica of this image and that container will also try to apply the same seeds and will probably fail since the entities already exist in the database. This happens because the "state" in which seeds are applied is not persisted.

Please add your comments/thoughts/suggestions

aimproxy:
May be this could do the trick, but I am not sure! Inside your docker-compose file add this line into your Hasura CLI Migrations service:  `command: sh -c "hasura-cli seed apply"`

mbrevda:
> that seeds are currently stateless

perhaps the lack of any migrations in the db can be used as a herustic to decide if seeds should be run or not? Alternatively, seeds can be written idempotently - this can be a limitation for those wanting seeds to be applied automatically.

PR #8721, opened 2022-07-27
add: Apply seeds in cli-migrations #7914 

preetjdp:
### Description

This PR adds the ability for the seeds to be auto-applied as part of the `cli-migrations` docker image ([Reference](https://hasura.io/docs/latest/graphql/core/migrations/advanced/auto-apply-migrations.html)

### Changelog

### Affected components

- [ ] Server
- [ ] Console
- [ ] CLI
- [x] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)
- [x] CLI Migrations

### Solution and Design

The PR follows the existing pattern of the [Docker Entrypoint file](https://github.com/hasura/graphql-engine/blob/master/packaging/cli-migrations/v3/docker-entrypoint.sh) and adds a check to apply the seeds if the directory is loaded/present in the file

### Steps to test and verify

The [test file](https://github.com/hasura/graphql-engine/blob/packaging/cli-migrations/v3/test/test.sh) has also been updated to show that the seeds have been updated

- Build the `cli-migrations` image locally
- Run the `test.sh` script

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


andrewlouis93:
Pretty big shift in the expectation for folks that use the image to run migrations, but not to re-seed.

Should be a behaviour that's disabled by default, and enabled on feature flag IMO.

scriptonist:
Thanks for the PR @preetjdp ❤️ 

I doubt we can accept this change. More information about the reason is written down here: https://github.com/hasura/graphql-engine/issues/7914#issuecomment-1202320800 

Couple of clarifications: 
- Seeds are typically meant for dev/staging type envs where some dummy data needs to be added to test things. Typically seeds are applied on all envs other than production
- seeds are not incrementally applied and hence can only be applied on fresh DBs or else data conflicts are likely to happen if the seed apply is run more than once
- if the users have data that need to be applied in all envs, even production, then they should be tracked as migrations instead of seeds. e.g. values in enum tables
- In case, users want to apply seeds in certain envs, they'll need to figure out a way to manually run the seeds apply command.

Issue #8299, opened 2022-03-09
JWT secret constantly being refreshed when using Firebase


aaka3207:
### Version Information

Server Version:
CLI Version (for CLI related issue): ^2.2.0

### Environment

Docker-Compose- WSL Ubuntu

### What is the expected behaviour?

Token is refreshed on the usual basis

### What is the current behaviour?

I'm using Firebase as my auth provider. I bumped my version in docker to 2.3.0 from 2.1.1 and now the engine is constantly refreshing the token. The API itself seems to still be working, but I've noticed points where it intermittently fails in between the gaps of not having a fresh token.

### How to reproduce the issue?

1.Create Firebase account, and auth provider
2. Use the following configuration for **HASURA_JWT_SECRET**: 

` HASURA_GRAPHQL_JWT_SECRET: '{"jwk_url":"https://www.googleapis.com/service_accounts/v1/jwk/securetoken@system.gserviceaccount.com","audience":"<project-name>","issuer":"https://securetoken.google.com/<project-name>", "allowed_skew": 3600}'`

### Screenshots or Screencast

![image](https://user-images.githubusercontent.com/16450271/157505825-5e47649a-e44f-4dca-a283-ca3b8d2f2f9d.png)

efstathiosntonas:
why this has to be logged constantly? our server logs are full with this message and [GCP charges](https://cloud.google.com/stackdriver/pricing?authuser=0#logging-pricing-summary) for this (not a big deal, just saying)

tirumaraiselvan:
This is a recent regression, will be fixed very soon.

daniel-chambers:
This has been fixed in https://github.com/hasura/graphql-engine/commit/69501b26572dc60bab48fcbb85a5cc405861f1e0 and will ship in the next release. 🙂

aaka3207:
@tirumaraiselvan was this released in one of the beta releases already or in a major release? I'm on v2.4.0-beta.3 and it's still happening.

efstathiosntonas:
Still happening on v2.4.0 with this config:

```
      HASURA_GRAPHQL_JWT_SECRET:
        '{ "type":"RS256", "jwk_url":
        "https://www.googleapis.com/service_accounts/v1/jwk/securetoken@system.gserviceaccount.com",
        "audience": "${FIREBASE_PROJECT_ID}", "issuer":
        "https://securetoken.google.com/${FIREBASE_PROJECT_ID}" }'
```

[...]

Issue #8611, opened 2022-06-21
ability to disable jwk-refresh-log via enabled log types


tirumaraiselvan:
### Is your proposal related to a problem?

Currently, we cannot disable `jwk-refresh-log` through enabled log types: https://hasura.io/docs/latest/graphql/core/deployment/logging/#log-types

Many JWK providers have a Cache-Control policy which makes Hasura refresh them every second and pollute the logs.

### Describe the solution you'd like

Add `jwk-refresh-log` as a configurable log type

CameronNemo:
Seeing the same with keycloak. https://github.com/keycloak/keycloak/issues/11320

pranshi06:
Closed via fd6aa0a2662632c867fcf16c1fca4e1636cd5cf2 and 7b7845f697d1b0014b75533e03a0958e5c41530b

PR #8730, opened 2022-07-29
Prevent log spamming when using a non-cached JWKS

lovasoa:
Multiple auth providers (keycloak, firebase, ...) do not allow caching on their JWK endpoints.
In this case, Hasura requests the endpoint every second.
This change avoids spamming the logs when this happens.

### Changelog

### Affected components

- [x] Server
- [ ] Console
- [ ] CLI
- [ ] Docs
- [ ] Community Content
- [ ] Build System
- [ ] Tests
- [ ] Other (list it)

### Related Issues

https://github.com/hasura/graphql-engine/issues/8299

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/Server/Auth/JWT.hs
+++ b/server/src-lib/Hasura/Server/Auth/JWT.hs
@@ -346,7 +346,7 @@ updateJwkRef ::
 updateJwkRef (Logger logger) manager url jwkRef = do
   let urlT = tshow url
       infoMsg = "refreshing JWK from endpoint: " <> urlT
-  liftIO $ logger $ JwkRefreshLog LevelInfo (Just infoMsg) Nothing
+  liftIO $ logger $ JwkRefreshLog LevelDebug (Just infoMsg) Nothing
   res <- try $ do
     req <- liftIO $ HTTP.mkRequestThrow $ tshow url
     let req' = req & over HTTP.headers addDefaultHeaders


SamirTalwar:
Hi @lovasoa, thanks for the patch!

This seems benign but there are definitely use cases where users might want to see these logs at INFO level too. We think it might be more prudent to make this configurable, so that you can customize it for yourself as needed, as outlined in #8611.

I see you've already spotted this issue. Would you be interested in submitting a patch along those lines?

lovasoa:
Hi, and thanks for the response. 
Sorry, but I don't have the bandwidth for that now. 
Are you sure it makes sense to log this as INFO by default for everyone ? I agree it should be made configurable, but by default, given that hasura will refresh keys every second for most jwks providers, I think it doesn't make sense to display something on every refresh.

SamirTalwar:
It really depends on the caching proposed by the auth provider. I took a look at a few (Auth0, AWS Cognito, Firebase) and they all seem to encourage some caching.

[The Firebase documentation I found](https://firebase.google.com/docs/auth/admin/verify-id-tokens) suggests that keys are cached, and uses a central, public keyset. I hit the URL, https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com, and got the following headers back:

```
cache-control: public, max-age=24087, must-revalidate, no-transform
date: Fri, 05 Aug 2022 13:40:48 GMT
expires: Fri, 05 Aug 2022 20:22:15 GMT
```

Can you explain your situation in greater detail so I can understand why it's different?

lovasoa:
Hey @SamirTalwar , I'm sorry I missed your message when you posted it. Thank you for investigating on this ! 
Our auth provider is the open-source [Keycloak](https://www.keycloak.org/). It serves JWKs with the following headers : 

```
< HTTP/2 200 
< date: Fri, 28 Oct 2022 13:48:49 GMT
< content-type: application/json
< content-length: 1466
< cache-control: no-cache
< x-xss-protection: 1; mode=block
< x-frame-options: SAMEORIGIN
< referrer-policy: no-referrer
< strict-transport-security: max-age=15724800; includeSubDomains
< x-content-type-options: nosniff
< x-forwarded-for: keycloak.staging.dezilec.com
< x-forwarded-proto: https
```

Unfortunately, that does not seem to be configurable, and we end up with very large amounts of logs from hasura, which are expensive to store and make debugging more difficult.

SamirTalwar:
That's interesting. My first thought was that we should probably let the auth provider specify the expiry, but a default of 1 second seems very arbitrary. I'll bring this up and see if we can make that configurable.

pranshi06:
Hi @lovasoa , thanks for taking the initiative to resolve this issue.
We have fixed it by making `jwk-refresh-log` type configurable as outlined in #8611 via fd6aa0a2662632c867fcf16c1fca4e1636cd5cf2 and 7b7845f697d1b0014b75533e03a0958e5c41530b.

lovasoa:
Great, thank you  @pranshi06  !

Issue #4678, opened 2020-05-05
Computed fields of functions which return a single row should return a single object


franciscolourenco:
### Actual:
Querying a computed field defined by a function which `RETURNS table_name`, results in an array of objects, same as if the function was defined with `RETURNS SETOF tablename`

```js
"computed_field": [
  {
    "id": 1
  }
]
```

### Expected:
A function which `RETURNS table_name` should result in a single object, because the function can only return a single row.

```js
"computed_field": {
    "id": 1
}

```

### Notes:
The [hasura docs](https://hasura.io/docs/1.0/graphql/manual/schema/computed-fields.html) say that only `RETURNS SETOF table_name` is supported, however hasura seems to treat `RETURNS table_name` as if it is had `SETOF`.

------------------
Related to https://github.com/hasura/graphql-engine/issues/4299

tirumaraiselvan:
Closing this in favour of https://github.com/hasura/graphql-engine/issues/4299

larsblumberg:
To my understanding this issue is a duplicate of #6562 (rather than #4299).

I am hoping for #6562 to get implemented as an object relationship for a computed field looks like a fundamental and yet missing feature for me. The array relationship yields an ugly workaround for relationships where there really can be just one object (or none).

Issue #6562, opened 2021-02-17
Create object relation through computed field


dariocravero:
Hi,
How're you?

Have you considered allowing relations through computed fields?

I'd like to skip making a view if possible to avoid the extra maintenance step.
I know that an alternative is to use a computed field to return a `setof <table-name>` but if I only need an object relation, how can I approach that?

Thanks!
Darío

tirumaraiselvan:
I think we should extend support for computed fields which return `<table-name>` instead of only `set of <table-name>`. We did this for top-level functions already: https://github.com/hasura/graphql-engine/issues/4299

dariocravero:
That'll be amazing @tirumaraiselvan!! Thanks. Let me know if I can be of any help.

dariocravero:
Have you also considered allowing filtering by computed fields in `where`? I think I saw this in another issue but couldn't find it.

esseswann:
@dariocravero It's tracked here https://github.com/hasura/graphql-engine/issues/3772

dariocravero:
Is this being considered as parts of the ongoing computed fields improvements?

dariocravero:
Also, somewhat related but since remote joins now support computed fields as part of their relations, would it be possible to use computed fields as part of relations too? Eg, a computed field returns an id, then we can use that in a relation to another table

jgoux:
This feature would be very helpful for us as well!

We enforce our authorization rule per node, for example, we have a "Project" entity that checks if the user is a member of the project, once the user is allowed to enter the node, they can access the project's fields.

In order to keep practical queries like all the `by_pk` ones, we add them when needed in those nodes as computed fields : 
```graphql
projects_by_pk("1") {
  task_by_pk("123") { name }
}
```

Here `task_by_pk` is in fact a computed field on the "projects" table. It works beautifully, but the fact that we get an array back is a real pain on the frontend as we have to unwrap the single entity returned by the function every time.

I hope it will be considered, as the feature already exists for top-level functions I guess most of the solution is already there. 😄

akhilrajvc:
Hi, Is there any update on this

mcsimps2:
Eager for this feature as well.  The resulting code without this functionality is very spaghetti.  Lots of developer confusion seeing `myComputedField[0].foo` in the codebase instead of simply just `myComputedField.foo`.

debagger:
[...]

PR #8524, opened 2022-05-24
Add _cast support for other postgres datatypes fix #6050 and fix #5426

abdullah2993:
### Description
Right now[ `_cast` only supports geometry, geography and jsonb](https://hasura.io/docs/1.0/graphql/core/queries/query-filters.html#cast-a-field-to-a-different-type-before-filtering-cast) types. This PR adds `_cast` support to almost *all PostgreSQL types.
`_cast` now supports following data types:

1. SmallInt
2. Integer
3. BigInt
4. Float
5. Double
6. Numeric
7. Money
8. Boolean
9. Char
10. Date
11. TimeStamp
12. TimeStampTZ
13. TimeTZ
14. JSON
15. UUID

### Changelog

- [x] `CHANGELOG.md` is updated with user-facing content relevant to this PR. If no changelog is required, then add the `no-changelog-required` label.

### Affected components

- [x] Server
- [x] Tests

### Related Issues
#6050 
#5426

### Steps to test and verify
you can now run queries like so
```gql
query {
  cast_test (where: {numeric_col: {_cast: {String: {_ilike: "123"}}}}){
    id
    numeric_col
  }
}
```

#### Catalog upgrade

Does this PR change Hasura Catalog version?
- [x] No
- [ ] Yes
  - [ ] Updated docs with SQL for downgrading the catalog 

#### Metadata
Does this PR add a new Metadata feature?
- [x] No
- [ ] Yes

#### GraphQL
- [ ] No new GraphQL schema is generated
- [x] New GraphQL schema is being generated:
   - [x] New types and typenames are correlated

#### Breaking changes

- [x] No Breaking changes
- [ ] There are breaking changes:

PR #8938, opened 2022-09-07
server: return single object from non-SETOF SQL functions

kraftwerk28:
### Description

When applying an SQL function for table computed field, the function might either return:
1. a scalar (`bool`, `int`, `decimal` etc)
2. `table_name`, where `table_name` is tracked by Hasura. The return value is a single object
3. `SETOF table_name`, where `table_name` is tracked by Hasura. The return value is an array of objects

The issue is that for 2nd case, Hasura generates `SETOF`-like schema, i.e. the return value is an array with a single object in it. This is somewhat inconvenient, as the code which accesses the computed field value, must always index into the array.

A simplified example:

```sql
create table "company" (
  company_id bigserial primary key,
  company_name text not null unique
);
create table "user" (
  user_id bigserial primary key,
  username text not null unique,
  company_id bigint not null references "company" (company_id)
);

-- This can be achieved using object relationships, but for the sake of example...
create function user_company("user")
  returns "company"
  language sql
  immutable
as $$
  select * from "company" c
  where c.company_id = $1.company_id
  limit 1;
$$;
```
```gql
mutation A {
  insert_company_one(object: {company_id: 1, company_name: "Nokia"}) {
    company_id
  }
  insert_user_one(object: {user_id: 1, username: "John", company_id: 1}) {
    user_id
  }
}
```

### Changelog

__Component__ : server / cli / console / build 

__Type__: bugfix / feature / enhancement 

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

This fixes https://github.com/hasura/graphql-engine/issues/4678 / https://github.com/hasura/graphql-engine/issues/6562


nosovk:
@rikinsk could you please approve CI builds for PR?
It will also allow us to build that one https://github.com/hasura/graphql-engine/pull/8728

plcplc:
Hi @kraftwerk28.

Thanks for making this contribution, and thanks for a very articulate PR description.

This looks like a change we would like to include. I'm going to be reviewing the PR and likely also push some additional commits that add tests of the feature.

plcplc:
> Is there any way to build&publish this branch on CI to test out my changes after I fix the issue? My main machine with 16GB RAM + 4GB swap is unable to build the graphql-engine, `ghc` is simply OOM-killed.

I'm going to have a look and see if we support CI for community-contributed PRs.

Unless you have already, I can recommend turning off compiler optimizations for the graphql-engine library by adding `flags: -optimize-hasura` at the top-level of `cabal.project.local`.

nosovk:
Thanks, `flags: -optimize-hasura` improved memory consumption and was able to run it locally

Stefmore02:
Hey! @kraftwerk28 

This is a friendly reminder to address the comments/changes made to proceed with this PR. If any questions, don't hesitate to reach out! 🤓

nosovk:
@Stefmore02 could you allow running CI on a branch?

server/src-lib/Hasura/Backends/Postgres/Translate/Select/Internal/Process.hs
```diff
@@ -342,10 +342,15 @@ processAnnFields sourcePrefix fieldAlias similarArrFields annFields tCase = do
               fieldName
               PLSQNotRequired
               sel
-          let computedFieldTableSetSource = ComputedFieldTableSetSource fieldName selectSource
+          let selectSourceWithoutNulls =
+                case selectSource of
+                  SelectSource {_ssFrom = S.FIFunc (S.FunctionExp {S.feAlias = Just (S.FunctionAlias {S._faIdentifier = fnIden})})} ->
```
kraftwerk28, line 347: @plcplc I think I need some help here, probably there is neater way of getting a function alias for adding it to a `WHERE` clause

plcplc: Don't worry about that - to me this does not look like an unreasonable way to get a function alias.

But is it correct to add a `WHERE function-result-alias IS NOT NULL` condition? It's been a while since I looked at this PR, but if you have a function which might return null (which it can in principle, since it returns a single (if composite) value) have that be reflected in the response you get?

kraftwerk28:
It now works as expected, non-SETOF computed fields don't have `where` / `limit` / `offset` / `order by` / `distinct` clauses and are single value, nullable.
image

plcplc:
@kraftwerk28 I think this is shaping up pretty good.

It would help it along if you could add a test of this in the `server/lib/api-tests` test suite. This should be relatively easy if you simply copy one of the existing tests of computed fields and adapt it.

kraftwerk28:
Note, that this PR introduces the following changes to the GraphQL schema of computed fields:

1. `SETOF <table>`: The return type changed from `[some_type!]` to `[some_type!]!`. That is, if a `SETOF` function returns `NULL`, it means an empty, non-nullable array (so, no `NULL / []` ambiguity)
2. `<table>`: The return type changed from `[some_type!]` to `some_type`

kraftwerk28:
@plcplc, I've added tests for Postgres and BigQuery. Because BigQuery doesn't support `SETOF` modifier, the computed fields for this backend type always act as if the function was `SETOF` (`[some_type!]!` in GQL).

plcplc:
@kraftwerk28 : Thanks for this.

As far as developing the actual feature I think this PR is in good shape.
However, since it does make potentially breaking changes to the schema which are not opt-in (because an instance having a non-setof computed field will see the type of that change from `[a]` to `a!` when it is upgraded), we need to be mindful of how we actually introduce such a change into the graphql-engine.

Currently our process for handling that is not particularly streamlined, and we're deliberating internally how to accommodate changes such as this one as well of [several](https://github.com/hasura/graphql-engine/pull/8524) [others](https://github.com/hasura/graphql-engine/pull/9414).

I don't know exactly how we're going to proceed. It may be that we'll add a means to make the change opt-in, but since we're yet undecided I won't ask you to come up with that :-)

nosovk:
It would be nice to have it in graphql-engine. We have some pain with that issue, the solution is not something out of nowhere, its an issue from real project.

plcplc:
> It would be nice to have it in graphql-engine. We have some pain with that issue, the solution is not something out of nowhere, its an issue from real project.

Yeah, I can appreciate that. There's no question that this is, by itself, purely a good change. I'll try and get back onto this as soon as we have made up our minds about how to proceed!

TRONJon:
I myself have ran into this issue with several new projects. If there's any chance of getting this PR through (without breaking backwards compatibility with old schemas) then my dev team will be incredibly grateful 🙇

dariocravero:
Hey all, it's been a while since the last conversations here. What's the likelihood of this getting merged at the end? Would love to see this in. Thanks!

PR #8940, opened 2022-09-07
Update Digital Ocean docs with caddy bug fix info

tehpsalmist:
### Description
https://caddy.community/t/way-to-cache-ssl-certificate-and-not-reissue-it-every-reload-docker/10385

This issue crops up if you reload the docker-compose file too often. This PR inserts a step in the docs to clarify how to avoid that issue.

### Changelog

__Component__ : docs

__Type__: bugfix

#### Short Changelog

inserts a step in the Digital Ocean guide for enabling SSL via Caddy.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes


robertjdominguez:
Thanks for this PR, @tehpsalmist 🚀 

Taking a look at this and getting back to you soon 👍

tehpsalmist:
> Thanks for this PR, @tehpsalmist 🚀 
> 
> 
> 
> Taking a look at this and getting back to you soon 👍 

If you need any insight from me while reviewing, I'm available any time: bstewardcodes@gmail.com

shahidhk:
Hi @tehpsalmist, the volumes are only removed if you use `-v` in docker compose down command, right? Why do we need to explicitly create the volume then? 
Let me know if I am missing something.

tehpsalmist:
> do we need to explicitly create the volume then?

Admittedly, I don't know enough about docker to say that this is an exclusively required step, but I do know that the existing commands and configurations given led to the volume being destroyed every time down/up was run. If you know how to configure the docker compose file so that the volume will persist without manually creating it, then that would also solve the problem. Then we can update the repo that holds the Digital Ocean configuration with that change.

mendeni:
@hasura/infra-tools could someone review this PR when possible?

nithindv:
@tehpsalmist Apologies for the delay. 

We tried to reproduce this but could not do so; it works as expected. The volumes continue to stay intact unless you run a `docker-compose down -v`.

 Could you share a detailed list of commands that we can use to replicate this?

Stefmore02:
I'm closing this PR due to lack of activity. Feel free to re-open if you want to pick it back up.

PR #9105, opened 2022-10-12
Update query.mdx for boolean expresssions - close #9056

### LINKED ISSUE #9056: Docs for Boolean expressions are incomplete
The API docs specify that a Boolean expression (e.g. for `where` clauses) is one of five options.
https://github.com/hasura/graphql-engine/blob/52873dba95df9aac8dfee174597c7967291e4828/docs/docs/api-reference/graphql-api/query.mdx#L366-L384
In particular, `ColumnExp` simply represents a comparison operator.
https://github.com/hasura/graphql-engine/blob/52873dba95df9aac8dfee174597c7967291e4828/docs/docs/api-reference/graphql-api/query.mdx#L578-L620

At first sight, this is matched by the internal Haskell representation: a `GBoolExp` is one of five options.
https://github.com/hasura/graphql-engine/blob/52873dba95df9aac8dfee174597c7967291e4828/server/src-lib/Hasura/RQL/IR/BoolExp.hs#L81-L90
However, in most cases, the `a` type parameter used by `BoolFld` is filled in by `AnnBoolExpFld`:
https://github.com/hasura/graphql-engine/blob/52873dba95df9aac8dfee174597c7967291e4828/server/src-lib/Hasura/RQL/IR/BoolExp.hs#L504-L517
This allows for not just a binary operator on a column (`AVColumn`), but also for deeper Boolean expressions in the case of relationships and computed fields.

Concretely, this means that something like
```graphql
query {
  articles(where: {author: {hometown: {_eq:"Amsterdam"}}}) {
    title
  }
}
```
is not considered a valid query by the documentation.

Was this part of the documentation possibly written before support for DB table relationships was added? In any case, it didn't get improved after computed fields were added.

jerempy:
Update to show that columnexp and booleans with can use relationships. Also removed the anchors on headers so that the links will work
close #9056

### Description

The docs concerning the boolean expressions were dated and so didn't express that you could use relationships as part of your queries.
### Changelog

__Component__ : server 

__Type__: bugfix 

#### Short Changelog

Clarify docs for relationships as part of boolean expressions in queries
#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/docs/docs/api-reference/graphql-api/query.mdx
+++ b/docs/docs/api-reference/graphql-api/query.mdx
@@ -345,8 +345,6 @@ enum article_select_column {
   <code>
     {`where: `}
     <a href='#boolexp'>BoolExp</a>
-    {` | `}
-      {`{ relation-name | computed-field-name: `}
   </code>
 </pre>
 
@@ -361,15 +359,6 @@ query {
     articles {
       title
     }
-  }
-}
-
-# And with relationships or computed fields:
-
-query {
-  author(where: { articles: { rating: { _gte: 4 } } }) {
-    id
-    name
   }
 }
 ```
@@ -594,12 +583,10 @@ query {
 
 <pre>
   <code>
-    {`
-    { computed-field-name | relation-name: }{`
-     {` | `}
-      `field-name: { `}
-        <a href='#operator'>Operator</a>
-        {`: Value }
+    {`{
+  field-name: { `}
+    <a href='#operator'>Operator</a>
+    {`: Value }
 }`}
   </code>
 </pre>
@@ -615,16 +602,32 @@ query {
     content
   }
 }
+```
 
-# can also use relationships and computed fields:
+:::info Note
 
+ColumnExp applies to more than just the column or the `field-name` but also apply to relationships and computed fields as well.
+
+:::
+
+**More examples:**
+
+```graphql
 query {
   articles(where: {author: {hometown: {_eq:"Amsterdam"}}}) {
     title
   }
 }
 
+# And also:
+
+query {
+  articles(where: {_not: {articles: {rating: {_gte: 4}}}}) {
+    title
+  }
+}
 ```
+
 
 ###### Operator {#operator}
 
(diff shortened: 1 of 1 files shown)


abooij:
Thanks for the PR, @jerempy.

It seems that there are some syntax errors in the code, as the docs build is failing. Perhaps @seanparkross can help you out with that? I'm not too familiar with the syntax of the docs.

As for the changes from a semantic PoV, I'm not so sure. This PR seems to be changing `WhereExp`, rather than `BoolExp`. This brings the docs only marginally closer to what the server actually supports. For instance, what about `where: {_not: {articles: {rating: {_gte: 4}}}}`? This is supported by the server, but not by the docs after (or before) this PR.

jerempy:
@abooij Thanks for the reply and insight! Makes sense. I took another shot at it. lemme know what you think.

abooij:
This new attempt adds a few helpful examples. That's good. However, it doesn't address #9056, because it doesn't change the grammar spelled out in the documentation - that is to say, it doesn't change `BoolExp`, `ColumnExp`, etc. There's something material missing in the actual grammar (namely the possibility to use relationships as part of boolean expressions), not just in the explanation.

@seanparkross could you please help trace down why the build of the docs isn't succeeding?

robertjdominguez:
@abooij, saw your comment via email. I pulled the branch and it ran fine locally. I went to the CI steps and scrolled down to find `Cloudflare` where it says deployed successfully. I don't know why the comment in the `mono` PR didn't update, but you can scroll down there to see the preview build URL.

jerempy:
@abooij thanks for the insight and patience. I took another shot - 🤞 i got it

seanparkross:
@jerempy Thanks for the Hacktoberfest submission from the Docs team! We're going to have some swag coming your way soon. 

In terms of moving forward on this PR we're going to loop back on it nearer the end of the month when our dev teams have more availability. 

Thanks

Stefmore02:
Hey! @jerempy 

We'd love to send you some Hasura swag as a token of appreciation. Could you please send me an email to estefani@hasura.io with a reference of this PR so I can send you the swag form? ✨

Issue #8643, opened 2022-06-29
Migrations fail if there exists a function that is added as a root field and as a computed field


harishnair96:
### Version Information

Server Version: 2.8.1

### Environment

OSS

### What is the expected behaviour?

Hasura allows tracking the same function as a root field and a computed field at the same time. Therefore it is expected that Hasura should continue to behave normally and migrations should continue to work.

### Keywords

function, computed field, overloaded function, tracked function

### What is the current behaviour?

If the same function is tracked as a root field and as a computed field, then migrations like adding a new column, adding a foreign key, creating a function etc. do not work. Certain migrations like creating a table still work.

### How to reproduce the issue?

1. Run the following SQL as a migration in Hasura:

```sql
CREATE TABLE BOOK(ID SERIAL PRIMARY KEY, AUTHOR_ID INTEGER);
CREATE TABLE AUTHOR(ID SERIAL PRIMARY KEY);
```

2. Track the tables _author_ and _book_.
3. Run the following SQL as a migration in Hasura:

```sql
CREATE FUNCTION authors(author_row author)
RETURNS SETOF author AS $$
    SELECT *
    FROM author
$$ LANGUAGE sql STABLE;
```

4. Track the function _authors_ as a root field.
5. Create a computed field _authors_comp_ in the _author_ table with the _authors_ function.
6. Try any of the following migrations and observe the error:
    1. Try to add a new column to any of the table
    2. Try to add a foreign key
    3. Try to add a new unique constraint on a key
    4. Try to create a new function

### Screenshots or Screencast

<img width="1512" alt="error" src="https://user-images.githubusercontent.com/20351766/176389156-9fe46f08-153a-4373-a933-4a719db47ead.png">

PR #9112, opened 2022-10-14
server: deduplicate Postgres function names with HashSet (close #8888)

### LINKED ISSUE #8888: server/refactor: use `Set`/`HashSet` to deduplicate function names in `runTxWithMetadataCheck`
The root cause of [graphql-engine#8643: Migrations fail if duplicate functions exist](https://github.com/hasura/graphql-engine/issues/8643) is that the list of functions to consider for metadata diff checks ends up with duplicates when a function is tracked both as root field and as computed field:

https://github.com/hasura/graphql-engine/blob/f80ec1d543e0e3e4711fe72bb8f19c097e6897c1/server/src-lib/Hasura/Backends/Postgres/DDL/RunSQL.hs#L303

The proper fix is to ensure that we use [sets](https://hackage.haskell.org/package/containers-0.6.6/docs/Data-Set.html)/[hashsets](https://hackage.haskell.org/package/unordered-containers-0.2.19.1/docs/Data-HashSet.html) in all the functions involved instead of lists in order to statically ensure no duplicates appear by accident.

daniel-shuy:
#### Long Changelog

There was a bug (#8643) where if a function is added as both a root field and computed field at the same time, the migrations would fail.

This was temporarily resolved in b0d2262 by using [Data.List#nub](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-List.html#v:nub) in `runTxWithMetadataCheck` to remove duplicates, however [Data.List#nub](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-List.html#v:nub) is really inefficient (`O(n^2)` time-complexity).

This PR refactors `runTxWithMetadataCheck` to use a `Set` instead of [Data.List#nub](https://hackage.haskell.org/package/base-4.16.3.0/docs/Data-List.html#v:nub) to deduplicate the Postgres function names.

### Related Issues

#8888

### Solution and Design

I used [Data.HashSet.InsOrd](https://hackage.haskell.org/package/insert-ordered-containers-0.2.5.1/docs/Data-HashSet-InsOrd.html) instead of [Data.HashSet](https://hackage.haskell.org/package/unordered-containers-0.2.19.1/docs/Data-HashSet.html) to maintain insertion order, so that the sorting is the same as now (using `List`), is that desired?

Alternately, should I use [Data.Set](https://hackage.haskell.org/package/containers-0.6.5.1/docs/Data-Set.html) to sort by alphabetical order instead?

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/server/src-lib/Hasura/Backends/Postgres/DDL/RunSQL.hs
+++ b/server/src-lib/Hasura/Backends/Postgres/DDL/RunSQL.hs
@@ -22,7 +22,6 @@ import Control.Monad.Trans.Control (MonadBaseControl)
 import Data.Aeson
 import Data.HashMap.Strict qualified as M
 import Data.HashSet qualified as HS
-import Data.HashSet.InsOrd qualified as OHS
 import Data.Text.Extended
 import Database.PG.Query qualified as PG
 import Hasura.Backends.Postgres.Connection.MonadTx
@@ -152,7 +151,7 @@ fetchTablesFunctionsMetadata ::
   ) =>
   TableCache ('Postgres pgKind) ->
   [TableName ('Postgres pgKind)] ->
-  OHS.InsOrdHashSet (FunctionName ('Postgres pgKind)) ->
+  HS.HashSet (FunctionName ('Postgres pgKind)) ->
   m ([TableMeta ('Postgres pgKind)], [FunctionMeta ('Postgres pgKind)])
 fetchTablesFunctionsMetadata tableCache tables functions = do
   tableMetaInfos <- fetchTableMetadata tables
@@ -299,8 +298,8 @@ runTxWithMetadataCheck source sourceConfig txAccess tableCache functionCache cas
 
         -- Before running the @'tx', fetch metadata of existing tables and functions from Postgres.
         let tableNames = M.keys tableCache
-            computedFieldFunctions = concatMap getComputedFieldFunctions (M.elems tableCache)
-            functionNames = OHS.fromList $ M.keys functionCache <> computedFieldFunctions
+            computedFieldFunctions = HS.fromList $ concatMap getComputedFieldFunctions (M.elems tableCache)
+            functionNames = M.keysSet functionCache <> computedFieldFunctions
         (preTxTablesMeta, preTxFunctionsMeta) <- fetchTablesFunctionsMetadata tableCache tableNames functionNames
 
         -- Since the @'tx' may alter table/function names we use the OIDs of underlying tables
@@ -400,7 +399,7 @@ fetchTablesFunctionsFromOids ::
   (MonadIO m) =>
   [OID] ->
   [OID] ->
-  PG.TxET QErr m ([TableName ('Postgres pgKind)], OHS.InsOrdHashSet (FunctionName ('Postgres pgKind)))
+  PG.TxET QErr m ([TableName ('Postgres pgKind)], HS.HashSet (FunctionName ('Postgres pgKind)))
 fetchTablesFunctionsFromOids tableOids functionOids =
   ((PG.getViaJSON *** PG.getViaJSON) . PG.getRow)
     <$> PG.withQE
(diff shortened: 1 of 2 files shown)


soupi:
@daniel-shuy Thanks for your contribution!
As far as I can tell the ordering doesn't matter, so using a regular `HashSet` is fine (@plcplc please correct me if I'm wrong).

server/src-lib/Hasura/Backends/Postgres/DDL/RunSQL.hs
```diff
@@ -300,7 +300,7 @@ runTxWithMetadataCheck source sourceConfig txAccess tableCache functionCache cas
         -- Before running the @'tx', fetch metadata of existing tables and functions from Postgres.
         let tableNames = M.keys tableCache
             computedFieldFunctions = concatMap getComputedFieldFunctions (M.elems tableCache)
-            functionNames = nub $ M.keys functionCache <> computedFieldFunctions
+            functionNames = OHS.fromList $ M.keys functionCache <> computedFieldFunctions
```
soupi, line 303: If we're using `HashSet` instead, this could be:

```suggestion
            functionNames = M.keysSet functionCache <> HS.fromList computedFieldFunctions
```

daniel-shuy:
@plcplc would love to get your feedback on this :)

plcplc (changes requested):
@daniel-shuy: Thanks for taking the time to make this PR, and sorry for taking a while to get to reviewing it!

Some thoughts:
* While it's true that `nub` on lists has worse asymptotics than hash sets, I'm wondering if this is actually any faster in most cases.
* I would expect the size of the input to be in the tens of elements for most cases and no larger than in the hundreds in any imaginable case.
* This code only runs when the server handles a `run_sql` api call, which crucially is not on the hot path of graphql query execution.

The real problem with the code in this module is that lists are used to model things that seem to be better modeled with sets, which complicates reading and reasoning about it.

Because of this it doesn't seem compelling to me to just change a one part of the processed data to a hash set to speculatively improve performance where performance isn't critical.

I'd almost wager that most of the data processed here could be better modeled by (unordered) sets. Implementing that would be a compelling change :-)

daniel-shuy:
@plcplc do you mean changing the return type for all the functions that return Postgres function names (e.g. `computedFieldFunctions`) to `HashSet`?

plcplc:
> @plcplc do you mean changing the return type for all the functions that return Postgres function names (e.g. `computedFieldFunctions`) to `HashSet`?

That's the gist of it yes. But it's been a while since I looked closely at this, and in certain places duplicates do indicate errors that need to abort execution, so we likely can't just make the change blindly.

daniel-shuy:
What about the functions that are implemented for other database vendors? Should I change it for all the other databases as well?

plcplc:
I think a sensible scope for this is just the Postgres backend :-)

server/src-lib/Hasura/Backends/Postgres/DDL/RunSQL.hs
```diff
@@ -300,7 +300,7 @@ runTxWithMetadataCheck source sourceConfig txAccess tableCache functionCache cas
         -- Before running the @'tx', fetch metadata of existing tables and functions from Postgres.
         let tableNames = M.keys tableCache
             computedFieldFunctions = concatMap getComputedFieldFunctions (M.elems tableCache)
```
daniel-shuy, line 302: Not sure if it makes sense to change `getComputedFieldFunctions` to return `HashSet` instead of list. There doesn't seem to be an equivalent of `concatMap`/`concat` for `HashSet` (coming from Scala, it is pretty frustrating that the Haskell containers don't implement common interfaces and are not interoperable).

soupi (approved):
Thank you for the contribution! I think this is an improvement over the current code, so I am approving it.

Stefmore02:
Hello @daniel-shuy 

Thanks so much for your contribution to Hasura OSS during the Hacktoberfest 2022! We'd love to send you some Hasura swag as a token of appreciation and you will be getting an email with the swag form shortly.✨

PR #9160, opened 2022-10-31
[WIP] chore(packaging): use nix flakes for building graphql-engine reproducibly

thenonameguy:
@SamirTalwar  I saw your work on adding the scaffolding for the development environment and decided to add Nix for building aarch64-darwin binaries out of graphql-engine. This project is a WIP branch that does everything without much regard to existing practices/best conventions to do a POC for Nix + Hasura => M1 native gql-engine.

### Changelog

__Component__ : server / cli / console / build 

__Type__: bugfix / feature / enhancement 

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [ ] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [ ] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


nix/overlays/packages.nix
```diff
@@ -0,0 +1,105 @@
...
+          graphql-server = super.haskell.lib.justStaticExecutables
+            ((super.haskell.lib.dontCheck (hsuper.callPackage ../../server {
+              hedgehog = hself.hedgehog_1_2;
+              immortal = hself.immortal_0_2_2_1;
+              pg-client = hself.pg-client-hs;
+              odbc = hself.hasura-odbc;
+              resource-pool = hself.hasura-resource-pool;
+            })));
+
+          # FIXME: for remote repos we have to calculate their default.nix until they get upstreamed/maintained
```
SamirTalwar, line 38: These seem legit. I suggest moving the actual fetches into the inputs in _flake.nix_ though, so they're pinned with _flake.lock_. (You can still specify the exact commit ref.)

server/CURRENT_VERSION
```diff
@@ -0,0 +1 @@
+2.14.0-beta.2
```
SamirTalwar, line 1: Please don't check this in. For some reason, this is missing from the _.gitignore_ file; I am remedying this.

For local development, you can write "12345" to this file.

server/default.nix
```diff
@@ -0,0 +1,112 @@
+{ mkDerivation, aeson, aeson-casing, aeson-ordered, aeson-pretty
```
SamirTalwar, line 1: Any reason not to use `callCabal2nix` for these?

If it's just for the version, you can override that later. I think this will work:

```nix
(callCabal2nix "graphql-engine" ./server {}).overrideAttrs(oldAttrs: {
  version = builtins.readFile ./server/CURRENT_VERSION;
})

File: server/graphql-engine.cabal [hunk 4]
```diff
@@ -158,7 +158,8 @@ flag ghci-load-test-with-lib
 common common-all
   ghc-options:
     -foptimal-applicative-do
-    -Wall -Wcompat -Wincomplete-record-updates -Wincomplete-uni-patterns -Wredundant-constraints
+    -Wall -Wcompat -Wincomplete-record-updates -Wincomplete-uni-patterns
+    -- -Wredundant-constraints
```
SamirTalwar, line 162: Why did you have to remove this? Did you get extra warnings?

server/graphql-engine.cabal
```diff
@@ -293,6 +294,7 @@ common lib-depends
                      , some
                      , split
                      , template-haskell
+                     , ghc-heap-view
```
SamirTalwar, line 297: I assume this change is just for testing.

SamirTalwar (commented):
Looks like an awesome start. I've left a few comments. Can I help with anything?

nix/overlays/packages.nix
thenonameguy: Yup, that's the plan. I need to do this anyway to not use [IFDs](https://nixos.wiki/wiki/Import_From_Derivation).

server/CURRENT_VERSION
thenonameguy: 1. Flake builds can only use files that are tracked in version control (as the build depends on this not-deterministic file)
2. I thought the version was used to fetch the CDN FE assets for the Console, is this not correct?

server/default.nix
thenonameguy: `callCabal2nix` is an anti-pattern in the flake/nixpkgs world, as it creates evaluations at runtime.
The computation of these files, while being more convenient, are done at eval time. 
https://github.com/input-output-hk/haskell.nix uses this approach extensively, I tried using that on a different branch in the past days and the IFD-less version of this branch is way more snappier to eval/build with. 
IMO for now it's worth to do this outside of Nix evalution, maybe automate it with a Makefile/CI script.

server/graphql-engine.cabal
thenonameguy: Yes, got a lot of compilation errors for redundant-constrains and typed holes, which I found strange. Temporary fix to get a build.

server/graphql-engine.cabal
thenonameguy: your CPP module was essentially upstreamed into the latest ghc-heap-view version. The surrounding code failed to compile until these changes were made. Of course the runtime dependency for this should be made optional, but Nix did not seem to find it using the conditional profiling include.

thenonameguy:
> Looks like an awesome start. I've left a few comments. Can I help with anything?

Thanks! Currently:
- x86-linux: builds
- aarch64-darwin: compiles but linking segfaults, same as https://github.com/NixOS/nixpkgs/issues/194367#issuecomment-1274564762
- x86-darwin: is compiling as we speak with Rossetta 2, should work.

server/default.nix
thenonameguy: Further reading:
https://github.com/input-output-hk/haskell.nix/blob/master/docs/tutorials/materialization.md

SamirTalwar:
We saw those same linking issues trying to build Ormolu with Nix, which was fixed by disabling the Template Haskell.

My guess is that it's similar here; the Template Haskell is causing serious problems _somewhere_. I'm not sure what to do about it, unfortunately. It definitely feels like an upstream bug, but I don't even know where. Is it with the version of `ld` shipped with nixpkgs on macOS/aarch64, or somewhere else?

What do you propose we do about it?

domenkozar:
This would be so nice to have, I'd like to support Hasura in devenv.sh: https://github.com/cachix/devenv/issues/308

SamirTalwar:
@domenkozar, I take it your main motivation is Linux support, not macOS?

If we were to reorient this PR to just care about Linux, it might be easier to get it in.

domenkozar:
Linux and macOS, I'm happy to help here if I can understand what are the blockers.

SamirTalwar:
The main macOS blocker is that `ld` chokes on HGE. The error is the same as https://github.com/NixOS/nixpkgs/issues/149692.

We would also need to clean this PR up a bit, but it's mostly small things; @thenonameguy has done a wonderful job.

server/CURRENT_VERSION
SamirTalwar: Apologies, I thought I'd replied to your comments.

I suggest adding a `preConfigure` step that writes the file with the contents of the package version then.

server/default.nix
SamirTalwar: Very helpful, thank you!

We'd need to add some CI to make sure this stays in sync with however it's generated, as well as include instructions on how to redo it.

server/graphql-engine.cabal
SamirTalwar: @jberryman: Do you think we can upgrade `ghc-heap-view` to take advantage of this?

thenonameguy:
Cheers, FYI at my company have been using this branch + Cachix binary cache for our local Hasura development ever since without hiccups. I propose you take over the necessary bits and pieces over to a new branch as I do not have the time for foreseeable future to improve this further.

Super stoked to see this initiative going forward!

SamirTalwar:
Thanks @thenonameguy, glad it's working well at your end. We'll do our best to incorporate it.

john-rodewald:
I picked up where @thenonameguy left off, transformed my fork into a monstrosity of patches that I don't _really_ want anyone to see, and eventually managed to build on `aarch64-darwin` with Nix. I'm not sure when or if I'll have time to clean it up, but it runs and may be helpful to someone.

https://github.com/john-rodewald/graphql-engine/tree/master-nix

server/default.nix
srid: > We'd need to add some CI to make sure this stays in sync with however it's generated, as well as include instructions on how to redo it.

You can include a pre-commit hook [that runs cabal2nix](https://github.com/cachix/pre-commit-hooks.nix/blob/bd38df3d508dfcdff52cd243d297f218ed2257bf/modules/hooks.nix#L1047-L1053).

Note that devs don't have to install the hooks themselves, but only have to run `pre-commit run` after staging the changes. Else, `nix flake check` should be expected to fail.

server/lib/error-message/default.nix
```diff
@@ -0,0 +1,17 @@
...
+  version = "1.0.0";
+  src = ./.;
+  libraryHaskellDepends = [
+    aeson base graphql-parser text unordered-containers
+  ];
+  testHaskellDepends = [
+    aeson base graphql-parser hashable hspec text unordered-containers
+    vector
+  ];
+  homepage = "https://github.com/hasura/github-engine#readme";
```
the-sun-will-rise-tomorrow, line 15: `s/github-engine/graphql-engine/` ?

the-sun-will-rise-tomorrow:
@john-rodewald FYI, `nix run github:john-rodewald/graphql-engine/master-nix` now fails with  `error: Package ‘openapi3-3.2.3’ in /nix/store/qmyflblrsa4qikr5q54cfk0hxqx2a5p6-source/pkgs/development/haskell-modules/hackage-packages.nix:213450 is marked as broken, refusing to evaluate.`.

the-sun-will-rise-tomorrow:
> The main macOS blocker is that `ld` chokes on HGE. The error is the same as [NixOS/nixpkgs#149692](https://github.com/NixOS/nixpkgs/issues/149692).

It looks like that problem may have been fixed since then.

PR #9319, opened 2022-12-19
Added depends_on to hasura container config

cmartin616:
Adding depends_on gdw_server to hasura container config will avoid start up race conditions when the agent is being queried by Hasura GraphQL Engine prior to being available which results in inconsistent metadata errors.

### Description
The current docker-compose does not have dependencies between containers.  This can result in Hasura GraphQL Engine starting up with inconsistent metadata errors as it attempts to reach the agent prior to the agent being available.  This adds a dependency that requires the agent to be started prior to Hasura GraphQL Engine starting.  This resolves the race condition resulting in inconsistent metadata.

#### Catalog upgrade

- [ ] No

- [ ] No

#### GraphQL

#### Breaking changes

- [ ] No Breaking changes


sordina (approved):
LGTM

PR #9396, opened 2023-01-23
cli console: differantiate API listening address and API accessing hostname (fix #7241)

### LINKED ISSUE #7241: Should separate --address flag from `hasura console` into --address and --bind-address
When I'm running hasura console, I'm only given a chance to configure --address once. I assume this is because hasura-cli is assumed to be running on localhost. However, address does two things:
1. Binds both the api server and the console server to be running on that hostname (using http.ListenAndServe, which ultimately uses net.Listen)
2. Points the served console page towards the served api server.

According to the net.Listen documentation: https://pkg.go.dev/net#Listen
```
The address can use a host name, but this is not recommended, because it will create a listener for at most one of the host's IP addresses.
```

My setup requires me to run hasura console on a remote dev machine, and my local needs to connect to that dev machine. However, when I run hasura console:
- if i add `--address 0.0.0.0`, then the served console page will attempt to access the api server via `http://0.0.0.0:9693`, which cannot be reached.
- if i add `--address myhostname`, then the net.Listen will bind to 127.0.0.1, which doesn't allow me to access it from my computer.

It would be ideal if I could get hasura console to bind to 0.0.0.0, while using myhostname to locate the migrate api server.

zigarn:
### Description

#7519 added the `--api-host` and #8039 fixed a bit of collision with `--address`, but the merge of #8039 (f530b3ac6ffee87a97dd07fef49c03e5a4bd24de) added the listening of API on `--api-host`.

All this create a confusion between listening host and accessing hostname.

E.g. when using containers, it is an issue (when it is not possible to use `--net host`):

```sh
# Launching GraphQL Engine
docker container run --name hasura -d -p 8080:8080 -p 9695:9695 -p 9693:9693 hasura/graphql-engine:v2.17.0.cli-migrations-v3

# Launching console
docker container exec hasura hasura-cli console --no-browser --address 0.0.0.0  # 0.0.0.0 is mandatory for expose even with '-p 9695:9695'
```

When opening http://localhost:9695, browser is trying to access http://0.0.0.0:9693/apis/migrate/settings, which doesn't work (it's OK on Linux as it's redirected to 127.0.0.1, but doesn't work on other systems).

To avoid this, the use of `--api-host` is necessary:

```sh
# Launching console
docker container exec hasura hasura-cli console --no-browser --address 0.0.0.0 --api-host http://localhost
```

But this time, the browser is trying to access http://localhost:9693/apis/migrate/settings, but it doesn't work as the API is only exposed on 127.0.0.1 inside the container.

This PR differentiate the listening address of API server (using the same as the console) from the accessing (or exposed) address to use in the browser.

### Changelog

__Component__ : cli

__Type__: bugfix

#### Short Changelog

Use `--address` to expose the migrate API with console, `--api-host` is only used for accessing it from console UI.

### Related Issues

- #7241
- #7519
- #8039/#8005

### Solution and Design

Make API listen on `--address` instead of `--api-host`.

If we may want to listen on different interfaces, then should introduce a `--api-address`.

### Steps to test and verify

Launch console:

```sh
hasura-cli console --address 0.0.0.0 --api-host http://localhost
```

Check port 9693 is open on 0.0.0.0 (e.g. via `ss`), and console UI is calling http://localhost:9693/apis/migrate/settings

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/cli/commands/console.go
+++ b/cli/commands/console.go
@@ -83,7 +83,7 @@ func NewConsoleCmd(ec *cli.ExecutionContext) *cobra.Command {
 	f := consoleCmd.Flags()
 
 	f.StringVar(&opts.APIPort, "api-port", "9693", "port for serving migrate api")
-	f.StringVar(&apiHost, "api-host", "http://localhost", "(PREVIEW: usage may change in future) host serving migrate api")
+	f.StringVar(&apiHost, "api-host", "http://localhost", "(PREVIEW: usage may change in future) host for accessing migrate api")
 	f.StringVar(&opts.ConsolePort, "console-port", "9695", "port for serving console")
 	f.StringVar(&opts.Address, "address", "localhost", "address to serve console and migration API from")
 	f.BoolVar(&opts.DontOpenBrowser, "no-browser", false, "do not automatically open console in browser")
@@ -136,7 +136,7 @@ func (o *ConsoleOptions) Run() error {
 		return errors.E(op, "cannot validate version, object is nil")
 	}
 
-	apiServer, err := console.NewAPIServer(o.APIHost.Host, o.APIPort, o.EC)
+	apiServer, err := console.NewAPIServer(o.Address, o.APIPort, o.EC)
 	if err != nil {
 		return errors.E(op, err)
 	}


bwobst:
Any update on this issue?

zigarn:
Looks like this will never go further, closing as there is no point to keep it open.

PR #9471, opened 2023-02-27
optimize typescript compiler performance

beerose:
### Description

This PR makes TypeScript compiler over 3x faster:

BEFORE:
CleanShot 2023-02-27 at 14 05 48@2x

AFTER:
CleanShot 2023-02-27 at 14 03 18@2x

Long story short, I was looking at some bigger TypeScript repos to play a bit with optimizing TS performance. I generated a trace for the console and found that `react-hook-form` is quite slow (the `Controller` components turned out to be a bottleneck), so I first tried to fix that, but then found this issue: https://github.com/microsoft/TypeScript/issues/46948, and I upgraded TS to see if that helps. Turns out it does! 🔥

I also updated the code in a few places to make it work with the newest TS.

### Changelog

N/A

### Steps to test and verify

Run extended diagnostics to see the improvement.

You could think about excluding Storybook's stories from the ts build (add it to tsconfig and possibly create a separate config for Storybook?). That would improve the time even more.

### DIFF
--- a/console/package-lock.json
+++ b/console/package-lock.json
@@ -256,7 +256,7 @@
         "terser-webpack-plugin": "4.2.3",
         "ts-jest": "^27.1.4",
         "tsconfig-paths-webpack-plugin": "^3.5.1",
-        "typescript": "4.5.2",
+        "typescript": "4.9.5",
         "url-loader": "1.1.2",
         "webpack": "4.42.0",
         "webpack-cli": "4.10.0",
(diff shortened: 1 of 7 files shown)


manasag:
Hi @beerose, thanks for opening this pull request. Please not that all the Console related code has been migrated to a new folder called `frontend` and uses NX based build system. Please refer to [Readme](https://github.com/hasura/graphql-engine/blob/master/frontend/README.md) in that folder on new steps to build Console locally. You probably need to rework this PR for the new system. Feel free to drop in your questions if you are still facing issue. cc @beaussan

beaussan:
Hey @beerose !

Nice to see you here =)

We migrated the console to Nx and we need to update the readme of the /console to inform contributors this, sorry for this !

You can find the new workspace at `/frontend` (with [documentation here](https://github.com/hasura/graphql-engine/blob/master/frontend/docs/from-previous-console.md) about the before / after files locations).

I was working today on a ts & dependancies updates, thanks for your findings ! I'll bump it to 4.9.5 (it was 4.8.4 in my PR).

> You could think about excluding Storybook's stories from the ts build (add it to tsconfig and possibly create a separate config for Storybook?). That would improve the time even more.

This is something we are doing in the new workspace ([tsConfig](https://github.com/hasura/graphql-engine/blob/master/frontend/libs/console/legacy-ce/.storybook/tsconfig.json) of storybook, vs the one of the [lib](https://github.com/hasura/graphql-engine/blob/master/frontend/libs/console/legacy-ce/tsconfig.lib.json)) ! It indeed improve the speed !

beerose:
Oh well, unnecessary PR then 😅 Closing it

PR #9695, opened 2023-06-05
docs: add group titles to navigation

lukaselmer:
### Description

Add group titles to navigation for better visual grouping

### Solution and Design

before:

image

after:

image

There are certainly "cleaner" implementations than "abusing" the existing categories. On the other hand, we don't change existing code, and only add a couple of new lines.

The following things could be improved:

* Add `aria-*` properties to the new div
* The inline styles could be converted to a CSS class name, and the styles could be moved to css / scss
* Add an additional group title "Other" or "Technical" to group the stuff after "Security tools"

#### Catalog upgrade

- [x] No

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/docs/docs/business-logic/_category_.json
+++ b/docs/docs/business-logic/_category_.json
@@ -1,0 +1,4 @@
+{
+  "label": "Business Logic",
+  "position": 79
+}
--- a/docs/docs/data-api/_category_.json
+++ b/docs/docs/data-api/_category_.json
@@ -1,0 +1,4 @@
+{
+  "label": "Data API",
+  "position": 21
+}
--- a/docs/docs/release/_category_.json
+++ b/docs/docs/release/_category_.json
@@ -1,0 +1,4 @@
+{
+  "label": "Release",
+  "position": 129
+}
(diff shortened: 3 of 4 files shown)


robertjdominguez:
Thanks for this contribution, @lukaselmer! We have an internal PR that solves it in a similar way:

For each heading desired, a new file is created in this format in the `/docs` directory:

```
_heading_<heading_name>.mdx
```

Then, frontmatter is added to the file:

```
---
sidebar_position: 10
sidebar_label: 'Basics'
sidebar_class_name: sidebar_heading
---
```

The `sidebar_heading` class is parsed in `/src/theme/DocSidebarItem/Link/index.js` to determine whether to render a `<Link />` or a `<li />`. Headings cannot be clicked; thus, they're returned simply as `<li />.`

Additionally, to prevent a user from navigating to an empty page (as all `_heading_<heading_name>.mdx` pages are), the `pathname` is parsed in `/src/components/CustomDocItem/index.tsx` and, if `_heading_` is included in the `pathname`, the user is redirected to the index of the docs site. **Navigating to one of these routes is only achievable by a user explicitly typing in the address.**

```jsx
 if (props.location.pathname.includes('_heading_')) {
    return <Redirect to="/docs/latest/index/" />;
  }
```

Altogether, it looks like [this](https://pr-9427---docusaurus-5kcgrtdh3a-wl.a.run.app/docs/latest/index/). We'll consider aspects of your approach, and many thanks for your ideas in #9696 on the two-column layout inside a docs page 🤙

robertjdominguez:
Thanks again @lukaselmer 🤙 

You can see our merged update [here](https://github.com/hasura/graphql-engine/commit/60ba0566db22cb96b0b4225b0027b7a0f0ef9c5f).

PR #9769, opened 2023-07-11
Omit `Server` Header from Server Responses by Default

andrewlouisx:
### Overview

This PR addresses an unnecessary and potentially harmful default setting in the Warp server that automatically includes the `Server` header in all responses, explicitly revealing the server version as `Warp/3.3.23`.

Screenshot 2023-07-11 at 1 05 59 PM

### Problem

The presence of the `Server` header could inadvertently aid attackers by exposing specific software versions, which could be exploited if known security holes are present. The potential risks are outlined in [RFC 2068](https://www.ietf.org/rfc/rfc2068.txt) and further discussed in this article by [Troy Hunt](https://www.troyhunt.com/shhh-dont-let-your-response-headers/). The offending default can be located in the [Warp server's source code here](https://github.com/yesodweb/wai/blob/bf60ed2c18670a932cd84252b6fddafa3f986f42/warp/Network/Wai/Handler/Warp.hs#L337-L344).

### Solution

This PR proposes to disable the Server header by default, significantly reducing the exposure of potentially sensitive information.

I can also revise this to make it configurable if need be; but I think this is a sane default.

### Impact

Removing the `Server` header will mainly impact transparency, and should not disrupt functionality. There's a slight chance that clients relying on the header for metrics or logging might be affected. But given that the `Server` header offers no functional benefit to client-server interactions, these cases should be rare.

### Changelog

__Component__ : server 

__Type__: bugfix 

#### Short Changelog

`Server` header in server responses are omitted.

#### Long Changelog

The `Server` HTTP header is no longer included in server responses by default. This change enhances security by withholding potentially sensitive information about the server version that could aid attackers in identifying known security vulnerabilities. This new default behavior aligns with the recommendations in [RFC 2068](https://www.ietf.org/rfc/rfc2068.txt). 

### Related Issues
* https://github.com/yesodweb/wai/issues/619

### Steps to test and verify

Hitting the healthcheck endpoint is the easiest way here :) 

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

On the off-change that clients rely on the header for metrics or logging - they may experience a degradation. This is a highly unlikely scenario though.


jflambert:
@andrewlouisx this actually broke my integration

>  can also revise this to make it configurable if need be

this should have been part of the PR. For now I'll downgrade to Hasura v2.34

PR #9864, opened 2023-09-02
fix: avoid wrapping server in shell

jbergstroem:
### Description
When passing a command to CMD as a string it wraps the command with `/bin/sh -c` which is unhelpful since passing signals to the container such as INT would be intercepted by shell instead of the server.

This should make a shutdown more clean.

Also, by splitting entrypoint and command; one can do things like `docker run -it engine version` and it would pass `version` to the hasura engine.

### Changelog

__Component__ : build

__Type__: bugfix

#### Short Changelog

Don't spin up a shell when running in a container.

### Steps to test and verify

*Important*: I wasn't able to figure out how you layer your builds and couldn't find shell scripts that assembles the final image, so this needs more testing.

#### Catalog upgrade
- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging

### DIFF
--- a/packaging/graphql-engine/Dockerfile
+++ b/packaging/graphql-engine/Dockerfile
@@ -9,4 +9,5 @@ COPY rootfs/srv/ /srv/
 
 HEALTHCHECK --start-period=10s CMD curl -f http://localhost:8080/healthz || exit 1
 
-CMD "${HGE_BINARY}" serve
+ENTRYPOINT [ "${HGE_BINARY}" ]
+CMD [ "serve" ]


packaging/graphql-engine/Dockerfile
```diff
@@ -9,4 +9,5 @@ COPY rootfs/srv/ /srv/
 
 HEALTHCHECK --start-period=10s CMD curl -f http://localhost:8080/healthz || exit 1
 
-CMD "${HGE_BINARY}" serve
+ENTRYPOINT [ "${HGE_BINARY}" ]
```
SamirTalwar, line 12: Does this actually work? What interprets the environment variable?

packaging/graphql-engine/Dockerfile
```diff
@@ -9,4 +9,5 @@ COPY rootfs/srv/ /srv/
...
+CMD [ "serve" ]
```
SamirTalwar, line 13: This is a breaking change for any scripts orchestrating a HGE Docker container. For example, they might provide their own entrypoint but use the command syntax rather than `--entrypoint`.

I suggest we keep both the command and the argument in the `CMD` directive.

```suggestion
CMD ["${HGE_BINARY}", "serve"]
```

That said, this is moot until my other question is answered.

packaging/graphql-engine/Dockerfile
jbergstroem: Right; after reading a bit closer I'm inclined to agree to not being expanded. If you know where the documentation / steps to reproduce the finished container is, point me in that direction. I built the ubuntu base container and pointed to it as a build arg but it wasn't enough since I don't know where `rootfs` comes from.

Does this have to be dynamic? I can create a shell entrypoint that traps the signal otherwise.

packaging/graphql-engine/Dockerfile
jbergstroem: Right; happy to revert to this after we choose a direction for above.

packaging/graphql-engine/Dockerfile
SamirTalwar: If it's just the shell traps that you want, you can also run the Docker image with `--init` (or `init: true` in Docker Compose).

jbergstroem: yes, I just thought I'd contribute back so I can improve the state for everyone. Feel free to close if you think that's what's most appropriate at this stage.

jbergstroem:
Gonna close based on discussion.

PR #10116, opened 2024-01-25
rfc for improvements to permissions in v3

0x777:
[Rendered](https://github.com/hasura/graphql-engine/blob/0x777-v3-permission-improvements/rfcs/permission-improvements-v3.md)

### DIFF
--- a/rfcs/permission-improvements-v3.md
+++ b/rfcs/permission-improvements-v3.md
@@ -1,0 +1,303 @@
+# Permissions roadmap in V3
+
+We would like to improve on these known permission issues:
+
+## 1. Reusing parent permissions
+
+Let me explain this with an example:
+
+```graphql
+user, user_channel, channel, message
+user_channel:                     { user_id: { _eq: x-hasura-user-id } }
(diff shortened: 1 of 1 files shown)


rfcs/permission-improvements-v3.md
```diff
@@ -0,0 +1,330 @@
...
+
+    ```yaml
+    conflict_rules:
+    # admin insert permissions should be preferred over user role's insert
+    # permissions on all tables
+    - admin.insert > user.insert
+
+    # manager's update permissions should be preferred over user role's select
+    # update permissions on employee_details table
+    - manager.update.employee_details > user.update.employee_details
```
codingkarthik, line 134: I think this shouldn't be part of the metadata (otherwise we will run into the problems we had with inherited roles in V2) and rather it should be part of the GraphQL request, akin to SQL's `on_conflict`. 

I am not sure how this information can be provided on a per-request basis, but one idea I can think of which may work is to accept a header that accepts a JSON object with the above mentioned rules.

PR #10141, opened 2024-02-24
RFC for supporting Hasura in Apollo Federation

abhinav-hasura:
### Description

RFC for metadata changes to enable building a GraphQL schema using Hasura v3 that can be brought into Apollo Federation


abhinav-hasura:
@0x777

PR #10149, opened 2024-02-29
DOCS: Update metadata database url env var for k8s guide

clinejj:
### Description
When reviewing the environment variables here: https://hasura.io/docs/latest/deployment/graphql-engine-flags/reference/#metadata-database-url

I noticed that the metadata database URL env var was not current on the kubernetes deployment guide.

#### Catalog upgrade
- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required
    - [ ] Yes
    - [x] Not required

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code

  2. GraphQL API

     Schema Generation:

     Schema Resolve:-

  3. Logging

### DIFF
--- a/docs/docs/deployment/deployment-guides/kubernetes.mdx
+++ b/docs/docs/deployment/deployment-guides/kubernetes.mdx
@@ -37,7 +37,7 @@ Edit `deployment.yaml` and set the right database url:
 
 ```yaml {2}
 env:
-  - name: HASURA_METADATA_DATABASE_URL
+  - name: HASURA_GRAPHQL_METADATA_DATABASE_URL
     value: postgres://<username>:<password>@hostname:<port>/<dbname>
 ```
 
--- a/docs/docs/deployment/deployment-guides/kubernetes.mdx
+++ b/docs/docs/deployment/deployment-guides/kubernetes.mdx
@@ -41,7 +41,7 @@ env:
     value: postgres://<username>:<password>@hostname:<port>/<dbname>
 ```
 
-Examples of `HASURA_METADATA_DATABASE_URL`:
+Examples of `HASURA_GRAPHQL_METADATA_DATABASE_URL`:
 
 - `postgres://admin:password@localhost:5432/my-db`
 - `postgres://admin:@localhost:5432/my-db` _(if there is no password)_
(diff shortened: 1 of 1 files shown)


robertjdominguez:
Thanks for this PR @clinejj 👋 

This was recently updated as `HASURA_METADATA_DATABASE_URL` has some side effects:

`HASURA_GRAPHQL_DATABASE_URL` is a legacy env var from v1 days and should be avoided if possible as it has some quirky behavior. E.g., using it will make the postgres db to be connected by default to HGE as a data source on HGE startup, which is unwanted for a manifest targeted towards MySQL, for example.

clinejj:
@robertjdominguez thanks for following up! I'm not sure I totally understand then - is the env var list at  https://hasura.io/docs/latest/deployment/graphql-engine-flags/reference/#metadata-database-url not the intention for the kubernetes setup documented here?

Or is `HASURA_METADATA_DATABASE_URL` the intended environment variable for the kubernetes guide?

robertjdominguez:
I've made this as confusing as possible by mistyping earlier 🤦 

`HASURA_GRAPHQL_METADATA_DATABASE_URL` is the environment variable that has unintended side effects. 

You should be able to set `HASURA_METADATA_DATABASE_URL` in conjunction with an environment variable for a data source (e.g.,`PG_DATABASE_URL`) and proceed as you'd expect to serve metadata from ``HASURA_METADATA_DATABASE_URL` and your data from `PG_DATABASE_URL` or whatever variable name you wish.

clinejj:
Got it, I can close out this PR then. What are the unintended side effects of using `HASURA_GRAPHQL_METADATA_DATABASE_URL`, and should the environments variable documentation page be updated to suggest the newer `HASURA_METADATA_DATABASE_URL` ?

robertjdominguez:
Primarily that it automatically enables prepared statements. Though, this hasn't been officially deprecated — our focus (as you saw in the docs) was updating the variable users implement.

But, you're right — the server reference page could use an update. We'll take care of this!

Thanks for your attention to detail — we'd like to send you some swag 🥳 

I'll hit you up on LinkedIn 🤙

PR #10173, opened 2024-03-19
Alter log level of garbage collector message (fix #10172)

### LINKED ISSUE #10172: Change log level for GC message from warn to info
### Version Information

Server Version: 2.37.1

### Environment

OSS

A warning is generated whenever the GC is forced to run

An info message should be generated when GC runs as there is no expectation for a user to take action

1. Run Hasura without a reprise for memory usage
2. Observe logs

### Keywords

Overdue for a major GC: forcing one even though we don't appear to be idle

typhonius:
### Description
The garbage collector runs periodically to free unused memory with the default operation set to occur during quiet periods. In the event that a quiet period does not arise, the garbage collector is forced and a warning generated to inform users this has occured.

As warnings require users to take action, this log message causes confusion. Therefore, the log level of the garbage collection message has been altered from LevelWarn to LevelInfo to provide information that this has occured without expectation of users to take action.

### Changelog

Instead of a warning when the GC is forced to run, an info message will be generated.

__Component__ : server

__Type__: bugfix

#### Short Changelog

Instead of a warning when the GC is forced to run, an info message will be generated.

#### Catalog upgrade

- [ ] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code

  2. GraphQL API

     Schema Generation:

     Schema Resolve:-

  3. Logging

### DIFF
--- a/server/src-lib/Hasura/GC.hs
+++ b/server/src-lib/Hasura/GC.hs
@@ -75,7 +75,7 @@ ourIdleGC (Logger logger) idleInterval minGCInterval maxNoGCInterval =
               else do
                 when (areOverdue && not areIdle)
                   $ logger
-                  $ UnstructuredLog LevelWarn
+                  $ UnstructuredLog LevelInfo
                   $ "Overdue for a major GC: forcing one even though we don't appear to be idle"
                 performMajorGC
                 startTimer >>= go (gcs + 1) (major_gcs + 1) True


danieljharvey (approved):
Great stuff, thanks!

i-am-tom (approved):

Issue #10227, opened 2024-05-08
Hasura fails because max_function_args in postgres is no enough


ammirator-administrator:
### Version Information

Server Version:
2.36.1

### Environment

OSS

### What is the current behaviour?
Hasura fails to retrieve data from a Table (From Citus), and that happens because it seems that max_function_args in PG is not enough for that, the table indeed has let's say 50columns, but I do not understand why hasura creates a Method with more than max_function_args and try to retrieve data after that
That may be due to json modules used in hasura or something like that
And to fix that you just hae to recompile PF to give a bigger max_function_args and that is not possible in most cases since no one wants to recompile PG) this value cantbe added to PG cofig it is a readonly value and postgres doesnt allow that

SO it seems that tables with more that 50columns are not supported in hasura at the mooment

This is the errror
```
statusCode: 54023
cannot pass more than 100 arguments to a function
```

ammirator-administrator:
Any help?? Citus is not usable with hasura because of this error
Can we do something about this?
It seems that just compiling Postgres may help, but that is not a option and no one is compiling his own version of postgres

ammirator-administrator:
Is there a way to configure hasura to use not json_(..,..,..,,) witha lot of arguments but use maybe json_array not json_object

ammirator-administrator:
More context about the issue I also fount heer https://postgrespro.com/list/thread-id/2486891
Basically the same issue

ammirator-administrator:
If i use in my databases.yaml
this

kind: postgres
instead of

kind: citus
Then everything works, but I just simply do not know if that is ok and what I lose by doing so

happyhunter7:
+1 This is supper needed and is a limitation from Hasura

Big companies do not use just postgres since it doesn't make sense in big apps but with some extension like Citus
And if the Hasura Citus adapter implementation has these limitations it is sad for big companies which may try to us eHasura

SamirTalwar:
Can you explain how you trigger this issue? Is it a problem when loading metadata, making a query, using the Console, or something else?

There aren't that many differences between the vanilla PostgreSQL and Citus backends, so I am surprised that switching the "kind" would help, but that does narrow it down a lot.

I would also recommend checking out [Hasura DDN](https://hasura.io/ddn), which fixes many of the issues in Hasura GraphQL Engine v2. Most notably in this case, the custom Citus support is strictly limited to introspection, and during queries, both PostgreSQL and Citus are treated identically.

[...]

PR #10229, opened 2024-05-08
Citus Fix json_build_object with tables with more than 50 columns

ammirator-administrator:
Postgres has by default a limitation for functions names max_function_args and is set to 100, which means you can't pass more that 100 arguments to a pg function otherwise it throws error

BTW I'm talking about postgres with Citus, seems that in postgres implementation you guys use row_json which is fine and that;s why there it doesn't happen

The implementation in hasura is that for every table column it adds 2 arguments in json_build_object object one lets say "a" and another "root.a" 
WHICH means that with hasura this limitation becomes 50 column maximum per table, which is not really friendly

To change max_function_args in Postgres you just have to recompile Postgres, which is something I would not sugest to someone to do) even Postgres team doesn't sugest to compile your own version

SInce postgres way is not a solution It needs to be fixed in Hasura definitelly because Postgres limitation is good 100 but because of how hasura adds args it just becomes 50 which is already a problem for most apps, I have a small table where I keep app config, and is normal that a config would have more that 50 columns but hasura doesn't allow to query this table at all because of this limitation

PS. I just opened this PR because I see there are 2k issues and the chance that the issue will be read soon is zero
But I consider this a serious limitation
SInce I do not know haskel I cant contribute so I would love some help from Hasura team to implement this 
I think this fix for them would take not much time since they know haskel and the fix most probably is already described here 
https://postgrespro.com/list/thread-id/2486891

### DIFF
--- a/server/src-lib/Hasura/Backends/Postgres/Execute/Subscription.hs
+++ b/server/src-lib/Hasura/Backends/Postgres/Execute/Subscription.hs
@@ -358,6 +358,16 @@ mkStreamingMultiplexedQuery userInfo (fieldAlias, resolvedAST) = do
         [resultIdAlias, resultVarsAlias]
 
     -- json_build_object('field1', field1.root, 'field2', field2.root, ...)
+    -- json_build_object should not be used without array https://postgrespro.com/list/thread-id/2486891
+    -- Please someone with proper Haskel knowledge change this) 
+    -- Postgres has a limitation of maximum 100 function arguments
+    -- If you have a table with 50 columns this code produces 100 arguments in the  json_build_object and postgres throws error
+     -- And you can't increase this limit in Postgres, the only way is to compile your own version of postgres which is stupid to do otherwise such configuraton can't be done
+     -- Same change needs to be done not just in subscription but in othe rplaces as well where json_build is used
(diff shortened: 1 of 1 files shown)


ammirator-administrator:
Also I do not understand why in postgres driverr/addaper you guys use row_json which is good but with citus addapter you use json_build_object
Citus is fully compatible with postgres since is just a postgre sextension, would it be correct to use same methods? this way we awoid these limitations
I guess in all adapters where you guys use json_build_object we should have this limitation of maximum 100 args because thats the default for postgres and cant be changed

ammirator-administrator:
If i use in my databases.yaml 
this 
```
kind: postgres
```
instead of 
```
kind: citus
```

Then everything works, but I just simply do not know if that is ok and what I lose by doing so

happyhunter7:
+1 This is supper needed and is a limitation from Hasura

Big companies do not use just postgres since it doesn't make sense in big apps but with some extension like Citus
And if the Hasura Citus adapter implementation has these limitations it is sad for big companies which may try to us eHasura

SamirTalwar:
This is not a change request but a feature request, already tracked in #10227. While it is important, I suggest we have the discussion and talk about fixing it in one place, not two, so as not to split the conversation.

PR #10237, opened 2024-05-14
[RFC][v3] Authorization Rules in V3

abhinav-hasura:
### Description

[Rendered](https://github.com/hasura/graphql-engine/blob/abhinav/v3-authorization-rules-rfc/rfcs/v3/authorization-rules.md)

### DIFF
--- a/rfcs/v3/authorization-rules.md
+++ b/rfcs/v3/authorization-rules.md
@@ -31,14 +31,14 @@ The condition must start at a boolean operator. These are the available boolean 
 - `or`: array of boolean expressions which are ORed together.
 - `not`: boolean exrepssion which is negated.
 - `equal`: Compares two value expressions are the same.
+- `contains`: Whether a given value is contained within a given array of values
 - `greaterThan`: Compares left > right.
 - `lessThan`: Compares left < right.
 - `greaterThanOrEqual`: Compares left >= right
 - `lessThanOrEqual`: Compares left <= right
 
-In the future, these may be expanded to:
+In the future, these may be expanded to more boolean operators. For example:
 - `regexMatch`: Whether a given value matches the given regular expression
-- `contains`: Whether a given value is contained within a given array of values
 
 #### Value Expressions
 
@@ -250,3 +250,43 @@ definition:
               left: sessionVariable: x-hasura-role
               right: literal: admin
 ```
+
+### Reusing Relationship Permissions
+
+In the existing authorization system, the entire permissions predicate needs to be specified for a model's permission. For related models, however, this can result in a duplicate and very nested permission predicates.
+
+For example, if I have models for `Channels`, `Threads`, `Replies`. Then my predicates would be as follows:
+- `Channels`: `members._contains(x-hasura-user-id)`
+- `Threads`: `channel.members._contains(x-hasura-user-id`
+- `Replies`: `thread.channel.members._contains(x-hasura-user-id)`.
+
+Ideally, you would be able to express the fact that an object should be selectable if the related object is also selectable.
+
+Hence, we propose the following addition to the permissions predicate:
+```
+allowObjects:
+  relationship:
+    name: 'my_relationship'
+    relatedObjectAllowed: true
+```
+
+
+## FAQs
+
+1. **Will existing metadata continue to be work?**  
+    
+    Yes, existing metadata (using the `v1` version of permissions objects) will continue to work and trivially maps to the new system where the condition is an equality check on the `x-hasura-role` session variable.
+
+2. **Does this subsume the inherited roles feature in v2?**
+    
+    Yes, this is the intended replacement for inherited roles in v2. So, instead of defining multiple base roles and an inherited you would simply define multiple rules which can activated by arbitrary conditions.
+
+    Note, however, that cell-level authorization will not be supported by this system, since you define type (column) permissions and model (row) permissions independently.
+
+3. **Are any changes required to the NDC protocol?**
+   
+   No.
+
+4. **Are any changes required to the user's authentication system?**
+
+   No changes are required to existing authentication systems. However, to fully utilize the power of authorization rules you would want to inject more granular claims (instead of simply roles) during authentication.
--- a/rfcs/v3/authorization-rules.md
+++ b/rfcs/v3/authorization-rules.md
@@ -7,6 +7,11 @@ Currently, Hasura uses role based access control, where you define the entire se
 - It's not possible to reuse permissions (allowed fields, model predicate) across roles.
 - For complicated permissions, it's hard to verify the correctness of a model's permissions predicate at a glance.
 
+Some Github issues addressed in this RFC:
+- https://github.com/hasura/graphql-engine/issues/3685
+- https://github.com/hasura/graphql-engine/issues/1919
+- https://github.com/hasura/graphql-engine/issues/3544
+
 ## Proposal
 
 This RFC proposes a new rule-based mechanism for defining permissions for types, models, and commands in the metadata. The following requirements have been considered as a part of this proposal:
@@ -26,18 +31,17 @@ The condition that may be attached to a rule is a boolean expression built using
 
 #### Boolean Operators
 
-The condition must start at a boolean operator. These are the available boolean operators:
+The condition must start at a boolean operator. These are some potential boolean operators:
 - `and`: array of boolean expressions which are ANDed together.
 - `or`: array of boolean expressions which are ORed together.
 - `not`: boolean exrepssion which is negated.
 - `equal`: Compares two value expressions are the same.
+- `isNull`: whether a given value is null
 - `contains`: Whether a given value is contained within a given array of values
 - `greaterThan`: Compares left > right.
 - `lessThan`: Compares left < right.
 - `greaterThanOrEqual`: Compares left >= right
 - `lessThanOrEqual`: Compares left <= right
-
-In the future, these may be expanded to more boolean operators. For example:
 - `regexMatch`: Whether a given value matches the given regular expression
 
 #### Value Expressions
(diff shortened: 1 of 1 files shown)


rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+- Deny access to certain fields
+
+If a field is both allowed and denied, it will be denied.
+
+#### Example
+
+```yaml
+kind: TypePermissions
+version: v2
+definition:
```
danieljharvey, line 95: What do rules like these mean for the generated GraphQL schema?

Will `internal_order_notes` always be visible, but we'll make it nullable?

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+
+### AuthConfig
+
+The notion of roles will be removed from AuthConfig. In particular, the `allow_role_emulation_by` key will be replaced by a `allow_session_emulation_if` key which specifies a condition which needs to evaluate to true for the request to allow emulating session variables.
+
+No `x-hasura-` session variables will now be required in the webhook response or in JWT claims. Users can choose arbitrary names for session variables.
+
+### Implementation
+
+The above proposal has major implications for engine, the `lang-graphql` crate in particular. Currently, `lang-graphql` allows constructing a single graphql schema with information about all roles embedded into the schema and allows query validation / IR generation to happen in the context of a particular role only. However, with the above proposal, there is no notion of roles anymore. So, `lang-graphql` will need to move to an architecture where when trying to access a graphql field, instead of looking up an allowed set of roles, it can evaluate an arbitrary expression that evaluates to a boolean. The details of this architecture are out of scope for this RFC.
```
danieljharvey, line 292: This presumably means losing the ability to generate a GraphQL schema (and thus generate types etc) for only the queries / types that a given role is allowed to access. This feels like quite a price to pay, are we satisfied that users don't need this?

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+- support_agent_with_pii_access_and_gov_access
+
+And the permissions would be very hard to manage. Eg:
+
+```yaml
+kind: TypePermission
+version: v1
+definition:
+  typeName: User
+  roles:
```
danieljharvey, line 366: I know this isn't the root concern of this RFC, but surely there are other ways to make such a permission less arduous to define, for instance?

```yaml
  permissions:
    - roles:
      - developer_base
      - developer_with_gov_access
      - support_agent_base
      - support_agent_with_gov_access
      output:
        allowedFields: [id]
    - roles:
      - developer_with_pii_access
      - developer_with_pii_access_and_gov_access
      - support_agent_with_pii_access
      - support_agent_with_pii_access_and_gov_access
      output:
        allowedFields: [id, name, email]
```

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+                  fieldComparison:
+                    field: agent_id
+                    operator: _eq
+                    value: sessionVariable: x-hasura-agent-id
+            - not:
+                fieldComparison:
+                  fieldName: is_gov
+                  operator: _eq
+                  value: literal: true
+      - role: support_agent_with_gov_access # Same for support_agent_with_pii_access_and_gov_access
```
danieljharvey, line 428: If we're concerned about the usability of metadata, then it doesn't seem wild that eventually we're going to want to abstract these predicates too and refer to them by name.

```yaml
.....
- role: support_agent_with_gov_access # Same for support_agent_with_pii_access_and_gov_access
  select:
    filter:
      relationship:
        name: tickets
        predicate: agent_predicate
---
kind: Predicate
version: v1
definition:
  name: agent_predicate
  fieldComparison:
    field: agent_id
    operator: _eq
    value: sessionVariable: x-hasura-agent-id
```

rfcs/v3/authorization-rules.md
danieljharvey: (or would accessing a field that one is not allowed trigger an error instead of a `null` field?)

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+version: v2
+definition:
+  typeName: User
+  authorizationRules:
+    # Rule 1: Allow id field for everyone
+    - allowFields: [id]
+    # Rule 2: Allow name and email if pii access is allowed
+    - allowFields: [name, email]
+      condition:
+        equals:
```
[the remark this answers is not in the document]
danieljharvey: I sketched out what this might look like here: https://gist.github.com/danieljharvey/2718cf1a8fc340a5b21063bc5e843b36

rfcs/v3/authorization-rules.md
abhinav-hasura: `internal_order_notes` will not be visible in the GraphQL schema if the `x-hasura-role` session variable is not set to `admin` in the introspection query.

(Note that there is a separate conversation going on around eliminating dynamic role based schemas altogether but for the purpose of this RFC, let's assume we still want the GraphQL schema to contain only what's accessible)

rfcs/v3/authorization-rules.md
abhinav-hasura: No, we would still generate the GraphQL schema only according to what's accessible to a given set of session variables.

When an introspection query comes in, we would evaluate which fields/types are accessible based on the session variables associated with the request and generate a schema containing only those types/fields.

rfcs/v3/authorization-rules.md
abhinav-hasura: Right, for type permissions itself, it's not as big a deal, but you would still need to model 8 different roles in your authn/authz system. And now imagine, another attribute like 'hippa_access' gets added, now I would have 16 roles to manage.

So, fundamentally the issue is that it's not always feasible to enumerate all possible authz states that can exist in my system since that can grow exponentially with each independent attribute in my system.

rfcs/v3/authorization-rules.md
abhinav-hasura: In a world where my rules are composable, the ability to reuse predicates isn't that important.
The fundamental problem that is being addressed is not the ability to reuse parts of my metadata, but not having to enumerate all possible authorization states in my system.

rfcs/v3/authorization-rules.md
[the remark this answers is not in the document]
abhinav-hasura: We can still have static guarantees with the current design, i.e., if a request has session variables `x-hasura-has-pii-access: true` and `x-hasura-role: developer`, then the schema will be like _this_.

Also, there is a subtle difference between role predicates and independent rules w.r.t how type permissions and model permissions interact.

Eg: With independent rules:
```
- allow row 1 of model X, if condition1
- allow all rows of model X, if condition 2
- allow all fields of model X, if condition1
- allow field A of model X, if condition 2
```
If both condition1 and condition2 are true, i can see that all rows and fields will be accessible.

But, if I used role predicates:
```
Roles:
- role1 if condition1
- role2 if condition2

Permissions:
- role1 can access all fields of row 1 of model X
- role2 can access field A of all rows of model X
```
Now if both condition1 and condition2 are true, then I would expect to have access to field A of all rows, and all fields of row 1.
This sort of cell-level authorization we explicitly don't want to support.

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+    
+    Yes, existing metadata (using the `v1` version of permissions objects) will continue to work and trivially maps to the new system where the condition is an equality check on the `x-hasura-role` session variable.
+
+2. **Does this subsume the inherited roles feature in v2?**
+    
+    Yes, this is the intended replacement for inherited roles in v2. So, instead of defining multiple base roles and an inherited you would simply define multiple rules which can activated by arbitrary conditions.
+
+    Note, however, that cell-level authorization will not be supported by this system, since you define type (column) permissions and model (row) permissions independently.
+
+3. **Are any changes required to the NDC protocol?**
```
paf31, line 321: But if row and column permissions are disjoint, would you not want to push down the cell-level predicates? Or are you proposing that we redact data on the engine side.

That would be simpler but result in more network overhead.

I think it'd be worth spelling this out here.

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+  - Arguments input validation
+  - Calling external systems
+  - Accessing sub-fields in value expressions
+  - Advanced boolean operators
+
+## FAQs
+
+1. **Will existing metadata continue to be work?**  
+    
+    Yes, existing metadata (using the `v1` version of permissions objects) will continue to work and trivially maps to the new system where the condition is an equality check on the `x-hasura-role` session variable.
```
daniel-chambers, line 313: I think we should consider letting the user pick which way they'd prefer to express their permissions, either this advanced system, or the existing role based system (which we just desugar into the the advanced system behind the scenes). This advanced system is very powerful, but also less obvious how to use it to recreate the simple use case and I contend that users who just need basic role-based permissions would find the old mechanism easier to author and understand. 

To be clear, this doesn't mean just retaining support for the v1 ModelPermissions and TypePermissions, it means including both options on v2. Just retaining the old v1 types is not as good as it implicitly discourages their use because they are an "old version".

rfcs/v3/authorization-rules.md
```diff
@@ -0,0 +1,568 @@
...
+        fieldComparison:
+          fieldName: is_hidden
+          operator: _eq
+          value: 
+            literal: true
+      condition:
+        not:
+          equals:
+            left: sessionVariable: x-hasura-role
+            right: literal: admin
```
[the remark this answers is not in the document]
paf31: I think the idea is that (for now at least) expressions are evaluated in the engine only, since they only depend on the values of session variables. I agree that pushing down subexpressions is not a tractable plan.

Issue #10247, opened 2024-05-21
Federated Mutations (orchestration framework)


manasag:

beepsoft:
This issue seems to combine issues like https://github.com/hasura/graphql-engine/issues/5773, https://github.com/hasura/graphql-engine/issues/1583, https://github.com/hasura/graphql-engine/issues/2657.

@manasag , @sordina do you have an idea when this will be implemented?

manasag:
The current thought about this is to expose some built in functions in the Typescript Connector that allows to call back into the Hasura Supergraph. These callbacks will make it possible to compose any number of queries or mutations in easy typescript function definitions. We will be sharing the RFC and maybe some early work soon about this.
This is not planned to be in production soon. Best case end of 2024.

manasag:
The RFC can be found here https://github.com/hasura/graphql-engine/blob/daniel/rfc-supergraph-client/rfcs/v3/supergraph-client.md

PR #10311, opened 2024-06-24
Supergraph Client RFC

daniel-chambers:
TODO

### DIFF
--- a/rfcs/v3/supergraph-client.md
+++ b/rfcs/v3/supergraph-client.md
@@ -1,0 +1,202 @@
+# Supergraph Client RFC
+
+Customers have a few use cases that could be served by enabling certain types of connectors (code-based connectors, such as the NodeJS Lambda Connector) to issue requests back into the Hasura Supergraph at runtime:
+
+* I want to issue multiple GraphQL queries/mutations and interleave business logic amongst them
+* I want to apply validation logic before an existing mutation in the supergraph
+* I want to run complex authorization logic before applying a mutation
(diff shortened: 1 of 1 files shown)


rfcs/v3/supergraph-client.md
```diff
@@ -0,0 +1,202 @@
...
+      type:
+        named: SupergraphClient
+  resultType:
+    named: Bool
+```
+
+The engine would send a JSON object with the URL to the current Supergraph build's GraphQL API (when running locally, this would just be a localhost engine URL):
+
+```json
+{ "supergraphUrl": "https://settling-impala-7722-50bb46e1d5.ddn.hasura.app/graphql" }
```
shahidhk, line 59: We need to work out how this would work in the case where there is a custom domain. Engine is only aware of the canonical build urls as custom domains are handled by the gateway. Also, we plan to introduce path based routing for build urls instead of hostname based, and gateway is going to be handling that as well.

A solution could be for the gateway to pass the actual hostname to engine as a separate header and then use that header as SupergraphURL.

daniel-chambers: That could work. I think we may need an environment variable setting for it on v3-engine as well, for people self-hosting v3-engine without a gateway injecting headers, or for use when running locally. 

Perhaps the env var setting would offer three options:
* Get it from a header from the gateway
* Get it as a value from the env var itself (for self-hosting)
* No env var set (default to localhost)

manasag:
#10247

PR #10475, opened 2024-07-24
[RFC] REST API output on Hasura v3

ecthiender:
## Description

This is our new proposal of auto-generating REST API output in Hasura v3.

[Rendered]

### DIFF
--- a/rfcs/v3/rest_json_api.md
+++ b/rfcs/v3/rest_json_api.md
@@ -1,0 +1,258 @@
+# REST API Output
+
+Hasura v3 will auto-generate a [JSON:API](https://jsonapi.org/) spec-compliant REST API output.
+Hasura v3 will only generate these for DDN models. Commands are out of scope of this RFC.
+
+## Fetching Data
+
+Data, including resources and relationships, can be fetched by sending a GET request to an endpoint.
+
+Responses can be further refined with the optional features described below.
+
+### Fetching Resources
+
+A server MUST support fetching resource data for every URL provided as:
+
+* a self link as part of the top-level links object
+* a self link as part of a resource-level links object
+* a related link as part of a relationship-level links object
+
+### Selecting
+
+JSON:API supports selecting specific fields of a resource and specific fields of its related
+resources.
+
+An endpoint **MAY** also support an `include` query parameter to allow the client to customize which
+related resources should be returned.
+
+A client **MAY** request that an endpoint return only specific
+[fields](https://jsonapi.org/format/#document-resource-object-fields) in the response on a per-type
+basis by including a `fields[TYPE]` query parameter. They are called [sparse
+fieldsets](https://jsonapi.org/format/#fetching-sparse-fieldsets).
+
+```
+GET /articles/1?include=comments
+
+GET /articles/1?include=comments.author
+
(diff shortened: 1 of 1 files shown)


rfcs/v3/rest_json_api.md
```diff
@@ -0,0 +1,258 @@
...
+
+GET /movies?filter={"$or":[{"name":{"$eq":"Braveheart"}},{"name":{"$eq":"Gladiator"}}]} // url encoded
+```
+
+Both the above syntax are equivalent. If you notice, the last syntax is equivalent to what Hasura
+generates for its GraphQL API.
+
+Hasura will also generate an OpenAPI schema for complex filtering on nested types. As a result, any
+auto-generated client would get full auto-complete and type-checking on any nested filtering
+clauses. Check out <https://github.com/hasura/json-api-experiments> to see this in action.
```
danieljharvey, line 73: I must have missed that we're able to get this working in a well-typed fashion. Nice!

rfcs/v3/rest_json_api.md
```diff
@@ -0,0 +1,258 @@
...
+      objects
+  - `links`: Top-level links object contains pagination links (`self`, `next`, `last`).
+  - `jsonapi:` Contains the version of the JSON
+  - being used and additional meta-information about the API, including a `describedby` URL.
+
+Read more on the [resource object](https://jsonapi.org/format/\#document-resource-objects).
+
+# OpenDD config
+
+For the first iteration, to keep things simple, we are not introducing any OpenDD configuration. Any
```
danieljharvey, line 257: I like this because we're keeping scope slim, but I am concerned about breakages - ie, if by default we expose all models as a JSON:API endpoint, and then we later make each model opt-in, does that constitute a breaking change?

Do we solve this by only offering this feature on a strictly "this is experimental" basis?

rfcs/v3/rest_json_api.md
```diff
@@ -0,0 +1,258 @@
...
+## Customizations
+
+If required, response transformation could be performed with a JSON:API output on Hasura.
+
+### Response Transforms Example
+
+#### Flatten Response JSON
+
+Say we want to flatten a field of the JSON response. We want to change `article.author.name` to
+`article.autor_name`.
```
danieljharvey, line 141:
```suggestion
`article.author_name`.
```

danieljharvey (approved):

rfcs/v3/rest_json_api.md
ecthiender: That's a good point. It will be a breaking change, and I guess we would have to incorporate this via Compatibility Date. But let me see if we can do anything better here.

danieljharvey: Oh of course. If we just say "everything before this date gets all models, everything after gets specific ones", we're fine.

ecthiender: Yeah, I'm thinking we can introduce a date after which it is still all models but user gets a warning. And after a second date, we switch to specific ones. Does this make sense?

danieljharvey: Yeah, sounds ideal!

Issue #8095, opened 2022-01-28
"allow_inconsistent_metadata" env var missing from cli-migrations image


jflambert:
### Version Information

Server Version: 2.1.1

### Environment

Linux

### What is the expected behaviour?

2.0.0 added the `allow_inconsistent_metadata` argument to the `replace_metadata` API call. Unless documentation is wrong, it doesn't seem to be available as an environment variable for the cli-migrations image.

It would also be nice to have a checkbox in the frontend when importing metadata.json there, but I don't personally need that.

### What is the current behaviour?

`allow_inconsistent_metadata` only seems available through API calls.

scriptonist:
@jflambert can you elaborate a bit on the behaviour that you expect here?

In the CLI migrations image at a high level, the following are the steps which are executed: 

```
$ hasura metadata apply
$ hasura migrate apply --all-databases
$ hasura metadata reload
```
if after the `metadata reload` the metadata is inconsistent will you expect the applied migrations to be rolled back and the metadata to be restored to the "old" state? ie whatever it was before this process started.

jflambert:
No

I expect the same behaviour as the API offers

fully applied migrations

broken metadata

scriptonist:
@jflambert Thanks. Can you clarify what you mean by "broken metadata"? 

If we call the new metadata that we are applying "new" and the current metadata on the server "old". Then if when trying to apply  "new" metadata results in an inconsistency you expect the metadata on the server to be unchanged ie the server will have "old" metadata. is that so?

chardo:
Hello, I know this is a very old thread, but I just wanted to chime in to say that exposing the CLI's `--disallow-inconsistent-metadata` flag via the `cli-migrations` image would be really helpful. Without that flag, rolling updates to `graphql-engine` deployments will apply metadata to _all_ running instances, even the old ones (because they listen for changes to metadata), even if that metadata is inconsistent. 

This setup breaks the standard mental model of rolling updates (in which a functioning deployment's replicas won't be torn down until the new deployment has available replicas) because the inconsistent metadata "leaks" into the old replicas. My team ran into this issue earlier this week, and while obviously we can do our best to avoid these situations in the future, a production deployment of Hasura relies on a lot of coordination that any team of human beings is bound to get wrong every once in a while. Ensuring that inconsistent metadata will fail a new deployment's rollout without affecting an existing deployment would go a long way towards making me feel safer about running Hasura in prod.

[...]

PR #10602, opened 2024-11-15
Support disallowing inconsistent metadata in cli-migrations images (close #10599)

### LINKED ISSUE #10599: Support "--disallow-inconsistent-metadata" flag in cli-migrations docker image

My team runs the community edition of the Hasura graphql-engine on Kubernetes in a horizontally replicated configuration. We use rolling updates to deploy, use the `cli-migrations` (v2) docker image as our base image in production (as described [here](https://hasura.io/docs/2.0/migrations-metadata-seeds/auto-apply-migrations/)), and use the `/healthz?strict=true` endpoint as our container's readiness/liveness check. This allows us to deploy Hasura in prod without exposing the console or metadata API, while still allowing for git-driven continuous delivery of metadata changes by version-controlling our metadata files and packaging them into new docker images that get deployed to our Kube cluster on each commit to main. This works great most of the time, allowing us to roll out changes to our GraphQL API just by merging metadata changes to main.

However, this configuration has one major drawback. The `cli-migrations` entrypoint script ([ref](https://github.com/hasura/graphql-engine/blob/5c7e20f83b83cc4126bebc8fc4d9f12efc128d72/packaging/cli-migrations/v2/docker-entrypoint.sh#L98)) runs `hasura-cli metadata apply` without the `--disallow-inconsistent-metadata` flag. This means that if we accidentally merge inconsistent metadata to main (eg, by applying metadata changes before a production DB migration has completed), the new pods for our updated Kubernetes deployment will apply that metadata and persist it to the database regardless of any issues. This metadata update in turn gets picked up by the existing/old pods, which listen for changes to the metadata DB, and "leaks" into them even though they (in theory) were built from an older, functioning revision of our code. 

In effect, this defeats the purpose of using a deployment tool like Kubernetes, because what should have been a failed deploy ends up taking down the existing pods, rather than simply failing to spin up new ones. This breaks from the standard mental model of rolling updates, in which a pre-existing stable deployment should stay available until new pods are ready to receive traffic. This scenario befell us earlier this week, leading to a full API outage instead of what we'd expect for any other API that can't become ready - a failed rollout of the new without affecting the old.

Related to #8095 

I'd love to have some way to configure the `cli-migrations` image to disallow the application of inconsistent metadata, the same way I know is possible using the CLI's `hasura metadata apply --disallow-inconsistent-metadata` command. I'm open to whether that should be possible via a global Hasura configuration (ie, configure this as a setting on the server itself) or just a flag/env-var that can be passed to the `cli-migrations` image similar to the `HASURA_GRAPHQL_METADATA_DIR` variable. I expect the latter would be simpler to implement. 

Doing this would allow us to have a fully safe CI/CD setup in Kubernetes by preventing pods which contain inconsistent metadata from poisoning the shared metadata DB and taking down the entire API.

Just noting - I would be very happy to work on this feature myself, in particular if we'd prefer to scope it to a change to the `cli-migrations` image only. This would be really helpful for my team, and I'd love a chance to help contribute back to Hasura given how much use we've gotten out of the graphql engine over the last year or two!

I see a few different ways of working around this, which I'll explain my POV on inline:
 - Stop using the `cli-migrations` image, and use the CLI to apply metadata in production directly. I'd really prefer not to do this for my team, because it would mean exposing the metadata API in prod, which would then make it accessible to anyone who has the admin key. This would be a pretty major step up in the risk/impact of leaking that key, which I'd rather not take on.
 - Don't enforce "strict=true" on the `/healthz` endpoint for our readiness and liveness checks, allow inconsistent metadata to applied to prod, and monitor for it some other way. I'd also prefer to avoid this - given it is possible to determine prior to deployment whether a batch of metadata files will be inconsistent (and thus lead to some amount of API degradation), I see no good reason we'd want our prod service to declare itself ready for traffic in those scenarios.
 - Improve our own testing/quality gates prior to merging to reduce the odds of inconsistent metadata making it into main in the first place. Of course I'd love to do this as well, but no testing is going to cover 100% of possible mistakes, and the change I'm recommending seems (to me) like a surefire way to prevent broken deploys anyways.

chardo:
### Description

The Hasura CLI's `hasura metadata apply` command supports a `--disallow-inconsistent-metadata` flag, which helps to prevent breaking changes to metadata before they're applied rather than discovering them through `hasura metadata ic list` or, worse, via runtime application errors. However, in production environments, it's common to deploy graphql-engine using the `cli-migrations` Docker image and avoid exposing the metadata API entirely. This means that CI/CD workflows have no guaranteed way of preventing inconsistent metadata from landing in production, which is made even riskier by the fact that metadata changes will be automatically picked up by any already-running instances, even if that metadata is inconsistent.

This change attempts to address that by exposing an optional `HASURA_GRAPHQL_DISALLOW_INCONSISTENT_METADATA` env variable that can be provided to the `cli-migrations` Docker images in order to activate the corresponding `--disallow-inconsistent-metadata` flag on the `hasura metadata apply` step. If this is set, metadata application will fail, the `docker-entrypoint.sh` script will exit early, and the container will fail to start up.

Also, just to say: I know that I opened this PR before I got any traction on the associated issue. If there's good reason for _not_ implementing this change, I will understand and won't mind throwing this work away.

### Changelog

__Component__ : build 

__Type__: feature 

#### Short Changelog

Add support for disallowing inconsistent metadata in `cli-migrations` image

### Related Issues

#10599 
#8095 

### Solution and Design

This follows the existing pattern for configuration env vars in the `docker-entrypoint.sh` script(s), though is strict in requiring the value of the new variable to be "true" (case insensitive) rather than it just being set to any value.

In this first draft, I've not made any effort to fail gracefully in the event of inconsistent metadata. Ideally I think we'd _probably_ capture the exit code, shut down the temporary graphql-engine server, and then exit with the original code. That said, this implementation is consistent with how the script already handles possible non-zero exit codes for the `hasura-cli` commands (eg if the server is unreachable, or the metadata contains invalid YAML).

Zooming out a little bit, it's maybe also worth mentioning that I've deliberately chosen to isolate this feature to the `cli-migrations` image, rather than making it a server-level config variable that would change the graphql-engine's default behavior when receiving new metadata updates. The latter seems a bit far-reaching, and I'd rather leverage an existing API than broaden/complicate its scope in a significant way.

### Steps to test and verify

I updated the existing test scripts so that, after confirming the "good" behavior works as intended, they also attempt to apply some inconsistent metadata and then confirm that the docker image is unable to start up.

I couldn't find any tools/docs for running tests locally, but I was able to get both test scripts running and passing locally by:
 - installing a copy of the hasura CLI into a local `hasura-cli`
 - building the `cli-migrations` images manually
 - setting the necessary test env vars
 - running the test scripts on local

I did set the new disallow-inconsistent-metadata flag to "true" on both of the test docker-compose.yaml files, so that I could just augment the existing test files. If you'd prefer to have this test run in a separate, isolated file with a different env configuration, I'm willing to do that too! Just thought this was a simpler first revision.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

n/a

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  2. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  3. Logging


packaging/cli-migrations/v3/docker-entrypoint.sh
```diff
@@ -77,7 +87,7 @@ if [ -d "$HASURA_GRAPHQL_METADATA_DIR" ]; then
     echo "version: 3" > config.yaml
     echo "endpoint: http://localhost:$HASURA_GRAPHQL_MIGRATIONS_SERVER_PORT" >> config.yaml
     echo "metadata_directory: metadata" >> config.yaml
-    hasura-cli metadata apply 
+    hasura-cli metadata apply $HASURA_GRAPHQL_DISALLOW_INCONSISTENT_METADATA
```
chardo, line 90: question for the codeowners - is there a good reason that the v3 image applies metadata updates before applying db migrations? v2 does them in the other order. I suppose either one might result in temporary metadata inconsistencies if DB updates and metadata updates are bundled into the same release. if you're dropping a column or table, you probably want metadata applied first; if you're adding a column or table, you probably want migrations applied first -- so we probably need to accept some unavoidable ICs either way. just checking to see if there's a good reason that we're picking one side for v2 and another for v3.

chardo: Update on this - I went ahead and modified this script so that it has the same ordering (and thus the same behavior) as v2. The need for this became more clear after my updates to the test scripts, which revealed that the v2 test could pass while the v3 test would fail under the same circumstances, because the current v3 setup relies on DB migrations being run before metadata can be applied, in order to remain strictly consistent.

scriptnull (commented):
Hi @chardo, Thanks for the PR! Sorry that we couldn't take a look at it earlier.

It looks good overall :+1: The one thing I am a bit doubtful about is the change of execution order of the metadata and migrations command in the v3 cli migrations image.

I feel like there was a reason to preserve the execution order when we originally introduced it but I am unable to find the reason behind it. So I am going to ask around internally (cc: @scriptonist) to see if I can find the original reason for this. Please give me a little more time to arrive at a conclusion on this matter.

scriptnull (commented):
The reason behind running the metadata command before the migrations command in the v3 cli migrations image seems to be:

- CLI uses the [run_sql API](https://hasura.io/docs/2.0/api-reference/schema-api/run-sql/) to apply migrations of a database.
- The information about "connected" databases is in hasura metadata.
- So, if we are trying to apply migrations on a database A, it should already be connected to hasura. Otherwise hasura will throw an error saying it does not know what A database is.
  - Because of this reason, we apply metadata first.
  - ie Let hasura know about A first, though there might be some metadata inconsistencies if we are tracking tables from A  etc
  - Now, migrations can be applied without any errors.

So let us avoid changing the order of the commands in the v3 cli migrations image.

chardo:
@scriptnull great, thanks so much for the review and the context! I can change the v3 docker entrypoint script on this branch so that it preserves the existing order.

Just want to point out, though, that this will probably result in different behavior between v2 and v3 when disallowing inconsistent metadata. In v2, running migrations first means you can create a new table and define metadata for that table in the same change (since the table will exist by the time the `metadata apply` command runs). But v3 applying metadata updates first will prevent you from applying both of those changes in one deploy - the metadata will be inconsistent since the table doesn't exist yet.

I'm totally fine with this (it preserves existing behavior, and the flag is opt-in rather than being on by default) but just wanted to mention it because it might be confusing for others. Do you think it's worth clarifying this difference in the docs in some way?

chardo:
Just to follow up on my previous comment: I just ran the test script on the v3 image inconsistent metadata disallowed, and confirmed that the graphql-engine container can't start up from scratch because of the ordering of metadata application/migrations (the engine can't apply metadata that references a table which hasn't been created yet).

So as is, the baseline v3 tests are failing with the step ordering reverted to the original state. I think I can get the tests working properly by starting up the test gql-engine container _without_ disallowing ICs, applying metadata and migrations, then restarting it with the `HASURA_GRAPHQL_DISALLOW...` flag and confirming that we get an error. 

Again, happy to do all this, but wanted to follow up with a confirmation that I'm seeing the predicted consequences of the current ordering and double check whether you have any concerns with my planned approach. I'll hold off on getting too deep into fiddling with tests for now, in case you have any other ideas.

hgiasac:
closed by https://github.com/hasura/graphql-engine/commit/f824d4d19aa1e2acdf403ecdcc66e4ae1160eab2

PR #10663, opened 2025-01-27
Update action-handlers.mdx to use dashes intead of underscores in the headers

matt-sdc:
### Description
Using underscores in headers can cause issues with nginx.  Given the number of  comments on this stackoverflow:

https://stackoverflow.com/questions/22856136/why-do-http-servers-forbid-underscores-in-http-header-names

And nginx being the a very common production server: https://nginx.org/en/docs/http/ngx_http_core_module.html#underscores_in_headers

I recommend you update the docs to use dashes instead of underscores.  While underscores are technically valid according to RFC, you are causing undo pain on devs finding this querk.

### Changelog

__Component__ : documentation

__Type__: enhancement

#### Short Changelog

Changed `ACTION_SECRET` to `ACTION-SECRET` on the action handlers documentation

### Solution and Design

Nginx and other server configs drop headers with underscores.  This isn't widely known, and as mentioned in the description, many developers have wasted many hours finding this out the hard way.

By using action-secret instead of action_secret, we implicitly guide developers to not deal with this issue.

### Steps to test and verify
This is a documentation enhancement.  To test nginx settings... I don't know, spin up a server?

You could go further and warn people about the underscore issue, but I don't think hasura docs is the place for that.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

### DIFF
--- a/docs/docs/actions/action-handlers.mdx
+++ b/docs/docs/actions/action-handlers.mdx
@@ -202,7 +202,7 @@ Update the definition of your action by adding the action secret as a header:
        handler: http://localhost:3000
        forward_client_headers: true
        headers:
-         - name: ACTION_SECRET
+         - name: ACTION-SECRET
            value_from_env: ACTION_SECRET_ENV
 ```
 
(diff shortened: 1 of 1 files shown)


robertjdominguez:
Thanks for this, @matt-sdc 🙏 

> While underscores are technically valid according to RFC...

As this is valid and nginx has methods of handling it, I'll instead add a note to this page to aid users who may stumble on this with their config.

PR #10677, opened 2025-02-07
Handle illegal JSON casting in `hdb_catalog.insert_event_log` (close #8498)

### LINKED ISSUE #8498: invalid input syntax for type json
hasura: v2.5.0
PostgreSQL 13.6 (Ubuntu 13.6-1.pgdg20.04+1+b1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0, 64-bit

still getting an error "invalid input syntax for type json" on update
if I remove event trigger, than there is no error

trivedi:
### Changelog

__Component__ : server

__Type__: bugfix

#### Long Changelog

If a Hasura Event Trigger is attached to a table, changes to that table will call the `hdb_catalog.insert_event_log` [function](https://github.com/hasura/graphql-engine/blob/88d446a6eed464ea4824efe1eb0459e3c4a8dde4/server/src-rsr/init_pg_source.sql#L78-L117) to insert the row’s payload into the `hdb_catalog.event_logs` table.

This function enriches the payload with information about the user via `current_setting('hasura.user', 't')`. It then builds a JSON object with this value—along with the changed row, operation, tracecontext, etc)—and attempts to insert this for the `hdb_catalog.event_logs.payload` JSON column. This creates an implicit casting to JSON.

This assumes that `current_setting('hasura.user', 't')` will always return JSON or NULL. **This is not a safe assumption**. 
Consider the following:

1. Client app opens a transaction, runs  `set_config(`hasura.user`,  'something', 't')`, commits
2. Client app updates a table (with an event trigger) directly via SQL
3. `hdb_catalog.insert_event_log()` is called
4. That function does something to the effect of inserting `current_setting('hasura.user', 't')` to a json column

This makes Postgres throw an error. The call to `current_setting('hasura.user', 't')` will return an empty string instead of NULL because of the modification made in the transaction[1].

There are legitimate reasons why someone may set `hasura.user` via `set_config` -- the biggest one is so that it can be grabbed for the audit table [as described in Hasura's documentation](https://hasura.io/docs/2.0/schema/postgres/postgres-guides/auditing-tables/). Queries that go though Hasura will have this information attached. But queries made directly to the database will not. In the latter case, we may still want the information available to the audit log function. And therefore need set the `hasura.user` manually.

[1] The effect of `set_config` is [accepted behavior](https://www.postgresql.org/message-id/flat/56842412.5000005@joeconway.com) in Postgres and the recommendation is that it should be handled by callers.

### Related Issues

#8498 

### Solution and Design

- Updates the function to fallback to NULL if it encounters an empty string for the configuration parameters.
- This is also consistent with the function’s behavior for old versions of Postgres. At the moment this actually means there's inconsistency in behavior.

### Steps to test and verify

1. Create an event trigger on some table
5. Manually update `hasura.user` in a transaction
  - BEGIN; set_config('hasura.user', {}); COMMIT;
6. Update the table that has the event trigger in some way via directly querying the database
7. See that an error is not thrown

   

#### Catalog upgrade

- [ ] No
- [x] Yes

#### Metadata

- [x] No
- [ ] Yes
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required
    - [ ] Yes
    - [ ] Not required

#### GraphQL
   
   

#### Breaking changes

- [x] No Breaking changes

  1. Metadata API

     Existing `query` types:
     - [ ] Change in error code
     

  8. GraphQL API

     Schema Generation:
     
     
     Schema Resolve:-
     
  9. Logging


rakeshkky:
Hi, thank you for your efforts! Migrations for Postgres sources are located in `src-rsr/pg_source_migrations`. We've made the necessary changes and pushed a commit ([reference](https://github.com/hasura/graphql-engine/commit/13dd2243bde1c0b55187777bc520eeb98a66f4de)) and will be included in next releases.
Closing this PR.

PR #10832, opened 2026-04-16
Fix 'occured' -> 'occurred' typos in v3 graphql validation error messages

SAY-5:
Three `#[error(...)]` thiserror attributes on `InternalError` variants in `v3/crates/graphql/lang-graphql/src/validation/error.rs` read `an internal error occured`:

- line 41: `an internal error occured during validation: type lookup failed for {type_name}`
- line 44: `an internal error occured during validation: field {field_name} lookup failed ...`
- line 108: `an internal error occured: expected {type_name} to be an input type but is of {actual_type}`

All three messages are surfaced to API clients as the `Display` impl when validation fails. String-literal-only change.


hgiasac (approved):
Hi @SAY-5. 

Thanks you for contributing.

PR #10835, opened 2026-04-29
fix(mssql-toquery): adjust ORDER BY rendering for NULL handling

b3hz4d:
### Description
This PR refactors the MSSQL `ORDER BY` query generation logic in `ToQuery.hs` to simplify NULL handling and improve SQL correctness and maintainability.

From a user perspective, this change does not affect GraphQL behavior but improves how SQL is generated internally for ordering queries involving `NULL` values.

### Changelog

__Component__ : server

__Type__: enhancement

#### Short Changelog

Refactored MSSQL ORDER BY NULL handling logic to use CASE expressions and simplify printer composition.

#### Long Changelog

- Replaced previous `IIF`-based NULL ordering logic with `CASE WHEN` expressions for better SQL Server compatibility and clarity.
- Updated `fromNullsOrder` to return `Maybe Printer` instead of always emitting SQL fragments.
- Simplified `fromOrderBy` by removing unnecessary type-based null handling (`wrapNullHandling` logic).
- Introduced `isDefinitelyNotNull` helper to avoid generating redundant NULL ordering expressions for constant expressions.
- Improved readability and composability of ORDER BY SQL generation logic.

These changes make the SQL generation more predictable and easier to extend.

### Related Issues

#<issue-no>

### Solution and Design
The previous implementation attempted to handle NULL ordering using `IIF` and additional type-based wrapping logic, which introduced complexity and inconsistent behavior across expression types.

This PR simplifies the design by:
- Delegating NULL ordering logic into a single `fromNullsOrder` function that returns `Maybe Printer`
- Using `CASE WHEN` expressions, which are more standard and flexible in SQL Server
- Removing type-specific wrapping logic in favor of a simpler predicate (`isDefinitelyNotNull`)

This results in clearer separation of concerns and more maintainable SQL printer logic.

### Steps to test and verify
- Generate queries involving `ORDER BY` with nullable columns
- Verify SQL output uses `CASE WHEN ... IS NULL` for NULL ordering
- Ensure queries without NULL-sensitive ordering behave unchanged
- Run regression tests for MSSQL backend query generation

None known.

#### Catalog upgrade
- [x] No
- [ ] Yes

#### Metadata
- [x] No
- [ ] Yes

#### GraphQL

#### Breaking changes
- [x] No Breaking changes

### DIFF
--- a/server/src-lib/Hasura/Backends/MSSQL/ToQuery.hs
+++ b/server/src-lib/Hasura/Backends/MSSQL/ToQuery.hs
@@ -666,21 +666,22 @@ fromOrderBys top moffset morderBys =
 
 fromOrderBy :: OrderBy -> [Printer]
 fromOrderBy OrderBy {..} =
-  [ fromNullsOrder orderByExpression orderByNullsOrder,
-    -- Above: This doesn't do anything when using text, ntext or image
-    -- types. See below on CAST commentary.
-    wrapNullHandling (fromExpression orderByExpression)
-      <+> " "
-      <+> fromOrder orderByOrder
-  ]
+  case fromNullsOrder orderByExpression orderByNullsOrder of
+    Nothing ->
+      [ fromExpression orderByExpression
+          <+> " "
+          <+> fromOrder orderByOrder
+      ]
+
+    Just nullExpr ->
+      [ nullExpr
+          <+> ", "
+          <+> fromExpression orderByExpression
+          <+> " "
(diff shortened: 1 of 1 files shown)


jberryman:
Thanks for contributing. Can you fix the compiler errors and make sure you've tested locally please? (Also the license compliance thing will need to be filled out before we could merge)

Closing for now, but feel free to try again

PR #10839, opened 2026-05-15
Fix space leak in withExpiringPGconn loop caused by exceptions

SapotaDA:
**Description**
This Pull Request addresses a sustained in-process memory leak (#10821) observed under steady workloads when `connection_lifetime` is enabled. 

**Root Cause Analysis**
Previously, `withExpiringPGconn` relied on throwing a synchronous exception (`PGConnectionStale` via `Exc.impureThrow`) to force the underlying `resource-pool` library to discard stale connections. Because this control flow was wrapped in a recursive `handleLifted` block, it captured and retained the execution context (including heavily-loaded monadic state thunks) on each exception cycle. In environments with a high connection turnover (e.g., a 120s `connection_lifetime` under steady load), these uncollected thunks accumulated, leading to unbounded growth in anonymous private memory.

**Resolution**
To eliminate the exception-driven control flow and the resulting space leak, `withExpiringPGconn` has been refactored to manually interface with `Data.Pool`. The function now utilizes a tail-recursive loop that explicitly calls `RP.takeResource`, `RP.destroyResource`, and `RP.releaseResource`. 

By proactively destroying stale connections and avoiding `Exc.impureThrow`, we prevent the monadic state from being captured and retained in exception handlers, ensuring stable memory consumption over the lifetime of the process.

**Related Issues**
- Fixes #10821

### DIFF
--- a/server/lib/pg-client/src/Database/PG/Query/Pool.hs
+++ b/server/lib/pg-client/src/Database/PG/Query/Pool.hs
@@ -344,32 +344,32 @@ sqlFromFile fp = do
 -- destroyed per the timeout policy in Data.Pool.
 withExpiringPGconn ::
   (MonadBaseControl IO m, MonadIO m) => PGPool -> (PGConn -> m a) -> m a
-withExpiringPGconn pool f = do
-  -- If the connection was stale, we'll discard it and retry, possibly forcing
-  -- creation of new connection:
-  old <- liftIO getCurrentTime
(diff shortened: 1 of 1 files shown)


jberryman:
Hi, thanks for the contribution, and it's an interesting theory but it doesn't build. Did you actually try this out and confirm it fixes a space leak and perhaps the commit got screwed up? Or is this speculative?

jberryman:
Closing for now, but feel free to comment here or open another PR that's tested

PR #10849, opened 2026-06-12
feat(server): add jsonb path filter operators _jsonb_path_exists and _jsonb_path_match (close #8685)

### LINKED ISSUE #8685: Support SQL/JSON Path Language for querying JSONB columns in Postgres
Currently, the ability to filter a query based on the contents of a JSONB column is extremely limited. Effectively you can only search for an exact string match, or if a given key is/is-not present at the top level of the object.

This means that any solution that wants to implement a JSONB column and needs even slightly flexible query support is effectively stuck. 

For example, we have an app that allows users to define surveys with arbitrary questions. We store the responses to the survey in a JSONB field. If a user wants to run a report of all the responses where some numerical value is > or < a certain value, that's currently impossible which is a major issue.

Postgres provided native functionality to solve this problem in PG12 with SQL/JSON Path Language (https://www.postgresql.org/docs/12/functions-json.html#FUNCTIONS-SQLJSON-PATH). But this is currently not supported by Hasura.

Hasura to implement some form of support for SQL/JSON Path Language. Specifically, the ability to pass a JSONB field Filter Expression as part of a first-class Hasura query filter. This would queries to be filtered on the values of a JSONB column significantly more flexibly than the current way which is extremely limited.

IMO - the ability to pass through native PG Filter Expressions would be fine. As long as it's possible to run the query and use the JSONB column in the filter.

Exposing the relevant values as computed fields and then filtering on them doesn't work for us because the schema of the data in the JSONB column is effectively "user-generated" for lack of a better word, so there is no reliable key to use for a computed field.

For this reason, we need the ability to execute effectively a "user generated" Filter Expression too. We've actually written a super user-friendly query language ourselves that takes really easy-to-understand syntax for business users, parses it, and converts it to the Hasura query filter syntax. We need the ability to extend this in-house query language to let people filter on the contents of the JSONB field.

### If the feature is approved, would you be willing to submit a PR?
Unfortunately, I don't know how to write Haskell. Otherwise, I totally would!

McHill007:
### Description
Adds support for SQL/JSON path language on JSONB columns via two new filter operators:

- `_jsonb_path_exists`: maps to PostgreSQL's `@?` operator — returns rows where the JSONPath returns any item
- `_jsonb_path_match`: maps to PostgreSQL's `@@` operator — returns rows where the JSONPath predicate evaluates to true

This enables flexible filtering on values within JSONB documents, including nested objects and arrays.

**Example:**
```graphql
# Filter orders where any item has quantity > 5
query {
  orders(where: { items: { _jsonb_path_exists: "$[*] ? (@.quantity > 5)" } }) {
    id
    items
  }
}

# Filter products where price < 50
query {
  products(where: { metadata: { _jsonb_path_match: "$.price < 50" } }) {
    id
    metadata
  }
}
```

### Changelog

__Component__: server

__Type__: feature

#### Short Changelog
Add `_jsonb_path_exists` and `_jsonb_path_match` filter operators for JSONB columns using PostgreSQL's SQL/JSON path language (`@?` and `@@`).

### Related Issues
#8685

### Solution and Design
Added two new backend-specific boolean operators for the Postgres backend:

- `AJsonbPathExists` / `SJsonbPathExists` → `@?`
- `AJsonbPathMatch` / `SJsonbPathMatch` → `@@`

The JSONPath expression is passed as a `String` in GraphQL and cast to `jsonpath` in the generated SQL. Changes span the standard operator pipeline: `Types/BoolExp.hs` → `DDL/BoolExp.hs` → `Instances/Schema.hs` → `Translate/BoolExp.hs` → `SQL/DML.hs`.

### Steps to test and verify
1. Create a table with a JSONB column
2. Insert rows with nested JSON objects or arrays
3. Use `_jsonb_path_exists` with a JSONPath filter expression, e.g. `"$[*] ? (@.quantity > 5)"`
4. Use `_jsonb_path_match` with a JSONPath predicate, e.g. `"$.price < 50"`
5. Verify only matching rows are returned

Requires PostgreSQL 12+ (SQL/JSON path language support).

- Requires PostgreSQL 12 or newer
- JSONPath expressions are passed as plain strings with no compile-time validation

#### Catalog upgrade
- [x] No

#### Metadata
- [x] No

#### GraphQL
- [x] New GraphQL schema is being generated:
  - [x] New types and typenames are correlated

#### Breaking changes
- [x] No Breaking changes


hgiasac:
Thanks for contributing. Your PR was merged in https://github.com/hasura/graphql-engine/commit/28c0366ff3c707cf54cb195d854952e0f3f0fb03

PR #10859, opened 2026-07-21
server/mssql: stop emitting redundant IIF null-handling sort keys in ORDER BY so indexes can be used (close #10844)

### LINKED ISSUE #10844: MSSQL: ORDER BY always wraps columns in IIF(col IS NULL, 0, 1) even for NOT NULL columns, causing full table scans
issue summarized by claude:
Environment

Hasura v2.40.0
Data source: Microsoft SQL Server (MSSQL)
Driver: ODBC Driver 18
Description

When querying an MSSQL table with order_by, Hasura wraps every sorted column in an IIF(col IS NULL, 0, 1) expression regardless of whether the column is NOT NULL. This appears to prevent SQL Server from using indexes on unfiltered queries, causing dramatically slower execution compared to an equivalent hand-written query.

GraphQL query:

{
  my_table(
    limit: 100
    order_by: [{ company_id: asc }, { record_id: asc }]
  ) {
    company_id
    record_id
    category
    quantity
    amount
  }
}
SQL Hasura generates:

SELECT ...
FROM [dbo].[my_table] AS [t1]
ORDER BY IIF([t1].[record_id] IS NULL, 0, 1),
         [t1].[record_id] ASC,
         IIF([t1].[company_id] IS NULL, 0, 1),
         [t1].[company_id] ASC
         OFFSET 0 ROWS FETCH NEXT 100 ROWS ONLY
FOR JSON PATH, INCLUDE_NULL_VALUES
SQL I would expect (or write by hand):

SELECT TOP 100 ...
FROM [dbo].[my_table]
ORDER BY company_id ASC, record_id ASC
Both company_id and record_id are defined NOT NULL in the schema. The IIF(col IS NULL, ...) expression is therefore a no-op logically, but it prevents SQL Server's optimizer from using an index on those columns since the sort key is a computed expression rather than the raw column.

Performance impact:

Query	Hasura (generated SQL)	Raw ODBC (hand-written)	Ratio
100 rows, no filter, order by 2 cols	~2,800 ms	~28 ms	~100× slower
100 rows, filtered on indexed col	~180 ms	~140 ms	~1.3× slower
The unfiltered case is the most affected. With a WHERE filter on an indexed column, SQL Server can seek first and then sort the small result set, making the IIF less costly. Without a filter, SQL Server must scan the entire table before sorting, and the IIF prevents early-termination via index seek.

Root cause (from reading the source)

In server/src-lib/Hasura/Backends/MSSQL/Instances/Schema.hs, msOrderByOperators maps plain asc to NullsFirst and desc to NullsLast:

( define Name._asc "in ascending order, nulls first",
  (MSSQL.AscOrder, MSSQL.NullsFirst)    -- always emits IIF
),
( define Name._desc "in descending order, nulls last",
  (MSSQL.DescOrder, MSSQL.NullsLast)    -- always emits IIF
),
This flows through fromAnnotatedOrderByItemG → fromOrderBy → fromNullsOrder in ToQuery.hs, which emits the IIF for anything other than NullsAnyOrder. There is no nullability check anywhere in the chain.

Possible fix

SQL Server's native NULL ordering already matches the intended behavior:

ORDER BY col ASC → NULLs sort first (NULL = lowest value in SQL Server)
ORDER BY col DESC → NULLs sort last
So NullsAnyOrder (which emits no IIF) would produce identical result ordering to the current NullsFirst/NullsLast behavior, while allowing the optimizer to use indexes normally.

The proposed change is to map plain asc/desc to NullsAnyOrder in msOrderByOperators, leaving the explicit asc_nulls_last, asc_nulls_first, desc_nulls_first, desc_nulls_last operators unchanged:

( define Name._asc "in ascending order, nulls first",
  (MSSQL.AscOrder, MSSQL.NullsAnyOrder)   -- defer to SQL Server native (= nulls first for ASC)
),
( define Name._desc "in descending order, nulls last",
  (MSSQL.DescOrder, MSSQL.NullsAnyOrder)  -- defer to SQL Server native (= nulls last for DESC)
),
Questions before I open a PR:

Is this analysis correct — is NullsAnyOrder on SQL Server guaranteed to produce NULLs-first for ASC and NULLs-last for DESC across all SQL Server versions and collations?
Would this be considered a breaking change, given that the generated SQL changes (even though result ordering should be identical)?
Is there a known reason the current behavior was chosen over NullsAnyOrder that I'm missing — e.g. a specific SQL Server edge case where native NULL ordering differs?
Are there existing integration tests for MSSQL ORDER BY NULL behavior that would catch a regression?
Happy to open a PR with the two-line change if this direction is confirmed.

b3hz4d:
### Description

On SQL Server, `NULLS FIRST` / `NULLS LAST` has no native syntax, so the MSSQL backend emulates every requested nulls order by prepending an `IIF(<column> IS NULL, 0, 1)` sort key to the `ORDER BY` clause — including for plain `asc` / `desc`. Because the sort key is a computed expression, SQL Server cannot use an index on the ordered column, which degrades unfiltered sorted queries to full scans (#10844 reports ~100× slowdowns).

The `IIF` is almost always unnecessary:

1. SQL Server natively sorts `NULL` as the lowest value — ascending order already places nulls first and descending order already places nulls last. So `asc`, `asc_nulls_first`, `desc` and `desc_nulls_last` (whose documented semantics exactly match the native placement) need no emulation at all. These operators now map to `NullsAnyOrder` and emit a bare `ORDER BY <col> ASC/DESC`.
2. For the two operators that invert the native placement (`asc_nulls_last`, `desc_nulls_first`), the IR-to-TSQL translation now tracks whether the ordering expression can ever evaluate to `NULL`, and skips the null-handling sort key when it provably cannot (`NOT NULL` column, no redaction).

After both changes, the `IIF` sort key survives only where it is semantically required: `asc_nulls_last` / `desc_nulls_first` on a genuinely nullable column.

Example from #10844, `order_by: [{ company_id: asc }, { record_id: asc }]`:

Before:

```sql
ORDER BY IIF([t1].[company_id] IS NULL, 0, 1), [t1].[company_id] ASC,
         IIF([t1].[record_id] IS NULL, 0, 1), [t1].[record_id] ASC
```

After:

```sql
ORDER BY [t1].[company_id] ASC, [t1].[record_id] ASC
```

Result ordering is unchanged in all cases.

### Changelog

__Component__ : server

__Type__: enhancement

#### Short Changelog

MSSQL: stop emitting redundant `IIF(.. IS NULL, ..)` sort keys in `ORDER BY` (use native null ordering for `asc`/`desc`; skip null handling for non-nullable columns) so indexes can be used

### Related Issues

close #10844

### Solution and Design

Two complementary changes:

**1. Schema level** (`Hasura.Backends.MSSQL.Instances.Schema.msOrderByOperators`): the operators whose requested null placement equals SQL Server's native placement (`asc`, `asc_nulls_first`, `desc`, `desc_nulls_last`) now map to `NullsAnyOrder`, which prints no null-handling sort key. SQL Server's treatment of `NULL` as the lowest sort value is documented, version-independent behavior (collations affect string comparison, not `NULL` placement), so result ordering is identical. Only `asc_nulls_last` and `desc_nulls_first` keep an explicit nulls order.

**2. Translation level** (`Hasura.Backends.MSSQL.FromIr.Query.unfurlAnnotatedOrderByElement`): now returns, alongside the ordering expression and its scalar type, a `Nullable` flag describing whether the expression may evaluate to `NULL`:

- `AOCColumn`: non-nullable when the column is declared `NOT NULL` **and** there is no redaction expression (an inline redaction `CASE` can turn any value into `NULL`).
- `AOCObjectRelation`: conservatively nullable — the relationship is realized as an `OUTER APPLY` join, so the joined row as a whole may be missing even if the targeted column is non-nullable.
- `AOCArrayAggregation`: conservatively nullable — aggregates other than `COUNT` (e.g. `MIN`, `SUM`) return `NULL` when no rows match.

`fromAnnotatedOrderByItemG` then forces the nulls order to `NullsAnyOrder` (which prints nothing) for provably non-null expressions, so `ToQuery.fromNullsOrder` never emits the `IIF` sort key for them. Nullable expressions behave exactly as before.

Change 1 implements the fix proposed in #10844 (extended to `asc_nulls_first` / `desc_nulls_last`, which are also native); change 2 additionally covers `asc_nulls_last` / `desc_nulls_first` on non-nullable columns, where dropping the sort key is unconditionally sound regardless of native ordering semantics.

### Steps to test and verify

- New unit tests in `server/src-test/Hasura/Backends/MSSQL/FromIr/QuerySpec.hs` cover: nullable column with an inverting nulls order (IIF kept), non-nullable column (IIF dropped), no/any nulls order (no IIF, what plain `asc`/`desc` now produce), redacted non-nullable column (IIF kept), and a mixed ORDER BY.
- Run with: `cabal run graphql-engine-tests -- --match "fromSelect: ORDER BY null handling"`
- Ordering semantics are unchanged in all cases: for a `NOT NULL` column the `IIF` was a constant, and for the native-matching operators SQL Server already sorts `NULL` lowest (first for ASC, last for DESC).

- `asc_nulls_last` / `desc_nulls_first` on a nullable column still emit the `IIF` sort key — there, the null placement genuinely inverts SQL Server's native order, so a computed sort key is unavoidable in a single query.
- Aggregate orderings (e.g. `_aggregate: {count: ...}`) with an inverting nulls order still emit the `IIF` sort key even though `COUNT` never returns `NULL`; this could be a follow-up optimization.

#### Catalog upgrade

- [x] No
- [ ] Yes

#### Metadata

- [x] No

#### GraphQL

#### Breaking changes

- [x] No Breaking changes

MSSQL emulated every requested nulls order with an IIF(<col> IS NULL, ..)
sort key, a computed expression that prevents index use when sorting.
SQL Server natively sorts NULL lowest (first for ASC, last for DESC), so
operators whose semantics match the native placement (asc, asc_nulls_first,
desc, desc_nulls_last) now emit no null-handling sort key at all. For the
two operators that invert the native placement, nullability is tracked
during IR translation and the sort key is dropped when the ordering
expression provably cannot be NULL.

### DIFF
--- a/server/graphql-engine.cabal
+++ b/server/graphql-engine.cabal
@@ -1141,6 +1141,7 @@ test-suite graphql-engine-tests
     , monad-control
     , mtl
     , network-uri
+    , odbc
     , openapi3
     , optparse-applicative
     , pg-client
@@ -1208,6 +1209,7 @@ test-suite graphql-engine-tests
     Hasura.Backends.DataConnector.API.V0.SchemaSpec
     Hasura.Backends.DataConnector.API.V0.TableSpec
     Hasura.Backends.DataConnector.API.V0.TargetSpec
+    Hasura.Backends.MSSQL.FromIr.QuerySpec
     Hasura.Backends.Postgres.Connection.VersionCheckSpec
     Hasura.Backends.Postgres.Execute.PrepareSpec
     Hasura.Backends.Postgres.NativeQueries.NativeQueriesSpec
--- a/server/src-lib/Hasura/Backends/MSSQL/FromIr/Query.hs
+++ b/server/src-lib/Hasura/Backends/MSSQL/FromIr/Query.hs
@@ -967,29 +967,44 @@ fromAnnotatedOrderByItemG ::
   IR.AnnotatedOrderByItemG 'MSSQL Expression ->
   WriterT (Seq UnfurledJoin) (ReaderT EntityAlias FromIr) OrderBy
 fromAnnotatedOrderByItemG IR.OrderByItemG {obiType, obiColumn = obiColumn, obiNulls} = do
-  (orderByExpression, orderByType) <- unfurlAnnotatedOrderByElement obiColumn
-  let orderByNullsOrder = fromMaybe NullsAnyOrder obiNulls
+  (orderByExpression, orderByType, nullability) <- unfurlAnnotatedOrderByElement obiColumn
+  let orderByNullsOrder = case nullability of
+        -- When the ordering expression can never evaluate to NULL, a nulls
+        -- ordering would only add a redundant @IIF(.. IS NULL, ..)@ sort key
+        -- to the generated query, so we drop it.
+        IR.NotNullable -> NullsAnyOrder
+        IR.Nullable -> fromMaybe NullsAnyOrder obiNulls
       orderByOrder = fromMaybe AscOrder obiType
   pure OrderBy {..}
 
 -- | Unfurl the nested set of object relations (tell'd in the writer)
 -- that are terminated by field name (IR.AOCColumn and
 -- IR.AOCArrayAggregation).
+--
+-- Also reports whether the resulting ordering expression may evaluate to
+-- NULL, so that null handling can be skipped when it provably cannot.
 unfurlAnnotatedOrderByElement ::
   IR.AnnotatedOrderByElement 'MSSQL Expression ->
-  WriterT (Seq UnfurledJoin) (ReaderT EntityAlias FromIr) (Expression, Maybe TSQL.ScalarType)
+  WriterT (Seq UnfurledJoin) (ReaderT EntityAlias FromIr) (Expression, Maybe TSQL.ScalarType, IR.Nullable)
 unfurlAnnotatedOrderByElement =
   \case
     IR.AOCColumn columnInfo redactionExp -> do
       fieldName <- lift (fromColumnInfo columnInfo)
       ex <- lift $ potentiallyRedacted redactionExp (ColumnExpression fieldName)
+      let nullability = case redactionExp of
+            -- Redaction wraps the column in a conditional that can produce
+            -- NULL even for non-nullable columns.
+            IR.NoRedaction
+              | not (IR.ciIsNullable columnInfo) -> IR.NotNullable
+            _ -> IR.Nullable
       pure
         ( ex,
           case IR.ciType columnInfo of
             IR.ColumnScalar t -> Just t
             -- Above: It is of interest to us whether the type is
             -- text/ntext/image. See ToQuery for more explanation.
-            _ -> Nothing
+            _ -> Nothing,
+          nullability
         )
     IR.AOCObjectRelation IR.RelInfo {riMapping = IR.RelMapping mapping, riTarget = IR.RelTargetNativeQuery nativeQueryName} annBoolExp annOrderByElementG -> do
       let name = T.toTxt (getNativeQueryName nativeQueryName)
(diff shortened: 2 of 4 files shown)


jberryman:
Your change has landed in main in our mono repo. Thanks!

PR #10869, opened 2026-08-09
fix(tracing): mark database query/mutation spans as client, not internal

### LINKED ISSUE #10868: Database spans incorrectly have `internal` span.kind instead of `client`
### Version Information

Server Version: 2.50.0

### Environment

EE

The OTEL span for when Hasura makes a DB query, e.g. `Postgres Query for root field "my_query"` currently has a `span.kind` of `internal`.

Because of this, and the way Datadog [maps OTEL span types](https://docs.datadoghq.com/opentelemetry/mapping/semantic_mapping/?tab=datadogexporter#map-opentelemetry-span-attribute-to-datadog-span-type) to DD span types, the span ends up with a type of `custom` which breaks OTEL trace => DD database monitoring correlation.

As per the [semantic conventions](https://opentelemetry.io/docs/specs/semconv/db/database-spans/), the `span.kind` for database related spans _should_ be `client`, as it's not making a call to an in-memory database.

> Span kind SHOULD be CLIENT. It MAY be set to INTERNAL on spans representing in-memory database calls. It’s RECOMMENDED to use CLIENT kind when database system being instrumented usually runs in a different process than its client or when database calls happen over instrumented protocol such as HTTP.

1. Configure https://hasura.io/docs/2.0/observability/opentelemetry/graphql-engine/
2. Add the `HASURA_GRAPHQL_ENABLE_QUERY_TRACING: "true"` env var
3. Execute a query and look at the trace in something like Jaeger

### Screenshots or Screencast

Image

### Any possible solutions/workarounds you're aware of?

Manually override `span.kind` to `client` via OTEL collector

### Keywords

- "span.kind"
- "internal span"

eeshsaxena:
Fixes #10868.

## Problem

Spans for database queries and mutations are created with `SpanKind` `SKInternal` ("internal"), but per the [OTel database span semantic conventions](https://opentelemetry.io/docs/specs/semconv/db/database-spans/) a database call is a `client` span. Because of the wrong kind, exporters that map OTel span kinds to their own types (e.g. Datadog) classify these as `custom` rather than a DB span, which breaks OTel-trace → DB-monitoring correlation.

The `DataConnector` backend already gets this right — it uses `SKClient` for both its query and mutation spans — so Postgres and MSSQL are simply out of line with it.

## Fix

Change the `newSpan ... SKInternal` to `SKClient` for the DB query/mutation spans in both backends, matching `DataConnector` and the semantic conventions.

The issue only mentions the Postgres query span, but the same defect is present on every sibling DB span, so all of them are corrected together:

- `Backends/Postgres/Instances/Transport.hs` — `Postgres Query for root field ...` and both `Postgres Mutation for root field ...` spans.
- `Backends/MSSQL/Instances/Transport.hs` — `MSSQL Query for root field ...` and `MSSQL Mutation for root field ...` spans.

These are the only `SKInternal` uses in the two transport modules, and each is a database span; no non-DB internal spans are affected. `SKClient` renders as `"client"` (see `Tracing.TraceId`).

### DIFF
--- a/server/src-lib/Hasura/Backends/MSSQL/Instances/Transport.hs
+++ b/server/src-lib/Hasura/Backends/MSSQL/Instances/Transport.hs
@@ -78,7 +78,7 @@ runQuery ::
 runQuery reqId query fieldName _userInfo logger _ sourceConfig tx genSql _ = do
   logQueryLog logger $ mkQueryLog query fieldName genSql reqId
   withElapsedTime
-    $ newSpan ("MSSQL Query for root field " <>> fieldName) SKInternal
+    $ newSpan ("MSSQL Query for root field " <>> fieldName) SKClient
     $ (<* attachSourceConfigAttributes @'MSSQL sourceConfig)
     $ fmap snd (run tx)
 
@@ -116,7 +116,7 @@ runMutation ::
 runMutation reqId query fieldName _userInfo logger _ sourceConfig tx _genSql _ = do
   logQueryLog logger $ mkQueryLog query fieldName Nothing reqId
   withElapsedTime
-    $ newSpan ("MSSQL Mutation for root field " <>> fieldName) SKInternal
+    $ newSpan ("MSSQL Mutation for root field " <>> fieldName) SKClient
     $ (<* attachSourceConfigAttributes @'MSSQL sourceConfig)
     $ run tx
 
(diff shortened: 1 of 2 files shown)


eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

Blacksmoke16:
This would be still good to have yes :pray:.

hgiasac:
Thanks @eeshsaxena for contributing. Your PR was merged

